PR #14, opened 2021-11-17
Query programs by name

kt474:
### Summary

Ported from https://github.com/Qiskit-Partners/qiskit-ibm/pull/194

### Details and comments
Fixes https://github.com/Qiskit-Partners/qiskit-ibm/issues/180


rathishcholarajan (changes requested):
@kt474 Could you please resolve conflicts.

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -843,13 +850,19 @@ def programs(
         """
         if skip is None:
             skip = 0
+        if name:
+            for program in list(self._programs.values())[skip : limit + skip]:
+                if program.name == name:
+                    self._programs[program.program_id] = program
+                else:
+                    del self._programs[program.program_id]
```
rathishcholarajan, line 858: 1. We should loop through all programs in self._programs.
2. If we find a program with given name then we should add it to a temporary list.
3. After we find all programs with the given name, we should then apply limit and skip on the temporary list and immediately return it.

All this local filtering should happen only if refresh is not set to True and self._programs is non empty.

We shouldn’t delete programs from self._programs since it’s usually lot of info that’ll have to be refetched from sever.

rathishcholarajan (changes requested):

rathishcholarajan:
@kt474 
1. The first time when no programs are there in cache, we should fetch all programs and store in cache. If name was passed in the first call and cache is empty we should fetch all programs without passing name to backend, do local filtering with name and then return the results.
2. If cache is non empty (second and further requests), if name is passed then do local filtering and if name and refresh are passed then fetch programs with name filter set, update programs local cache based on returned results, and then return the results.

kt474:
How would limit/skip work in this? 
1. The first time the cache is empty (which is usually going to be the first call) - regardless of parameters, ignore skip/limit and fetch all programs so the cache has all programs. Then if there are filters, filter locally and return those results without updating the whole cache 
2. Afterwards, when the cache is non empty, if `refresh=False`, then just do local filtering, and only update the cache when `refresh=True`

rathishcholarajan:
> I assume limit/skip will remain working the same way
> 
> 1. The first time the cache is empty (which is usually going to be the first call) - fetch all programs within the limit/skip parameters. Then if there is a name given, filter locally and return those results without updating the whole cache.
> 2. Afterwards, when the cache is non empty, if `refresh=False`, then just do local filtering, and only update the cache when `refresh=True`

The first call should fetch all programs, no matter limit/skip and store in cache. And then limit/skip should be applied always at the end and return results. This is how it works today and it shouldn’t change.

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -171,6 +171,7 @@ def __init__(
         )
         self._backends: Dict[str, "ibm_backend.IBMBackend"] = {}
         self._api_client = None
+        self._programs_first_call = True
```
rathishcholarajan, line 174: Do we need this additional variable? If self._programs is empty doesn’t that mean it’s the first call?

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -870,11 +880,22 @@ def programs(
...
+                    if not self._programs_first_call:
+                        refreshed = True
+                    self._programs_first_call = False
                     # Stop if there are no more programs returned by the server.
                     break
                 offset += len(program_page)
         if limit is None:
             limit = len(self._programs)
+        if name:
+            if refreshed:
```
rathishcholarajan, line 892: Do we need this additional variable? If refresh is passed can’t we just make another call here to API to get programs by name?

For second time onwards, if refresh is True and if no name is set then repopulate entire cache,
If refresh is True and name is set, then repopulate partial cache only with items fetch for that name.

So in the call above where you populate entire cache you don’t have to pass name.
In the call below where you populate partial cache you can pass name. It’ll be a lot easier to understand the code that way.

rathishcholarajan (changes requested):

qiskit_ibm_runtime/ibm_runtime_service.py
kt474: If refresh is passed then we wouldn't need to make another api call here right? Just refresh on the first call and whenever refresh is True and then handle all name querying locally

rathishcholarajan: We could do that but we would end up not using the API at all. If name and refresh are passed it’d be faster to make the API call to just update the cache partially.

kt474: If we don't pass name in the first call and only in the second - if both name and refresh are passed wouldn't there be two api calls? Don't we have to use an additional variable/logic and pass in name in the first call?

rathishcholarajan: For the first time when cache is empty once we populate all programs using first call, set refresh to false, so when it goes to the latter call only local filtering will happen. From second time onwards first call won’t execute but the latter call will execute when name and refresh are set. So at any given time only one API call should be made.

qiskit_ibm_runtime/ibm_runtime_service.py
rathishcholarajan, line 174:
```suggestion
```

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -908,6 +909,31 @@ def program(self, program_id: str, refresh: bool = False) -> RuntimeProgram:
 
         return self._programs[program_id]
 
+    def _retrieve_programs(self, name: str = "") -> None:
+        """Make an API call to fetch programs.
+
+        Args:
+            name: Name of program.
+        """
+        self._programs = {}
```
rathishcholarajan, line 918: This function should not modify self._programs right? If we call this function with name parameter then we will lose all programs already in cache that don’t match the name. This function should return programs and they should be stored in self._programs in the calling function above when called without name and stored in temp variable when called with name.

rathishcholarajan (changes requested):

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -846,35 +852,30 @@ def programs(
...
             limit = len(self._programs)
+        if name:
+            if refresh and not already_retrieved:
+                temp_programs = self._retrieve_programs(name)
+                return list(temp_programs.values())[skip : limit + skip]
+            matched_programs = []
+            for program in list(self._programs.values()):
+                if program.name == name:
+                    matched_programs.append(program)
+            return matched_programs[skip : limit + skip]
```
rathishcholarajan, line 878:
```suggestion
            matched_programs = []
            if refresh and not already_retrieved:
                matched_programs = list(self._retrieve_programs(name).values())
            else:
                for program in list(self._programs.values()):
                    if program.name == name:
                        matched_programs.append(program)
            return matched_programs[skip : limit + skip]
```
Can we do something like this?

rathishcholarajan (changes requested):

kt474:
Closing and reopening https://github.com/Qiskit/qiskit-ibm-runtime/pull/77 to avoid switching between forks

PR #18, opened 2021-11-17
Use Black for code formatting 

### LINKED ISSUE #9: Enable Black Code Style
Use [Black Code Style](https://black.readthedocs.io/en/stable/the_black_code_style/index.html) instead of `pycodestyle`.

Increased development efficiency, because
- less time is needed to discuss if personal code style preferences do not match
  - `pycodestyle` leaves room for that
- code style changes are applied or fixed automatically
  - `pycodestyle` requires developers to fix identified style issues manually

Inspired from https://github.com/Qiskit/qiskit-terra/pull/6361. 

**Technical Implementation**
- install `black`
- reformat existing code via `black folder1 folder 2`
- subsequently fail verification if `black --check folder1 folder2` returns a non-zero exit code (to force contributors to adhere to the new code style)

daka1510:
### Summary
Use Black for code formatting. 

### Details and comments
Fixes #9

### Related Links and Discussions
- https://black.readthedocs.io/en/stable/integrations/editors.html
- https://github.com/Qiskit/qiskit-terra/pull/6361
- https://github.com/Qiskit/qiskit-terra/pull/5883

My main motivation to raise this here matches the statement from @levbishop in https://github.com/Qiskit/qiskit-terra/pull/5883:
> Currently we rely on pycodestyle and pylint for nagging us to write well-styled code.
That's a bit annoying because it can take a couple iterations to get those tools happy (especially for inexperienced contributors) and also because pylint especially is really slow.


rathishcholarajan:
@daka1510 I've fixed the broken tests. Could you please rebase and run black again and update this PR. Thanks.

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -117,10 +127,7 @@ class IBMRuntimeService:
     """
 
     def __init__(
-            self,
-            token: Optional[str] = None,
-            url: Optional[str] = None,
-            **kwargs: Any
+        self, token: Optional[str] = None, url: Optional[str] = None, **kwargs: Any
```
rathishcholarajan, line 130: I am not particularly fond of black doing this when it’s clearer on separate lines. Is there a setting to disable this?

qiskit_ibm_runtime/test/ibm_provider_mock.py
```diff
@@ -33,7 +33,9 @@ def mock_get_backend(backend):
     """
     mock_ibm_provider = MagicMock()
     if not hasattr(backend_mocks, backend):
-        raise NameError("The specified backend name is not a valid mock from " "qiskit.test.mock")
+        raise NameError(
+            "The specified backend name is not a valid mock from " "qiskit.test.mock"
```
rathishcholarajan, line 37:
```suggestion
            "The specified backend name is not a valid mock from qiskit.test.mock"
```

daka1510:
@rathishcholarajan done. All tests passed. Can you take a look if you agree with the proposed changes?

.pylintrc
```diff
@@ -212,7 +213,7 @@ max-nested-blocks=5
 [FORMAT]
 
 # Maximum number of characters on a single line.
-max-line-length=100
+max-line-length=105
```
rathishcholarajan, line 216: Is there a significance to 105? Just curious.

rathishcholarajan (changes requested):
Looks good for the most part. Left few comments.

.pylintrc
daka1510: It accomodates the fact that default the line length limit in Black is not a hard limit. This change makes sure that linting doesn't fail with:

```
************* Module qiskit_ibm_runtime.credentials.hub_group_project_id
qiskit_ibm_runtime/credentials/hub_group_project_id.py:49:0: C0301: Line too long (102/100) (line-too-long)
```

Background: 
- https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length
> [...] Black defaults to 88 characters per line [...] 
> Black will try to respect that. However, sometimes it won’t be able to without breaking other rules. In those rare cases, **auto-formatted code will exceed your allotted limit.**

qiskit_ibm_runtime/ibm_runtime_service.py
daka1510: As Black is an opinionated code formatter, it does not leave much room for configuration.  You can read more about how Black wraps lines at https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines. In this case all arguments fit in a single line (<88 chars) and have therefore been updated.

As a final fallback we can always make use of 
>  It [Black] doesn’t reformat blocks that start with # fmt: off and end with # fmt: on, or lines that ends with # fmt: skip. 

In this case I personally would prefer to stick with the defaults.

daka1510:
Thanks for your quick feedback, @rathishcholarajan. Added comments and included your proposals. Also, you indirectly made me realize I missed to clean up referecnes to `pycodestyle`. This is now addressed via https://github.com/Qiskit/qiskit-ibm-runtime/pull/18/commits/7b3bdb415e50cbdd8131b28117d99918e69e11dc.

rathishcholarajan (approved):
Looks good! Thanks @daka1510 and congratulations on your first contribution here! 🙂🎉

PR #33, opened 2021-11-29
Use websocket stream to retrieve runtime job results

kt474:
### Summary

Ported from https://github.com/Qiskit-Partners/qiskit-ibm/pull/175

### Details and comments
Fixes https://github.com/Qiskit-Partners/qiskit-ibm/issues/117


kt474:
Failing integration test [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/ibm/runtime/test_runtime_integration.py#L520) where `called_back` is true instead of false. My understanding is that if `stream_results` is force called on a completed job then the callback function would still be called which sets `called_back` to true?

jyu00:
@kt474 can we raise the priority of this? Polling every 5 seconds is not helping our cron job connection issues nor our performance numbers :(

kt474:
> @kt474 can we raise the priority of this? Polling every 5 seconds is not helping our cron job connection issues nor our performance numbers :(

Yeah the functionality is there, but for the test cases I wasn't sure if we reached a decision on 
- using a mock wait function every time `wait_for_final_state` is used (`test_integration_interim_results`, `test_integration_job`, `test_integration_retrieve_job`, `test_job_retrieval`, `test_jobs`)
or
- using the `MockWsServer` for all these test suites?

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -262,14 +249,12 @@ def stream_results(
             RuntimeInvalidStateError: If a callback function is already streaming results or
                 if the job already finished.
         """
+        if self._status in JOB_FINAL_STATES:
```
kt474, line 252: Not sure if/how this logic needs to be updated? If wait_for_final_state times out and stream_results is called, the error messages would be inaccurate

kt474: also there are still a couple failing integration tests in push-test

test/test_job_retrieval.py
```diff
@@ -183,8 +185,11 @@ def test_jobs_filter_by_program_id(self, service):
 
         job = run_program(service=service, program_id=program_id)
         job_1 = run_program(service=service, program_id=program_id_1)
-        job.wait_for_final_state()
-        job_1.wait_for_final_state()
+        with mock.patch.object(
+            RuntimeJob, "wait_for_final_state", side_effect=time.sleep(0.5)
```
rathishcholarajan, line 189: Why do we have to put different sleep times in different functions? This can get very flaky. I think we should write a function that returns status in BaseFakeRuntimeJob and any call to wait_for_final_state should be patched with a function that constantly checks status method in BaseFakeRuntimeJob to see if the job status is final.

rathishcholarajan (changes requested):

test/test_integration_interim_results.py
```diff
@@ -83,16 +83,16 @@ def test_stream_results_done(self, service):
...
 
-        called_back = False
+        called_back_count = 0
         job = self._run_program(service, interim_results="foobar")
         job.wait_for_final_state()
         job._status = JobStatus.RUNNING  # Allow stream_results()
         job.stream_results(result_callback)
         time.sleep(2)
-        self.assertFalse(called_back)
+        self.assertEqual(1, called_back_count)
```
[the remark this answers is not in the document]
kt474: `stream_results` will be called in `wait_for_final_state` so the callback will always be called once

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -222,29 +218,20 @@ def error_message(self) -> Optional[str]:
...
+    def wait_for_final_state(self, timeout: Optional[float] = None) -> None:
         """Poll the job status until it progresses to a final state such as ``DONE`` or ``ERROR``.
 
         Args:
             timeout: Seconds to wait for the job. If ``None``, wait indefinitely.
-            wait: Seconds between queries.
 
         Raises:
             JobTimeoutError: If the job does not reach a final state before the
                 specified timeout.
```
daka1510, line 229: You removed the import but did not update the docstring. Should the code continue to raise a `JobTimeoutError` given that only an implementation detail (socket vs. polling) changed?

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -222,29 +218,20 @@ def error_message(self) -> Optional[str]:
         self._set_status_and_error_message()
         return self._error_message
 
-    def wait_for_final_state(
-        self, timeout: Optional[float] = None, wait: float = 5
-    ) -> None:
+    def wait_for_final_state(self, timeout: Optional[float] = None) -> None:
         """Poll the job status until it progresses to a final state such as ``DONE`` or ``ERROR``.
```
daka1510, line 222: "Poll the job status" is no longer accurate. Can you shortly document how the code determines when the job reaches a final state?

I don't understand this looking at the code changes alone. Maybe there are some underlying implicit assumptions (e.g. `/stream/jobs/{id}` does not open a stream until a job succeeds) that are not obviously clear.

daka1510 (commented):
Also took a short look to get an understanding of the upcoming changes. Left 2 additional comments.

daka1510:
test

test/test_job_retrieval.py
daka1510: Agree that we should be very carefuly with any types of timeouts in our unit tests. @rathishcholarajan, @kt474 has updated the code. Does that fit your proposal?

test/unit/mock/fake_runtime_client.py
```diff
@@ -451,6 +455,14 @@ def job_delete(self, job_id):
         self._get_job(job_id)
         del self._jobs[job_id]
 
+    def wait_for_final_state(self, job_id):
+        """Wait for the final state of a program job."""
+        final_states = ["COMPLETED", "FAILED", "CANCELLED", "CANCELLED - RAN TOO LONG"]
+        status = self._get_job(job_id).status()
+        while status not in final_states:
+            time.sleep(1)
```
daka1510, line 463: Is there any reason for putting in "1 second" here? We should try to keep the duration of our unit tests low and 1s is a long time...

daka1510 (changes requested):
@kt474 - as you updated code paths that may affect the integration tests, please make sure to configure the correct environment secrets in your fork so that the integration tests will not fail due to configuration issues

image

Apart from that, I just added one more question and will leave it to @rathishcholarajan to check the mock part as he initially proposed how to change it.

test/unit/mock/fake_runtime_client.py
kt474: No particular reason, the original `wait_for_final_state` had a sleep but I can try removing it

test/unit/test_jobs.py
```diff
@@ -259,8 +289,13 @@ def test_get_result_twice(self, service):
         job_cls.custom_result = custom_result
 
         job = run_program(service=service, job_classes=job_cls)
-        _ = job.result()
-        _ = job.result()
+        with mock.patch.object(
+            RuntimeJob,
+            "wait_for_final_state",
+            side_effect=self._fake_wait_for_final_state(service, job),
+        ):
```
rathishcholarajan, line 296: This is too verbose to repeat in all tests, could you move to a function in utils and then use it in tests?

rathishcholarajan (changes requested):

test/unit/test_jobs.py
kt474: Since `mock.patch.object` is used as a context manager how can we move it to a separate function?

rathishcholarajan: Could you try if something like this works?
```
with mock_wait_for_final_state(service, job):
    job.wait_for_final_state()
```

```
def mock_wait_for_final_state(service, job):
    return mock.patch.object(
            RuntimeJob,
            "wait_for_final_state",
            side_effect=fake_wait_for_final_state(service, job),
        )
```

rathishcholarajan (approved):
LGTM!

rathishcholarajan:
@kt474 Could you cherry pick this to qiskit-ibmq-provider as well?

PR #40, opened 2021-11-30
Drop support for Python 3.6 and use Python 3.9 by default.

daka1510:
### Summary
Drop support for Python 3.6 and use Python 3.9 by default. Also update pylint to latest version.

Why?
- Python 3.6 will reach end of life soon (Dec' 2021)
  - I don't think we need to wait in this case until qiskit-terra drops support for Python 3.6 ([[1]](https://github.com/Qiskit-Partners/qiskit-ibm/pull/183#issuecomment-954996658))
- Python 3.9 is the latest version that qiskit-terra currently supports
  - I'm a fan of using latest stable versions by default. As we heavily depend on qiskit-terra, I believe we need to wait until it supports Python 3.10.

### Details and comments
- Related: https://github.com/Qiskit-Partners/qiskit-ibm/pull/183


.pylintrc
```diff
@@ -1,265 +1,313 @@
 [MASTER]
 
-# Specify a configuration file.
```
daka1510, line 3: Regenerated this file via `pylint --generate-rcfile` (using the most recent pylint version).

Subsequent modifications
- disabled rules that we currently do not fully adhere to (marked with TODO or another comment)
- reintroduced docparams and docstyle plugins we used before

requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 mypy==0.812
-pylint==2.4.4
+pylint==2.12.1
```
daka1510, line 2: `pylint 2.4.4` was released about 2 years ago. I believe some of the best practices have changed since that time as the APIs have evolved.

Used the update to Python 3.9 as a trigger to switch to the most recent version. Was forced to do so because `pylint 2.4.4` reported many false  positives with `Python 3.9` (see https://github.com/PyCQA/pylint/issues/3882 for details).

.pylintrc
```diff
@@ -1,265 +1,313 @@
...
+        too-many-instance-attributes,   # too verbose
+        too-many-lines,                 # too verbose
+        too-many-locals,                # too verbose
+        too-many-nested-blocks,         # too verbose
+        too-many-public-methods,        # too verbose
+        too-many-statements,            # too verbose
+        unnecessary-pass,               # allow for methods with just "pass", for clarity
+        unspecified-encoding,           # TODO: investigate / re-enable
+        use-dict-literal,               # TODO: investigate / re-enable
+        use-list-literal,               # TODO: investigate / re-enable
```
daka1510, line 103: @rathishcholarajan - Do you agree if I would open a follow-up issue to investigate and/or re-enable some of the rules again?

My feeling is that using an up-to-date linter will help developers that are not Python professionals yet (like myself) to stick with best practices.

daka1510:
All push tests in my fork succeeded
- https://github.com/daka1510/qiskit-ibm-runtime/actions/runs/1525660680

rathishcholarajan (approved):
Looks good @daka1510!

PR #46, opened 2021-12-06
Rest adapter for IBM Cloud backends

rathishcholarajan:
### Summary
Created a REST adapter for accessing the IBM Cloud backends, backend properties, configuration, status and pulse defaults.

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/api/rest/cloud_backend.py
+++ b/qiskit_ibm_runtime/api/rest/cloud_backend.py
@@ -56,7 +56,6 @@ class CloudBackend(RestAdapterBase):
         Returns:
             JSON response of backend properties.
         """
-        # pylint: disable=redefined-outer-name
         url = self.get_url("properties")
         response = self.session.get(url).json()
         # Adjust name of the backend.
--- a/qiskit_ibm_runtime/api/rest/cloud_backend.py
+++ b/qiskit_ibm_runtime/api/rest/cloud_backend.py
@@ -1,0 +1,99 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""IBM Cloud Backend REST adapter."""
+
+from typing import Dict, Any
+
+from .base import RestAdapterBase
+from ..session import RetrySession
+
+
+class CloudBackend(RestAdapterBase):
+    """Rest adapter for IBM Cloud backend related endpoints."""
+
+    URL_MAP = {
+        "configuration": "/configuration",
+        "properties": "/properties",
+        "pulse_defaults": "/defaults",
+        "status": "/status",
+    }
+
+    def __init__(
+        self, session: RetrySession, backend_name: str, url_prefix: str = ""
+    ) -> None:
+        """Backend constructor.
+
+        Args:
+            session: Session to be used in the adaptor.
+            backend_name: Name of the backend.
+            url_prefix: Base URL.
+        """
+        self.backend_name = backend_name
+        super().__init__(session, "{}/devices/{}".format(url_prefix, backend_name))
+
+    def configuration(self) -> Dict[str, Any]:
+        """Return backend configuration.
+
+        Returns:
+            JSON response of backend configuration.
+        """
+        url = self.get_url("configuration")
+        return self.session.get(url).json()
+
+    def properties(self) -> Dict[str, Any]:
+        """Return backend properties.
+
+        Returns:
+            JSON response of backend properties.
+        """
+        # pylint: disable=redefined-outer-name
+        url = self.get_url("properties")
+        response = self.session.get(url).json()
+        # Adjust name of the backend.
+        if response:
+            response["backend_name"] = self.backend_name
+        return response
+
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/api/rest/cloud_backend.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class CloudBackend(RestAdapterBase):
+    """Rest adapter for IBM Cloud backend related endpoints."""
+
+    URL_MAP = {
+        "configuration": "/configuration",
+        "properties": "/properties",
+        "pulse_defaults": "/defaults",
+        "status": "/status",
```
daka1510, line 28: Where can I find the related API docs? https://us-east.quantum-computing.cloud.ibm.com/openapi/ doesn't list those endpoints.

qiskit_ibm_runtime/api/rest/runtime.py
```diff
@@ -185,6 +191,31 @@ def logout(self) -> None:
         url = self.get_url("logout")
         self.session.post(url)
 
+    # IBM Cloud only functions
```
daka1510, line 194: Why do you use separate files for backend related adapters (`/rest/backend.py` vs. `/rest/cloud_backend.py`) and a single file for runtime related adapters?

qiskit_ibm_runtime/api/rest/cloud_backend.py
```diff
@@ -0,0 +1,99 @@
...
+        }
+        # 'pending_jobs' is required, and should be >= 0.
+        if "length_queue" in response:
+            ret["pending_jobs"] = max(response["length_queue"], 0)
+        else:
+            ret["pending_jobs"] = 0
+        # Not part of the schema.
+        if "busy" in response:
+            ret["dedicated"] = response["busy"]
+        return ret
```
daka1510, line 99: This is mostly a duplicate of https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/api/rest/backend.py#L85-L112. Just a little detail appears to differ (`length_queue` vs. `lengthQueue`).

Do we expect both code paths to evolve independently from here on or should we strive for more reuse?

qiskit_ibm_runtime/api/rest/cloud_backend.py
```diff
@@ -0,0 +1,99 @@
...
+        url = self.get_url("configuration")
+        return self.session.get(url).json()
+
+    def properties(self) -> Dict[str, Any]:
+        """Return backend properties.
+
+        Returns:
+            JSON response of backend properties.
+        """
+        # pylint: disable=redefined-outer-name
```
daka1510, line 59: Are you sure this is needed? `make lint` does not fail for me without this.

qiskit_ibm_runtime/api/rest/cloud_backend.py
rathishcholarajan: It's available only on the test env for now https://us-east.quantum-computing.test.cloud.ibm.com/openapi, will be pushed to prod with next release.

qiskit_ibm_runtime/api/rest/runtime.py
rathishcholarajan: 3 reasons:
1. `rest/backend.py` connects to API provided by the old API service (not the runtime service) vs `/rest/cloud_backend.py` connects to the `/devices` endpoints in the runtime service deployed on IBM Cloud. So they both connect to entirely different APIs offered by two different API services. Whereas the runtime related adapters connect to the same runtime API service, just hosted on two different platforms IQX vs IBM Cloud.
2. Since this is not a public facing class I thought it'd be clearer for the team to have separate files to connect to two distinct API services, the JSON responses are also slightly different.
3. We also have a new `/devices/{id}/configuration` endpoint in the runtime service on IBM Cloud which is not available in the old API service.

I am happy to put it all in one class if you all think it'd be easy to maintain and easily understandable for the team. @jyu00 @kt474 Thoughts?

qiskit_ibm_runtime/api/rest/cloud_backend.py
rathishcholarajan: For the reasons mentioned in this comment https://github.com/Qiskit/qiskit-ibm-runtime/pull/46#discussion_r763095193, I have this function copied in a separate file. If we decide to put it all in one file then it'll address above comment. Else we can discuss how to reuse. But since they connect to two different API services it would be nice to keep these separate.

qiskit_ibm_runtime/api/rest/cloud_backend.py
rathishcholarajan: Good catch! Fixed here https://github.com/Qiskit/qiskit-ibm-runtime/pull/46/commits/f1b20231ec23eb22262b4b9d3c78ace4cd30ab82

qiskit_ibm_runtime/api/rest/runtime.py
daka1510: Thanks for the details. That convinced me for my part that using separate backend adapters (`/rest/backend.py`, `/rest/cloud_backend.py`) is a good choice. I'm also fine with sticking to a single file for runtime related adapters.

qiskit_ibm_runtime/api/rest/cloud_backend.py
daka1510: Convinced it makes sense to keep separate files. Maybe you find a simple way to delegate calls from `cloud_backend.py#status` to `backend.py#status` to avoid duplicating many lines as long as the APIs are very similar.

rathishcholarajan: I don't want to combine the two paths, but yes I promise to get back to this once we get everything working.

PR #47, opened 2021-12-06
Fix broken if conditions in setup_test_logging

daka1510:
### Summary
Noticed that every test run creates many empty files in my filesystem even though I do not set `FILE_LOG=True`.

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -39,13 +39,13 @@ def setup_test_logging(logger: logging.Logger, filename: str):
     )
     formatter = logging.Formatter(log_fmt)
 
-    if os.getenv("STREAM_LOG", "true"):
+    if os.getenv("STREAM_LOG", "true").lower() == "true":
         # Set up the stream handler.
         stream_handler = logging.StreamHandler()
         stream_handler.setFormatter(formatter)
         logger.addHandler(stream_handler)
 
-    if os.getenv("FILE_LOG", "false"):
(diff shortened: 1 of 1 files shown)


test/utils.py
```diff
@@ -39,13 +39,13 @@ def setup_test_logging(logger: logging.Logger, filename: str):
     )
     formatter = logging.Formatter(log_fmt)
 
-    if os.getenv("STREAM_LOG", "true"):
+    if os.getenv("STREAM_LOG", "true").lower() == "true":
```
daka1510, line 42: **Problem:**
When `STREAM_LOG=False`, it previously evaluated to `True` because `bool("False") == True`

test/utils.py
```diff
@@ -39,13 +39,13 @@ def setup_test_logging(logger: logging.Logger, filename: str):
...
         # Set up the stream handler.
         stream_handler = logging.StreamHandler()
         stream_handler.setFormatter(formatter)
         logger.addHandler(stream_handler)
 
-    if os.getenv("FILE_LOG", "false"):
+    if os.getenv("FILE_LOG", "false").lower() == "true":
```
daka1510, line 48: **Problem:**
When `FILE_LOG=False` or not set, it previously evaluated to `True` because `bool("False") == True`

rathishcholarajan (approved):

PR #50, opened 2021-12-07
Move Program and ProgramJob REST adapters to separate files

rathishcholarajan:
### Summary
Since this is a runtime only repo we can move Program and ProgramJob REST adapters to separate files instead of having them crammed in one file.

### Details and comments
Fixes #


daka1510 (changes requested):
Looks good in general but lint checks failed. You can fix it via `make black`

```
Oh no! 💥 💔 💥
1 file would be reformatted, 98 files would be left unchanged.
make: *** [Makefile:24: style] Error 1
```

Not sure if the test failures are also related.

PR #51, opened 2021-12-08
Add backend requirements to program __str__

kt474:
### Details and comments
Fixes https://github.com/Qiskit-Partners/qiskit-ibm/issues/198

Was only able to find `input_allowed` and `min_number_qubits` as properties of `backend_requirements` - are there others?


qiskit_ibm_runtime/runtime_program.py
```diff
@@ -101,6 +101,15 @@ def _format_common(schema: Dict) -> None:
...
                 formatted.append(" " * 4 + "Type: {}".format(str(schema["type"])))
+            if "input_allowed" in schema:
+                formatted.append(
+                    " " * 4 + "Input Allowed: {}".format(str(schema["input_allowed"]))
+                )
+            if "min_num_qubits" in schema:
+                formatted.append(
+                    " " * 4
+                    + "Minimum number qubits: {}".format(str(schema["min_num_qubits"]))
+                )
```
rathishcholarajan, line 112: description, type and properties don't exist in backend_requirements like the other three properties (interim_results, return_values and parameters). Also input_allowed and min_num_qubits don't exist in interim_results, return_values and parameters. So we shouldn't add this to the method named _format_common since it's not common. You can create another private method to format backend_requirements.

rathishcholarajan (changes requested):
1. Could you paste the output of printing a sample program by running this code?
2. Could we add some tests?

kt474:
`service.pprint_programs(detailed=True)`

Example outputs
```
qiskit-test:
  Name: qiskit-test
  Description: Qiskit test program
  Creation date: 2021-10-10T18:48:27Z
  Update date: 2021-10-10T18:48:27Z
  Max execution time: 600
  Backend requirements:
    none
  Input parameters:
    none
  Interim results:
    none
  Returns:
    none
```
```
qasm3-runner:
  Name: qasm3-runner
  Description: A runtime program that takes one or more circuits, compiles them, converts them to OpenQASM3, and executes them. The input can also be one or more OpenQASM3 strings. Note that only one circuit/string can be passed if using a simulator, and this program can only run on a backend that supports OpenQASM3. This program is currently in beta mode and will likely be updated frequently.
  Creation date: 2021-11-01T17:57:48.28061Z
  Update date: 2021-12-10T23:33:14.828172Z
  Max execution time: 14400
  Backend requirements:
    Supported features: qasm3
  Input parameters:
    Properties:
    ... 
```

qiskit_ibm_runtime/runtime_program.py
```diff
@@ -115,6 +115,23 @@ def _format_common(schema: Dict) -> None:
...
+            if "supported_features" in schema:
+                formatted.append(
+                    " " * 4
+                    + "Supported features: {}".format(str(schema["supported_features"]))
+                )
```
rathishcholarajan, line 133: For any key that we have not handled here, we should call the sentence_case function. You might have to loop through the schema dict in this function.

rathishcholarajan (changes requested):

qiskit_ibm_runtime/runtime_program.py
```diff
@@ -115,6 +115,28 @@ def _format_common(schema: Dict) -> None:
...
+        def _format_backend_requirements(schema: Dict) -> None:
+            """Add backend requirements details to `formatted`."""
+            if "input_allowed" in schema:
+                formatted.append(
+                    " " * 4 + "Input allowed: {}".format(str(schema["input_allowed"]))
+                )
+            if "min_num_qubits" in schema:
+                formatted.append(
+                    " " * 4
+                    + "Minimum number qubits: {}".format(str(schema["min_num_qubits"]))
```
rathishcholarajan, line 127:
```suggestion
                    + "Minimum number of qubits: {}".format(str(schema["min_num_qubits"]))
```

qiskit_ibm_runtime/runtime_program.py
```diff
@@ -115,6 +115,28 @@ def _format_common(schema: Dict) -> None:
...
+                formatted.append(
+                    " " * 4
+                    + "Supported features: {}".format(str(schema["supported_features"]))
+                )
+            for key, value in schema.items():
+                if key not in ["input_allowed", "min_num_qubits", "supported_features"]:
+                    formatted.append(
+                        " " * 4 + "{}: {}".format(sentence_case(key), str(value))
+                    )
+
```
rathishcholarajan, line 139: Input allowed and supported features will get handled by this right? Do we want to special case them?

rathishcholarajan (changes requested):

qiskit_ibm_runtime/runtime_program.py
kt474: Not exactly, `sentence_case` converts camelCase to sentence case so the `_` would still remain.
`input_allowed` -> `Input_allowed`
`supported_features` -> `Supported_features`

I think it's better without the `_` but the properties section includes them so maybe we should just handle it here?

rathishcholarajan: Aah right we can’t use the sentence_case function. We need a new function that converts snake_case to Sentence case.

qiskit_ibm_runtime/runtime_program.py
```diff
@@ -107,15 +107,40 @@ def _format_common(schema: Dict) -> None:
...
+            if "min_num_qubits" in schema:
+                formatted.append(
+                    " " * 4
+                    + "Minimum number of qubits: {}".format(
+                        str(schema["min_num_qubits"])
+                    )
+                )
+            for key, value in schema.items():
+                if key not in ["min_num_qubits"]:
+                    print(key)
```
rathishcholarajan, line 130:
```suggestion
```

rathishcholarajan (approved):
LGTM!

PR #55, opened 2021-12-10
Add backend related functions in RuntimeClient

rathishcholarajan:
### Summary
This PR adds the following backend related functions to `RuntimeClient`.
- list_backends
- backend_configuration
- backend_status
- backend_properties
- backend_pulse_defaults

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/api/clients/runtime.py
+++ b/qiskit_ibm_runtime/api/clients/runtime.py
@@ -13,7 +13,7 @@
 """Client for accessing IBM Quantum runtime service."""
 
 import logging
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from qiskit_ibm_runtime.credentials import Credentials
 from qiskit_ibm_runtime.api.session import RetrySession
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/api/clients/runtime.py
```diff
@@ -278,3 +278,60 @@ def job_logs(self, job_id: str) -> str:
     def logout(self) -> None:
         """Clear authorization cache."""
         self.api.logout()
+
+    # IBM Cloud only functions
+
+    def list_backends(self, timeout: Optional[float] = None) -> List[Dict[str, Any]]:
```
daka1510, line 284: This parameter appears to make sense for most calls. Other calls seem to use the [default](https://github.com/Qiskit/qiskit-ibm-runtime/blob/f847c86841f8d1f69423e83d9d429508daece672/qiskit_ibm_runtime/api/session.py#L148) value set during session initialization. Why is it explicitly needed in this case?

daka1510 (approved):
Looks good :).

qiskit_ibm_runtime/api/clients/runtime.py
rathishcholarajan: This API used to take long since all the backends and their configuration were fetched in one call. So the user wanted the ability to pass an increased timeout in case backends() call fails. Now that the configuration is fetch in separate calls, we should no longer need this timeout. Have removed it.

daka1510 (approved):

jyu00 (approved):
I think it'd be better if we had just a single `client` for both cloud and legacy, but use different rest adaptor under the cover. But this is good enough for now.

PR #56, opened 2021-12-10
Add support for viewing backends available on IBM Cloud

rathishcholarajan:
### Summary
This PR builds upon https://github.com/Qiskit/qiskit-ibm-runtime/pull/55 (will rebase this once #55 is merged) to make the following functionality available to IBM Cloud users. (Currently works only with staging env)

- backends()
- get_backend()
- backend.status()
- backend.properties()
- backend.defaults()
- backend.configuration()

Example:
```python
from qiskit_ibm_runtime import IBMRuntimeService
service = IBMRuntimeService(auth="cloud", token="<staging_api_key>", locator="<staging_locator>")

# List backends
service.backends()

# Get a backend by name
simulator = service.get_backend("ibmq_qasm_simulator")

# Get a backend's status
simulator.status()

# Get a backend's properties
simulator.properties()

# Get a backend's configuration
simulator.configuration()

# Get a backend's pulse defaults
simulator.defaults()
```

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/ibm_backend.py
+++ b/qiskit_ibm_runtime/ibm_backend.py
@@ -69,7 +69,8 @@ class IBMBackend(Backend):
         configuration: Union[QasmBackendConfiguration, PulseBackendConfiguration],
         service: "ibm_runtime_service.IBMRuntimeService",
         credentials: Credentials,
-        api_client: Union[AccountClient, RuntimeClient],
+        account_client: Optional[AccountClient] = None,
+        runtime_client: Optional[RuntimeClient] = None,
     ) -> None:
         """IBMBackend constructor.
 
@@ -81,7 +82,8 @@ class IBMBackend(Backend):
         """
         super().__init__(provider=service, configuration=configuration)
 
-        self._api_client = api_client
+        self._account_client = account_client
+        self._runtime_client = runtime_client
         self._credentials = credentials
         self.hub = credentials.hub
         self.group = credentials.group
@@ -150,9 +152,12 @@ class IBMBackend(Backend):
             datetime = local_to_utc(datetime)
 
         if datetime or refresh or self._properties is None:
-            api_properties = self._api_client.backend_properties(
-                self.name(), datetime=datetime  # type: ignore
-            )
+            if self._account_client:
+                api_properties = self._account_client.backend_properties(
+                    self.name(), datetime=datetime
+                )
+            elif self._runtime_client:
+                api_properties = self._runtime_client.backend_properties(self.name())
             if not api_properties:
                 return None
             decode_backend_properties(api_properties)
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -233,7 +233,7 @@ def reservations(
         """
         start_datetime = local_to_utc(start_datetime) if start_datetime else None
         end_datetime = local_to_utc(end_datetime) if end_datetime else None
-        raw_response = self._api_client.backend_reservations(
+        raw_response = self._api_client.backend_reservations(  # type: ignore
```
daka1510, line 236: I don't understand why the constructor accepts two different clients (`Union[AccountClient, RuntimeClient]`) and this line invokes a function that only one of the clients implements. There might be implicit underlying assumptions but can those be made clearer?  E.g. by passing in `account_client` / `runtime_client` instead of `api_client` in the constructor?

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -249,6 +248,58 @@ def _resolve_credentials(
...
+                continue
+            try:
+                decode_backend_configuration(raw_config)
+                try:
+                    config = PulseBackendConfiguration.from_dict(raw_config)
+                except (KeyError, TypeError):
+                    config = QasmBackendConfiguration.from_dict(raw_config)
+                backend_cls = (
+                    ibm_backend.IBMSimulator
+                    if config.simulator
```
daka1510, line 284: The `GateConfig` object does not have a `simulator` attribute. Is that just a typing issue or a bug?
image

rathishcholarajan: I don't see this issue, functionality seems to work fine.

qiskit_ibm_runtime/ibm_backend.py
rathishcholarajan: Thanks, I've split the parameters as suggested to make it more clear.

qiskit_ibm_runtime/ibm_runtime_service.py
daka1510: Do you happen to know in which cases `config.simulator` is truthy? I don't see any assignments of `self.config = ` in `PulseBackendConfiguration` or `QasmBackendConfiguration` (or their parent classes).

daka1510 (approved):

qiskit_ibm_runtime/ibm_runtime_service.py
rathishcholarajan: It’s truthy for simulators like ibmq_qasm_simulator and false for real backends. Not sure if that’s what you are asking?

daka1510: I was trying to understand where in the code `config.simulator` gets assigned (e.g. via `self.simulator`) in the backend classes or their parents.

Issue #2, opened 2021-11-08
Runtime 0.1 - Authentication


rathishcholarajan:
A user should be able to login using either IBM Cloud API key or IQX API token, save account, view saved account, view active account and also delete saved account for runtime 0.1 release.

```python
class IBMRuntimeService(RuntimeService):
    def __init__(self, auth=None, token=None, locator=None):
        """Initialize a new client to use the runtime service.
        
        Args:
            auth: Authentication type. `cloud` or `legacy`. If not specified, the saved default is used.
                If there is no default value, and both accounts were saved on disk, the cloud type is used.
            token: Token used for authentication. If not specified, the saved token is used.
            locator: The authentication url, if `auth=legacy`. Otherwise the CRN.
        """
        pass

    @staticmethod
    def delete_account(auth=None) -> None:
       """Delete account information on disk.
       Args:
           auth: Authentication type. `cloud` or `legacy`. If auth is not passed and if there is one account saved, that is 
               deleted. If auth is not passed and if there are two accounts saved then `cloud` is deleted first.
       """
       pass

    @staticmethod
    def save_account(token, locator, as_default) -> None:
        """Save account information on disk."""
        pass

   @staticmethod
    def saved_account() -> None:
        """Get saved account information on disk."""
        pass

    def active_account() -> None:
        """Get account currently in use for the session."""
        pass

    def logout(self) -> None:
        """Clears authorization cache on the server.

        For better performance, the runtime server caches each user's
        authorization information. This method is used to force the server
        to clear its cache.

        Note:
            Invoke this method ONLY when your access level to the runtime
            service has changed - for example, the first time your account is
            given the authority to upload a program.
        """
```

daka1510:
With https://github.com/qiskit/qiskit-ibm-runtime/issues/44 the last related issue has been closed. Therefore also closing the Epic.

Issue #44, opened 2021-12-02
Allow user to specify human readable service instance name during authentication


rathishcholarajan:
**What is the expected feature or enhancement?**
As per the current approved design, we can pass IBM Cloud API Key (token) and CRN (locator) parameters to authenticate.
Since the CRN is long and not easy to remember, we could also provide another option for the user to specify the human readable service instance name in the `locator` parameter.

For example:
```
service = IBMRuntimeService(token=<token>, locator="Quantum Services-og")
```

**Suggested Solution:**
1. Get IAM Access Token by passing API Key
```
curl -X POST 'https://iam.test.cloud.ibm.com/identity/token' -H 'Content-Type: application/x-www-form-urlencoded' -d 'grant_type=urn:ibm:params:oauth:grant-type:apikey&apikey=<api_key>'
```
2. Use IAM Access token from step 1 and service instance name (Ex: Quantum Services-og) to get the resource.
```
curl -X GET "https://resource-controller.test.cloud.ibm.com/v2/resource_instances?name=Quantum%20Services-og" -H "Authorization: Bearer <access_token>"
```
Then read the CRN from the resource data returned which looks like below, to make further calls to other runtime APIs.
```
{
  "rows_count": 1,
  "next_url": null,
  "resources": [
    {
      "id": "crn:v1:staging:public:quantum-computing:us-east:a/78d9efd23fdb4894837f663626efb044:7d4e1dd0-b3e1-415b-b601-5fc8b38dd8b2::",
      "guid": "7d4e1dd0-b3e1-415b-b601-5fc8b38dd8b2",
      "url": "/v2/resource_instances/7d4e1dd0-b3e1-415b-b601-5fc8b38dd8b2",
      "created_at": "2021-10-25T20:58:21.141685368Z",
      "updated_at": "2021-10-25T20:58:22.668667668Z",
      "deleted_at": null,
      "created_by": "IBMid-55000AJFHA",
      "updated_by": "",
      "deleted_by": "",
      "scheduled_reclaim_at": null,
      "restored_at": null,
      "scheduled_reclaim_by": "",
      "restored_by": "",
      "name": "Quantum Services-og",
      "region_id": "us-east",
      "account_id": "78d9efd23fdb4894837f663626efb044",
      "reseller_channel_id": "",
      "resource_plan_id": "88f4f591-8b8f-4fc7-94ba-7328d5027508",
[...]

PR #58, opened 2021-12-13
Acceptance Criteria for Authentication Epic

rathishcholarajan:
### Summary
I've added acceptance criteria for the Runtime 0.1 - Authentication https://github.com/Qiskit/qiskit-ibm-runtime/issues/2 epic. Will update this file for other epics in separate PRs.

### Details and comments
Fixes #

### DIFF
--- a/docs/test/Runtime 0.1.ipynb
+++ b/docs/test/Runtime 0.1.ipynb
@@ -1,0 +1,562 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "7f409f1f",
+   "metadata": {},
+   "source": [
+    "# Acceptance Criteria for Runtime 0.1 release"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3d039414",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qiskit_ibm_runtime import IBMRuntimeService"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a7300dad",
+   "metadata": {},
+   "source": [
+    "# Authentication\n",
+    "\n",
+    "## Legacy\n",
+    "Using IBM Quantum API token.\n",
+    "\n",
+    "### Save Legacy Account"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "0326e78b-b3df-40ad-9adb-db7a03a29f9c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Save default legacy account\n",
+    "IBMRuntimeService.save_account(\n",
+    "    auth=\"legacy\", \n",
+    "    token=\"xyz\", \n",
+    "    url=\"https://auth.quantum-computing.ibm.com/api\", \n",
+    "    instance=\"qiskit-hub/qiskit-group/qiskit-project\",\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ca5c496d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Save legacy account with a name\n",
+    "IBMRuntimeService.save_account(\n",
+    "    name=\"my-legacy-account\"\n",
+    "    auth=\"legacy\", \n",
+    "    token=\"abc\", \n",
+    "    url=\"https://auth.quantum-computing.ibm.com/api\", \n",
+    "    instance=\"qiskit-hub-1/qiskit-group-1/qiskit-project-1\",\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5232022a",
+   "metadata": {},
+   "source": [
+    "### Load Legacy Account from Saved Account"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "2f45dccc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Load default legacy account\n",
+    "service = IBMRuntimeService(auth=\"legacy\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bfcc2cec",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Load legacy account with name\n",
+    "service = IBMRuntimeService(name=\"my-legacy-account\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "76d4c05a",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


daka1510 (approved):
Thanks for writing this up. Matches my understanding. Looks good.

rathishcholarajan:
> Questions about `save_account`
> 
> 1. The examples only show `save_account` with `instance` explicitly specified. Presumably that parameter is optional?
Yes it’s optional. Will add another example.
> 2. For legacy runtime, `save_account(instance="hgp1")` means `hgp1` will be the default h/g/p to use, right? But the _actual_ h/g/p used is inferred from the backend selected (e.g. if I want to run on `backend1` which is only accessible under `hgp2`, not `hgp1`, `hgp2` will be used?
Correct. I’ll make a note about it.
> 3. Similar to above, except for cloud runtime, whatever instance I've saved will always be used?
Correct.
> 4. If I do `save_account()` for a `name` that already exists but with different values (e.g. different `instance`), it'd essentially update that entry?
Correct. Will mention that with an example.
> 5. `IBMRuntimeService()` will use legacy, not cloud, account if I only have legacy account saved, right?
Correct. Will add an example.

docs/test/Runtime 0.1.ipynb
```diff
@@ -0,0 +1,562 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "f76d6841",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Enable legacy account with token and url parameters\n",
+    "service = IBMRuntimeService(token=\"xyz\", url=\"https://auth.quantum-computing.ibm.com/api\")"
```
daka1510, line 124: Based on the latest [implementation](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/ibm_runtime_service.py#L228-L249) from @jyu00 this needs to be changed. 

```python
# ...
# Let's not infer based on these attributes as they may change in the future.
raise ValueError(
  "'auth' is required if 'token', or 'url' is specified but 'name' is not."
)
# ...
```

docs/test/Runtime 0.1.ipynb
```diff
@@ -0,0 +1,562 @@
...
+   "execution_count": null,
+   "id": "4e63a883",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Save default cloud account\n",
+    "IBMRuntimeService.save_account(\n",
+    "    auth=\"cloud\", \n",
+    "    token=\"xyz\", \n",
+    "    url=\"https://cloud.ibm.com\",\n",
```
daka1510, line 203: This proposal assumes `https://cloud.ibm.com` but the current [implementation](https://github.com/Qiskit/qiskit-ibm-runtime/blame/0b6deb742d41153252d9114a0866c0c3e0cf8196/qiskit_ibm_runtime/accounts/account.py#L27) expects `https://us-east.quantum-computing.cloud.ibm.com`.

I would vote for sticking to the proposed URL in the design document and derive all other required API URLs from it (e.g. `https://resource-controller.cloud.ibm.com`, `https://us-east.quantum-computing.cloud.ibm.com`, ...).

docs/test/Runtime 0.1.ipynb
```diff
@@ -0,0 +1,562 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "688a2f9b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Enable cloud account with token, url and instance parameters\n",
+    "service = IBMRuntimeService(token=\"xyz\", url=\"https://cloud.ibm.com\", instance=\"<crn:v1:bluemix:public:quantum-computing:us-east:a/...> or <Quantum Services-bg>\") # full CRN or service instance name"
```
daka1510, line 316: Same as https://github.com/Qiskit/qiskit-ibm-runtime/pull/58/files#r777471342 

Current implementation yields `'auth' is required if 'token', or 'url' is specified but 'name' is not.`

docs/test/Runtime 0.1.ipynb
jyu00: Yeah I'd prefer _not_ infer the auth type based on url, especially since we may have additional cloud provider in the future.

docs/test/Runtime 0.1.ipynb
jyu00: How do you plan on deriving that from `https://cloud.ibm.com`? For example, in the future my account might have both `us-east` and `us-south`, but I prefer using `us-east` since that's closer to me. How would the code know about my preference?

daka1510: It appears to me that the location specific API endpoint can be derived from the Cloud Resource Name (CRN), e.g.
```
crn:v1:bluemix:public:quantum-computing:us-east:a/41d9bde3ef128894870dfbeaca4a5260:0ddc92a3-00be-4ec8-987e-cafda1b086fb:: 
-> 
https://us-east.quantum-computing.cloud.ibm.com
```

With https://github.com/Qiskit/qiskit-ibm-runtime/issues/44 we will leverage the global [resource controller API](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#list-resource-instances) that also returns this information if users pass in a service name instead of the CRN via the `instance` parameter.

jyu00: I personally don't like inferring the url based on the CRN either, since we have 0 control over the relationship between the two (not that it's likely to change), unless there is an official API that'd do the translation for us.

daka1510: There's no official API as far as I can tell but the CRN format in general is fixed and [publicly documented](https://cloud.ibm.com/docs/account?topic=account-crn#format-crn). 

My feeling is it's more convenient for users if they do not need to pass in information manually that we can infer internally.

But let's assume for a bit that we will stick with `https://us-east.quantum-computing.cloud.ibm.com`.
- How do you suggest should we infer the resource controller API URL? Would this be an additional parameter users need to pass?

jyu00: ok if the CRN format is fixed, I suppose it's not so bad to infer the url from it. 

> How do you suggest should we infer the resource controller API URL? Would this be an additional parameter users need to pass?

I'd prefer if we don't add yet another parameter, given how many we already have. I suppose we'll need to hard code this one 😞 

All this inference / hard coding should only apply if the url is `https://cloud.ibm.com`.

docs/test/Runtime 0.1.ipynb
rathishcholarajan: Fixed!

daka1510 (approved):
Looks good from my side. We can run another sanity check once this PR and https://github.com/Qiskit/qiskit-ibm-runtime/pull/66 are integrated.

PR #49, opened 2021-12-07
Add account management functionality

daka1510:
### Summary
This PR groups changes related to a streamlined account management approach that works across legacy and Cloud accounts and stores configuration data in a JSON file.

### Details
This initial PR enables below minimal code paths as per https://github.com/Qiskit/qiskit-ibm-runtime/pull/58. 

```python
from qiskit_ibm_runtime import IBMRuntimeService

# Legacy
IBMRuntimeService.save_account(
    auth="legacy",
    name="my-legacy-account",
    token="...",
)

service = IBMRuntimeService(name="my-legacy-account")
service.pprint_programs()
service.active_account()

IBMRuntimeService.saved_accounts()
IBMRuntimeService.delete_account(name="my-legacy-account")

# Cloud
IBMRuntimeService.save_account(
    auth="cloud",
    name="my-cloud-account",
    token="...",
    instance="crn:v1:bluemix:public:quantum-computing:us-east:a/...::",
)

service = IBMRuntimeService(name="my-cloud-account")
service.pprint_programs()
service.active_account()

IBMRuntimeService.saved_accounts()
IBMRuntimeService.delete_account(name="my-cloud-account")
```

### Next Steps
Given that the library is not released yet, I would favor to integrate these changes to enable for more parallelism and avoid additional merge conflicts. In a next step, I will
- add proper unit test coverage for all account management related code paths
- refactor the environment variable names and enable integration tests against a cloud environment 

### Details and comments
Relates to
- https://github.com/Qiskit/qiskit-ibm-runtime/issues/48
- https://github.com/Qiskit/qiskit-ibm-runtime/issues/31

PR #62, opened 2021-12-15
Reorganize tests and make backend name optional for cloud

jyu00:
### Summary

This PR includes the following (apologize for the size, but I wanted to get this done quickly):

Mainline
1. Updated `IBMRuntimeService.run()` to make `backend_name` optional for auth=cloud (not fully implemented since it's not quite supported by the server yet).
2. Changed h/g/p parameters to `instance` to match the constructor
3. Updated `IBMRuntimeService._hgps` to use `h/g/p` as the key since that's what gets passed in now
3. Remove passing `service` to `ibm_backend` since it's not used and is not a provider
4. Replace `Credentials` with `ClientParameters` when passing to api clients. 
6. Remove all reservation related methods. Reservation is only available on legacy, is not really part of runtime, and I'm not too convinced people actually use those methods. We can always add them back later if requested.
7. Remove `run_circuits` as it's going to be replaced by `sampler`.
9. Move `least_busy` into `IBMRuntimeService` and make it accept same parameters as `backends()`
10. Support credentials via env variables
11. Support loading account from disk. If only one is saved, that's used, otherwise default to cloud.
11. Remove backend names as attributes to `IBMRuntimeService` since they are not runtime service instances.
11. Various bug fixes.

Tests
1. Moved all the fake clients to `test.mock`
7. Created `fake_account_client`
8. Created `fake_runtime_service` to replace the old context manager to make it more customizable
9. Broke `test_runtime.py` into `test_data_serialization.py`, `test_jobs.py`, `test_programs.py`, and `test_retrieve_jobs.py`
10. Moved some utility functions to a common modules so they can be shared by all tests and moved them to `test.utils`. 
1. Stop inheriting Qiskit tese case 
2. Add decorator to run test cases using both cloud and legacy services
5. Move all tests out of `test/ibm` now that we are not in the qiskit namespace.
6. Cleanup all tests so they should all pass now (although cloud runtime is a bit flaky right now)
10. Update CI to run new tests

### Details and comments
Fixes #


qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -162,30 +162,27 @@ def __init__(
         account_credentials, account_preferences = self._resolve_credentials(
             token=token, locator=locator, **kwargs
         )
+        self._auth = auth
+
+        # Common attributes
+        self._backends: Dict[str, "ibm_backend.IBMBackend"] = {}
+        self._programs: Dict[str, RuntimeProgram] = {}
```
daka1510, line 169: This will also conflict with https://github.com/Qiskit/qiskit-ibm-runtime/pull/49. Can you take a look and consider merging #49 first?

test/ibm/runtime/test_runtime.py
```diff
@@ -1,907 +0,0 @@
-# This code is part of Qiskit.
-#
-# (C) Copyright IBM 2021.
```
daka1510, line 3: Thanks, this refactoring looks great. 

I personally also find another file name (`test/ibm/runtime/test_ibm_provider`) confusing. Why is it called provider while it includes tests for the `IBMRuntimeService` class? Maybe we can get rid of some legacy here as well?!

jyu00: oh definitely, I just haven't got to there yet. Although it'll probably be another PR as this one is already larger than I'd like.

PR #75, opened 2022-01-06
Don't use ibmq_berlin

jyu00:
### Summary

`ibmq_berlin` has been retired but continue to be reported by the server as a valid backend. This PR skips `ibmq_berlin`, so tests won't keep tripping over it until that's fixed. 

### Details and comments

Since I'm trying to get all tests to pass, this PR also
- closes https://github.com/Qiskit-Partners/qiskit-ibm/issues/226 ,  by deleting the malformed url from the test case. The `http:{}:{}` format was no longer supported in `requests` 2.26+, and I don't see why we need to support it especially since this is a new repo (sorry @daka1510, I know you were dying to fix this one lol)
-  fixes `test_save_get`


rathishcholarajan (approved):
LGTM!

PR #76, opened 2022-01-06
Fix staging cron job env var

jyu00:
### Summary

Fixes the environment variable used to set the staging cron job. 

### Details and comments
Fixes #


rathishcholarajan (approved):
LGTM!

PR #77, opened 2022-01-07
Query programs by name

kt474:
### Summary
Ported from Qiskit-Partners/qiskit-ibm#194

### Details and comments
Fixes Qiskit-Partners/qiskit-ibm#180


rathishcholarajan:
@kt474 Could you please set appropriate env variables in your fork for legacy and cloud so we can verify if push tests pass.

QISKIT_IBM_API_TOKEN
QISKIT_IBM_API_URL
QISKIT_IBM_CLOUD_CRN
QISKIT_IBM_CLOUD_TOKEN
QISKIT_IBM_CLOUD_URL
QISKIT_IBM_HGP
QISKIT_IBM_STAGING_API_TOKEN
QISKIT_IBM_STAGING_API_URL
QISKIT_IBM_STAGING_CLOUD_CRN
QISKIT_IBM_STAGING_CLOUD_TOKEN
QISKIT_IBM_STAGING_CLOUD_URL
QISKIT_IBM_STAGING_HGP

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -658,7 +660,11 @@ def pprint_programs(
                 print(f"  Description: {prog.description}")
 
     def programs(
-        self, refresh: bool = False, limit: int = 20, skip: int = 0
+        self,
+        refresh: bool = False,
+        name: Optional[str] = "",
+        limit: int = 20,
+        skip: int = 0,
```
daka1510, line 667: This way, this change is not backward-compatible. Users may rely on the order of the positional arguments (limit, skip). An easy way to fix that is to add the optional name parameter at the end.

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -667,35 +673,30 @@ def programs(
...
             limit = len(self._programs)
+        if name:
+            matched_programs = []
+            if refresh and not already_retrieved:
+                matched_programs = list(self._retrieve_programs(name).values())
+            else:
+                for program in list(self._programs.values()):
+                    if program.name == name:
+                        matched_programs.append(program)
+            return matched_programs[skip : limit + skip]
```
daka1510, line 699: I personally have problems to understand this deeply nested code because filtering by name either happens on the client-, or on the server-side. In the most likely path where a user runs `service.programs(name="sampler")` **all** programs are loaded, cached and the filtering happens on the client-side. 

I suggest to make the following simplifications
- if no `name` is provided continue to load and cache all programs as before (use previous code as a baseline)
- if a `name`  parameter is provided do not use caching and simply return the results from the API
  - this is based on the simplifying assumption that the API can quickly return programs that match a given name (no caching needed) but returning all programs is expensive (that is why caching was previously introduced)

qiskit_ibm_runtime/ibm_runtime_service.py
```diff
@@ -729,6 +730,35 @@ def program(self, program_id: str, refresh: bool = False) -> RuntimeProgram:
...
+            # there was no limit or skip
+            count = response.get("count", 0)
+            for prog_dict in program_page:
+                program = self._to_program(prog_dict)
+                programs[program.program_id] = program
+            if len(programs) == count:
+                # Stop if there are no more programs returned by the server.
+                break
+            offset += len(program_page)
+        return programs
```
daka1510, line 760: Given that this PR highly overlaps with https://github.com/Qiskit/qiskit-ibm-runtime/pull/78 why not combine them into this PR and avoid merge conflict upfront? 

Positive side-effect: makes it easier to review your changes in one effort and allows to see if it's consistently solved.

daka1510 (changes requested):

qiskit_ibm_runtime/ibm_runtime_service.py
kt474: I think this is also a good approach - originally we wanted to minimize the number of API calls, and just use the cache for every call after the first one that didn't specifically set `refresh`. That's why the logic is currently a bit complicated. My understanding is that other than the first call to `programs`, we want to retrieve from the cache as much as possible - @rathishcholarajan thoughts?

qiskit_ibm_runtime/ibm_runtime_service.py
daka1510: Cool, thanks!

PR #79, opened 2022-01-07
Fix test_retrieve_jobs_limit

jyu00:
### Summary

`test_retrieve_jobs_limit` [occasionally fails](https://github.com/Qiskit/qiskit-ibm-runtime/runs/4733818448?check_suite_focus=true#step:5:2306). I think it's because the test case was getting jobs submitted by other tests and therefore the retrieved jobs were not among those submitted by it. 

This PR adds `program_id` as a filter to ensure it only gets its jobs (`program_id` is unique for each test class).

### Details and comments
Fixes #


rathishcholarajan (approved):
LGTM!

Issue #48, opened 2021-12-06
Replace **kwargs with proxies and verify in IBMRuntimeService constructor


jyu00:
**What is the expected feature or enhancement?**

`IBMRuntimeService` constructor has a `**kwargs` which is documented as 

```
            **kwargs: Additional settings for the connection:
                * proxies (dict): proxy configuration.
                * verify (bool): verify the server's TLS certificate.
```

If `proxies` and `verify` are the only ones allowed, why not spell them out?

rathishcholarajan:
document proxies

PR #80, opened 2022-01-07
Improve test coverage and documentation for proxy configuration

daka1510:
### Summary
- Added type hints for the proxies dictionary passed to various APIs
- Improved account validation and added related test cases to fail fast when users try to save or load invalid accounts  

### Details and comments
Fixes https://github.com/Qiskit/qiskit-ibm-runtime/issues/48


qiskit_ibm_runtime/accounts/account.py
```diff
@@ -13,16 +13,32 @@
...
+    All items in the dictionary are optional. When ``urls`` are provided, they must contain a dictionary
+    mapping protocol or protocol and host to the URL of the proxy. Refer to
+    https://docs.python-requests.org/en/latest/api/#requests.Session.proxies for details and examples.
+
+    NTLM user authentication can be enabled by setting ``username_ntlm`` and ``password_ntlm``.
+    """
+
+    urls: Dict[str, str]
+    username_ntlm: str
+    password_ntlm: str
```
daka1510, line 39: The proxy configuration dictionary is passed to multiple public and private APIs. I did not want to duplicate more detailed documentation in the docstrings across all places.

I therefore decided to add a type hint for the existing dictionary (see [PEP-0589](https://www.python.org/dev/peps/pep-0589/)) that will also add type safety internally.

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -48,17 +64,53 @@ def _assert_valid_url(url: str) -> None:
...
             )
-    # TODO: add validation for legacy instance when adding test coverage
+    if auth == "legacy":
+        if instance is not None:
+            try:
+                from_instance_format(instance)
+            except:
+                raise ValueError(
+                    f"Invalid `instance` value. Expected hub/group/project format, got {instance}"
+                )
```
daka1510, line 113: Required updates in other test cases that passed in invalid values indirectly, see https://github.com/Qiskit/qiskit-ibm-runtime/pull/80/commits/90942c78d8a7164e96035dba03e58a8d9eb48b24

qiskit_ibm_runtime/accounts/account.py
jyu00: Making it a `TypedDict` is an interesting idea. However, you'd still still need to document it in the user facing function (`IBMRuntimeService` constructor) _or_ make the hyperlink work so a user can click on the link and see the docstring for this class.

An alternative idea is to make it a normal class. You document the usage in `IBMRuntimeService` constructor and then convert it to your new class. The benefit of that is you can then add methods like `validate()` and `to_request_params()`, which encapsulate all the work related to proxy specifications. Unfortunately it doesn't seem custom methods are supported for `TypedDict` subclasses. 

In addition, I'd suggest making it a separate module instead of lumping into `account.py` since its content is used by both `Account` and `ClientParams`.

daka1510: Thanks for your thoughtful and crisp suggestions. Updated code accordingly and think/hope it got much cleaner, @jyu00. Please take a look.

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -48,17 +49,30 @@ def _assert_valid_url(url: str) -> None:
     try:
         urlparse(url)
     except:
-        raise ValueError(f"Inappropriate `url` value. Failed to parse '{url}' as URL.")
+        raise ValueError(f"Invalid `url` value. Failed to parse '{url}' as URL.")
+
+
+def _assert_valid_proxies(config: ProxyConfiguration) -> None:
```
jyu00, line 55: Why did you decide to use private functions in `account.py` instead of making them part of the `Account` class?

qiskit_ibm_runtime/api/clients/utils.py
```diff
@@ -61,8 +61,8 @@ def ws_proxy_params(client_params: ClientParameters, ws_url: str) -> Dict:
 
     if "auth" in conn_data:
         out["http_proxy_auth"] = (
-            client_params.proxies["username_ntlm"],
-            client_params.proxies["password_ntlm"],
+            client_params.proxies.username_ntlm,
+            client_params.proxies.password_ntlm,
```
jyu00, line 65: This function was created to retrofit the `proxies` parameters, which is in `requests` format, into websockets format. Since we now have a nice new `ProxyConfiguration` class, this should probably go into that class (e.g. as `to_ws_params()`) so we don't have proxy-related functions in multiple places.

daka1510: Good catch, didn't understand that indirection initially. Fixed via https://github.com/Qiskit/qiskit-ibm-runtime/commit/d94daf50f4e7b5a2b4698b14b1ba0cebfa72c205.

rathishcholarajan (changes requested):
@daka1510 Quite a few proxy related tests have failed. Could you please check?
https://github.com/daka1510/qiskit-ibm-runtime/runs/4766501390?check_suite_focus=true#step:5:5927

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -134,3 +99,58 @@ def __eq__(self, other: object) -> bool:
...
         )
+
+    def validate(self) -> "Account":
+        self._assert_valid_auth(self.auth)
+        self._assert_valid_token(self.token)
+        self._assert_valid_url(self.url)
+        self._assert_valid_instance(self.auth, self.instance)
+        self._assert_valid_proxies(self.proxies)
+        return self
+
```
daka1510, line 110: Had to move the validation part into a separate function and not include it in the constructor (appears to be a cleaner pattern anyway). 

Reason: in some cases we first instantiate an invalid account and make it valid by patching additional properties, see
https://github.com/Qiskit/qiskit-ibm-runtime/blob/a624ddf8eceded51ee24cdfa44c1ac57fccc41e9/qiskit_ibm_runtime/ibm_runtime_service.py#L250-L260

daka1510:
@rathishcholarajan good catch, fixed failing integration test cases via https://github.com/Qiskit/qiskit-ibm-runtime/pull/80/commits/de874a9d56f863f919202716b86ef6b7498aa315 and https://github.com/Qiskit/qiskit-ibm-runtime/pull/80/commits/68149286ec054183887b1abef6cfc688efcc2d4b

daka1510:
(Still waiting for push test results in my fork, will check again once https://github.com/daka1510/qiskit-ibm-runtime/runs/4774200812?check_suite_focus=true completes).

rathishcholarajan (approved):
LGTM!

PR #82, opened 2022-01-07
Don't raise if cancel failed because job is done

jyu00:
### Summary

Test cases that cancel a running job can [fail](https://github.com/Qiskit/qiskit-ibm-runtime/runs/4739824168?check_suite_focus=true#step:5:2651) if the job has already finished. This PR adds a utility function that logs a warning and exits the test case if the cancel failed due to job already finished. 

### Details and comments
Fixes #


rathishcholarajan (approved):
LGTM!

PR #83, opened 2022-01-07
Separating slow test terra main

rathishcholarajan:
### Summary
Slow test is usually more stable than slow test against terra main, having them both under same file gives impression that slow tests are always failing in the github UI, so separating them gives better visibility.

### DIFF
--- a/.github/workflows/cron-slow-terra-main.yml
+++ b/.github/workflows/cron-slow-terra-main.yml
@@ -10,13 +10,13 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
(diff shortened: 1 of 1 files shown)


daka1510 (approved):
Looks good.

**Related Remarks**
I think we should revisit the naming scheme of our workflows as a whole. Getting this "right" requires additional thoughts that I do not want to push into this PR :).

_Why do I think so?_
- I find the current naming scheme inconsistent because we only prefix some of the scheduled jobs with `cron-`
- The name "slow test" alone does not hint at what tests are run (performance tests, integration tests, unit tests, ...)

PR #84, opened 2022-01-08
Fix documentation issues

rathishcholarajan:
### Summary
- Hide dashboard for initial release
- Remove hub, group and project parameters for IBMRuntimeService in docs
- No need to make utils public in docs

### Details and comments
Fixes #

### DIFF
--- a/docs/apidocs/ibm-runtime.rst
+++ b/docs/apidocs/ibm-runtime.rst
@@ -1,5 +1,5 @@
 *****************************************
-Qiskit IBM Quantum Provider API Reference
+Qiskit IBM Runtime API Reference
 *****************************************
 
 .. toctree::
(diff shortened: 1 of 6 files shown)


docs/apidocs/ibm-runtime.rst
```diff
@@ -1,11 +1,10 @@
...
 *****************************************
 
 .. toctree::
    :maxdepth: 1
 
    ibm_runtime
    ibm_visualization
-   ibm_jupyter
-   ibm_utils
+   ibm_jupyter
```
daka1510, line 10: Side-note: why do we continue to need `ibm_visualization`, `ibm_jupyter`?

daka1510 (approved):
Looks good. Just added one related question.

docs/apidocs/ibm-runtime.rst
rathishcholarajan: Visualizing a backend can still be useful. But we can consider if we want to remove in a separate PR.

PR #87, opened 2022-01-10
Display job and program id in service.jobs()

daka1510:
### Summary
Print program and job id in service.jobs()

image


rathishcholarajan (approved):
LGTM!

PR #92, opened 2022-01-12
Split test_integration_job

jyu00:
### Summary

Many test cases in `test_integration_job` were failing on Windows with connection issues. We had a similar problem in `qiskit-ibmq-provider` wherein connections would just drop if a single test job on Windows makes too many requests. This PR moves job retrieval and interim result test cases into their own modules, and run a `test3` separate. Hopefully this will get past the connection issue in cron jobs.

### Details and comments
Fixes #


rathishcholarajan (approved):
LGTM!

PR #94, opened 2022-01-12
Allow running a program without backend on cloud

jyu00:
### Summary

This PR adds code to allow a program to run on cloud runtime without having a backend specified. The server will automatically pick a backend and return the name. 

### Details and comments
Fixes #


rathishcholarajan (approved):
LGTM!

PR #95, opened 2022-01-12
Disallow filtering by instance if using cloud

jyu00:
### Summary

As discussed, filtering by `instance` when using cloud runtime will not be supported in the initial release.

### Details and comments
Fixes #


rathishcholarajan (approved):
LGTM!

PR #98, opened 2022-01-12
Fix missing program in test

jyu00:
### Summary

`test_retrieve_unauthorized_program_data` uses a hard coded program id `sample-program`, which is no longer available on cloud runtime. This PR fixes that by selecting a public program. 

### Details and comments
Fixes #


rathishcholarajan (approved):
LGTM!

PR #99, opened 2022-01-12
Allow users to set a logging level when running a program

### LINKED ISSUE #11: Specify logging level when running a program
Ported from https://github.com/Qiskit-Partners/qiskit-ibm/issues/192.

### What is the expected enhancement?
Users should be able to specify the logging level of a specific runtime execution instance when running a program. `logging_level` query parameter is being added to `POST /jobs`. Possible values for logging_level: critical, error, warning, info, debug, CRITICAL, ERROR, WARNING, INFO, DEBUG. Default is warning/WARNING.

jyu00:
### Summary

Allow users to set a logging level when running a program.
@daka1510 Sorry for taking another one from you, but I need this to get some performance numbers. 

### Details and comments
Fixes #11

This PR adds a new `RuntimeOptions` class for the execution options, including `backend_name`, `image`, and now `log_level`. Since we already have `options` as a parameter to `run()`, it made more sense to me to have both `image` and `log_level` in it instead of creating more parameters, especially since they're both related to the execution environment.

### DIFF
--- a/qiskit_ibm_runtime/runtime_options.py
+++ b/qiskit_ibm_runtime/runtime_options.py
@@ -30,7 +30,7 @@ class RuntimeOptions:
             the form of ``image_name:tag``. Not all accounts are
             authorized to select a different image.
         log_level: logging level to set in the execution environment. The default
-            level is WARNING.
+            level is ``WARNING``.
     """
 
     backend_name: Optional[str] = None
--- a/qiskit_ibm_runtime/runtime_options.py
+++ b/qiskit_ibm_runtime/runtime_options.py
@@ -29,8 +29,9 @@ class RuntimeOptions:
         image: the runtime image used to execute the program, specified in
             the form of ``image_name:tag``. Not all accounts are
             authorized to select a different image.
-        log_level: logging level to set in the execution environment. The default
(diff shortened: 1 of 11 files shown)


qiskit_ibm_runtime/runtime_options.py
```diff
@@ -0,0 +1,66 @@
...
+@dataclass
+class RuntimeOptions:
+    """Class for representing runtime execution options.
+
+    Args:
+        backend_name: target backend to run on. This is required for legacy runtime.
+        image: the runtime image used to execute the program, specified in
+            the form of ``image_name:tag``. Not all accounts are
+            authorized to select a different image.
+        log_level: logging level to set in the execution environment. The default
```
rathishcholarajan, line 32: Should we list the logging levels here so the user doesn't have to try out something and then learn the logging levels from the invalid input message?

rathishcholarajan (approved):
LGTM! Left a minor doc suggestion...

qiskit_ibm_runtime/runtime_options.py
jyu00: oops I thought I had them. Fixed in e5f05cb

PR #97, opened 2022-01-12
Create tutorials

jyu00:
### Summary

First pass of the tutorials. Some require code not yet in main (`BackendV2` and #94) but should be enough for initial review.  

### Details and comments
Fixes #

PR #102, opened 2022-01-14
Add support for BackendV2

### LINKED ISSUE #73: Runtime 0.1 - BackendV2
qiskit-ibm-runtime is right now working fine with BackendV1. We need to upgrade it to support BackendV2.

**Related Terra PRs:**
https://github.com/Qiskit/qiskit-terra/pull/5885
https://github.com/Qiskit/qiskit-terra/pull/7323

**Acceptance Criteria:**
WIP (Rathish)

rathishcholarajan:
### Summary
Add support for BackendV2

TODO:
- [ ] Fix perf issue with loading pulse defaults

### Details and comments
Fixes #73


qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -0,0 +1,161 @@
...
+from qiskit.circuit.gate import Gate
+from qiskit.circuit.measure import Measure
+from qiskit.circuit.reset import Reset
+from qiskit.providers.models.pulsedefaults import PulseDefaults
+from qiskit.test.mock.utils.json_decoder import decode_pulse_defaults
+
+
+def convert_to_target(
+    configuration: Dict, properties: Dict = None, defaults: Dict = None
+) -> Target:
```
daka1510, line 31: Consider using https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors for the commit that includes this file so that the original author (Matthew) is also represented in the history.

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -73,13 +79,61 @@ def __init__(
...
+        except AttributeError:
+            pass
+        try:
+            return self._configuration.__getattribute__(name)
+        except AttributeError:
+            raise AttributeError(
+                "'{}' object has no attribute '{}'".format(
+                    self.__class__.__name__, name
+                )
+            )
```
daka1510, line 110: I did a bit on research myself (e.g. [[1]](https://stackoverflow.com/questions/3278077/difference-between-getattr-vs-getattribute)) but am still struggling to understand these lookup fallbacks. Can you elaborate a bit (or, ideally add a code comment that explains a little background here)?

daka1510:
Added a minor comment, other than that it looks good to me.

This PR will likely conflict with (at least) https://github.com/Qiskit/qiskit-ibm-runtime/pull/97 and https://github.com/Qiskit/qiskit-ibm-runtime/pull/114.

qiskit_ibm_runtime/ibm_backend.py
rathishcholarajan: I've added doc string and comments.

daka1510: Thanks, that makes it much clearer.

daka1510 (approved):
Looks good from my side.

jyu00:
Don't forget to update [this tutorial](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/docs/tutorials/03_backends.ipynb), which has a giant "TODO - finish when BackendV2 is done" lol

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -74,13 +158,73 @@ def __init__(
...
+            if api_defaults:
+                self._defaults = defaults_from_server_data(api_defaults)
+
+    def _convert_to_target(self) -> None:
+        """Converts backend configuration, properties and defaults to Target object"""
+        if not self._target:
+            self._target = convert_to_target(
+                configuration=self._configuration.to_dict(),
+                properties=self._properties.to_dict() if self._properties else None,
+                defaults=self._defaults.to_dict() if self._defaults else None,
```
mtreinish, line 226: Why is `to_dict()` necessary here? It's adding extra overhead to the construction. Basically you're going from JSON string response body -> python dict -> `qiskit.providers.models` object -> python dict -> Target and you end up traversing the de-serialized response body 3 times (once for the model object creation, once for the `to_dict()` and once to create the target).

I wrote the converter code assuming it'd be done with the deserialized json payload bodies directly since that'd be the most efficient. But, if that's not feasible here we should modify `convert_to_target` to work with the models objects directly (via attribute access) instead of the output of `to_dict`. This way you at least eliminate one traversal and intermediate object.

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -74,13 +158,73 @@ def __init__(
...
+
+    def __getattr__(self, name: str) -> Any:
+        """Gets attribute from self or configuration
+
+        This magic method executes when user accesses an attribute that
+        does not yet exist on IBMBackend class.
+        """
+        # Lazy load properties and pulse defaults and construct the target object.
+        self._get_properties()
+        self._get_defaults()
```
mtreinish, line 185: Did you look into lazy loading this via an `InstructionProperties` subclass like we discussed offline? I'm concerned about the performance overhead if we're required to fetch and parse the defaults payload for ever use of a backend with the qiskit compiler.

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -74,13 +158,73 @@ def __init__(
...
-        # Attributes used by caching functions.
+        self._configuration = configuration
         self._properties = None
+        self._qubit_properties = None
         self._defaults = None
+        self._target = None
+        self._max_circuits = configuration.max_experiments
+        if not self._configuration.simulator:
+            self.options.set_validator("noise_model", type(None))
+            self.options.set_validator("seed_simulator", type(None))
```
mtreinish, line 175: Do you want to set validators based on `max_shots`, `rep_delay_range, `qubit_lo_range`, and `meas_lo_range`. Something like:

```suggestion
        if not self._configuration.simulator:
            self.options.set_validator("noise_model", type(None))
            self.options.set_validator("seed_simulator", type(None))
        self.options.set_validator("shots", (1, configuration.max_shots)
        self.options.set_validator(
            "rep_delay", (configuration.rep_delay_range[0], configuration.configuration.rep_delay_range[1])
        )
        self.options.set_validator(
            "qubit_lo_freq", (configuration.qubit_lo_range[0], configuration.configuration.qubit_lo_range[1])
        )
        self.options.set_validator(
            "measure_lo_freq", (configuration.measure_lo_range[0], configuration.configuration.measure_lo_range[1])
        )
```

mtreinish (commented):
Just some initial comments from a quick first pass over the PR

qiskit_ibm_runtime/ibm_backend.py
rathishcholarajan: Merging this as is for the rc2 release, I will handle this in a separate PR.

Issue #74, opened 2022-01-06
Exiting Python session generates TypeError


jyu00:
**Describe the bug**

After using `IBMRuntimeService` in a Python CLI session, when exiting it generated a bunch of `TypeError`:

```
Exception ignored in: <function RetrySession.__del__ at 0x13fd5cd30>
Traceback (most recent call last):
  File "/Users/jessieyu/Documents/Q/Github/qiskit-ibm-runtime/qiskit_ibm_runtime/api/session.py", line 173, in __del__
  File "/Users/jessieyu/.pyenv/versions/qiskit-ibm-runtime/lib/python3.8/site-packages/requests/sessions.py", line 747, in close
  File "/Users/jessieyu/.pyenv/versions/qiskit-ibm-runtime/lib/python3.8/site-packages/requests/adapters.py", line 325, in close
  File "/Users/jessieyu/.pyenv/versions/qiskit-ibm-runtime/lib/python3.8/site-packages/urllib3/poolmanager.py", line 222, in clear
  File "/Users/jessieyu/.pyenv/versions/qiskit-ibm-runtime/lib/python3.8/site-packages/urllib3/_collections.py", line 100, in clear
  File "/Users/jessieyu/.pyenv/versions/qiskit-ibm-runtime/lib/python3.8/site-packages/urllib3/poolmanager.py", line 173, in <lambda>
  File "/Users/jessieyu/.pyenv/versions/qiskit-ibm-runtime/lib/python3.8/site-packages/urllib3/connectionpool.py", line 494, in close
TypeError: catching classes that do not inherit from BaseException is not allowed
Exception ignored in: <function RetrySession.__del__ at 0x13fd5cd30>
Traceback (most recent call last):
```

**Steps to reproduce**

```
>>> from qiskit_ibm_runtime import IBMRuntimeService
>>> service = IBMRuntimeService()
>>> backend = service.least_busy()
<ctrl-D>
```

**Expected behavior**

**Suggested solutions**

**Additional Information**

- **Qiskit IBM Runtime version**:
- **Python version**: 3.8.1
- **Operating system**:

daka1510:
That's a tricky one. Nailed it down as below and continue the investigation early next week.

#### good-case.py
```python
from requests import Session


class CustomSession(Session):
    def __del__(self) -> None:
        self.close()


session = CustomSession()
result = session.get("https://auth.quantum-computing.ibm.com/api/version").json()
print(result)
```

[...]

PR #103, opened 2022-01-17
Ignore errors during session cleanup

daka1510:
### Summary
See https://github.com/Qiskit/qiskit-ibm-runtime/issues/74#issuecomment-1014537885 for background information on the proposed solution.

### Details and comments
https://github.com/Qiskit/qiskit-ibm-runtime/issues/74, i.e. avoids `TypeErrors` when scripts finish or interactive python sessions are aborted.

### DIFF
--- a/qiskit_ibm_runtime/api/session.py
+++ b/qiskit_ibm_runtime/api/session.py
@@ -170,7 +170,11 @@ class RetrySession(Session):
 
     def __del__(self) -> None:
         """RetrySession destructor. Closes the session."""
-        self.close()
+        try:
+            self.close()
+        except Exception:  # pylint: disable=broad-except
+            # ignore errors that may happen during cleanup
+            pass
 
     def _initialize_retry(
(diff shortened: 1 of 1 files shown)


daka1510:
@jyu00, @rathishcholarajan please check the referenced comment for details. Not the most satisfying solution but one that should work.

Let me know if you see other reliable options that would allow us to get rid of triggering session cleanup from within the class destructor.

jyu00:
My only concern about this solution is that `close()` in `connectionpool.py` loops through and close all connections. By tripping this exceptions there could be connections not cleaned up. 

Having said that, I think your solution is good enough for now given the schedule. However, I'd prefer leaving the issue open for further investigation in the future, especially since this was not an issue with `qiskit-ibmq-provider`.

rathishcholarajan (approved):
LGTM! Agree with @jyu00 on leaving the issue open for further investigation.

PR #106, opened 2022-01-18
Update encoder/decoder to handle instructions

kt474:
### Summary

As previously discussed, serializing a single instruction is a pretty niche use case and this seems like a hacky fix. Thoughts on this approach - @mtreinish @rathishcholarajan 

### Details and comments
Fixes https://github.com/Qiskit-Partners/qiskit-ibm/issues/190


qiskit_ibm_runtime/utils/json.py
```diff
@@ -210,10 +211,13 @@ def default(self, obj: Any) -> Any:  # pylint: disable=arguments-differ
             )
             return {"__type__": "ParameterExpression", "__value__": value}
         if isinstance(obj, Instruction):
+            # Append instruction to empty circuit
+            quantum_register = QuantumRegister(obj.num_qubits)
+            quantum_circuit = QuantumCircuit(quantum_register)
```
mtreinish, line 216: You technically don't need to create a separate register here, you can just do:

```suggestion
            quantum_circuit = QuantumCircuit(obj.num_qubits)
```

test/test_data_serialization.py
```diff
@@ -214,6 +214,19 @@ def test_encoder_datetime(self):
             decoded = json.loads(encoded, cls=RuntimeDecoder)
             self.assertEqual(decoded, obj)
 
+    def test_encoder_instruction(self):
+        """Test encoding and decoding instructions"""
+        subtests = (
+            {"instruction": CXGate()},
+            {"instruction": PhaseGate(theta=1)},
+            {"instruction": U2Gate(phi=1, lam=1)},
```
mtreinish, line 222: It'd probably be good to test parameterized instructions too, just to make sure those are handled correctly. If you're really handling submission of standalone instruction objects to runtimes (I still don't know what the 
use case for this could be) having unbound parameters on the instruction seems worth testing here. Something like:

```suggestion
            {"instruction": CXGate()},
            {"instruction": PhaseGate(theta=1)},
            {"instruction": U2Gate(phi=1, lam=1)},
            {"instruction": U2Gate(phi=Parameter('phi'), lam=Parameter('lambda'))},
```

and maybe one with a `ParameterVector` element and a `ParameterExpression` too.

qiskit_ibm_runtime/utils/json.py
```diff
@@ -270,9 +274,8 @@ def object_hook(self, obj: Any) -> Any:
                     obj_val, self.__read_parameter_expression, False
                 )
             if obj_type == "Instruction":
-                return _decode_and_deserialize(
-                    obj_val, qpy_serialization._read_instruction, False
-                )
+                circuit = _decode_and_deserialize(obj_val, qpy_serialization.load)[0]
+                return circuit.data[0][0]
```
mtreinish, line 278: I think having a comment here on what this is doing would be useful. Something like:

```
# Standalone instructions are encoded as the sole instruction in a QPY serialized circuit
# to deserialize load qpy circuit and return first instruction object in that circuit.
```

mtreinish (commented):
Overall this looks fine to me. I have small inline comments, but nothing major

daka1510 (approved):
LGTM, thanks @kt474.

@rathishcholarajan - do you agree to merge it that way?

rathishcholarajan (approved):
LGTM! @kt474 Could you please cherry pick this to qiskit-ibmq-provider?

PR #108, opened 2022-01-19
Update README and API doc

jyu00:
### Summary

Fix README and documentation, such as using `sampler` instead of `circuit-runner`; `hub/group/project` instead of `provider` etc.

### Details and comments

Note that I deleted 2 examples from the visualization package since the code doesn't work anymore (no idea how doc was able to build before). They should be added back in when we are ready to include the visualization package.


rathishcholarajan (approved):
LGTM! Left a minor comment...

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2022.
```
daka1510, line 3: This raises a general question:
- Should the year in the copyright header be updated to 2022 whenever we modify files?

Background: in past projects I learned one should not replace but extend the list of years (i.e. 2021, 2022) whenever copyrightable changes are added. Not sure how important those changes are but let's make sure we all use a consistent approach here. I'm fine with either way.


**Related**
https://opensource.stackexchange.com/questions/6389/how-do-the-years-specified-in-a-copyright-statement-work

daka1510 (approved):

qiskit_ibm_runtime/runtime_job.py
jyu00: Yes the convention here is also `<start year>, <last modified year>` if there is more than one, or just `<start/ last modified year>` if they both happen in the same year. 

Now the question is, which year is considered the _first published_. In my old job (proprietary) it was always the year it's _released_ to the users (in our case 2022). But since this is an open source project, perhaps it should be when it's first added to Github (i.e. 2021)? I'm happy to change this to `2021, 2022` if people feel strongly about it. Not that we are at all good at keep it updated lol

daka1510: Thanks for the background and clarifications. That helps. No strong feelings about this change from my side. I'm fine with keeping it at 2022 for now!

PR #119, opened 2022-01-24
Allow to trigger workflow that publishes the documentation manually

daka1510:
### Summary
Allow to trigger workflow that publishes the documentation manually. When this PR is integrated into the main branch, all contributors with write access to the repository will be able to run it like so
image

### Details and comments
N/A

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - "*"
+  workflow_dispatch:
 
 jobs:
   deploy:


.github/workflows/docs.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - "*"
+  workflow_dispatch:
```
daka1510, line 7: cf. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

rathishcholarajan:
@daka1510 Good to know. We shouldn't have to do this anymore since we are planning to a pre-release tag soon (for internal testing purposes) which should deploy the docs.

daka1510:
> Also it's too risky to have this enabled even after release since anyone on the team can accidentally trigger it and break production docs

Let's close this PR then. If we ever need it, it's easy to recreate.

PR #122, opened 2022-01-24
/jobs API returns 400 when program is not found

rathishcholarajan:
### Summary
/jobs API returns 400 when program is not found and not 404. 400 is returned for other bad requests as well and hence checking the message as well.

### Details and comments
Fixes #


jyu00:
Shouldn't this be fixed on the server side? 404 is a pretty standard "not found" return code.

rathishcholarajan:
I've created a bug with API team, if they fix it then we can close this PR, else we need this to raise appropriate exception.

daka1510 (commented):
Agree with @jyu00 that we should not attempt to work around an issue that can be fixed on the API side.

PR #136, opened 2022-01-31
Fix broken docs links 

### LINKED ISSUE #131: Tutorial source links don't work
**Describe the bug**

The tutorial docs on qiskit.org have links to the actual notebooks to allow people to run them, but the links don't work. 

image

**Steps to reproduce**

**Expected behavior**

The links should work. 

**Suggested solutions**

**Additional Information**

- **Qiskit IBM Runtime version**:
- **Python version**:
- **Operating system**:

### LINKED ISSUE #132: Link to max execution doc doesn't work
**Describe the bug**

Tutorials on qiskit.org doc point to the max execution doc, but the link doesn't work (`this documentation` is supposed to be a hyperlink):

image

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

**Additional Information**

- **Qiskit IBM Runtime version**:
- **Python version**:
- **Operating system**:

kt474:
### Details and comments
Fixes #131 
Fixes #132


docs/conf.py
```diff
@@ -91,7 +91,7 @@
     .. note::
         This page was generated from `docs/{{ docname }}`__.
 
-        __"""
+    __ https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/docs/{{ docname }}"""
```
rathishcholarajan, line 94: @kt474 It won't be always accurate if we point to the main branch link. We need to point to a stable branch or a tag. So if we do something like https://github.com/Qiskit/qiskit-ibm-runtime/tree/{{ release }}/docs/{{ docname }} would it work? Can you test if we can use the `release` variable from this file?

rathishcholarajan (changes requested):

docs/conf.py
kt474: Yeah we can do something similar to whats done in [qiskit-nature](https://github.com/Qiskit/qiskit-nature/blob/stable/0.3/docs/conf.py) like this 

```
link_str = f" https://github.com/Qiskit/qiskit-ibm-runtime/blob/{release}/docs/"
nbsphinx_prolog += link_str + "{{ docname }}"
```

where it'll use the release variable (the link still wouldn't work at the moment bc 0.1.0 doesn't exist)

docs/conf.py
```diff
@@ -93,6 +93,9 @@
 
         __"""
 
+link_str = f" https://github.com/Qiskit/qiskit-ibm-runtime/blob/{release}/docs/"
+nbsphinx_prolog += link_str + "{{ docname }}"
```
rathishcholarajan, line 97:
```suggestion
vers = release.split(".")
link_str = f" https://github.com/Qiskit/qiskit-ibm-runtime/blob/{vers[0]}.{vers[1]}/docs/"
nbsphinx_prolog += link_str + "{{ docname }}"
```

rathishcholarajan (changes requested):
It's better to point to the stable branch

docs/conf.py
rathishcholarajan: Could you check if the URL is getting formed properly with above change?

kt474: is the stable branch going to be `0.1` or `0.1.0`?

rathishcholarajan: Stable branch will be 0.1

rathishcholarajan (approved):
LGTM!

docs/conf.py
```diff
@@ -93,6 +93,11 @@
 
         __"""
 
+link_str = f" https://github.com/Qiskit/qiskit-ibm-runtime/blob/{release}/docs/"
```
kt474, line 96: we can remove this line

Issue #127, opened 2022-01-27
Create happy path E2E test suite to run against simulators or real devices every night


rathishcholarajan:
**What is the expected feature or enhancement?*
We used to run slow tests against real devices every night. PR https://github.com/Qiskit/qiskit-ibm-runtime/pull/114 removes running slow tests.

We should create a critical path minimal E2E test suite tailored to run against simulators as well as real devices. (We can discuss in comments what that critical path is).

**Acceptance criteria**
1. Runs nightly against a least busy simulator or real backend.
1. Tutorial notebooks are automatically executed as test case and they run through
 
Bonus points:
Certain people can trigger this manually against a specific simulator or backend of choice on demand. (Selectable from a dropdown list of simulators or backends). So if a new simulator or backend is sprung up and if we are asked to do a sanity check, we can run this suite.

daka1510:
Basic e2e test coverage is enabled and tests run on a daily basis (https://github.com/Qiskit/qiskit-ibm-runtime/actions/workflows/e2e-tests.yml). Closing.

PR #139, opened 2022-02-03
Split by test type

daka1510:
### Summary
Move test cases to dedicated root folders based on their type (`unit`, `integration`, `e2e`). Thereby
- make it obvious what kind of tests are run (vs. deciding under the covers based on environment configuration)
- no longer run unit tests as part of the integration test suite (saves 3-4 minutes per integration test run)
- add dedicated e2e test suite that verifies supported tutorial notebooks run through in a first step

Also clarified how to run each of the tests and how they differ in `Contributing.md`.

### Details and comments
Split test cases by type (integration, unit, e2e) in preparation for https://github.com/Qiskit/qiskit-ibm-runtime/issues/127


test/types/integration/test_auth_client.py
```diff
@@ -14,77 +14,11 @@
...
-from qiskit_ibm_runtime.api.exceptions import ApiError, RequestsApiError
 from qiskit_ibm_runtime.api.client_parameters import ClientParameters
-
-from .ibm_test_case import IBMTestCase
-from .mock.http_server import SimpleServer, ClientErrorHandler
-from .utils.account import custom_envs, no_envs
-from .utils.decorators import integration_test_setup, IntegrationTestDependencies
-
-
-class TestAccountClient(IBMTestCase):
```
daka1510, line 27: We had mixed unit and integration tests in a single file. This clas was therefore moved to a separate file in the `/test/types/unit` directory.

CONTRIBUTING.md
```diff
@@ -198,50 +198,54 @@ section of the Qiskit documentation.
...
+To execute all integration tests, run
 ``` {.bash}
-$ python -m unittest test/test_something.py
+$ python -m unittest discover --verbose --top-level-directory . --start-directory test/types/integration
 ```
 
-Windows:
+##### 3. E2E tests
+
+Executed against an external system configured via a (token, instance, url) tuple. Basic coverage of most important user-facing happy paths. Test suite runs faster than integration but slower than unit tests and is stable.
```
rathishcholarajan, line 221:
```suggestion
Executed against an external system configured via a (token, instance, url) tuple. Basic coverage of most important user-facing happy and non-happy paths. Test suite runs faster than integration but slower than unit tests and is stable.
```

CONTRIBUTING.md
```diff
@@ -198,50 +198,54 @@ section of the Qiskit documentation.
...
+##### 1. Unit tests
+Run locally without connecting to an external system. They are short-running, stable and give a basic level of confidence during development.
 
-For executing a simple python test manually, you can just run this
-command:
-
-Linux and Mac:
+To execute all unit tests, run:
+``` {.bash}
+$ python -m unittest discover --verbose --top-level-directory . --start-directory test/types/unit
```
rathishcholarajan, line 209:
```suggestion
$ make unit-test
```
Since this is super long, can we promote using the make commands?

CONTRIBUTING.md
```diff
@@ -198,50 +198,54 @@ section of the Qiskit documentation.
...
+```
+##### 2. Integration tests
+Executed against an external system configured via a (token, instance, url) tuple. Detailed coverage of happy and non-happy paths. They are long-running and unstable at times. A successful test run gives a high level of confidence that client and APIs work well together.
 
+To execute all integration tests, run
 ``` {.bash}
-$ python -m unittest test/test_something.py
+$ python -m unittest discover --verbose --top-level-directory . --start-directory test/types/integration
```
rathishcholarajan, line 216:
```suggestion
$ make integration-test
```

CONTRIBUTING.md
```diff
@@ -198,50 +198,54 @@ section of the Qiskit documentation.
...
 
-Windows:
+##### 3. E2E tests
+
+Executed against an external system configured via a (token, instance, url) tuple. Basic coverage of most important user-facing happy paths. Test suite runs faster than integration but slower than unit tests and is stable.
 
+To execute all e2e tests, run
 ``` {.bash}
-C:\..\> python -m unittest test/test_something.py
+$ python -m unittest discover --verbose --top-level-directory . --start-directory test/types/e2e
```
rathishcholarajan, line 225:
```suggestion
$ make e2e-test
```

test/types/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Tests by type."""
```
rathishcholarajan, line 13: Why do we need the `types` directory? Could we just have `test/unit`,`test/integration`, `test/e2e` and `test/utils` to keep the directory structure shallow?

rathishcholarajan (changes requested):
Thanks @daka1510! Left a few suggestions, also there are some merge conflicts that needs to be resolved...

test/types/__init__.py
daka1510: Good question. It's a tradeoff. 

Used what you proposed in a first shot, too, and then decided to introduce an additional `types` directory because I didn't like that root folders for different _types_ were at the same level with other common test folders (e.g. `mock`, `utils`) 

```
|-test
|--e2e
|--integration
|--mock
|--unit
|--utils
```

What do you think? Other proposals that would combine "the best of both worlds" ;)?

CONTRIBUTING.md
daka1510: My understanding (and the current status quo) is different here.

We have extensive test coverage for integration tests (with happy, non-happy paths). We therefore accept them to be instable at times.

For our e2e tests the goal should be to keep them simple and focus on the most common user-facing paths without covering non-happy paths.

I'm fine with updating the definition if we reach another agreemetn.

test/types/__init__.py
daka1510: Thought a bit more about how to combine both ideas and came up with a solution via https://github.com/Qiskit/qiskit-ibm-runtime/pull/139/commits/4e489194c793b796600ea65c35f5445fda10468b that leads to below shallow structure:

image

daka1510:
Thanks @rathishcholarajan - addressed your feedback, only one open comment remains. Please take a look :)

rathishcholarajan (approved):
LGTM! Thanks @daka1510!

Issue #124, opened 2022-01-26
Revive or discontinue HTML representation for IBM Backends


daka1510:
**What is the expected feature or enhancement?**
In qiskit-ibmq-provider users could opt-in to HTML representation of IBM backends by importing `qiskit.providers.ibmq.jupyter`
```python
# cf. https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/docs/tutorials/2_jupyter_tools.ipynb
import qiskit.providers.ibmq.jupyter  # This is the where the magic happens (literally).
from qiskit.providers.ibmq import least_busy

backend = least_busy(...)
backend
```
As a result the backend information was displayed as an HTML widget
![image](https://user-images.githubusercontent.com/16404496/151140714-2958b05f-b48a-4e32-8f24-7fee6ee9aef1.png)

The related code has been ported to `qiskit-ibm-runtime` but not been updated to work with latest code changes. Therefore displaying a backend fails when `qiskit_ibm_runtime.jupyter` is imported. As this is not a documented API, it's not an issue that users are likely to hit.

```python
import qiskit_ibm_runtime.jupyter
backend = # ...
backend
```

**Acceptance criteria**
- Option 1 - HTML representation of IBM backends is documented, tested and works
- Option 2 - HTML representation of IBM backends is discontinued, dead code paths have been removed

Issue #112, opened 2022-01-19
Revive IQX dashboard widget


jyu00:
**What is the expected feature or enhancement?**

In qiskit-ibmq-provider there was an IQX dashboard that displayed backends as well as legacy jobs. It'd be nice to have something similar, especially for cloud runtime since they don't show up on IQX. 

**Acceptance criteria**

A user can see runtime backend and job information on a dashboard.

daka1510:
**Related background information**
- The IQX dashboard widget is implemented via the [jupyter](https://github.com/Qiskit/qiskit-ibm-runtime/tree/main/qiskit_ibm_runtime/jupyter) and [visualization](https://github.com/Qiskit/qiskit-ibm-runtime/tree/main/qiskit_ibm_runtime/visualization) modules.
- I have removed it from the public API documentation of qiskit-ibm-runtime via this [PR](https://github.com/Qiskit/qiskit-ibm-runtime/pull/123).
- Its usage in qiskit-ibmq-provider is documented in this [notebook](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/docs/tutorials/2_jupyter_tools.ipynb)
  - when a user runs the Jupyter magic command (`%iqx_dashboard`), below dashboard is shown
![image](https://user-images.githubusercontent.com/16404496/151133188-238bc3b5-bbb0-4cae-ac3d-9250040746a4.png)
![image](https://user-images.githubusercontent.com/16404496/151133269-245a1e9e-dc52-4fd6-8f99-f0d4104cfca3.png)

**Open Questions**
Some initial questions we may want to clarify when we pick up this work item:
1. Should the name stay as-is?
1. Should it work transparently across Cloud (https://cloud.ibm.com/) and Legacy (https://quantum-computing.ibm.com/)?
1. Are the hard-coded device layouts up to date and complete?
1. What are reasons and benefits for implementing a dashboard _widget_ within the client library vs. implementing and linking to a dedicated UI?

jyu00:
My 2 cents:
1. No, we should not be using `IQX` anymore. We can change it to something like `Qiskit Runtime Dashboard`
2. I think it's ok to have separate tabs for cloud and legacy jobs
3. I don't know if we even need to display backends, since there are dedicated UI for that on both IQX and cloud (but not for jobs). 
4. Linking is probably good enough.

PR #143, opened 2022-02-09
Enable strict mypy type checking 

### LINKED ISSUE #88: Mypy does not check types in test code
Mypy should also check test code and fail if typing issues are found. Would have avoided https://github.com/Qiskit/qiskit-ibm-runtime/pull/80#pullrequestreview-848478338 where tests had to fail to uncover that tests have not been updated to the latest method signatures.

kt474:
### Summary

I assume the purpose of #88 is to enable `show_none_errors` which will shows errors related to strict `None` checking - not sure if this would've caught the issue [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/80#pullrequestreview-848478338). Everything else in mypy is already enabled.

Enabling `show_none_errors` gives 52 "errors" - I'll have to go through each of them but there will be merge conflicts with the current PRs and I'm not sure all of the errors are worth fixing

### Details and comments
Fixes #88


daka1510:
> I assume the purpose of https://github.com/Qiskit/qiskit-ibm-runtime/issues/88 is to enable show_none_errors which will shows errors related to strict None checking - not sure if this would've caught the issue https://github.com/Qiskit/qiskit-ibm-runtime/pull/80#pullrequestreview-848478338. Everything else in mypy is already enabled.

No, the intent of https://github.com/Qiskit/qiskit-ibm-runtime/issues/88 is to ensure that `mypy` also checks for type errors in our test code. The command we currently use does not run mypy for all files in the `test` directory. 
```
mypy:
	mypy --module qiskit_ibm_runtime
```

setup.cfg
```diff
@@ -15,3 +16,6 @@ strict_optional = False
 show_none_errors = False
 show_error_codes = True
 no_site_packages = True
+
+[mypy-test.*]
+ignore_errors = True
```
kt474, line 21: Which specific rules do we want to enable here? When enabled, there are over 400 errors - mostly `no-untyped-def`, and `no-untyped-call`

Also, if `test_jupyter` is not excluded, there are a lot errors in `qiskit_ibm_runtime/visualization` and `qiskit_ibm_runtime/jupyter`

daka1510: `qiskit_ibm_runtime/visualization` and `qiskit_ibm_runtime/jupyter` are dead code paths that we will follow up on via https://github.com/Qiskit/qiskit-ibm-runtime/issues/112, https://github.com/Qiskit/qiskit-ibm-runtime/issues/124. So I agree to exclude it from type checking.

I suggest to start with
```
[mypy-test.*]
disallow_untyped_calls = False
disallow_untyped_defs = False
```
and investigate/fix all remaining 60 issues. 41 issues can be fixed very easily by adding an explicit return type (e.g. `-> None`).
```
# appears like disallow_untyped_defs does not apply here. If function parameters are partially typed, mypy expects the return type to be explicitly set
test/integration/test_proxies.py:120: error: Function is missing a return type annotation  [no-untyped-def]
```

test/ibm_test_case.py
```diff
@@ -1,3 +1,4 @@
+# mypy: ignore-errors
```
daka1510, line 1: Why is this needed?

test/unit/mock/fake_runtime_service.py
```diff
@@ -52,11 +52,11 @@ def __init__(self, *args, **kwargs):
         ):
             super().__init__(*args, **kwargs)
 
-    def _authenticate_legacy_account(self, client_params: ClientParameters):
+    def _authenticate_legacy_account(self, client_params: ClientParameters) -> Any:
```
daka1510, line 55: Does `-> FakeAuthClient` not work?

daka1510 (commented):
Thanks for following up, @kt474. Added 2 questions, otherwise looks good.

test/unit/mock/fake_runtime_service.py
kt474:
```
Return type "FakeAuthClient" of "_authenticate_legacy_account" incompatible with return type "AuthClient" in supertype "IBMRuntimeService"
```

There is a mypy override error if `FakeAuthClient` is used as a return type so we can have `Any` as a return type or do something like ```-> 'FakeAuthClient' # type: ignore```

test/ibm_test_case.py
kt474: There are `attr-defined` errors in the setup and teardown class methods - I can ignore the specific lines instead of ignoring errors for the whole file

daka1510: Can we also try to fix it in the test code? Would something like below work instead?

```
class IBMTestCase(unittest.TestCase):
    """Custom TestCase for use with the Qiskit IBM Runtime."""

    log: logging.Logger
    dependencies: IntegrationTestDependencies
    service: IBMRuntimeService
```

daka1510: That way we can benefit from auto-completion in subclasses that access the class attributes :)

image

daka1510 (approved):
Thanks for following up on my proposals, @kt474. Looks good to me. @rathishcholarajan any objections from your side to integrate it?

rathishcholarajan (approved):
LGTM!

PR #150, opened 2022-02-16
Run e2e tests on a daily basis

daka1510:
### Summary
As of today the [E2E tests](https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/1852419954) only run when manually triggered. They all succeeded today and therefore I think we can enable them to run on a daily basis.

image

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -12,6 +12,8 @@
 
 name: E2E Tests
 on:
+  schedule:
+    - cron: '0 16 * * *'
   # allow everyone with write access to the repository to trigger the workflow run manually
   workflow_dispatch:
 jobs:


.github/workflows/e2e-tests.yml
```diff
@@ -12,6 +12,8 @@
 
 name: E2E Tests
 on:
+  schedule:
+    - cron: '0 16 * * *'
```
daka1510, line 16: Integration tests use `0 4 * * *` and use the same account. I suggest to split load and start the e2e tests 12 hours later.

rathishcholarajan: We should avoid running during US daytime preferably. Maybe start a couple hours after integration tests start?

rathishcholarajan (changes requested):

.github/workflows/e2e-tests.yml
rathishcholarajan, line 16:
```suggestion
    - cron: '0 8 * * *'
```

rathishcholarajan (approved):
LGTM!

PR #172, opened 2022-02-28
Allow docs publish manual trigger, not deploy

rathishcholarajan:
### Details and comments
Fixes #


daka1510:
@rathishcholarajan why is this needed?

I'm totally fine with this change, just want to understand why we need it now and it was too risky to add back in January :) (https://github.com/Qiskit/qiskit-ibm-runtime/pull/119#issuecomment-1020487748).

PR #179, opened 2022-03-02
Update jobs to use state instead of status 

### LINKED ISSUE #168: Switch to using `state` object instead of status `string`
**Describe the bug**
We have been using the `status` string so far which could look like this when the job gets cancelled - "Cancelled" (general) or "Cancelled - Ran too long". The latter is used whenever a job gets cancelled since it ran past maximum execution time.

These two different cancelled states have now been merged and a new `state` field is being returned by the API with `status` and `reason`. The `status` string is also being returned for backwards compatibility until provider and UI switch to it. So provider should hereafter ignore `status` string and always read from `state` field going forward.

```
{
  "id": "c7ncn08ihcmlom131fm0",
  "hub": "defaultHub",
  "group": "defaultGroup",
  "project": "defaultProject",
  "backend": "ibmq_qasm_simulator",
  "state": {
    "status": "Cancelled",
    "reason": "Ran too long"
  },
  "params": {},
  "program": {
    "id": "test-state-prog-8B4QozJRkg"
  },
  "created": "2022-01-24T15:53:38.68766Z",
  "runtime": "qiskit-program:latest",
  "status": "Cancelled - Ran too long"
}
```

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

**Additional Information**
Jira NTC-129
PR NTC-912

- **Qiskit IBM Runtime version**:
- **Python version**:
- **Operating system**:

kt474:
### Details and comments
Fixes #168


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -307,9 +307,17 @@ def _set_status(self, job_response: Dict) -> None:
...
         try:
-            self._status = API_TO_JOB_STATUS[job_response["status"].upper()]
+            if "reason" in job_response["state"]:
+                status = (
+                    job_response["state"]["status"]
+                    + " - "
+                    + job_response["state"]["reason"]
+                )
+            else:
+                status = job_response["state"]["status"]
```
daka1510, line 317: This change retrofits the new response format to the existing code in `qiskit-ibm-runtime`.

I would prefer if we could keep our implementation aligned with the API changes, i.e. use separate properties to store the returned `reason` and `status` values instead of string concatenation.

As a result we would need to infer the job status in a slightly different way. Instead of accessing the [key-value maps](https://github.com/Qiskit/qiskit-ibm-runtime/blob/9f8b89e5de71a93b1ebc49d94bac6fdcea6af58f/qiskit_ibm_runtime/constants.py#L20-L33) there could be a dedicated helper functions to compute the job status or job error message given the returned state (reason/status) from the API.

test/unit/mock/fake_runtime_client.py
```diff
@@ -117,6 +117,7 @@ def __init__(
         """Initialize a fake job."""
         self._job_id = job_id
         self._status = final_status or "QUEUED"
+        self._state = {"status": final_status or "QUEUED"}
```
daka1510, line 120: This appears redundant. The API needs this redundancy to not break existing clients. In our own mocks I think we should no longer include `self._status`.  This gives us more confidence that we no longer depend on a soon-to-be-removed response property.

daka1510 (changes requested):

qiskit_ibm_runtime/runtime_job.py
daka1510: Competing opinions, @kt474 or @rathishcholarajan?

rathishcholarajan: Yes I'd prefer storing `status` and `reason` as separate properties as well. We might expand to giving reasons for other sorts of failures.

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -111,7 +111,8 @@ def __init__(
         self._params = params or {}
         self._creation_date = creation_date
         self._program_id = program_id
-        self._status = JobStatus.INITIALIZING
+        self._state = {"status": JobStatus.INITIALIZING}
```
daka1510, line 114: `self._status` does not map to the API response format directly. Why do you introduce a more complicated data structure here? Can't we leave it as before?

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -308,24 +309,48 @@ def _set_status(self, job_response: Dict) -> None:
...
+        Args:
+            status: Job status returned from the API.
+            msg_type: Dictionary to use to convert API message.
+            reason: Job status reason returned from the API.
+
+        Returns: A job status or errror message.
+        """
+        if reason:
+            status = status + " - " + reason
+        return msg_type[status.upper()]
```
daka1510, line 352: I believe this value should not return a concatenated string but one of the [JobStatus enum values](https://github.com/Qiskit/qiskit-terra/blob/2eee56616d50a9e26756f855ef4aa0135920ad78/qiskit/providers/jobstatus.py#L18-L27)

daka1510: Here's pseudo code that outlines the key changes that I believe are required:

```python
# change 1 - update mapping table that relies on the old concatenated status property
API_TO_JOB_STATUS = {
    "QUEUED": JobStatus.QUEUED,
    "RUNNING": JobStatus.RUNNING,
    "COMPLETED": JobStatus.DONE,
    "FAILED": JobStatus.ERROR,
    "CANCELLED": JobStatus.CANCELLED,
    # removed key "CANCELLED - RAN TOO LONG"
}

# change 2 - add dedicated function to derive status enum value from job response
self._status = status_from_job_response(response)


# change 3 - implement mapping code
def status_from_job_response(response):
  mapped_job_status = API_TO_JOB_ERROR_MESSAGE[response.state.status]
  if mapped_job_status == JobStatus.CANCELLED and response.reason = "Ran too long"
     mapped_job_status = JobStatus.ERROR
```

daka1510 (changes requested):

daka1510:
@kt474 please rerequest a review after you made changes (or tag me know if you have objections). Otherwise I might miss your review request :).

qiskit_ibm_runtime/runtime_job.py
kt474: How would this work if there are different reasons for different statuses in the future? Wouldn't we have to add more if statements?

daka1510: Yes, if we stick with the existing pattern to map the response from the API to a Qiskit `JobStatus` enum, we have to maintain some sort of mapping code. Ideally this mapping does not depend on a reason string but I don't know if that's feasible. My proposal was meant to map the existing code in a backward-compatible way.

kt474: I made some updates - I think getting the [error message](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/constants.py#L29) from the job response will still require the reason string unless we also want to refactor that logic

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -320,12 +323,34 @@ def _set_error_message(self, job_response: Dict) -> None:
...
+                error_msg = (
+                    job_response["state"]["status"]
+                    + " - "
+                    + job_response["state"]["reason"]
+                )
+            else:
+                error_msg = job_response["state"]["status"]
+            self._error_message = API_TO_JOB_ERROR_MESSAGE[error_msg.upper()].format(
+                self.job_id, job_result_raw
+            )
```
daka1510, line 336: I'm personally not a fan of using string concatenation here as well. It appears to be done only to retrofit the "new" response format on our "old" datastructure.

Why not use the same approach to compute the error messages that you now use to compute the status?

e.g. 

```
self._error_message = _error_message_from_job_response(...)
```

Maybe a simple mapping table is no longer the best approach...

test/unit/mock/fake_runtime_client.py
```diff
@@ -117,7 +117,7 @@ def __init__(
     ):
         """Initialize a fake job."""
         self._job_id = job_id
-        self._status = final_status or "QUEUED"
+        self._state = {"status": final_status or "QUEUED"}
```
daka1510, line 120: Why do you set `self._state` in the fake job and other instance variables in the "real" job?
```
self._status = JobStatus.INITIALIZING
self._reason: Optional[str] = None
```

daka1510: Shouldn't it be in sync?

daka1510 (changes requested):

qiskit_ibm_runtime/runtime_job.py
kt474: I agree, it's best to not use the `API_TO_JOB_ERROR_MESSAGE` constant and just handle the error msg in a separate method

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -320,12 +323,42 @@ def _set_error_message(self, job_response: Dict) -> None:
         """
         if self._status == JobStatus.ERROR:
             job_result_raw = self._api_client.job_results(job_id=self.job_id)
-            self._error_message = API_TO_JOB_ERROR_MESSAGE[
-                job_response["status"].upper()
-            ].format(self.job_id, job_result_raw)
+            self._error_message = self._error_msg_from_job_response(
+                job_response
+            ).format(self.job_id, job_result_raw)
```
daka1510, line 328: I would find it cleaner to remove the `.format(...)` part and include this in the `_error_msg_from_job_response` function. Previously where we relied on a simple mapping this was not possible but the dedicated function you introduced allows for that flexibility.

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -320,12 +323,42 @@ def _set_error_message(self, job_response: Dict) -> None:
...
+        """Returns the error message from an API response
+
+        Args:
+            response: Job response from the runtime API.
+
+        Returns:
+            Error message.
+        """
+        status = response["state"]["status"].upper()
+        if status == "CANCELLED" and self._reason == "Ran too long":
```
daka1510, line 342: Previously we used a case insensitive check. Do you think this important or can we rely on the API to use this exact capitalization?

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -320,12 +323,42 @@ def _set_error_message(self, job_response: Dict) -> None:
         """
         if self._status == JobStatus.ERROR:
             job_result_raw = self._api_client.job_results(job_id=self.job_id)
-            self._error_message = API_TO_JOB_ERROR_MESSAGE[
```
daka1510, line 323: Do we still need the `API_TO_JOB_ERROR_MESSAGE` mapping?

daka1510 (changes requested):
Thanks for following up, @kt474.  Three minor remarks/questions. Other than that it looks good.

qiskit_ibm_runtime/runtime_job.py
kt474: I think we'll still need it for places like [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/integration/test_job.py#L91) but then there won't be a single source of truth for these values so maybe it's better to still use it in `_error_msg_from_job_response` method 

```
        error_msg = API_TO_JOB_ERROR_MESSAGE["FAILED"]
        if status == "CANCELLED" and self._reason == "Ran too long":
            error_msg = API_TO_JOB_ERROR_MESSAGE["CANCELLED - RAN TOO LONG"]
        return error_msg.format(self.job_id, job_result_raw)
```

qiskit_ibm_runtime/runtime_job.py
kt474: To be safe, I think we can do a case insensitive check, the capitalization may be inconsistent in the future

daka1510 (approved):
LGTM

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -319,13 +323,40 @@ def _set_error_message(self, job_response: Dict) -> None:
...
+
+        Args:
+            response: Job response from the runtime API.
+
+        Returns:
+            Job status.
+        """
+        mapped_job_status = API_TO_JOB_STATUS[response["state"]["status"].upper()]
+        if mapped_job_status == JobStatus.CANCELLED and self._reason == "RAN TOO LONG":
+            mapped_job_status = JobStatus.ERROR
```
rathishcholarajan, line 357: Why is cancelled ran too long mapped to error? I think it should be mapped to Cancelled status.

rathishcholarajan (changes requested):

qiskit_ibm_runtime/runtime_job.py
rathishcholarajan: Hmm didn't realize that, thanks, fine with me.

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -319,13 +323,40 @@ def _set_error_message(self, job_response: Dict) -> None:
...
-            job_result_raw = self._api_client.job_results(job_id=self.job_id)
-            self._error_message = API_TO_JOB_ERROR_MESSAGE[
-                job_response["status"].upper()
-            ].format(self.job_id, job_result_raw)
+            self._error_message = self._error_msg_from_job_response(job_response)
         else:
             self._error_message = None
 
+    def _error_msg_from_job_response(self, response: Dict) -> str:
+        """Returns the error message from an API response
```
rathishcholarajan, line 331:
```suggestion
        """Returns the error message from an API response.
```

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -319,13 +323,40 @@ def _set_error_message(self, job_response: Dict) -> None:
...
+        """
+        status = response["state"]["status"].upper()
+        job_result_raw = self._api_client.job_results(job_id=self.job_id)
+        error_msg = API_TO_JOB_ERROR_MESSAGE["FAILED"]
+        if status == "CANCELLED" and self._reason == "RAN TOO LONG":
+            error_msg = API_TO_JOB_ERROR_MESSAGE["CANCELLED - RAN TOO LONG"]
+        return error_msg.format(self.job_id, job_result_raw)
+
+    def _status_from_job_response(self, response: Dict) -> str:
+        """Returns the job status from an API response
```
rathishcholarajan, line 347:
```suggestion
        """Returns the job status from an API response.
```

rathishcholarajan (changes requested):

rathishcholarajan (approved):

rathishcholarajan:
@kt474 Could you please check the integration test failures in your fork?

kt474:
> > I believe it was just an intermittent fail - passing now [here](https://github.com/kt474/qiskit-ibm-runtime/runs/5760661725?check_suite_focus=true)
> 
> Thanks how about cloud integration tests? https://github.com/kt474/qiskit-ibm-runtime/runs/5760661798?check_suite_focus=true

Was using the wrong cloud token, all passing now 
https://github.com/kt474/qiskit-ibm-runtime/runs/5764090893?check_suite_focus=true

Issue #177, opened 2022-03-01
Create nightly CI workflow to run unit tests against terra main branch


rathishcholarajan:
**What is the expected feature or enhancement?**
There used to be a CI workflow that runs all tests against qiskit-terra `main` branch. We should create a CI workflow that runs at least unit tests against qiskit-terra `main` branch to begin with so we can learn about breaking changes early.

**Acceptance criteria**
Unit tests should run nightly with terra `main` branch.

PR #193, opened 2022-03-07
Test integration with latest code of qiskit-terra on a nightly basis

daka1510:
### Summary
See title.

### Details and comments
Fixes https://github.com/Qiskit/qiskit-ibm-runtime/issues/177


rathishcholarajan (changes requested):
@daka1510 It is confusing to run unit tests in a file name integration tests. Can we add a separate workflow file for this and not combine this with anything? This has the potential to fail frequently since we are running against terra main.

Also could you check if you have to install Rust compiler to install terra from source. https://github.com/Qiskit/qiskit-terra/blob/main/CONTRIBUTING.md#installing-qiskit-terra-from-source
(Matthew posted recently in qiskit-dev channel about this change).

daka1510:
@rathishcholarajan - I updated my PR and triggered a sample test run. Installation from source succeeded without additional tweaks (see logs to verify).
https://github.com/daka1510/qiskit-ibm-runtime/runs/5463652831?check_suite_focus=true

I get the confusion of running unit tests in a file named integration tests. Conceptually I still think it's an integration test (with the primary purpose to tests the integration with another component, qiskit-terra). It's an implementation detail that we use the unit tests for that.

Overall, I think there are pros and cons for both approaches (separate workflow file vs. adding to the existing one). I followed your proposal to add a separate one now :). Please take a look.

rathishcholarajan (approved):
LGTM!

PR #194, opened 2022-03-07
Update wait for final state

kt474:
### Summary

If a job is already finished, calling `wait_for_final_state` will result in NoneType error. We should only use the websocket server to wait for the final states if the results are still streaming. 

### Details and comments
Fixes #


rathishcholarajan (changes requested):
I think this will fix it, could you add an integration test case that tests this scenario and also add a release note?

rathishcholarajan (approved):
LGTM!

PR #199, opened 2022-03-09
Integration tests for Sampler

rathishcholarajan:
### Summary
I don't know how to serialize ParameterView, so I am not passing parameters in the tests now.

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_ibm_sampler.py
+++ b/test/integration/test_ibm_sampler.py
@@ -45,21 +45,43 @@ class TestIntegrationIBMSampler(IBMIntegrationTestCase):
         with sampler_factory(circuits=[bell], parameters=[[]]) as sampler:
             self.assertIsInstance(sampler, BaseSampler)
 
-            result = sampler(circuit_indices=[0], parameter_values=[[]])
+            circuit_indices = [0]
+            result = sampler(circuit_indices=circuit_indices, parameter_values=[[]])
             self.assertIsInstance(result, SamplerResult)
+            print(result)
+            self.assertEqual(len(result.quasi_dists), len(circuit_indices))
+            self.assertEqual(len(result.metadata), len(circuit_indices))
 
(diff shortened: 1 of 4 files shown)


test/integration/test_ibm_sampler.py
```diff
@@ -0,0 +1,87 @@
...
+        pqc = RealAmplitudes(num_qubits=2, reps=2)
+        pqc.measure_all()
+        pqc2 = RealAmplitudes(num_qubits=2, reps=3)
+        pqc2.measure_all()
+
+        theta1 = [0, 1, 1, 2, 3, 5]
+        theta2 = [1, 2, 3, 4, 5, 6]
+        theta3 = [0, 1, 2, 3, 4, 5, 6, 7]
+
+        with sampler_factory(circuits=[pqc, pqc2]) as sampler:
```
rathishcholarajan, line 81: I am not passing parameters here yet since I don't know how to serialize ParameterView.

test/integration/test_ibm_sampler.py
```diff
@@ -0,0 +1,87 @@
...
+            self.assertIsInstance(sampler, BaseSampler)
+
+            result = sampler(
+                circuit_indices=[0, 0, 1], parameter_values=[theta1, theta2, theta3]
+            )
+            self.assertIsInstance(result, SamplerResult)
```
ajavadia, line 87: maybe assert that the returned result has 3 elements in it. better if we can assert that the quasi-probabilities match ideal simulation within some tolerance. 

Qiskit has a function under `qiskit.test.base` called `assertDictAlmostEqual` which may be useful here.

rathishcholarajan: I am checking the length now. Will open a separate PR to check the values with `assertDictAlmostEqual`.

Issue #147, opened 2022-02-15
Avoid temporary version inconsistencies between client and server


daka1510:
**Problem**
Today end-users hit the issue below when running a sampler program because a new version of `qiskit-terra` was released and it was picked up via `qiskit-ibm-runtime` from end users, **before** required server-side changes to support it were rolled out to production.

```
UserWarning: The qiskit version used to generate the provided QPY file, 0.19.2, is newer than the current qiskit version 0.19.1. This may result in an error if the QPY file uses instructions not present in this current qiskit version

...
File "/usr/local/lib/python3.8/site-packages/qiskit/circuit/qpy_serialization.py", line 1572, in load
2022-02-15T04:06:47.910338877-06:00     circuits.append(_read_circuit(file_obj, qpy_version))
2022-02-15T04:06:47.910338877-06:00   File "/usr/local/lib/python3.8/site-packages/qiskit/circuit/qpy_serialization.py", line 1609, in _read_circuit
2022-02-15T04:06:47.910338877-06:00     raise QiskitError("Duplicate register bits found")
2022-02-15T04:06:47.910338877-06:00 qiskit.exceptions.QiskitError: \'Duplicate register bits found\'\n'
```

**Acceptance Criteria**
- [ ] Agreed on corrective actions to ensure this cannot happen again
- [ ] Rolled out changes as needed

PR #202, opened 2022-03-10
Add Parameter serialization

rathishcholarajan:
### Summary
Parameter is a sub class of ParameterExpression and that's why Parameter was getting converted to ParameterExpression.

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/utils/json.py
+++ b/qiskit_ibm_runtime/utils/json.py
@@ -38,6 +38,7 @@ except ImportError:
 
 from qiskit.circuit import (
     Instruction,
+    Parameter,
     ParameterExpression,
     ParameterVector,
     QuantumCircuit,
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/utils/json.py
```diff
@@ -205,6 +206,13 @@ def default(self, obj: Any) -> Any:  # pylint: disable=arguments-differ
                 serializer=lambda buff, data: qpy_serialization.dump(data, buff),
             )
             return {"__type__": "QuantumCircuit", "__value__": value}
+        if isinstance(obj, Parameter):
+            value = _serialize_and_encode(
+                data=obj,
+                serializer=qpy_serialization._write_parameter,
```
mtreinish, line 212: This won't work starting in 0.20.0 as this private method has moved on `main`, we added aliases for the other private functions the provider was using previously to avoid a hard breakage, but this wasn't in the list and I don't think we want to add more use of private methods here. This raises https://github.com/Qiskit/qiskit-ibm-runtime/issues/147#issuecomment-1040736485 again and that's how we should be fixing this.

rathishcholarajan: Thanks Matthew! I will merge this PR right now to unblock testing primitives and then immediately start working on copying qpy module so we can have a copy here before 0.20.0.

PR #203, opened 2022-03-11
Update max_time.rst

renier:
Add other limitations to new limitations document.

Will help with frequent questions on slack on this subject.

This type of information used to be here https://github.com/Qiskit-Partners/qiskit-runtime/blob/main/docs/limitations.rst
But that repo has been archived.

### DIFF
--- a/docs/max_time.rst
+++ b/docs/max_time.rst
@@ -6,7 +6,7 @@ Maximum Execution Time
 
 To ensure fairness, there is a maximum execution time for each Qiskit Runtime job. If
 a job exceeds this time limit, it is forcibly cancelled. This is represented in the job
-status as `Canceled - Ran too long`. The maximum execution time is the
+status as `Cancelled - Ran too long`. The maximum execution time is the
 smaller of 1) the system limit and 2) the ``max_execution_time`` defined by the program.
 The system limit is defined below:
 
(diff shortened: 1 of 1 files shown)


docs/max_time.rst
```diff
@@ -5,7 +5,8 @@ Maximum Execution Time
 ======================
 
 To ensure fairness, there is a maximum execution time for each Qiskit Runtime job. If
-a job exceeds this time limit, it is forcibly terminated. The maximum execution time is the
+a job exceeds this time limit, it is forcibly canceled. This is represented in the job
+status as `Canceled - Ran too long`. The maximum execution time is the
```
daka1510, line 9: With https://github.com/Qiskit/qiskit-ibm-runtime/issues/168 we will move away from the deprecated job status string and instead use the new `state` property

```
"state": {
    "status": "Cancelled",
    "reason": "Ran too long"
}
```

I assume this change will be transparent for users and we can keep it as-is here...

daka1510 (approved):
LGTM.

docs/max_time.rst
renier: Yes, since the UI will continue to present it as one thing.

rathishcholarajan (changes requested):
Looks good mostly, suggested minor changes…

rathishcholarajan (approved):

PR #204, opened 2022-03-11
Make circuit_indices, observable_indices and parameter_values optional

rathishcholarajan:
### Details and comments
Fixes #


jyu00:
While we are at this, we can make the `__init__()` function match the implementation as well, e.g. 

```
class Estimator(BaseEstimator):
    """Estimator primitive."""
    def __init__(
        self,
        circuits: Union[QuantumCircuit, Iterable[QuantumCircuit]], 
        ...
    ):
```
This allows one to do `estimator(circuits=bell)` instead of `estimator(circuits=[bell])` (which incidentally is closer to Qiskit convention)

PR #207, opened 2022-03-14
[Don't merge] Add Primitives tutorials

rathishcholarajan:
### Summary
Add tutorials for Estimator and Sampler (IQX and Cloud)
(This PR is only for the reference of community team as they are writing nicer tutorials and shouldn't be merged as is)

### Details and comments
Adds a new Primitives section under Tutorials with Estimator and Sampler.
Screen Shot 2022-03-15 at 1 02 25 AM

PR #216, opened 2022-03-16
Update exisiting tutorials

rathishcholarajan:
### Summary
This PR updates existing tutorials to fix E2E runs. These updates were separated out from PR #207.

### Details and comments
Fixes #


docs/tutorials/01_introduction_cloud_runtime.ipynb
```diff
@@ -228,44 +228,48 @@
...
       "  Creation date: 2021-10-26T14:41:57Z\n",
-      "  Update date: 2022-02-24T22:17:10.144954Z\n",
+      "  Update date: 2022-03-14T17:12:06.298047Z\n",
       "  Max execution time: 10000\n",
       "  Backend requirements:\n",
       "    none\n",
       "  Input parameters:\n",
       "    Properties:\n",
+      "        - circuit_indices:\n",
+      "            Description: Indexes of the circuits to evaluate.\n",
```
kt474, line 238: Pretty minor but I'd suggest being consistent with the plural of "index"

kt474 (approved):

docs/tutorials/01_introduction_cloud_runtime.ipynb
daka1510: I agree, likely nothing to fix in this PR because that's just output returned from the server.

docs/tutorials/01_introduction_cloud_runtime.ipynb
```diff
@@ -373,14 +377,18 @@
...
+      "<IBMBackend('ibmq_qasm_simulator')>\n"
      ]
     }
    ],
    "source": [
     "# Specify the program inputs here.\n",
-    "program_inputs = {\"circuits\": qc}\n",
+    "program_inputs = {\n",
+    "    \"circuits\": qc,\n",
+    "    \"circuit_indices\": [0],\n",
```
daka1510, line 390: I don't really understand the value of this parameter and why it's mandatory for this use-case. But given that my context is limited and the decision appears to be fixed already, I won't block it here :)

daka1510 (approved):

docs/tutorials/01_introduction_cloud_runtime.ipynb
rathishcholarajan: Fixed the tutorial as well as opened a PR upstream to fix this.

docs/tutorials/01_introduction_cloud_runtime.ipynb
rathishcholarajan: We couldn't settle on a sensible default for circuit_indices when multiple circuits are passed, so this was made mandatory just to be consistent with estimator.

PR #218, opened 2022-03-17
hello-world program return "Hello, World!"

1ucian0:
I think the `hello-world` program should... well.. return `"Hello, World!"`. This is [kinda standard](https://en.wikipedia.org/wiki/%22Hello,_World!%22_program). I understand that, in this context, the program is promoted as a sanity check that everything is fine. Even something like `"Hello, World! All done"` is fine to me :)


rathishcholarajan (changes requested):
@1ucian0 Thank you! There are a few other occurrences in this repo that need to be updated. Would you be able to update those as well?

rathishcholarajan:
Thanks for bring this up @1ucian0! Once your upstream PR merges I will open a separate PR here to fix all occurrences.

1ucian0:
I changed the rest of the instances in 888f8f910013672fa53ad19e4916cfff11f06813

PR #250, opened 2022-03-30
Add QPY module

### LINKED ISSUE #147: Avoid temporary version inconsistencies between client and server

rathishcholarajan:
### Summary
Copy QPY module from qiskit-terra. (CI works fine with terra main, this should be merged only after qiskit-terra 0.20.0 is released.)

### Details and comments
Fixes #147


qiskit_ibm_runtime/qpy/__init__.py
```diff
@@ -0,0 +1,637 @@
...
+###########################################################
+QPY serialization (:mod:`qiskit.qpy`)
+###########################################################
+
+.. currentmodule:: qiskit.qpy
+
+*********
+Using QPY
+*********
+
```
mtreinish, line 23: I think in this module you'll want to make it clear that this is a downstream fork of the terra implementation (it'd be nice to preserve the git history if possible too) and set expectations around when and how to update and to pull updates from terra. That's particularly important for this as the reason to keep a local copy in the runtime client is to manage when the qpy version is bumped in sync with the service offering.

qiskit_ibm_runtime/qpy/__init__.py
```diff
@@ -0,0 +1,637 @@
...
+        double real;
+        double imag;
+    }
+
+this matches the internal C representation of Python's complex type. [#f3]_
+
+
+.. [#f1] https://tools.ietf.org/html/rfc1700
+.. [#f2] https://numpy.org/doc/stable/reference/generated/numpy.lib.format.html
+.. [#f3] https://docs.python.org/3/c-api/complex.html#c.Py_complex
```
mtreinish, line 623: You don't need any of this, the protocol format specification is part of terra, you can just link to that in https://qiskit.org/documentation/

```suggestion
```

mtreinish, line 623: You don't need any of this, the protocol format specification is part of terra, you can just link to that in https://qiskit.org/documentation/

```suggestion
```

mtreinish, line 623: You don't need any of this, the protocol format specification is part of terra, you can just link to that in https://qiskit.org/documentation/

```suggestion
```

mtreinish, line 623: You don't need any of this, the protocol format specification is part of terra, you can just link to that in https://qiskit.org/documentation/

(I tried to do a suggestion deleting it, but github internally timed out and returned a 502 trying to generate the comment).

daka1510 (commented):
I'm personally not a fan of this duplication for two reasons
1. This comes with an overhead to keep the different implementations in sync. As we've learned with https://github.com/Qiskit/qiskit-ibm-runtime/pull/226, there might be unexpected requirements that force us to change code that we thought can be copied across repositories to save efforts.
2. Core maintainers of `qiskit-ibm-runtime` are not familiar with the copied code that now falls into their responsibilities (without dedicated test cases this lack of knowledge introduces a risk to introduce bugs)

Not saying I won't approve but wonder if we have carefully weighed these drawbacks against other options. Wouldn't it make sense to introduce a new python package for core functionality that multiple projects in the qiskit organization depend on?  
As far as I understand this should allow us to overcome the timing issue that leads to the problems outlined in https://github.com/Qiskit/qiskit-ibm-runtime/issues/147.

mtreinish:
That core package is qiskit-terra. The problem is it moves at a different speed than server side deployment of that core package causing the sync issues.

mtreinish:
As for your specific point 2, the solution for that is you don't touch the qpy code here. Just treat it as a vendored dependency and any changes from it have to go through terra first. The point of vendoring it here is to match the qpy version used by the runtime client to something that is known compatible with the server side (which is using a fixed version of terra to deserialize it).

rathishcholarajan:
> I'm personally not a fan of this duplication for two reasons
> 
> 1. This comes with an overhead to keep the different implementations in sync. As we've learned with #226, there might be unexpected requirements that force us to change code that we thought can be copied across repositories to save efforts.
> 2. Core maintainers of `qiskit-ibm-runtime` are not familiar with the copied code that now falls into their responsibilities (without dedicated test cases this lack of knowledge introduces a risk to introduce bugs)
> 
> Not saying I won't approve but wonder if we have carefully weighed these drawbacks against other options. Wouldn't it make sense to introduce a new python package for core functionality that multiple projects in the qiskit organization depend on? As far as I understand this should allow us to overcome the timing issue that leads to the problems outlined in #147.

The short term problem is the provider is importing private functions from terra now and with terra 0.20 (due today) we won't be able to import a couple of those private functions. If we don't do this users will have serialization issues when using parameterized circuits with Sampler or Estimator. So we need to copy this qpy module from qiskit in order to fix this issue short term. That way we are no longer relying on private functions from terra which I think is an improvement.

The serialization related test cases that we currently have are passing when I run locally. So I believe that is good for now. In the next iteration I will copy over qpy related test cases from terra. I think like Matthew said the key is to not deviate too much in this fork and make any changes only through terra.

rathishcholarajan:
@mtreinish @daka1510 
As for a process to follow for making changes to this, does the following make sense?
1. Make any changes to qpy module in terra and terra releases the changes.
2. Copy those qpy module changes to qiskit-ibm-runtime and release qiskit-ibm-runtime.
4. ntc-provider picks up the latest qiskit-ibm-runtime once it's released. (ntc-provider is not yet using qiskit-ibm-runtime yet).

Please correct me if my understanding is wrong.

PR #251, opened 2022-03-30
Fix account migration code so that it updates accounts with non-default names

daka1510:
### Summary
Accounts with custom names are not updated by the current migration code because `hasattr({"auth": "whatever"}, "auth")` alway returns False.

### Details and comments
Follow-up for https://github.com/Qiskit/qiskit-ibm-runtime/pull/226


daka1510:
Found during manual tests. Will check why it was not covered by unit tests and update this PR accordingly in a second step (FYI: @rathishcholarajan).

qiskit_ibm_runtime/accounts/management.py
```diff
@@ -205,7 +205,7 @@ def migrate(cls) -> None:
                     overwrite=False,
                 )
             else:
-                if hasattr(value, "auth"):
```
daka1510, line 208: Missed to find this during my review. Background: https://stackoverflow.com/a/10724829

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -77,10 +77,8 @@ def to_saved_format(self) -> dict:
     def from_saved_format(cls, data: dict) -> "Account":
         """Creates an account instance from data saved on disk."""
         proxies = data.get("proxies")
-        auth = data.get("auth")
-        channel = "ibm_quantum" if auth == "legacy" else "ibm_cloud"
         return cls(
-            channel=data.get("channel", channel),
```
daka1510, line 83: Given is is guaranteed that `from_saved_format` is only called after `migrate`, we don't need this code.

It was hiding a bug in the migration code and prevented the test cases from failing.

qiskit_ibm_runtime/accounts/management.py
daka1510: Without this change we would not migrate named accounts. Thus we could never remove the "auth" support from our code.

qiskit_ibm_runtime/accounts/management.py
```diff
@@ -205,7 +205,7 @@ def migrate(cls) -> None:
...
+                if isinstance(value, dict) and "auth" in value:
```
daka1510, line 208: The `isinstance` check is just a sanity check that should not be needed for a proper configuration file.

test/account.py
```diff
@@ -155,7 +155,7 @@ def get_account_config_contents(
         }
     }
     if verify is not None:
-        out["verify"] = verify
+        out[name]["verify"] = verify
     if proxies is not None:
-        out["proxies"] = proxies
+        out[name]["proxies"] = proxies
```
daka1510, line 160: I don't think we ever supported "global" configuration values for the new account management approach. Setting up the configuration file in an invalid way from our test cases lead to test issues.

PR #261, opened 2022-04-04
Deprecate IBMRuntimeService, IBMEstimator, IBMSampler classes

### LINKED ISSUE #260: Rename IBMRuntimeService/IBMEstimator/IBMSampler classes
1. Rename IBMRuntimeService -> QiskitRuntimeService (Deprecate IBMRuntimeService class name)
2. Deprecate IBMEstimator/IBMSampler classes and have the users directly use the current Estimator/Sampler classes. (with only change of backend will now be passed under options parameter)

Old:
```python
from qiskit_ibm_runtime import IBMRuntimeService
service = IBMRuntimeService(channel="ibm_cloud", token="abc", instance="crn")

estimator_factory = IBMEstimator(service=service, backend="ibm_algiers")

with estimator_factory(circuits=[qc], observables="...", parameters="...") as estimator:
    # estimator here is instance of Estimator class
```

New:
```python
service = QiskitRuntimeService(channel="ibm_cloud", token="abc", instance="crn")

with Estimator(circuits=[qc], observables="...", parameters="...", options={ "backend": str | Backend }, service=service) as estimator:
    result = estimator(circuit_indices=[0], ...)
```

**Acceptance criteria**

rathishcholarajan:
### Summary
See [issue](https://github.com/Qiskit/qiskit-ibm-runtime/issues/260) description.

### Details and comments
Fixes #260

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -19,12 +19,108 @@ from qiskit.quantum_info import SparsePauliOp
 
 # TODO import BaseEstimator and EstimatorResult from terra once released
 from .qiskit.primitives import BaseEstimator, EstimatorResult
+from .exceptions import IBMInputValueError
+from .ibm_backend import IBMBackend
 from .ibm_runtime_service import QiskitRuntimeService
 from .runtime_session import RuntimeSession
 
 
 class Estimator(BaseEstimator):
-    """Estimator primitive."""
+    """Class for interacting with Qiskit Runtime Estimator primitive service.
+
+    Qiskit Runtime Estimator primitive service estimates expectation values of quantum circuits and
+    observables.
+
+    Estimator can be initialized with the following parameters.
+
+    * circuits: a (parameterized) :class:`~qiskit.circuit.QuantumCircuit` or
+        a list of (parameterized) :class:`~qiskit.circuit.QuantumCircuit`.
+
+    * observables: a list of :class:`~qiskit.quantum_info.SparsePauliOp`.
+
+    * parameters: a list of parameters of the quantum circuits.
+        (:class:`~qiskit.circuit.parametertable.ParameterView` or
+        a list of :class:`~qiskit.circuit.Parameter`) specifying the order
+        in which parameter values will be bound.
+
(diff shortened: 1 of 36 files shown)


qiskit_ibm_runtime/estimator.py
```diff
@@ -19,21 +19,117 @@
...
+    Qiskit Runtime Estimator primitive service estimates expectation values of quantum circuits and
+    observables.
+
+    Estimator can be initialized with the following parameters.
+
+    * circuits: a (parameterized) :class:`~qiskit.circuit.QuantumCircuit` or
+        a list of (parameterized) :class:`~qiskit.circuit.QuantumCircuit`.
+
+    * observables: a list of :class:`~qiskit.quantum_info.SparsePauliOp`.
+
```
daka1510, line 40: The source of the changes in this file (and the relation to the scope of the PR) is not clear to me. Is that copied from [qiskit-terra](https://github.com/Qiskit/qiskit-terra/tree/main/qiskit/primitives)

test/integration/test_estimator.py
```diff
@@ -0,0 +1,88 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Integration tests for Estimator primitive."""
```
daka1510, line 13: I'm not sure about the difference between `test_estimator` and `test_ibm_estimator` (likely due to lack of context on my side). Why do we need both?

qiskit_ibm_runtime/estimator.py
rathishcholarajan: Estimator becomes the primary interface for users instead of IBMEstimator. Issue description has more details.

rathishcholarajan: This is copied from IBMEstimator class.

PR #265, opened 2022-04-06
Fix failing integration tests

rathishcholarajan:
### Summary
Adding check for `ibm_quantum` since `instance` is only supported in `ibm_quantum` for filtering backends.

### Details and comments
Fixes a couple of failing integration tests https://github.com/Qiskit/qiskit-ibm-runtime/runs/5844881493?check_suite_focus=true#step:5:1520

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -1303,11 +1303,8 @@ class QiskitRuntimeService:
             hub = raw_data.get("hub")
             group = raw_data.get("group")
             project = raw_data.get("project")
-            instance = (
-                to_instance_format(hub, group, project)
-                if all([hub, group, project])
-                else None
-            )
+            if all([hub, group, project]):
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1298,14 +1298,16 @@ def _decode_job(self, raw_data: Dict) -> RuntimeJob:
...
+        instance = None
+        if self._channel == "ibm_quantum":
+            hub = raw_data.get("hub")
+            group = raw_data.get("group")
+            project = raw_data.get("project")
+            instance = (
+                to_instance_format(hub, group, project)
+                if all([hub, group, project])
+                else None
+            )
```
kt474, line 1310:
```suggestion
            if all([hub, group, project]):
                instance = to_instance_format(hub, group, project)
```

Since `instance` is already `None` it might be cleaner to do it this way

kt474 (approved):

PR #326, opened 2022-05-16
Project name update

marcolincs:
Qiskit Runtime IBM Quantum Client -> Qiskit Runtime IBM Client

### Summary

Updating the project name for naming alignment. No impact on end user.


rathishcholarajan (approved):
LGTM! Thanks @marcolincs! 

For more context to readers on why this change was done:
There are two channels as of today, "IBM Quantum" and "IBM Cloud" through which Qiskit Runtime can be accessed. So having "IBM Quantum" in the title was confusing and meant like it was referring to just one of the channels, but that is not the case.

PR #348, opened 2022-06-01
Add service property to IBMBackend class

### LINKED ISSUE #300: Runtime backends should have their service as an attribute / method

In many cases external code is wrapping calls to things like `backend.run()` that need to be replaced with `Sampler`.  However doing so, and maintaining compatibility is difficult.  Namely, I must change the external API to take a service instance as well.  Instead, it would be nice if the backend instance itself could be used to set the service from which it originated.  In that case the external API can remain unchanged.

**Acceptance criteria**

rathishcholarajan:
### Summary
The ``service`` object which is an instance of `QiskitRuntimeService`
class can now be accessed from `IBMBackend` class using the ``service``
property.

Ex:
```python
backend = service.get_backend("ibmq_qasm_simulator")
backend.service  # QiskitRuntimeService instance used to instantiate the backend
```

### Details and comments
Fixes #300

### DIFF
--- a/qiskit_ibm_runtime/hub_group_project.py
+++ b/qiskit_ibm_runtime/hub_group_project.py
@@ -18,6 +18,7 @@ from typing import Any, Dict, Optional
 
 from qiskit_ibm_runtime import (  # pylint: disable=unused-import
     ibm_backend,
+    qiskit_runtime_service,
 )
 
 from .api.clients import AccountClient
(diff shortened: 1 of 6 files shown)


rathishcholarajan:
FYI @tmittal947

kt474 (approved):

nonhermitian (approved):
This is very helpful to have.  Makes integrating code that wraps backends, e.g our own `QuantumInstance` much easier to do in a backward compatible manner.  Many thanks!

Issue #358, opened 2022-06-07
Remove deprecated IBMEstimator class


rathishcholarajan:
**What is the expected feature or enhancement?**
The IBMEstimator class was deprecated in favor of the Estimator class. It can be now be removed.

Note: IBMSampler was removed in this PR https://github.com/Qiskit/qiskit-ibm-runtime/pull/342 and can be used as an example.

bopardikarsoham:
Hello @rathishcholarajan . Can you please assign me to these issues #359 and #364 as well for the future. I know that these issues are currently on hold due to the Qiskit version policy but I surely would love to solve these issues as well.

rathishcholarajan:
@bopardikarsoham I am not able to assign external collaborators. Thank you very much. You could create the PRs and we'll leave it on hold until 0.6.0.

rathishcholarajan:
Fixed by https://github.com/Qiskit/qiskit-ibm-runtime/pull/363

PR #342, opened 2022-05-29
(fix): deprecate class qiskit_ibm_runtime.IBMSampler for the next release

dtquai:
### Summary

Class  qiskit_ibm_runtime.IBMSampler has been deprecated. This PR fixes IBMSampler's deprecation for the next release, discards _init_.py import for IBMSampler class, and changes 0.4.0 release notes for notify.

### Details and comments

Deprecation of IBMSampler class is completed.

PR #363, opened 2022-06-08
Removed deprecated qiskit_ibm_runtime.IBMEstimator class

bopardikarsoham:
I have removed the deprecated IBMEstimator class as per Issue #358.


mtreinish (changes requested):
Per the qiskit deprecation policy this is too early to remove the deprecated functionality See 3 and 3a on: https://qiskit.org/documentation/contributing_to_qiskit.html#deprecation-policy

mtreinish:
It's not a deprecation if you remove the thing < 7 days after the PR initially deprecating and both are included in the same release. That's just a breaking API change. The deprecation policy is there so that users can rely on the code we're shipping and get fair warning before we make changes. Even if you didn't start promoting the library until 0.4.0 that was > 1 month ago and you don't know how many people have started using it already and could be relying on the deprecated class. If you want people to start using the library in earnest and relying on it you need to provide a migration path for changes like this, if you just break things without warning nobody will want to use this as they can't rely on anything in the library (that being said #342 also was done incorrectly and is a breaking API change too).

Doing the deprecation properly is low overhead, you just have to keep the class working as is for another couple of release and at least 3 months. That will give any existing users sufficient time to migrate to whatever the replacement is.

rathishcholarajan:
> It's not a deprecation if you remove the thing < 7 days after the PR initially deprecating and both are included in the same release. That's just a breaking API change.

Not sure what you are referring to here, we are not deprecating and removing anything in the same release. We are removing something in 0.5.0 three months after it was deprecated in 0.4.0.

But I understand the other pieces of your concern, and will play nice by 3a.

rathishcholarajan:
> If you want people to start using the library in earnest and relying on it you need to provide a migration path for changes like this, if you just break things without warning nobody will want to use this as they can't rely on anything in the library (that being said #342 also was done incorrectly and is a breaking API change too).

Just to be clear, we are not breaking anything without warning. We have already deprecated in 0.4.0 and have provided warning. We are removing it in 0.5.0 which is planned for three months after. But I understand 3a and we can skip a release and remove this in 0.6.0.

rathishcholarajan (approved):
LGTM! Thanks @bopardikarsoham! Now that 0.5.0 is released, we can include this in 0.6.0 planned for July.

PR #368, opened 2022-06-09
Primitives interface for error resilience and transpilation settings

### LINKED ISSUE #346: Interface for error mitigation
TBD

**Acceptance criteria**

rathishcholarajan:
### Summary
First pass at the primitives interface for error resilience and transpilation settings.

TODO:
- [ ] Do client side validation of passed settings
- [ ] Write tests

### Details and comments
Fixes #346

### DIFF
--- a/qiskit_ibm_runtime/sampler.py
+++ b/qiskit_ibm_runtime/sampler.py
@@ -114,7 +114,9 @@ class Sampler(BaseSampler):
             print(result)
     """
 
-    @deprecate_arguments({"skip_transpilation": "transpilation_settings.skip_transpilation"})
+    @deprecate_arguments(
+        {"skip_transpilation": "transpilation_settings.skip_transpilation"}
+    )
     def __init__(
         self,
         circuits: Union[QuantumCircuit, Iterable[QuantumCircuit]],
--- a/qiskit_ibm_runtime/sampler.py
+++ b/qiskit_ibm_runtime/sampler.py
@@ -15,6 +15,7 @@
 from typing import Dict, Iterable, Optional, Sequence, Any, Union
 
 from qiskit.circuit import QuantumCircuit, Parameter
+from qiskit.utils.deprecation import deprecate_arguments
 
 # TODO import BaseSampler and SamplerResult from terra once released
 from .qiskit.primitives import BaseSampler, SamplerResult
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/sampler.py
```diff
@@ -146,6 +153,40 @@ def __init__(
...
+                * optimization_settings:
+
+                    * level: How much optimization to perform on the circuits.
+                        Higher levels generate more optimized circuits,
+                        at the expense of longer transpilation times.
+                        * 0: no optimization
+                        * 1: light optimization
+                        * 2: heavy optimization
+                        * 3: even heavier optimization
+                        If ``None``, level 1 will be chosen as default.
```
nonhermitian, line 176: So this default `None` -> `1` is what Qiskit does, but it is not the optimal setting.  The primitives, in my mind at least, should by default give the user the best results on average that we know how provided that doing so is not overly expensive.  In this case `optimization_level=3` is by far a better default that level 1, and is not much more expensive than the former.  This is also important because most users never adjust any settings.

qiskit_ibm_runtime/sampler.py
```diff
@@ -146,6 +153,40 @@ def __init__(
...
+                leveraging the state of the art error suppression, mitigation and correction techniques.
+
+                * level: How much resilience to build against errors.
+                    Higher levels generate more accurate results,
+                    at the expense of longer processing times.
+                    * 0: no resilience
+                    * 1: light resilience
+                    * 2: heavy resilience
+                    * 3: even heavier resilience
+                    If ``None``, level 1 will be chosen as default.
```
nonhermitian, line 188: So for the sampler this should probably not be set by default.  The reason is because then the returned results can contain negative values.  I can not use these values in many places where a true probably distribution can.  One would have to consume them for expectation values, or map onto the closet probability distribution.    Most our users are not accustom to having to do this.  Additionally there is no good way to do the expectation values from the quasi-distributions in Qiskit at present.  So taken together, the default should probably be 0 here.

nonhermitian (commented):
Overall looks fine to me.  Just a few comments on default settings that would make usability better (in my opinion).

qiskit_ibm_runtime/sampler.py
rathishcholarajan: Fixed in https://github.com/Qiskit/qiskit-ibm-runtime/pull/368/commits/c2227151eab657431f7a64c03f5cdb5930ff3328

PR #370, opened 2022-06-13
Use qiskitrc credentials if default config doesn't exist 

### LINKED ISSUE #301: Runtime should load IQX credentials from Qiskit if available
The two different accounts for the provider and runtime make it difficult to work with code in a backward compatible manner.  The IQX account information used by Runtime (i.e the token) is the same in both cases.  As such, Runtime should be able to init a service instance from the Qsikit config credentials if none are present in the Runtime json

**Acceptance criteria**

kt474:
### Summary

If the `qiskit-ibm.json` config file doesn't exist or doesn't have any valid accounts, credentials will be copied over from a `qiskitrc` file, given that it exists. 

### Details and comments
Fixes #301


qiskit_ibm_runtime/accounts/storage.py
```diff
@@ -40,6 +42,26 @@ def save_config(filename: str, name: str, config: dict, overwrite: bool) -> None
...
+        name="default-ibm-quantum",
+        overwrite=True,
+        config=Account(
+            token=account_data["token"],
+            url=account_data["url"],
+            channel="ibm_quantum",
+        )
+        .validate()
+        .to_saved_format(),
+    )
```
rathishcholarajan, line 62: Saving should not be done from this storage file, read from the file here and then return the read data and then from management file call save_config to save the data.

rathishcholarajan (changes requested):

qiskit_ibm_runtime/accounts/management.py
```diff
@@ -163,6 +164,25 @@ def get(
             if account_name in all_config:
                 return Account.from_saved_format(all_config[account_name])
 
+        if os.path.isfile(_QISKITRC_CONFIG_FILE):
+            qiskitrc_data = read_qiskitrc(_QISKITRC_CONFIG_FILE)
+            save_config(
+                filename=_DEFAULT_ACCOUNT_CONFIG_JSON_FILE,
+                name="default-ibm-quantum",
```
rathishcholarajan, line 171:
```suggestion
                name=_DEFAULT_ACCOUNT_NAME_IBM_QUANTUM,
```

qiskit_ibm_runtime/accounts/management.py
```diff
@@ -163,6 +164,25 @@ def get(
...
+                overwrite=True,
+                config=Account(
+                    token=qiskitrc_data["token"],
+                    url=qiskitrc_data["url"],
+                    channel="ibm_quantum",
```
rathishcholarajan, line 176: @kt474 what about provider/instance? we need to copy over that too...

rathishcholarajan (changes requested):

rathishcholarajan (changes requested):
@kt474 Looks good so far, just remembered that we need to copy these kwargs as well, proxies and verify https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/ibmqfactory.py#L240

qiskit_ibm_runtime/accounts/management.py
```diff
@@ -163,6 +165,33 @@ def get(
...
+            save_config(
+                filename=_DEFAULT_ACCOUNT_CONFIG_JSON_FILE,
+                name=_DEFAULT_ACCOUNT_NAME_IBM_QUANTUM,
+                overwrite=True,
+                config=Account(
+                    token=qiskitrc_data.get("token", None),
+                    url=qiskitrc_data.get("url", None),
+                    instance=qiskitrc_data.get("default_provider", None),
+                    verify=bool(qiskitrc_data.get("verify", None)),
+                    proxies=proxies,
```
kt474, line 184: `verify` is True by default in the qiskitrc and if initialized as False, there will be no entry in the qiskitrc file 

`proxies` needs to be converted to a dict first - it's a bit awkward to use `json.loads` here because the str returned has the incorrect format eg: 

```{'https': 'localhost:8080'}```  instead of ```'{"https": "localhost:8080"}'```

File: qiskit_ibm_runtime/accounts/management.py [hunk 5]
```diff
@@ -163,6 +165,33 @@ def get(
...
+            qiskitrc_data = read_qiskitrc(_QISKITRC_CONFIG_FILE)
+            proxies = (
+                ProxyConfiguration(ast.literal_eval(qiskitrc_data["proxies"]))
+                if "proxies" in qiskitrc_data
+                else None
+            )
+            save_config(
+                filename=_DEFAULT_ACCOUNT_CONFIG_JSON_FILE,
+                name=_DEFAULT_ACCOUNT_NAME_IBM_QUANTUM,
+                overwrite=True,
```
rathishcholarajan, line 178: I think we should remove this overwrite, just to be careful not to overwrite existing qiskit-ibm.json in any scenario.

rathishcholarajan (changes requested):

qiskit_ibm_runtime/accounts/management.py
kt474: we can just set `overwrite=False` since it's a required parameter

rathishcholarajan (approved):
LGTM! Good job @kt474!

PR #405, opened 2022-07-01
Fix custom_max_execution_time test 

kt474:
### Summary

Job no longer queues if an invalid `max_execution_time` is given 

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_job.py
+++ b/test/integration/test_job.py
@@ -127,20 +127,10 @@ class TestIntegrationJob(IBMIntegrationJobTestCase):
             service, max_execution_time=max_execution_time
         )
         inputs = {"iterations": 1, "sleep_per_iteration": 60}
-        job = self._run_program(
-            service, program_id=program_id, inputs=inputs, max_execution_time=1
-        )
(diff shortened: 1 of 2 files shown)


test/integration/test_job.py
```diff
@@ -122,25 +122,20 @@ def test_run_program_failed_ran_too_long(self, service):
...
-                job.job_id, job_result_raw
-            ),
-            job.error_message(),
+        self.assertEqual(
+            job._api_client.job_get(job.job_id)["cost"], job_max_execution_time
         )
-        with self.assertRaises(RuntimeJobFailureError):
-            job.result()
+        with self.assertRaises(IBMRuntimeError):
+            self._run_program(service, program_id=program_id, max_execution_time=1)
```
rathishcholarajan, line 138: I think it's better if we split this into 2 separate tests, one to check if the job cost is set and the other to test if it fails when atleast a minimum of 300 is not given...

rathishcholarajan (approved):
LGTM! Thanks @kt474!

PR #420, opened 2022-07-14
Support job metrics endpoint

### LINKED ISSUE #224: User should be able to get the time taken for a runtime job programatically.
There should be a way to recover the full elapsed time a job takes from the job and/or result instance.

**Acceptance criteria**
And user sees the total elapsed time, i.e the time the user would be billed for on the IBM Cloud

kt474:
### Summary

Added a new method, `job.metadata()` that returns the data from `/jobs/{id}/metrics`

### Details and comments
Fixes #224


rathishcholarajan:
@kt474 I don't really like how some things are named in the returned dictionary from the API, like for example what does bss mean? If someone uses this then they might ask. It'd be better if the naming was self explanatory. So I was hoping you would talk to @tmittal947 and @renier about this before working on it. So I am not comfortable merging this right now. But if @renier or @kwoane think this is good enough user experience for now, I'll let them approve and merge it.

renier:
> @kt474 I don't really like how some things are named in the returned dictionary from the API, like for example what does bss mean? If someone uses this then they might ask. It'd be better if the naming was self explanatory. So I was hoping you would talk to @tmittal947 and @renier about this before working on it. So I am not comfortable merging this right now. But if @renier or @kwoane think this is good enough user experience for now, I'll let them approve and merge it.

Based on our last conversation, we are just going to pass that through to the user as a dict returned from `.metadata()` as they are named. Then we sit and wait for questions to come, instead of trying to predict what they'll be after a first pass, to then change the method names with a second pass to address those questions.

renier (approved):

PR #423, opened 2022-07-14
Add support for testing SDK against experimental development cluster

daka1510:
### Summary
Allow developers to run the SDK against internal development clusters like so:
```
# Option 1
service1 = QiskitRuntimeService(
    url='https://api-ntc-....experimental-us-....us-east.containers.appdomain.cloud',
    channel='ibm_cloud',
    token='...',
    instance='crn:v1:staging:public:quantum-computing:us-east:a/...',
    verify=False
)
service1.backends()

# Option 2
service2 = QiskitRuntimeService.save_account(
    name='experimental'
    url='https://api-ntc-....experimental-us-....us-east.containers.appdomain.cloud',
    channel='ibm_cloud',
    token='...',
    instance='crn:v1:staging:public:quantum-computing:us-east:a/...',
    verify=False
)
service2 = QiskitRuntimeService(name='experimental')
service2.backends()
``` 

In this case, the resolved runtime url should not be updated like in the regular flow where the location identifier from the CRN is merged into the provided url.

### Details and comments
This change should be transparent for most users except for internal developers working on Qiskit runtime. For that reason I did not add a release note either.

### DIFF
--- a/qiskit_ibm_runtime/utils/utils.py
+++ b/qiskit_ibm_runtime/utils/utils.py
@@ -108,7 +108,7 @@ def get_runtime_api_base_url(url: str, instance: str) -> str:
     api_host = url
 
     # cloud: compute runtime API URL based on crn and URL
-    if is_crn(instance):
+    if is_crn(instance) and not _is_experimental_runtime_url(url):
         parsed_url = urlparse(url)
         api_host = (
             f"{parsed_url.scheme}://{_location_from_crn(instance)}"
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/utils/utils.py
```diff
@@ -118,6 +118,16 @@ def get_runtime_api_base_url(url: str, instance: str) -> str:
...
 
 
+def _is_experimental_runtime_url(url: str):
+    """Checks if the provided url points to an experimental runtime cluster. This type of URLs is used for internal
+    development purposes only.
+
+    Args:
+        url: The URL.
+    """
+    return isinstance(url, str) and "experimental" in url and url.endswith(".cloud")
```
daka1510, line 128: I don't like the dependency on the internal cluster name here but cannot think of a better solution that does not require a bigger rework in the library and changes to the existing public interface.

PR #431, opened 2022-07-19
Update VQE estimator tutorial to be compatible with the changes to qiskit optimizers

lockwo:
### Summary

Fixes the tutorial to no longer break with the latest version of qiskit, specially a small change in the optimizers were made so that change had to be incorporated here. @rathishcholarajan 

### Details and comments
Fixes #


lockwo:
Seems to be some sort of gateway timeout (not sure why or what that is from)

rathishcholarajan:
> Seems to be some sort of gateway timeout (not sure why or what that is from)

This has been happening past couple days, having problem reporting test coverage to coveralls.io, sometimes restarting failed builds work, I've restarted them.

PR #335, opened 2022-05-24
Raise RuntimeJobTimeoutError

rathishcholarajan:
### Summary
Raise `RuntimeJobTimeoutError` when `timeout` expires for `result()` or `wait_for_final_state()`.

### Details and comments
Fixes #316

PR #432, opened 2022-07-20
Raise RuntimeJobMaxTimeoutError if a program fails for running too long

kwoane:
### Summary

Raise RuntimeJobMaxTimeoutError if a program fails for running too long, so that programs can handle it.

Changes to handle an expired session error will be in a separate PR.

### Details and comments
Fixes #


rathishcholarajan (changes requested):
@kwoane Right now we use RuntimeJobTimeoutError https://github.com/Qiskit/qiskit-ibm-runtime/pull/335 when waiting for result times out. Do we want to distinguish this from that and use a different name here?

Also could you please add a release note using the command `reno new your-note-name`?

kwoane:
@rathishcholarajan  The name of this error made sense to me because the job timed out during its execution. I am open to suggestions for other names.

rathishcholarajan:
> @rathishcholarajan The name of this error made sense to me because the job timed out during its execution. I am open to suggestions for other names.

@kwoane Maybe something like ``RuntimeJobMaxTimeoutError``.

releasenotes/notes/exception-when-program-runs-too-long-cf80743eda83d601.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
+  - |
+    Raise an exception when a job runs for too long so that it can be handled appropriately by programs.
```
rathishcholarajan, line 4: @kwoane this description can be a bit more specific since it will go on the release notes page here https://qiskit.org/documentation/partners/qiskit_ibm_runtime/release_notes.html and people should not have to read the code changes to understand it.

Something like:
```suggestion
    Raise ``RuntimeJobMaxTimeoutError`` when a job runs for too long so that it can be handled appropriately by programs.
```

rathishcholarajan (changes requested):
@kwoane Almost looks good, one last thing is if you could follow these instructions https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/CONTRIBUTING.md#configuration to setup the ibm-cloud-production and ibm-quantum-production environment secrets like shown in this repo https://github.com/Qiskit/qiskit-ibm-runtime/settings/environments, in your fork, and also enable github actions in your fork, so we can see if the integration tests pass in your fork before merging PRs. You should be using your personal API tokens in your fork.

rathishcholarajan (approved):
LGTM! Thanks @kwoane!

PR #454, opened 2022-08-02
Convert `EstimatorResult.values` to ndarray if necessary

t-imamichi:
### Summary

This PR converts `ExpectationValues.results` into ndarray if necessary to be compatible with the [definition](https://github.com/Qiskit/qiskit-terra/blob/cf951202c77b49d28e7d23f43088b191638f282a/qiskit/primitives/estimator_result.py#L43).

### Details and comments

The job page of IQX dashboard displays binary data as expectation values.
image

I'm working on make the values as a list of float values. So, we need to convert them back to ndarray to construct `EstiamtorResult`.

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -14,6 +14,7 @@
 
 from typing import Iterable, Optional, Dict, Sequence, Any, Union
 
+import numpy as np
 from qiskit.circuit import QuantumCircuit, Parameter
 from qiskit.quantum_info import SparsePauliOp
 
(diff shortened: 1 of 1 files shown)


rathishcholarajan (changes requested):
@t-imamichi Looks good. Could you please add a release note since this is a breaking change?

t-imamichi:
This actually does not break API. Currently, the server returns ndarray as `values` and pass it to users.

I made an internal PR so that the server returns a list of float as `values` instead of ndarray, which allows the IQX dashboard to display the values.

This PR avoids API breakage by converting a list of float to ndarray. If the server returns ndarray (until the internal PR is merged), this PR just passes it to users.

rathishcholarajan (approved):

PR #457, opened 2022-08-04
Remove unnecessary qubit property conversions 

kt474:
### Summary

The `apply_prefix` method [here](https://github.com/Qiskit/qiskit-terra/blob/beea6f0f2de0aa8f0484fb14c5024f22c0b5b715/qiskit/providers/models/backendproperties.py#L200) in `BackendProperties` is already converting to seconds and hz

### Details and comments
Fixes #


rathishcholarajan (approved):
Thanks @kt474!

PR #461, opened 2022-08-11
Add backend.run()

### LINKED ISSUE #459: Support backend.run()
- Call `circuit-runner` program behind the scenes in the qiskit runtime api.
- Use undocumented param for program_id where the program can be overriden

kt474:
### Summary

```
backend = service.backend('ibmq_qasm_simulator')
backend.run(circuits = ...)
```

### Details and comments
Fixes #459

### DIFF
--- a/qiskit_ibm_runtime/ibm_backend.py
+++ b/qiskit_ibm_runtime/ibm_backend.py
@@ -14,9 +14,12 @@
 
 import logging
 
-from typing import Iterable, Union, Optional, Any, List
+from typing import Iterable, Union, Optional, Any, List, Callable, Type, Dict
 from datetime import datetime as python_datetime
 
+
+from qiskit.circuit import QuantumCircuit
+from qiskit.pulse import Schedule, LoConfig
 from qiskit.qobj.utils import MeasLevel, MeasReturnType
 from qiskit.providers.backend import BackendV2 as Backend
 from qiskit.providers.options import Options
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -475,11 +481,92 @@ def control_channel(self, qubits: Iterable[int]) -> List[ControlChannel]:
...
+        init_qubits: Optional[bool] = None,
+        use_measure_esp: Optional[bool] = None,
+        callback: Optional[Callable] = None,
+        result_decoder: Optional[Type[ResultDecoder]] = None,
+        instance: Optional[str] = None,
+        session_id: Optional[str] = None,
+        job_tags: Optional[List[str]] = None,
+        max_execution_time: Optional[int] = None,
+        start_session: Optional[bool] = None,
+    ) -> RuntimeJob:
```
kt474, line 509: Not sure if all of these parameters are required, I took all the ones from the old provider `backend.run` that work in runtime, left the rest out.

renier (approved):

taalexander:
My understanding was that we might expect this to be used for the `qasm3-runner` program ID as well, which may take separate program runtime arguments. Currently the arguments to `backend.run` are hardcoded as kwargs. We may need to instead accept a separate config data structure for program runtime arguments?

renier:
You should still be able to do this. The program id is overridable. @kt474 can we allow arbitrary kwargs?

kt474:
> You should still be able to do this. The program id is overridable. @kt474 can we allow arbitrary kwargs?

Yeah i'll create an issue and take care of it

Issue #462, opened 2022-08-12
Flexible session test and doc update


jyu00:
**What is the expected feature or enhancement?**

#453 introduced flexible session, but there are some follow up work not included in the PR:

- [x] More testing on the new interface
- [x] Refactoring old tests to use the new interface
- [x] Update docstrings (API reference)
- [x] Update tutorials to use the new interface
- [x] Update how-to doc
- [x] Some TODOs in the code, such as options that require serialization support
- [x] Release note

**Acceptance criteria**

jyu00:
This is done

PR #465, opened 2022-08-15
Update sampler/estimator tests 

kt474:
### Summary

Fixes all the failing integration tests 

### Details and comments
Related to #462

### DIFF
--- a/test/integration/test_sampler.py
+++ b/test/integration/test_sampler.py
@@ -120,11 +120,8 @@ class TestIntegrationIBMSampler(IBMIntegrationTestCase):
         circ.measure(0, 0)
 
         with Session(service) as session:
-            sampler = Sampler(
-                session=session,
-                options=self.options,
-                skip_transpilation=True,
-            )
+            sampler = Sampler(session=session, options=self.options)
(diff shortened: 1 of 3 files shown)


test/integration/test_sampler.py
```diff
@@ -119,10 +119,11 @@ def test_sampler_skip_transpile(self, service):
...
         circ.measure(0, 0)
 
-        with Session(service=service) as session:
-            sampler = session.sampler(
+        with Session(service) as session:
+            sampler = Sampler(
+                session=session,
                 options=self.options,
-                transpilation_settings={"skip_transpilation": True},
+                skip_transpilation=True,
```
jyu00, line 126: `skip_transpilation` is deprecated, so can we do something like

```
        with Session(service) as session:
            sampler = Sampler(session=session, options=self.options)
            sampler.options.transpilation.skip_transpilation = True
```

jyu00 (commented):
Thanks @kt474 for fixing these!

jyu00 (approved):
LGTM! Thanks for fixing this!

PR #466, opened 2022-08-16
Always update default session if session is closed

kt474:
### Summary

When using the sampler/estimator as a session, the default session should always be updated because the current one could be closed 

### Details and comments
Fixes #


qiskit_ibm_runtime/estimator.py
```diff
@@ -170,8 +170,7 @@ def __init__(
         if session:
             self._session = session
         else:
-            if session_pkg._DEFAULT_SESSION is None:
-                session_pkg._DEFAULT_SESSION = Session(service=service)
+            session_pkg._DEFAULT_SESSION = Session(service=service)
```
jyu00, line 173: The idea of `_DEFAULT_SESSION` is to reuse the same session if possible. So this should open a new one _only_ if one doesn't exist or if the existing is closed.

jyu00 (changes requested):

jyu00 (approved):
LGTM!

PR #468, opened 2022-08-16
Update API reference for flexible sessions

jyu00:
### Summary

- Update API reference to show flexible session and to promote the use of primitives + session while hiding the extra information (such as uploading a program) in drop downs.
- Fix the huge connection error that was mistakenly included in the turorial
- Add release note for flexible session 

### Details and comments
Related to #462


kt474 (approved):
Thanks for taking care of this @jyu00 - LGTM

Issue #351, opened 2022-06-02
Add support for setting max_execution_time at job level


rathishcholarajan:
**What is the expected feature or enhancement?**

The [POST /jobs API](https://us-east.quantum-computing.cloud.ibm.com/openapi#/Jobs/create_job) now supports setting max execution time for a program using the `cost` parameter.

`service.run` method should be updated to accept a `max_execution_time` parameter.

PR #484, opened 2022-08-18
[WIP] Adding cost estimate to backend.run circuit-runner jobs

### LINKED ISSUE #479: Add estimation for circuit-runner jobs
Now that we have `backend.run()` that will run circuit jobs with the `circuit-runner` program, we need to be able to estimate the time-cost of executing that circuit on the system (in the client) so that when the job is submitted. We can use the same method used in `qiskit-services`

kt474:
### Details and comments
Fixes #479

### DIFF
--- a/qiskit_ibm_runtime/api/rest/cloud_backend.py
+++ b/qiskit_ibm_runtime/api/rest/cloud_backend.py
@@ -72,6 +72,15 @@ class CloudBackend(RestAdapterBase):
         url = self.get_url("pulse_defaults")
         return self.session.get(url).json()
 
+    def cost_parameters(self) -> Dict[str, Any]:
+        """Return backend cost parameters.
+
+        Returns:
+            JSON response of cost parameters.
+        """
+        response = self.session.get(f"/backends/{self.backend_name}").json()
+        return response.get("costParameters")
+
     def status(self) -> Dict[str, Any]:
         """Return backend status.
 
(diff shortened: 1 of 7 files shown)


kt474:
@renier in #351 we set `max_execution_time` as the `cost` payload [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/api/rest/runtime.py#L176) - are `max_execution_time` and `cost` interchangeable here? Because then wouldn't the cost be way too low for the majority of cases (unless there is something wrong with my conversion)?

`IBMRuntimeError: 'Failed to run program: \'400 Client Error: Bad Request for url: https://runtime-us-east.quantum-computing.ibm.com/jobs. {"errors":[{"code":"bad_request","message":"Key: \\\'ProgramRunParams.Cost\\\' Error:Field validation for \\\'Cost\\\' failed on the \\\'min\\\' tag","more_info":"https://cloud.ibm.com/apidocs/quantum-computing#error-handling"}],"trace":"cbv94m76q7nnhok4ioo0"}\''`

renier:
Yes, they are interchangeable. For now, make the minimum 300 on the client side for the cost sent.

qiskit_ibm_runtime/api/rest/cloud_backend.py
```diff
@@ -72,6 +72,15 @@ def pulse_defaults(self) -> Dict[str, Any]:
         url = self.get_url("pulse_defaults")
         return self.session.get(url).json()
 
+    def cost_parameters(self) -> Dict[str, Any]:
+        """Return backend cost parameters.
+
+        Returns:
+            JSON response of cost parameters.
+        """
+        response = self.session.get(f"/backends/{self.backend_name}").json()
```
jyu00, line 81: `cloud_backend` uses Qiskit Runtime API. I don't think it has a `/backends/{backend name}` endpoint.

kt474:
We are removing `backend.run` support from qiskit-ibm-runtime so this is no longer needed

Issue #474, opened 2022-08-17
Backend option should be at session level, not job level


jyu00:
**Describe the bug**

The new `Options` class added `backend` as an option that one can specify when initializing a primitive class or running a primitive job. This implies that one can switch to using a different backend in the middle of a session, which is not currently supported by the scheduler. 

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

Make `backend` a session parameter and not part of `Options`. 

**Additional Information**

- **qiskit-ibm-runtime version**:
- **Python version**:
- **Operating system**:

nonhermitian:
Yes, and then this can be extended to a list, if needed in the future.

kt474:
How exactly do we want this done? Completely remove `backend` from `Options` and add it to `Session` instead? 

Example:
```
with Session(backend="ibmq_qasm_simulator") as session:
    sampler = Sampler(session=session)
```

cc @renier

jyu00:
Since `session` is an optional parameter to `Sampler` and `Estimator`, we'll need to add `backend` to them as well, so we know which backend to use if `session` is not specified. I'd like to avoid having the additional `service` parameter as well if possible. So perhaps `Sampler` and `Estimator` constructors can now take

- `session` (optional)
- `backend`, also optional, and can be either an `IBMBackend` instance or str. 
    - if a default session exists and the backend name matches, use that default session
    - otherwise close the default session and open a new one with the new backend
- if both `session` and `backend` are specified
   - error if they point to different backend
   - otherwise use the `session`

Also `RuntimeOptions` was deprecated because `backend` was moved to the job level. Since that's no longer true, we can potentially un-deprecate it. At the very least the deprecation message needs to be changed as `Options` is no longer a suitable alternative (since it won't have `backend`).

PR #485, opened 2022-08-18
Prepare release 0.7.0rc1

kt474:
### Summary

Move all reno to 0.7 folder 

### Details and comments
Fixes #


jyu00:
Can we wait for https://github.com/Qiskit/qiskit-ibm-runtime/issues/474 before doing the release? Since the fix will be a breaking change (which also means fix should wait for the workshop lol).

kt474:
> Can we wait for #474 before doing the release? Since the fix will be a breaking change (which also means fix should wait for the workshop lol).

Sure, I can work on that - fyi @renier

renier:
We can do multiple pre-releases as needed. Prereleases can break the interface still. Do we really need to wait for #474 ?

jyu00:
> We can do multiple pre-releases as needed

Sure, but it does take time (even if none significant). I don't see any benefit of doing a pre-release now, but it's up to you.

renier:
@kt474 we promised one this week, so let's not hold up.

renier (approved):

PR #498, opened 2022-08-22
Add deprecation warning for low usage programs 

### LINKED ISSUE #495: Add deprecation warning for low usage programs on IQX
We need to deprecate the following programs with a 7-day deprecation warning(both in the UX and programmatically):

QR-torch-train
QR-torch-infer
QR-sample-expval
QR-quantum-kernel-alignment

kt474:
### Details and comments
Fixes #495


renier (approved):

renier:
we'll need one like this in qiskit-ibm-provider

PR #510, opened 2022-08-30
Add job_tags and max_execution_time to options 

### LINKED ISSUE #473: Options should include job tags and max time

The current `Options` class is missing `job_tags` and `max_execution_time` that one can specify on `service.run()`. 

**Acceptance criteria**

kt474:
### Details and comments
Fixes #473


jyu00 (changes requested):

qiskit_ibm_runtime/options.py
```diff
@@ -104,6 +104,12 @@ class Options:
     """Options for the primitive programs.
 
     Args:
+        max_execution_time: Maximum execution time in seconds. This overrides
+            the max_execution_time of the program and cannot exceed it.
```
jyu00, line 108: I find this description a bit confusing. What does "overrides the max_execution_time of the program" mean? Does it update the `max_execution_time` metadata of a program? I don't think that's how this works?

qiskit_ibm_runtime/options.py
```diff
@@ -104,6 +104,12 @@ class Options:
...
+
+        job_tags: Tags to be assigned to the job. The tags can subsequently be used
+                as a filter in the :meth:`jobs()` function call.
```
jyu00, line 111:
```suggestion
                as a filter in the :meth:`qiskit_ibm_runtime.qiskit_runtime_service.jobs()` function call.
```

There is no `jobs()` method in this file so the reference link needs to be fully qualified.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -860,6 +860,7 @@ def run(
             inputs: Program input parameters. These input values are passed
                 to the runtime program.
             options: Runtime options that control the execution environment.
+                ``job_tags`` and ``max_execution_time`` can also be set here.
```
jyu00, line 863: I think your intent is to move `job_tags` and `max_execution_time` to be part of `RuntimeOptions` class. If that's the case, then 1) the use of them as separate input args should be deprecated and 2) docstring should list them as bullet points, similar to `backend`, `image`, and `log_level`.

kt474: Is this something we want to do? Deprecate `job_tags` and `max_execution_time` as input args and have them passed in through in `options` instead?

jyu00: Yeah since I think it's rather confusing to have some options inside `options` and some outside.

qiskit_ibm_runtime/options.py
```diff
@@ -191,6 +198,8 @@ class Options:
               Default: ``True``.
     """
 
+    max_execution_time: int = None
+    job_tags: List[str] = None
```
jyu00, line 202: I created a new `EnvironmentOptions` in my last PR as an attempt to organize less commonly used options (such as `image`) that is not related to transpilation or execution. I think `job_tags` can go there too, but `max_execution_time` can stay at level 1 as it's more likely to be used.

kt474:
closing in favor of #565

PR #514, opened 2022-08-31
Make backend a session parameter

### LINKED ISSUE #474: Backend option should be at session level, not job level

jyu00:
### Summary

- Make `backend` a session parameter instead of job parameter
- Create new default session if backend or service changes
- Un-deprecate `RuntimeOptions` since it no longer conflicts with `Options`
- Allow users to specify a backend as the `session` parameter to `Sampler` and `Estimator`, so `session` can continue to be optional

### Details and comments
Fixes #474


kt474 (approved):
LGTM, ran integration tests too - Thanks @jyu00!

PR #519, opened 2022-09-02
Add input data check for Sampler

t-imamichi:
### Summary

Add a validation check of circuits for Sampler. It raises an error if there is no classical bits in the input circuits.

I added the same test to Terra https://github.com/Qiskit/qiskit-terra/pull/8678 https://github.com/Qiskit/qiskit-terra/pull/8708
So, if we use the base class of primitives of Terra, we can remove these checks in qiskit-ibm-runtime.


qiskit_ibm_runtime/sampler.py
```diff
@@ -180,14 +181,25 @@ def run(
...
-                "The circuits parameter has to be instances of QuantumCircuit."
-            )
+        if isinstance(circuits, QuantumCircuit):
+            circuits = [circuits]
+
+        for circ in circuits:
+            if not isinstance(circ, QuantumCircuit):
+                raise ValueError(
+                    "The circuits parameter has to be instances of QuantumCircuit."
+                )
```
Cryoris, line 191: Minor comment, and I don't know what the standard in qiskit-ibm-provider is, but in terra we always assume the type hint is satisfied, so this check could be omitted since by the type hint only `QuantumCircuit`s are allowed 🙂

qiskit_ibm_runtime/sampler.py
```diff
@@ -180,14 +181,25 @@ def run(
...
+            if circ.num_clbits == 0:
+                raise ValueError("The circuits should have at least one classical bit.")
+            if self.options.resilience_level >= 1 and circ.num_clbits > len(
```
Cryoris, line 194: Is `>` enough to check or do we need to check `!=` for the lengths?

Cryoris (commented):
Two small comments, otherwise LGTM!

qiskit_ibm_runtime/sampler.py
t-imamichi: Yes, but I just left the original code as is.
https://github.com/Qiskit/qiskit-ibm-runtime/blob/af308caeb7c261a1fb1a7ca7a45c49f55df02215/qiskit_ibm_runtime/sampler.py#L186-L188

qiskit_ibm_runtime/sampler.py
t-imamichi: Yes, it's impossible to set more measurements than classical bits in the final measurements (duplicate measurements are removed).

t-imamichi: But, it's clear to use `!=`. So, I update it.

Cryoris (approved):
LGTM! For the Terra primitives we can essentially apply the same update, right?

t-imamichi:
Yeah, I will port this change to Terra next week.

t-imamichi:
I updated this PR and made the same checks to terra https://github.com/Qiskit/qiskit-terra/pull/8678.

t-imamichi:
Note that this check exists in Terra main branch's `BaseSampler`. So, if we plan to replace `BaseSampler` with Terra's, we don't need this PR.

test/integration/test_sampler.py
```diff
@@ -166,3 +166,21 @@ def test_sampler_primitive_as_session(self, service):
             self.assertIsInstance(result, SamplerResult)
             self.assertEqual(len(result.quasi_dists), len(circuits0))
             self.assertEqual(len(result.metadata), len(circuits0))
+
+    @run_integration_test
+    def test_sampler_raise_error(self, service):
```
jyu00, line 171: Can you move this test to `unit/test_ibm_primitives.py`? It doesn't need to be an integration test since the check is done locally and integration test is expensive. You can just mock the `Session`.

jyu00: Actually, presumably you have tests in terra, so the test here isn't really necessary as we'll eventually use terra code.

jyu00:
> Note that this check exists in Terra main branch's `BaseSampler`. So, if we plan to replace `BaseSampler` with Terra's, we don't need this PR.

We will use terra's `BaseSampler` eventually... but we need to terra to release first, or the two need to be released together (which is harder to coordinate...)

test/integration/test_sampler.py
t-imamichi: OK. I then remove the test.

t-imamichi:
Yes, it's hard to sync with primitives in terra because it has been actively updated. Maybe copying a snapshot sometimes is a compromise.

jyu00 (approved):

Issue #482, opened 2022-08-18
Use new base primitives


jyu00:
**What is the expected feature or enhancement?**

The base primitive classes `Sampler` and `Estimator` were updated in https://github.com/Qiskit/qiskit-terra/pull/8382 to support flexible session, and the child classes in this repo should be updated to use them. 

Note that one significant change is that the base class `run()` returns a qiskit-terra `Job` class. `qiskit-ibm-runtime` has its own `RuntimeJob` class, so it could be difficult integrating the two (e.g. `job_id()` vs `job_id`).

**Acceptance criteria**

PR #522, opened 2022-09-07
Use updated base primitives

jyu00:
### Summary

Update the base primitives to the (almost) latest version.

The main change was for `RuntimeJob` to inherit the `Job` class in qiskit-terra, which means
- `job_id` and `backend` need to be methods, not attributes. This was done by adding `__call__` to those attributes so both `job.job_id` and `job.job_id()` would return the same thing.
- rename `metadata()` to `metrics()`. Luckily the method was introduced after the last stable release, so it's not really a breaking change
- `wait_for_final_state()` has a different signature from the base class because QRT API doesn't support getting job status from ws streaming. Will open a separate issue on that.

### Details and comments
Related to #482. There are new changes coming to the base primitives in qiskit-terra, so I'll leave that issue open.


kt474 (approved):
LGTM Thanks @jyu00! - just a minor change with updating the job metrics integration test

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -951,6 +953,18 @@ def run(
         if not backend:
             backend = self.backend(name=response["backend"])
 
+        global _JOB_DEPRECATION_ISSUED  # pylint: disable=global-statement
+        if not (start_session or session_id) and not _JOB_DEPRECATION_ISSUED:
+            # No need to issue warning if session is used since the old style session
+            # didn't return a job, and new style returns new job.
+            _JOB_DEPRECATION_ISSUED = True
+            issue_deprecation_msg(
```
mriedem, line 961: This is confusing, see #633.

PR #455, opened 2022-08-02
Convert `SamplerResult.quasi_dists` to list of `QuasiDistribution`

t-imamichi:
### Summary
Convert `SamplerResult.quasi_dists` to list of `QuasiDistribution` to be compatible with the [definition](https://github.com/Qiskit/qiskit-terra/blob/cf951202c77b49d28e7d23f43088b191638f282a/qiskit/primitives/sampler_result.py#L42)

### Details and comments
Fixes #362

PR #524, opened 2022-09-08
Move conversion of QuasiDistribution to result decoder

jyu00:
### Summary

#455 converted `SamplerResult.quasi_dists` from dict to `QuasiDistribution` to match the `SamplerResult` description. That PR was written before we deprecated `_call` and moved to `_run`, so it only applied to the old path. It is also why the latest integration test failed.

This PR moved that code to `SamplerResultDecoder`, which does result processing when `job.result()` is called. 

@t-imamichi Are you ok if I use the old code in the old path - i.e. the result returned by `_call()` will be `SamplerResult` with a list of dict instead of `QuasiDistribution`? This way it's no longer a breaking change, and people using the new `run()` method will get the correct output (list of `QuasiDistribution`). 

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/sampler.py
+++ b/qiskit_ibm_runtime/sampler.py
@@ -263,24 +263,10 @@ class Sampler(BaseSampler):
             program_id=self._PROGRAM_ID,
             inputs=inputs,
             options=Options._get_runtime_options(combined),
-            result_decoder=SamplerResultDecoder,
         ).result()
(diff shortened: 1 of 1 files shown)


kt474 (approved):
LGTM, can wait for @t-imamichi's approval

t-imamichi:
Thanks. LGTM. We need to simplify the code when we remove the old path though.

PR #525, opened 2022-09-10
Use context manager session as default session

jyu00:
### Summary

If a session is open as a context manager, then that session should be the default session. In addition, if a primitive instance is created under the context but uses a different backend, it should not auto-close the cm session. For example:

```
with Session(service, backend="foo") as cm_session:
    s1 = Sampler(session="bar")  <---- this should not close cm_session
    s2 = Sampler()  <---- this should use cm_session
```

### Details and comments
Fixes #


kt474 (approved):
LGTM

PR #531, opened 2022-09-15
Update integration test to use simulator on staging

### LINKED ISSUE #521: Integration tests time out on staging
Integration tests have been timing out on IQX staging for almost a week - the issue seems to be with websockets and `test_backend_run`

https://github.com/Qiskit/qiskit-ibm-runtime/runs/8182374892?check_suite_focus=true

kt474:
### Summary

fixes #521

### Details and comments
On staging integration tests `test_backend_run` is timing out because the least busy backend chosen does not appear to support the `circuit-runner` program 

Update: turns out `backend1` (the least busy backend) is not configured correctly to handle runtime jobs so we should just use simulator on the appropriate staging integration tests.


jyu00:
> On staging integration tests `test_backend_run` is timing out because the least busy backend chosen does not appear to support the `circuit-runner` program

Do we know which backend that is? All backends should support `circuit-runner`. Maybe it's because the staging cluster is down?

kt474:
> > On staging integration tests `test_backend_run` is timing out because the least busy backend chosen does not appear to support the `circuit-runner` program
> 
> Do we know which backend that is? All backends should support `circuit-runner`. Maybe it's because the staging cluster is down?

I tested locally on `backend1` and the test case timed out but you're right it may just be because the staging cluster is down

jyu00 (approved):
LGTM!

PR #533, opened 2022-09-19
Allow kwargs in Options

### LINKED ISSUE #518: Accept arbitrary options for primitives

This new `Options` class allows auto-complete, but it's very rigid and only allows fixed attributes. It should instead accept arbitrary keywords so that users don't have to upgrade their Qiskit just because a new keyword becomes available in the primitive programs.  

**Acceptance criteria**

`Options` would accept arbitrary keyword values.

jyu00:
### Summary

Allow users to specify arbitrary kwargs in `Options`. This PR also creates a new `environment` section for `log_level`, `image`, and `instance`, as an attempt to reduce the number of first level options.  

### Details and comments
Fixes #518

In order for auto-complete work as before, `Options` remains a dataclass. The new `_flexible` decorator dynamically creates a new dataclass with the additional fields. The new dataclass also inherits the original, so any `isinstance` call would continue to work.


kt474 (approved):
LGTM!

PR #546, opened 2022-09-26
Update for flexible session with data caching

ChitsimranGill:
### Summary
Calculate circuit id by serializing and hashing the circuit key. Pass `circuits_map` and `circuit_ids` instead of `circuit_indices`.

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -275,7 +275,7 @@ class Estimator(BaseEstimator):
             **kwargs: Individual options to overwrite the default primitive options.
 
         Returns:
-            Submitted job.
+            Submitted job
         """
         circuits_map = {}
         circuit_ids = []
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -26,6 +26,7 @@ from qiskit.quantum_info.operators.base_operator import BaseOperator
 
 # TODO import BaseEstimator and EstimatorResult from terra once released
 from .qiskit.primitives import BaseEstimator, EstimatorResult
+from .qiskit.primitives.utils import _circuit_key
 from .qiskit_runtime_service import QiskitRuntimeService
 from .utils.estimator_result_decoder import EstimatorResultDecoder
 from .runtime_job import RuntimeJob
@@ -140,6 +141,13 @@ class Estimator(BaseEstimator):
             observables=observables,
             parameters=parameters,
         )
+
+        self._first_run = True
+        self._circuits_map = {}
+        if self.circuits:
+            for circuit in self.circuits:
+                circuit_id = str(hash(_circuit_key(circuit)))
+                self._circuits_map[circuit_id] = circuit
 
         if skip_transpilation:
             deprecate_arguments(
@@ -269,9 +277,23 @@ class Estimator(BaseEstimator):
         Returns:
             Submitted job.
         """
+        circuits_map = {}
+        circuit_ids = []
+        for circuit in circuits:
+            circuit_id = str(hash(_circuit_key(circuit)))
+            circuit_ids.append(circuit_id)
+            if circuit_id in self._circuits_map:
+                continue
+            self._circuits_map[circuit_id] = circuit
+            circuits_map[circuit_id] = circuit
+
+        if self._first_run:
+            self._first_run = False
+            circuits_map.update(self._circuits_map)
+
         inputs = {
-            "circuits": circuits,
-            "circuit_indices": list(range(len(circuits))),
+            "circuits": circuits_map,
+            "circuit_ids": circuit_ids,
             "observables": observables,
             "observable_indices": list(range(len(observables))),
             "parameters": parameters,
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -144,15 +144,6 @@ class Estimator(BaseEstimator):
             parameters=parameters,
         )
 
-        self._first_run = True
-        self._circuits_map = {}
-        if self.circuits:
-            for circuit in self.circuits:
-                circuit_id = _hash(
-                    json.dumps(_circuit_key(circuit), cls=RuntimeEncoder)
-                )
-                self._circuits_map[circuit_id] = circuit
-
         if skip_transpilation:
             deprecate_arguments(
                 "skip_transpilation",
@@ -201,6 +192,14 @@ class Estimator(BaseEstimator):
         else:
             backend = session or backend
             self._session = get_default_session(service, backend)
+
+        self._first_run = True
+        if self.circuits:
+            for circuit in self.circuits:
+                circuit_id = _hash(
+                    json.dumps(_circuit_key(circuit), cls=RuntimeEncoder)
+                )
+                self._session._circuits_map[circuit_id] = circuit
 
     def run(
         self,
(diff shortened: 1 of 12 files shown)


qiskit_ibm_runtime/qiskit/primitives/utils.py
```diff
@@ -25,3 +27,49 @@
...
+        tuple(  # circuit.data
+            (
+                _bits_key(data.qubits, circuit),  # qubits
+                _bits_key(data.clbits, circuit),  # clbits
+                data.operation.name,  # operation.name
+                tuple(data.operation.params),  # operation.params
+            )
+            for data in circuit.data
+        ),
+        None if circuit._op_start_times is None else tuple(circuit._op_start_times),
```
jyu00, line 68: @ikkoham Why did you use the private `_op_start_times` instead of `op_start_times`?

jyu00:
@ChitsimranGill Can you add some test cases? Both unit and integration. 

Also caching doesn't work with simulators, so you'll need to check that (which may be a bit tricky for cloud channel where backend can be omitted).

ChitsimranGill:
> @ChitsimranGill Can you add some test cases? Both unit and integration.
> 
> Also caching doesn't work with simulators, so you'll need to check that (which may be a bit tricky for cloud channel where backend can be omitted).

I think the existing integration tests are good enough, I will write unit tests for this tomorrow.

jyu00:
> I think the existing integration tests are good enough, I will write unit tests for this tomorrow.

Does any of the existing integration tests actually test data caching?

ChitsimranGill:
We added unit tests for data caching in sampler and estimator. We are not sure how to assert this in integration tests. Hopefully these unit tests will sufficient.

test/unit/test_estimator.py
```diff
@@ -0,0 +1,99 @@
...
+
+from .mock.fake_runtime_service import FakeRuntimeService
+
+
+class TestEstimator(IBMTestCase):
+    """Class for testing the Estimator class."""
+
+    @classmethod
+    def setUpClass(cls):
+        return super().setUpClass()
```
jyu00, line 36: You don't really need to overwrite `setUpClass` if you're just calling the parent's method.

test/unit/test_estimator.py
```diff
@@ -0,0 +1,99 @@
...
+        with Session(
+            service=FakeRuntimeService(channel="ibm_quantum", token="abc"),
+            backend="ibmq_qasm_simulator",
+        ) as session:
+            estimator = Estimator(session=session)
+
+            # calculate [ <psi1(theta1)|H1|psi1(theta1)> ]
+            with patch.object(estimator._session, "run") as mock_run:
+                estimator.run([psi1, psi2], [H1, H2], [theta1, theta2])
+                mock_run.assert_called_once_with(
```
jyu00, line 61: You can simplify this by quite a bit with something like

```
estimator = cls(session=MockSession)
estimator.run(...)
_, kwargs = session.run.call_args
inputs = kwargs["inputs"]
self.assertDictEqual(inputs["circuits", {psi1_id: psi1, psi2_id: psi2})
self.assertEqual(inputs["circuit_ids", [psi1_id, psi2_id])
```

jyu00 (commented):
I think there can be a few more unit tests. For example, in addition to 

```
estimator.run([psi1, psi2], [H1, H2], [theta1, theta2]) 
estimator.run([psi2], [H2], [theta2])
```

You can also do
```
estimator.run([psi3], [H3], [theta3])  # Pass in a new circuit in a later call
estimator.run([psi4, psi1], ...)  # A new circuit plus an old one
```

jyu00:
> We are not sure how to assert this in integration tests. Hopefully these unit tests will sufficient.

You can do something similar to what you did in unit test, e.g. 

```
with Session(...):
    sampler.run([circ1, circ2], ...)
    sampler.run([circ1], ...)
    sampler.run([circ3], ...)
    sampler.run([circ1, circ4], ...)
```
where all the circuits all have different number of qubits, so you can tell from the result whether the correct circuit was used. Also different transpilation options - i.e. if you send in `circ1 + skip_transpilation=False` and `circ1+skip_transpilation=True`, it should not be reused. `skip_transpilation` is the easiest to test with because `skip_transpilation=True` will fail the job. 


Also don't forget this comment:

> Also caching doesn't work with simulators, so you'll need to check that (which may be a bit tricky for cloud channel where backend can be omitted).

test/unit/test_estimator.py
ChitsimranGill: removed it

test/unit/test_estimator.py
ChitsimranGill: Tried using mock session, but got errors while trying to access variable from session class. Did change the assertions and changed parameters to ANY.

ChitsimranGill:
> 

@jyu00 we wrote integration tests for sampler circuit caching.

@jyu00 @ikkoham @t-imamichi could you provide four sample circuits and observables for estimator with 2,3,4, and 5 qubits, which provide consistent results which we can assert. 

@jyu00 we tried `skip_transpilation` example that you suggested, but it did not fail on simulator. Right now we can only test on simulators from qiskit-ibm-runtime. Do you have an example that might actually fail on simulator without transpilation.

qiskit_ibm_runtime/estimator.py
```diff
@@ -189,6 +193,17 @@ def __init__(
             backend = session or backend
             self._session = get_default_session(service, backend)
 
+        self._first_run = True
+        self._circuits_map = {}
+        if self.circuits:
+            for circuit in self.circuits:
+                circuit_id = _hash(
```
jyu00, line 200: Why do you need to hash it when the output of `json.dumps()` is already a string?

jyu00:
> @jyu00 we wrote integration tests for sampler circuit caching.

Excellent! 
 
> @jyu00 @ikkoham @t-imamichi could you provide four sample circuits and observables for estimator with 2,3,4, and 5 qubits, which provide consistent results which we can assert.

You can use different gates instead of different qubit counts. For example, 

```
qc1 = QuantumCircuit(2)
qc1.x(range(2))
qc1.measure_all()
qc2 = QuantumCircuit(2)
qc2.measure_all()
qc3 = QuantumCircuit(2)
qc3.h(range(2))
qc3.measure_all()
```
And observable `H = SparsePauliOp.from_list([("IZ", 1)])`  will give you values of -1 for qc1, 1 for qc2, and somewhere in between for qc3. 


> @jyu00 we tried `skip_transpilation` example that you suggested, but it did not fail on simulator. Right now we can only test on simulators from qiskit-ibm-runtime. Do you have an example that might actually fail on simulator without transpilation.

Right, I forgot the simulator supports all gates. Try specifying `initial_layout=list(range(x))`, where x is smaller than the number of qubits you have. It should give you an error saying the length of the list needs to match that of the qubit count.

qiskit_ibm_runtime/estimator.py
ChitsimranGill: we used `_hash` because the value returned from `json.dumps()` was too long. We can remove it if you want.

ChitsimranGill:
> > @jyu00 we wrote integration tests for sampler circuit caching.
> 
> Excellent!
> 
> > @jyu00 @ikkoham @t-imamichi could you provide four sample circuits and observables for estimator with 2,3,4, and 5 qubits, which provide consistent results which we can assert.
> 
> You can use different gates instead of different qubit counts. For example,
> 
> ```
> qc1 = QuantumCircuit(2)
> qc1.x(range(2))
> qc1.measure_all()
> qc2 = QuantumCircuit(2)
> qc2.measure_all()
> qc3 = QuantumCircuit(2)
> qc3.h(range(2))
> qc3.measure_all()
> ```
> 
> And observable `H = SparsePauliOp.from_list([("IZ", 1)])` will give you values of -1 for qc1, 1 for qc2, and somewhere in between for qc3.
> 
> > @jyu00 we tried `skip_transpilation` example that you suggested, but it did not fail on simulator. Right now we can only test on simulators from qiskit-ibm-runtime. Do you have an example that might actually fail on simulator without transpilation.
> 
> Right, I forgot the simulator supports all gates. Try specifying `initial_layout=list(range(x))`, where x is smaller than the number of qubits you have. It should give you an error saying the length of the list needs to match that of the qubit count.

@jyu00 added integration tests to test circuit caching with transpilation options for both sampler and estimator.

jyu00 (approved):
Excellent work! Thank you!

PR #554, opened 2022-10-04
Fix environment options in from_dict

jyu00:
### Summary

Fix a bug in `Options._from_dict` where it only looked at `image` and ignored the other options. 

### Details and comments
Fixes #


kt474 (approved):
LGTM

PR #557, opened 2022-10-05
Pass in correct instance in service.run()

kt474:
### Summary

`qrt_options.instance` should be passed in instead of `instance`

Currently, this wouldn't use the correct instance 

```python
options = Options(environment= {'instance':'ibm-q/open/main'})
    sampler = Sampler(session=session, options=options)
    job = sampler.run(...)
```


jyu00 (approved):
Good catch, thanks!

PR #562, opened 2022-10-06
Remove backend.run

jyu00:
### Summary

After some discussion with @tmittal947 we thought it's better not add `backend.run()` in `qiskit-ibm-runtime` for a number of reasons:

1. It's not supported by all channels
2. `qiskit-ibm-runtime` is supposed to focus on the "managed services" experience, and `backend.run()` is the opposite of that. Instead, any need of direct access should be done using the new `qiskit-ibm-provider`.
3. It's confusing to users when we have `backend.run()` in both `qiskit-ibm-runtime` and `qiskit-ibm-provider`. 

Sorry @kt474 

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/ibm_backend.py
+++ b/qiskit_ibm_runtime/ibm_backend.py
@@ -14,12 +14,9 @@
 
 import logging
 
-from typing import Iterable, Union, Optional, Any, List, Callable, Type, Dict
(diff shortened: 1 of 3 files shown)


kt474 (approved):
No worries, LGTM - this will make things easier in the future

PR #565, opened 2022-10-07
Add job_tags and max_execution_time to options

jyu00:
### Summary

Adds `job_tags` and `max_execution_time` to `Options`. Replaces #510 (although I did copy most of the code so credits to @kt474).

### Details and comments
Fixes #473

PR #566, opened 2022-10-07
Remove excess options

jyu00:
### Summary

After talking to @tmittal947 we decided to keep the number of options to a minimum, in order to keep the interface simple. 

Note that this is based on #565, so that one should be reviewed/merged first. 

### Details and comments
Fixes #


kt474 (approved):
LGTM

PR #569, opened 2022-10-10
Update base primitives

### LINKED ISSUE #482: Use new base primitives

jyu00:
### Summary

Update the base primitives to the latest and fix options merging. 

### Details and comments
Fixes #482 

The latest primitives added a `_run_options` attribute, which is an instance of `qiskit.providers.Options`. This conflicts with the `qiskit_ibm_runtime.Options` class, which is a dataclass that allows categorizing options and auto-complete (e.g. one can specify `options.transpilation.skip_transpilation=True` instead of using a nested dictionary). As a result, this PR updated the primitives to largely ignore `_run_options` but kept the `options` attribute be the type of `qiskit.providers.Options` as defined by the base primitives.


kt474 (approved):
LGTM

PR #582, opened 2022-10-18
Prepare release 0.7

jyu00:
### Details and comments
Fixes #


kt474 (approved):
There's one more release note from [580](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/releasenotes/notes/upgrade-terra-80a543c559e259ce.yaml) that should be moved into the 0.7 folder - other than that lgtm

Issue #568, opened 2022-10-09
Problems retrieving results when submitting two jobs under the same session


pedrorrivero:
**Describe the bug**
Problems retrieving results when submitting two jobs under the same session: the first job reports results from the second one.

**Steps to reproduce**
```python
## IMPORTS
from qiskit.circuit import QuantumCircuit, Parameter
from qiskit.quantum_info import SparsePauliOp
from qiskit.primitives import Estimator as TerraEstimator
from qiskit_ibm_runtime import Estimator, QiskitRuntimeService, Session
import numpy as np
import matplotlib.pyplot as plt

## CIRCUIT
cir = QuantumCircuit(2)
cir.h(0)
cir.cx(0, 1)
cir.ry(Parameter("theta"), 0)
display(cir.draw("mpl"))

## THETAS
theta_vec = np.linspace(-np.pi, np.pi, 15)

## OBSERVABLE
obs1 = SparsePauliOp(["ZZ", "ZX", "XZ", "XX"], [1, -1, +1, 1])
obs2 = SparsePauliOp(["ZZ", "ZX", "XZ", "XX"], [1, +1, -1, 1])

## TERRA ESTIMATOR
estimator = TerraEstimator()

job1 = estimator.run(
    circuits = [cir] * len(theta_vec), 
    observables = [obs1] * len(theta_vec), 
    parameter_values = [[v] for v in theta_vec],
)
job2 = estimator.run(
    circuits = [cir] * len(theta_vec), 
    observables = [obs2] * len(theta_vec), 
    parameter_values = [[v] for v in theta_vec],
)

CHSH1_terra = job1.result()
CHSH2_terra = job2.result()

## RUNTIME ESTIMATOR
service = QiskitRuntimeService()
options = {"backend": "ibmq_qasm_simulator"}

with Session(service=service, backend="ibmq_qasm_simulator") as session:
    estimator = Estimator(session=session)
    
    job1 = estimator.run(
        circuits = [cir] * len(theta_vec), 
        observables = [obs1] * len(theta_vec), 
        parameter_values = [[v] for v in theta_vec],
    )
    job2 = estimator.run(
        circuits = [cir] * len(theta_vec), 
        observables = [obs2] * len(theta_vec), 
        parameter_values = [[v] for v in theta_vec],
    )
    
    CHSH1_runtime = job1.result()
    CHSH2_runtime = job2.result()
```

[...]

PR #583, opened 2022-10-18
Add test for 568

jyu00:
### Summary

Add a test for the use case described in #568, since our existing tests didn't catch that. 

### Details and comments
Fixes #


kt474 (approved):
LGTM

PR #592, opened 2022-10-26
Disable data caching temporarily

jyu00:
### Summary

Data caching is causing significant performance issue right now. This PR disables it until that's fix. 

### Details and comments
Fixes #


kt474 (approved):
LGTM

PR #597, opened 2022-10-28
Allow separate decoders

### LINKED ISSUE #589: Decoder for Estimator interim result callback
Currently, Estimator uses `EstimatorResultDecoder` which decodes to `EstimatorResult` but we also need to support a more general class for Estimator interim results.

jyu00:
### Summary

Closes #589. 

Allows one to specify separate decoders for interim and final results.

### Details and comments
Fixes #


kt474 (approved):
LGTM

PR #600, opened 2022-11-01
Add Python 3.10 support

kt474:
### Details and comments
Fixes #


jyu00:
Doesn't `tox.init` need to be updated as well? Also can you add a release note.

jyu00 (approved):

PR #601, opened 2022-11-01
Use backend service in session if known

jyu00:
### Summary

If I pass an `IBMBackend` instance to `Session` constructor, the service used to initialize the backend instance should be used for the session, if no other `service` parameter is passed. 

### Details and comments
Fixes #


kt474 (approved):
LGTM

PR #612, opened 2022-11-10
Add back language switcher and fix formatting issues

SooluThomas:
### Summary

Added back code that was deleted in https://github.com/Qiskit/qiskit-ibm-runtime/pull/599

### Details and comments
Fixes https://github.com/Qiskit/qiskit-ibm-runtime/issues/608 and https://github.com/Qiskit/qiskit-ibm-runtime/issues/611

Tested locally,

image

image

PR #613, opened 2022-11-11
Fix error mitigation notebook

### LINKED ISSUE #610: Image link broken in notebook "Get started with error suppression and error mitigation with Qiskit Runtime"
**Describe the bug**
One of the cells under the title "dynamical decoupling" is used to plot a graph, but it is not plotted. Instead "../_images/tutorials_Error-Suppression-and-Error-Mitigation_13_0.png" is written. Checked the images folder, no image with this name, so I think this is the reason the image is not showing. 

**Steps to reproduce**
Go to https://qiskit.org/documentation/partners/qiskit_ibm_runtime/tutorials/Error-Suppression-and-Error-Mitigation.html
Under title "dynamical decoupling" the second cell should contain an image for the plot, but the image is not shown.

**Expected behavior**
Image should be shown

**Suggested solutions**
Add the image to the folders images

runtime2

beckykd:
A problem with a heading was causing a plot to not work
Fixes #610 

### Details and comments
Fixes #


docs/_templates/versions.html
```diff
@@ -1,8 +1,4 @@
 <div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="versions">
-  <span class="rst-current-version" data-toggle="rst-current-version">
-    <span class="rst-current-version-label">{{ version_label }}</span>
-    <span class="rst-versions-dropdown-icon"></span>
-  </span>
   <div class="rst-other-versions">
```
kt474, line 2: I think we want to keep this right? It's what was added back in https://github.com/Qiskit/qiskit-ibm-runtime/pull/612

beckykd: Oh yeah; it didn't show up as a merge conflict, so I didn't notice it was missing.  thanks!

kt474 (approved):

PR #617, opened 2022-11-15
Update docstring for session_id

mriedem:
### Summary

The Session.session_id is None until the first job in the session is run since it implicitly creates the session and sets the session_id from that. This is a bit confusing though if you do something like this and expect something other than None:

```python
with Session(service=service) as session:
    print(f"Session id: {session.session_id}")
```

### Details and comments

n/a

### DIFF
--- a/qiskit_ibm_runtime/session.py
+++ b/qiskit_ibm_runtime/session.py
@@ -185,7 +185,7 @@ class Session:
         """Return the session ID.
 
         Returns:
-            Session ID.
+            Session ID. None until a job runs in the session.
         """
         return self._session_id
 


qiskit_ibm_runtime/session.py
```diff
@@ -185,7 +185,7 @@ def session_id(self) -> str:
         """Return the session ID.
 
         Returns:
-            Session ID.
+            Session ID. None until a job runs in the session.
```
mriedem, line 188: Because: https://github.com/Qiskit/qiskit-ibm-runtime/blob/baf911a510a4a55d887c4f6593b0009ca2ed5c1c/qiskit_ibm_runtime/session.py#L162

kt474 (approved):

PR #623, opened 2022-11-17
Set backend to none if not returned in job response

### LINKED ISSUE #622: KeyError: 'backend' in _decode_job
**Describe the bug**

I submitted a job to cloud runtime (staging if that matters) without specifying a backend so that the scheduler would pick one. The job actually completed on the backend (I can see from the logs) but when I get details about the job I get a KeyError:

```
  Input In [8] in <cell line: 1>
    job = service.job('cdr4n5es1c6fcn0blivg')
  File /opt/conda/lib/python3.8/site-packages/qiskit_ibm_runtime/qiskit_runtime_service.py:1278 in job
    return self._decode_job(response)
  File /opt/conda/lib/python3.8/site-packages/qiskit_ibm_runtime/qiskit_runtime_service.py:1408 in _decode_job
    backend = self.backend(raw_data["backend"], instance=instance)
KeyError: 'backend'
```

That's this code:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/5ef5711716b52a95b8a0211385f3b4450da82f74/qiskit_ibm_runtime/qiskit_runtime_service.py#L1408

In my case the job completed so I'd expect the backend to be set on the job, so it seems there might be a bug on the server side (I've also opened an issue for that internally), but also looking at the Swagger doc for GET /jobs/{id}:

https://us-east.quantum-computing.cloud.ibm.com/openapi/#/Jobs/get_job_details_jid

It says `backend` is optional in the response:

image

So the qiskit client code should be handling that and avoid the KeyError.

**Steps to reproduce**

Submit a cloud job without specifying an explicit backend.

**Expected behavior**

I should be able to get the cloud job details, like what backend it ran on once it's complete.

**Suggested solutions**

Handle the KeyError but not requiring `backend` to be in the raw JSON response, at least for cloud jobs (channel=`ibm_cloud`).

**Additional Information**

- **Python version**: 3.8
- **Operating system**: ubuntu focal

### LINKED ISSUE #633: DeprecationWarning about job_id and backend attributes which aren't used
**Describe the bug**

I have this program which runs successfully:

```python

options = {
    "backend": "",
    "job_tags": ["proxy_mcs", "use_runtime_workers", "rs-target"]
}

job = service.run(program_id="hello-world",
                  options=options,
                  inputs=program_inputs)
print(f'submitted job: {job.job_id()}')
print('waiting for results')
print(f'results: {job.result()}')
print(f'job ran on backend: {job.backend()}')
```

However, I always get this DeprecationWarning:

>DeprecationWarning: Note that the 'job_id' and 'backend' attributes of a runtime job have been deprecated as of qiskit-ibm-runtime 0.7 and will be removed no sooner than 3 months after the release date. Please use the job_id() and backend() methods instead.

The thing is, I'm not using the `job_id` and `backend` attributes directly, I'm using the methods as noted. So I don't think I should be seeing the warning. Maybe those methods are hitting the deprecated attributes indirectly and that's triggering the warning?

**Steps to reproduce**

See above.

**Expected behavior**

Don't warn on something that's not being directly used.

**Suggested solutions**

n/a

**Additional Information**

- **Python version**: 3.8
- **Operating system**: Ubuntu Focal LTS

kt474:
### Details and comments
Fixes #622 
Fixes #633


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1405,7 +1405,11 @@ def _decode_job(self, raw_data: Dict) -> RuntimeJob:
                 instance = to_instance_format(hub, group, project)
         # Try to find the right backend
         try:
-            backend = self.backend(raw_data["backend"], instance=instance)
+            if "backend" in raw_data:
+                backend = self.backend(raw_data["backend"], instance=instance)
+            else:
+                if self.channel == "ibm_cloud":
```
mriedem, line 1411: What happens if this is false? backend will not be defined and you'll get a weirder error down below on L1430. It _shouldn't_ happen for quantum premium jobs, but it does appear to be a logic bug here.

mriedem (commented):
Should probably have a test for the cloud case where you submit a job without a specific backend and make sure you can get it back.

kt474:
> Should probably have a test for the cloud case where you submit a job without a specific backend and make sure you can get it back.

There's already a test for that [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/integration/test_job.py#L61)

mriedem:
> > Should probably have a test for the cloud case where you submit a job without a specific backend and make sure you can get it back.
> 
> There's already a test for that [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/integration/test_job.py#L61)

So why didn't it fail with the KeyError as reported in the issue?

mriedem:
> > > Should probably have a test for the cloud case where you submit a job without a specific backend and make sure you can get it back.
> > 
> > 
> > There's already a test for that [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/integration/test_job.py#L61)
> 
> So why didn't it fail with the KeyError as reported in the issue?

Is this maybe because of #625 where if a backend isn't returned the SDK grabs the first one available even though maybe that's not the one selected to run the job?

Also, that test you pointed at is asserting the job has a backend set, but the code fix makes it OK to return None:

https://github.com/Qiskit/qiskit-ibm-runtime/pull/623/files#diff-4dd87aa2d19eac69b8f4ae234fd835fb5e777be546e019d7c32502217f352598R1411

So I'd think the test would actually to make sure `job.backend()` can return None, not the `assertTrue` being used in the existing test which doesn't cover this case. If there is a race condition, then that other test might intermittently fail?

mriedem:
Also, when do those integration tests run, I see it was skipped for this PR:

https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/3568046328/jobs/5996688133

Do those run on a cron?

mriedem:
> Also, when do those integration tests run, I see it was skipped for this PR:
> 
> https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/3568046328/jobs/5996688133
> 
> Do those run on a cron?

Yup: https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/.github/workflows/integration-tests.yml#L16

kt474:
Left a comment in the issue [here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/622#issuecomment-1356980836)

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -224,6 +223,20 @@ def cancel(self) -> None:
...
+            IBMRuntimeError: If a network error occurred.
+        """
+        if not self._backend:  # type: ignore
+            try:
+                response = self._api_client.job_get(self.job_id())
+                self._backend = response.get("backend")
+            except RequestsApiError as err:
+                raise IBMRuntimeError(f"Failed to get job backend: {err}") from None
+        return self._backend
+
```
kt474, line 239: @jyu00 copying this comment from the issue 
> I suppose that's one way of getting updated job data. But it'd be nicer for `backend()` to refresh under the cover if backend is empty, instead of the user having to know to re-retrieve the job again using `service.job()`. For example,
> 
> ```
> job = sampler.run()
> job.backend()  # fetch new data if backend is empty
> ```
> 
> is nicer than
> 
> ```
> job = sampler.run()
> rjob = service.job(job.job_id())
> rjob.backend()
> ```

I'm not sure how we could get the actual `IBMBackend` object when refreshing without calling the service somehow. This current method would only return a string if the 'backend' is initially `None`.

jyu00: Yeah I think you can just pass in the service instance when initializing the job and use that to get `IBMBackend`.

kt474:
@jyu00 when you get the chance can you take another look?

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -250,6 +255,21 @@ def cancel(self) -> None:
         self.cancel_result_streaming()
         self._status = JobStatus.CANCELLED
 
+    def backend(self) -> Backend:
```
jyu00, line 258: The return type is technically `Optional[Backend]` since it can still be None

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -157,7 +163,6 @@ def __init__(
         # For backward compatibility. These can be removed once 'job_id' and 'backend'
         # as attributes are no longer supported.
         self.job_id = CallableStr(job_id)
-        self.backend = self._backend
```
jyu00, line 160: If you are removing `backend` as an attribute, might as well remove `job_id` as an attribute and close [this issue](https://github.com/Qiskit/qiskit-ibm-runtime/issues/633). You'd also need to update the release note on that.

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -97,6 +100,7 @@ def __init__(
             Union[Type[ResultDecoder], Sequence[Type[ResultDecoder]]]
         ] = None,
         image: Optional[str] = "",
+        service: "qiskit_runtime_service.QiskitRuntimeService" = None,
```
jyu00, line 103: Does `service` need to be optional? If it indeed needs to be optional, then line 268 needs to check if it's None.

merav-aharoni (approved):
I have nothing more to add beyond Jessie's comments. LGTM.

PR #628, opened 2022-11-19
Choose correct backend within a session 

### LINKED ISSUE #627: Backend specified on job submitted as part of a session is not honored
**Describe the bug**

I've got a script which tries to create 5 jobs in a session, it submits the jobs via `QiskitRuntimeService.run` and the `options` on the jobs specify the `alt_algiers` backend. However, the jobs are run on the `alt_canberra` backend.

```python

QiskitRuntimeService.save_account(
    channel="ibm_cloud", token=apikey, instance=crn,
    url="https://test.cloud.ibm.com")
options = {
    "backend": "alt_algiers",
    "job_tags": ["proxy_mcs", "use_runtime_workers", "sessions"]
}

print('submitting 5 jobs in a session')
with Session(service=service) as session:
    for i in range(5):
        print(f'{i+1} submitting session job')
        job = session.run(program_id="hello-world",
                          options=options,
                          inputs=program_inputs)
        print(f"submitted job: {job.job_id()}")
        print(f"job results: {job.result()}")
    print(f'all jobs run for session {session.session_id}')
```

The first job created is `cdru53vj3bs1jpmpdmfg` which ran on `alt_canberra`, not `alt_algiers`.

Checking the server logs this is because it looks like the job was sent to the runtime API without a backend specified even though it's in the `options` passed to `run()`.

My guess this has something to do with the `Session` constructor in the context manager and how it doesn't have `backend` supplied, so maybe that's overwriting what's passed to submit the job to `run()`? From the docs:

https://qiskit.org/documentation/partners/qiskit_ibm_runtime/stubs/qiskit_ibm_runtime.Session.html#qiskit_ibm_runtime.Session

image

I guess that was a surprise to me, is it intentional that it'd overwrite the backend specified in `options` to the `run()` method?

**Steps to reproduce**

see the script above

**Expected behavior**

I would expect that if the `options` passed to `run()` have a backend specified but the `Session` doesn't, then the backend from `options` is used when submitting the job rather than `Session.backend` overwriting it when `Session.backend` is None.

If this is working as designed and my script is invalid, I'd expect an error when trying to submit the first job in the session because the usage is wrong - an explicit error would be easier to sort out from a user perspective rather than wonder why the backend I specified wasn't used.

**Suggested solutions**

See above.

**Additional Information**

- **Python version**: 3.8

kt474:
### Details and comments
Fixes #627


mriedem:
Test for this? It's not clear to me how this resolves the issue without a test.

test/integration/test_session.py
```diff
@@ -71,3 +71,33 @@ def test_estimator_sampler(self, service):
...
+
+        backend = "ibmq_qasm_simulator"
+        inputs = {"iterations": 1}
+        options = {
+            "backend": backend,
+        }
+        with Session(service=service, backend="statevector_simulator") as session:
+            job = session.run(program_id="hello-world", options=options, inputs=inputs)
+            self.assertEqual(backend, job.backend().name)
+
```
kt474, line 87: @jyu00 my understanding is that in a scenario like this the backend passed in through options should be honored?

jyu00: Yes, since the docstring says `options` can be anything in `RuntimeOptions` (which includes `backend`). We should change that though since you can't have a session cross multiple backends.

qiskit_ibm_runtime/session.py
```diff
@@ -141,7 +141,8 @@ def run(
         """
 
         options = options or {}
-        options["backend"] = self._backend
+        if "backend" not in options:
+            options["backend"] = self._backend
```
jyu00, line 145: Can you raise an exception if a session is already active, has a backend, and the backend passed in `options` doesn't match.

qiskit_ibm_runtime/session.py
```diff
@@ -138,9 +140,30 @@ def run(
...
+                self._session_id
+                and self._backend
+                and options["backend"] != self._backend
+            ):
+                raise IBMInputValueError(
+                    f"The backend '{options['backend']}' is different from",
+                    f"the session backend '{self._backend}'",
+                )
+
         options["backend"] = self._backend
```
jyu00, line 167:
```suggestion
        else:
            options["backend"] = self._backend
```

I think you accidentally deleted the original fix

qiskit_ibm_runtime/session.py
```diff
@@ -138,9 +140,30 @@ def run(
...
         options = options or {}
+        if "backend" in options:
+            issue_deprecation_msg(
+                "'backend' is no longer a supported option within a session",
+                "0.8",
+                "Sessions do not support multiple backends.",
+                3,
+            )
+            if (
+                self._session_id
```
jyu00, line 158: I don't think we need to compare session id here? If I do

```
session = Session(backend="foo")
session.run(options={"backend": "bar"})
```
^ this should fail because it's ambiguous which backend I really want.

qiskit_ibm_runtime/session.py
```diff
@@ -138,9 +140,30 @@ def run(
```
jyu00, line 140: Can you also update the docstring to something like

```
                See :class:`qiskit_ibm_runtime.RuntimeOptions` for all available options, EXCEPT ``backend``, which should be specified during session initialization.
```

qiskit_ibm_runtime/session.py
```diff
@@ -137,15 +139,33 @@ def run(
...
-        options["backend"] = self._backend
+        if "backend" in options:
+            issue_deprecation_msg(
+                "'backend' is no longer a supported option within a session",
+                "0.9",
+                "Instead, specify a backend when creating a Session instance.",
+                3,
+            )
+            if self._backend and options["backend"] != self._backend:
+                raise IBMInputValueError(
```
mriedem, line 163: There is no test coverage for this is there?

kt474: added test case here https://github.com/Qiskit/qiskit-ibm-runtime/pull/763

PR #661, opened 2022-12-20
Honor the instance chosen when selecting a backend

### LINKED ISSUE #630: `QiskitRuntimeService.backend` ignores the `instance` parameter (ibm_quantum channel)
**Describe the bug**
When I want to specify which instance to use when loading a backend before using it in a session (available through multiple providers with my account), I noticed on the job page in iqx that it completely ignores which instance is specified. 
Also noticed that when I don't specify anything, it will choose the same premium provider by default each time.

**Steps to reproduce**
This doesn't work : 
```
# doesn't work
from qiskit.test.reference_circuits import ReferenceCircuits
from qiskit_ibm_runtime import QiskitRuntimeService, Session, Sampler

service = QiskitRuntimeService(channel="ibm_quantum")
bell = ReferenceCircuits.bell()

backend=service.backend(name='ibmq_quito', instance='ibm-q/open/main')

    job = sampler.run(bell, shots=1024)
    print(f"Job ID: {job.job_id()}")
```

This does : 
```
# works
from qiskit.test.reference_circuits import ReferenceCircuits
from qiskit_ibm_runtime import QiskitRuntimeService, Session, Sampler

service = QiskitRuntimeService(channel="ibm_quantum", instance='ibm-q/open/main')
bell = ReferenceCircuits.bell()

backend=service.backend('ibmq_quito')

    job = sampler.run(bell, shots=1024)
    print(f"Job ID: {job.job_id()}")
```

**Expected behavior**
When I load `backend=service.backend(name='ibmq_quito', instance='ibm-q/open/main')` , the session should run through the right specified provider.

**Suggested solutions**

**Additional Information**

- **Python version**: 3.9.12

kt474:
### Details and comments
Fixes #630


qiskit_ibm_runtime/options/options.py
```diff
@@ -161,6 +161,9 @@ def _get_runtime_options(options: dict) -> dict:
             if fld.name in environment:
                 out[fld.name] = environment[fld.name]
 
+        if "instance" in options:
+            out["instance"] = options["instance"]
+
```
kt474, line 166: I believe this was another bug? Users should be able to pass the instance through options. 

```python
options = Options(instance='ibm-q/open/main')

    sampler = Sampler(session=session, options=options)
    job = sampler.run(bell, shots=1024)
```

jyu00: It feels to me that `instance` should be a session option and not a primitive option. I'm also not sure if the scheduler supports cross-instance session.

kt474: `instance` is a runtime options [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/runtime_options.py#L34) so shouldn't it be supported in [_get_runtime_options](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/options/options.py#L151)? Or am I understand the options incorrectly?

On this branch I am able to run jobs with different instances within a single session so it seems like it's supported by the scheduler.

kt474: will just revert this for now as it doesn't relate to the original issue

jyu00 (commented):
Code looks good, but can we have some tests?

PR #674, opened 2023-01-11
Remove `close()` when exiting session

### LINKED ISSUE #653: Jobs which have not run before session closes are not run as part of session
**Describe the bug**

When using the `Session` context manager, it exits and closes the session:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/85d429d64a0a1e2f5f0c0d86046444d2da5bf283/qiskit_ibm_runtime/session.py#L205

Per my discussion with @jyu00 my understanding of the original intent with this is it means no more jobs should be accepted for that session. What actually happens in the scheduler is that the jobs which did not run before the session was closed are not run as part of the session once it goes active, meaning they don't benefit from priority access to the backend.

In addition, in the scheduler we plan on making a change such that any jobs which did not run before the session closed will go into an error state.

**Steps to reproduce**

Submit several jobs using the `Session` context manager but don't wait for their results before exiting the context manager. Depending on when those jobs run in the scheduler, they will not run as part of the session if the session is explicitly closed by the exit of the context manager.

**Expected behavior**

The jobs should run as part of the session (once activated by the first job in the session running) and get the benefit of priority access to the backend. The session should close automatically when its max TTL is reached.

**Suggested solutions**

Remove the automatic `close()` call in the exit of the context manager.

**Additional Information**

There are docs and tutorials which use the the Session context manager example and show the job result blocking call being made outside of the context manager, which would be the bug discussed here. For example:

https://qiskit.org/documentation/partners/qiskit_ibm_runtime/how_to/run_session.html#full-example

When that job actually runs, it might not have been part of an active session. Furthermore, in the future when the scheduler changes, that will result in an error because the job was requested to run in a session but if it didn't run before the session closed, it will be put into an error state.

- **Python version**: 3.8
- **Operating system**: Ubuntu Focal 20.04

kt474:
### Details and comments
Fixes #653


nonhermitian:
But doesn't this mean that there will be a 5 minute idle time after each session unless an user remembers to call `session.close()` after all their jobs are done?  That idle time can add up quickly

jyu00:
> But doesn't this mean that there will be a 5 minute idle time after each session unless an user remembers to call `session.close()` after all their jobs are done? That idle time can add up quickly

Yes, and the remedy for that is parallelizing classical computation of multiple jobs to keep the system busy. Unlike reservation, a session does not have guaranteed exclusivity to a system.

docs/how_to/run_session.rst
```diff
@@ -64,8 +64,7 @@ This example starts a session, runs an Estimator job, and outputs the result:
   with Session(service=service, backend="ibmq_qasm_simulator") as session:
       estimator = Estimator(session=session, options=options)
       job = estimator.run(circuit, observable)
-
-  result = job.result()
+      result = job.result()
```
jyu00, line 67:
```suggestion
      result = job.result()
      session.close()
```
Add in the sample code to encourage people to close session manually

docs/how_to/run_session.rst
```diff
@@ -88,7 +87,6 @@ After this time limit is reached, the session is permanently closed.
 
 Additionally, there is an *interactive* timeout value. If there are no session jobs queued within that window, the session is temporarily deactivated and normal job selection resumes. After a session is deactivated, a subsequent job could start an additional session.  Jobs for the new session would then take priority until the new session deactivates or is closed. After the new session becomes inactive, if the job scheduler gets a job from the original session and its maximum timeout value has not been reached, the session is reactivated until its maximum timeout value is reached.
 
-When using primitives with their context managers as previously described, the session is closed automatically when the block is exited.
```
jyu00, line 91: Can you add some words to say something like "You are encouraged to use `session.close()` to close the session when you are done submitting jobs. This allows the scheduler to run the next job without waiting for session timeout. Keep in mind, however, that you cannot submit more jobs to a closed session."

jyu00 (approved):
LGTM, thanks!

PR #676, opened 2023-01-12
Don't override `max_execution_time` in options

### LINKED ISSUE #675: don't override explicit max_execution_time of first job
https://github.com/Qiskit/qiskit-ibm-runtime/blob/88c06a1935eac62c102b47e624d6daf35118b44f/qiskit_ibm_runtime/session.py#L146-L149

currently during `session.run()` the cost of the first job is set to the session max_time regardless of if `max_execution_time` was explicitly set  in the job options

```
 if not self._session_id and "max_execution_time" not in options: 
     options["max_execution_time"] = self._max_time 
```

this may be ultimately be replaced by fixes for #670 if we don't use `max_time` as `max_execution_time` but instead as `session_limit`. However, these changes should still be fixed since any first job session time will have it's max execution overwritten (even if `max_time` is None, if `max_execution_time` is set on the job, it will be replaced by `None`.

kt474:
### Details and comments
Fixes #675

### DIFF
--- a/qiskit_ibm_runtime/session.py
+++ b/qiskit_ibm_runtime/session.py
@@ -143,7 +143,7 @@ class Session:
         options = options or {}
         options["backend"] = self._backend
 
-        if not self._session_id:
+        if not self._session_id and not options.get("max_execution_time"):
             # TODO: What happens if session max time != first job max time?
             # Use session max time if this is first job.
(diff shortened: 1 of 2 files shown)


kt474:
closing for now, as this is probably not necessary

PR #577, opened 2022-10-14
Update qpy version to terra 0.23.1

kt474:
### Summary

Updating qpy to match terra 0.23.1

co-authored by @nkanazawa1989 @jakelishman @mtreinish 
https://github.com/Qiskit/qiskit-terra/pull/8278
https://github.com/Qiskit/qiskit-terra/pull/8928
https://github.com/Qiskit/qiskit-terra/pull/8927
https://github.com/Qiskit/qiskit-terra/pull/9314
https://github.com/Qiskit/qiskit-terra/pull/9002
https://github.com/Qiskit/qiskit-terra/pull/9391
https://github.com/Qiskit/qiskit-terra/pull/8949

PR #677, opened 2023-01-12
Use terra qpy

ChitsimranGill:
Use terra qpy


kt474 (commented):
LGTM, importing qpy from terra makes things a lot easier. If there are no objections to importing this way, I think this can be merged 

Also we can close #577

jyu00:
Well, there's a good reason we put a copy of qpy in qiskit-ibm-runtime. qpy only supports backward compatibility, which means if I use qpy version N+1 to serialize, it _cannot_ be deserialized by qpy version N. In the old days we constantly had the issue wherein terra releases qpy N+1, users serialize their circuits with that, and their jobs fail because runtime server is still at qpy N. Ideally runtime server would update their qpy version immediately after a new terra release, but that's not always possible. So if we go back to terra's qpy we're going to run into the same issues.

PR #701, opened 2023-02-07
Change default optimization_level

### LINKED ISSUE #643: Different defaults for error mitigation/suppression for noiseless backends

Current default options for the primitives are:

```
optimization_level = 3
resilience_level = 1
```

These options add a processing overhead for error suppression and error mitigation. When using noiseless devices, the overhead doesn't bring any additional improvements. We should consider using different defaults when using noiseless devices (i.e. simulator without noise model)

**Acceptance criteria**

merav-aharoni:
### Summary
Change default `optimization_level` to None. If the backend is a simulator without noise, `optimization_level` will be set to 1. Otherwise it will be set to 3.

### Details and comments
Fixes #643


merav-aharoni:
@kt474 , @jyu00 - I implemented as we discussed. 
There is one problem - in `test_ibm_primitives.py` several tests have an assertion `self.assertDictEqual(inst.options.__dict__, asdict(Options()))`. 
This assertion now fails because the default `optimization_level` is None , but in `inst.options` it is 1 (or 3). 
Any suggestions on how to resolve? One options is to check the fields one at a time and skip `optimization_level`.

kt474:
Looks great so far! 

What you suggested would work for the test cases - since there are only 3 tests that are failing you could something like the following: 

`test_run_multiple_different_options`, `test_run_overwrite_options` - just update the assert statement 
```python 
self.assertDictEqual(inst.options.__dict__, asdict(Options(optimization_level=1)))
```

`test_dict_options` same thing for this one 
```python
expected = asdict(Options(optimization_level=1))
```

merav-aharoni:
Thank you, @kt474 . Good idea! I will complete this and then I'll ask for your review.

merav-aharoni:
@kt474 - I think this PR is ready for review now.

test/integration/test_options.py
```diff
@@ -68,6 +68,7 @@ def test_noise_model(self, service):
     @run_integration_test
     def test_simulator_transpile(self, service):
         """Test simulator transpile options."""
+        print("servie = " + str(dir(service)))
```
kt474, line 71: This can be removed

releasenotes/notes/default-optimization-level-3197ff933dfcfa4f.yaml
```diff
@@ -0,0 +1,7 @@
+---
+
```
kt474, line 2:
```suggestion
```
The extra space may mess with the docs formatting 
This release note can be categorized as "upgrade"

test/integration/test_options.py
```diff
@@ -98,3 +99,28 @@ def test_simulator_transpile(self, service):
                         job2.result()
                     # TODO: Re-enable when ntc-1651 is fixed
                     # self.assertIn("TranspilerError", err.exception.message)
+
+    @run_integration_test
+    def test_optimization_level(self, service):
+        """Test various definitions for optimization_level."""
+
+        backend = service.backends(simulator=True)[0]
```
kt474, line 107: nit: ```service.get_backend('ibmq_qasm_simulator")```

kt474 (commented):
Looks pretty good to me! Some minor comments - also I think the original issue #643 mentioned different default values for both `optimization_level` and `resilience_level`

test/integration/test_options.py
merav-aharoni: Of course...

merav-aharoni:
Regarding `resilience_level`, what I understood from @jyu00 was that I should only take care of the `optimization_level`. But I'll be happy to do something similar for `resilience_level`. What do you think, @jyu00 , @kt474 ? Should the conditions be the same (0 for simulation with no noise model, 1 otherwise)?

test/integration/test_options.py
merav-aharoni: No problem, but this line appears in several tests in this file. Shall I change them all to be consistent?

kt474: ah I didn't notice that - you can leave it as is then

jyu00:
> Regarding `resilience_level`, what I understood from @jyu00 was that I should only take care of the `optimization_level`. But I'll be happy to do something similar for `resilience_level`. What do you think, @jyu00 , @kt474 ? Should the conditions be the same (0 for simulation with no noise model, 1 otherwise)?

Sorry if I wasn't clear before. We definitely should do `resilience_level` as well, since the overhead is even bigger than `optimization_level`

merav-aharoni:
Thanks, @jyu00 , my misunderstanding. I will add to the PR.

merav-aharoni:
Hi @kt474 , I addressed your comments and added support for the `resilience_level`. Can you please review again?

kt474 (approved):
LGTM!

qiskit_ibm_runtime/options/options.py
```diff
@@ -81,7 +81,11 @@ class Options:
             :class:`SimulatorOptions` for all available options.
     """
 
-    optimization_level: int = 3
+    # Defaults for optimization_level and for resilience_level will be assigned
+    # in Sampler/Estimator
+    _DEFAULT_OPTIMIZATION_LEVEL = 3
+    _DEFAULT_RESILIENCE_LEVEL = 1
+    optimization_level: Optional[int] = None
     resilience_level: int = 1
```
kt474, line 89: `resilience_level` should also be optional right?

qiskit_ibm_runtime/sampler.py
```diff
@@ -169,6 +169,18 @@ def __init__(
...
+                not hasattr(_options.simulator, "noise_model")
+                or asdict(_options.simulator)["noise_model"] is None
+            ):
+                _options.optimization_level = 1
+                _options.resilience_level = 0
+            else:
+                _options.optimization_level = Options._DEFAULT_OPTIMIZATION_LEVEL
+                _options.resilience_level = Options._DEFAULT_RESILIENCE_LEVEL
+
         self._options: dict = asdict(_options)
```
kt474, line 184: Sorry, didn't see this initially - users can set optimization level or resilience level separately so the logic would have to something like: 

```python
if _options.optimization_level is None:
      # set values appropriately 
            
if _options.resilience_level is None:
      # set values appropriately 
            
```

qiskit_ibm_runtime/options/options.py
merav-aharoni: Certainly. It was actually this way in my clone, but I guess I forgot to commit.

qiskit_ibm_runtime/sampler.py
merav-aharoni: Of course you are right - I don't know how I missed this. Will fix.

merav-aharoni:
@kt474 - I made the fixes you pointed out and also improved one of the tests. Please have another (hopefully final) review.

PR #721, opened 2023-02-20
Return a short error message

### LINKED ISSUE #415: Return only last exception traceback in error_message()

`job.error_message()` basically just returns the job log. However, since the default logging level is now INFO, this tends to generate a really long message with most of them irrelevant to the error (e.g. transpilation generates a ton of info logs). 

It would be really nice if the provider can read from the tail of log and have `job.error_message()` return only the last exception traceback. `job.logs()` can continue to return the full logs if users need them. Similarly, `job.result()` would return the short error message if the job failed. 

At the very least, `job.error_message()` can return the last n lines of the log message, where n is configuration with a default of say 20. 

**Acceptance criteria**

`job.error_message()` returns only a small set of data most relevant to the error.

merav-aharoni:
### Summary
Since error messages from a failing job may be long, we shortened them so that they begin from the last  `Traceback` in the message.

### Details and comments
Fixes #415


merav-aharoni:
Hi @kt474 - I just made a small change in the way the error message is passed on. I still have work to do on this, but I wanted your opinion before I continue:
1. As a basic approach, do you think this is a good approach?
2. Should we allow the user control over the length of the error message (and have a short default?)

kt474:
> Hi @kt474 - I just made a small change in the way the error message is passed on. I still have work to do on this, but I wanted your opinion before I continue:
> 
> 1. As a basic approach, do you think this is a good approach?
> 2. Should we allow the user control over the length of the error message (and have a short default?)

We did something similar in qiskit-ibm-provider [here](https://github.com/Qiskit/qiskit-ibm-provider/pull/459/files). I think it would be easier if we updated [error_msg_from_job_response](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/runtime_job.py#L413) to not return the whole raw results returned from the server. I don't think it's necessary to let users control the length of the error msg, if they need to see the full msg, they could use `job.logs()`

merav-aharoni:
@kt474 - I did it as you suggested above. I couldn't think of any suitable test. Do you think a test is necessary?

kt474 (approved):
LGTM, I agree no need for a test here

PR #722, opened 2023-02-20
Skip unsupported tests on staging

kt474:
### Summary

Skipping all failing test on staging due to fake backend incompatibilities, or custom programs that don't exist

example: https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/4229174184/jobs/7345309669

### Details and comments
Fixes #

### DIFF
--- a/test/decorators.py
+++ b/test/decorators.py
@@ -28,7 +28,7 @@ def production_only(func):
 
     @wraps(func)
     def _wrapper(self, *args):
-        if "dev" in os.environ.get("QISKIT_IBM_URL", ""):
+        if "dev" in self.dependencies.url:
             raise SkipTest(
                 f"Skipping integration test. {self} is not supported on staging."
             )
--- a/test/decorators.py
+++ b/test/decorators.py
@@ -27,12 +27,12 @@ def production_only(func):
     """Decorator that runs a test only on production services."""
 
     @wraps(func)
-    def _wrapper(self, *args):
+    def _wrapper(self, *args, **kwargs):
         if "dev" in self.dependencies.url:
             raise SkipTest(
                 f"Skipping integration test. {self} is not supported on staging."
             )
-        func(self, *args)
+        func(self, *args, **kwargs)
 
     return _wrapper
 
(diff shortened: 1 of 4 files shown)


test/decorators.py
```diff
@@ -23,6 +23,20 @@
...
+    @wraps(func)
+    def _wrapper(self, *args):
+        if "dev" in os.environ.get("QISKIT_IBM_URL", ""):
+            raise SkipTest(
+                f"Skipping integration test. {self} is not supported on staging."
+            )
+        func(self, *args)
+
+    return _wrapper
+
```
kt474, line 38: Couldn't find another way to check if the environment was staging - open to suggestions

test/decorators.py
```diff
@@ -23,6 +23,20 @@
 from .unit.mock.fake_runtime_service import FakeRuntimeService
 
 
+def production_only(func):
+    """Decorator that runs a test only on production services."""
+
+    @wraps(func)
+    def _wrapper(self, *args):
+        if "dev" in os.environ.get("QISKIT_IBM_URL", ""):
```
kt474, line 31: Couldn't find another way to check if the environment was staging - open to suggestions

merav-aharoni: I like the idea of adding this as a decorator. I don't have any insight regarding your method of checking. I am just wondering if the URL is necessarily taken from the environment, because I think I recall that the URL can be given directly in the interface. But I may be wrong.

merav-aharoni (commented):
Generally looks fine to me, but I have two questions, one I wrote above and in addition
I am just wondering if this should apply to every test that uses `get_real_device`, for example `test_run_program_real_device`.

test/decorators.py
kt474: Ah you're right, i'll have to find another way to do this

kt474: Updated to use the service dependency url instead of an environment variable

kt474:
> Generally looks fine to me, but I have two questions, one I wrote above and in addition I am just wondering if this should apply to every test that uses `get_real_device`, for example `test_run_program_real_device`.

`test_run_program_real_device` is actually already marked as a slow test so it doesn't run anyway in our CI

merav-aharoni (approved):

PR #725, opened 2023-02-26
Code cleaning of small issues

merav-aharoni:
### Summary
Cleaning small issues in the code.

### Details and comments
Removed redundant assignments to count in `fake_runtime_client.py`.
Changed `notifyAll` to `notify_all` to avoid warning in `lint`.


test/unit/mock/fake_runtime_client.py
```diff
@@ -439,38 +439,33 @@ def jobs_get(
         skip = skip or 0
         jobs = list(self._jobs.values())
         count = len(self._jobs)
```
kt474, line 441:
```suggestion
```
I think this can be removed as well

kt474 (approved):

Issue #685, opened 2023-01-25
Collect usage data


jyu00:
**What is the expected feature or enhancement?**

Having usage data would help us understand and improve common use cases. Since Qiskit Runtime is designed for algorithm developers, we'd like to know which algorithms are being used. As a starting point, we can collect 2 data points:

1. The modules used by the users
2. The specific calling class

1 is easily achievable using something like

```
import sys

interesting_keywords = ["qiskit_nature", "qiskit_machine_learning", "qiskit_optimization", "qiskit_finance", "qiskit.algorithms", "qiskit.primitives"]  # add more as you please

found_keywords = [key for key in interesting_keywords if key in sys.modules]
```

2 requires looking into the stack and find the calling submodule (e.g. `qiskit.algorithms.minimum_eigensolvers.vqe`)

We can save this data in the header, similar to Qiskit version. There should also be an option to op out. 

**Acceptance criteria**

johannesgreiner:
I'd think IBM's privacy laws would actually require users to explicitly opt in. Is it anonymous data or can it be related to the user?

jyu00:
> I'd think IBM's privacy laws would actually require users to explicitly opt in. Is it anonymous data or can it be related to the user?

We are only collecting module names, so nothing related to the users. I've already cleared with the security/privacy team.

PR #733, opened 2023-03-02
feat: Add support for the qiskit-version-2 header format

JakWai01:
Closes https://github.ibm.com/IBM-Q-Software/ntc/issues/2058
Closes https://github.com/Qiskit/qiskit-ibm-runtime/issues/685

### Summary

This PR adds support for the `qiskit-version-2` header format. The new header formats includes proper versioning and the "caller" of a job. 

#### Sample header
`qiskit-version-2/qiskit-terra-0.23.0*,qiskit-aer-0.11.2/qiskit-terra-algorithms/random_client_header`


qiskit_ibm_runtime/api/session.py
```diff
@@ -51,20 +54,35 @@
...
-        pass
-
-    qiskit_pkgs = ["qiskit-terra", "qiskit-aer", "qiskit-ignis", "qiskit-aqua"]
-    pkg_versions = {"qiskit-ibm-runtime": ibm_runtime_version}
+
+    qiskit_pkgs = [
+        "qiskit-terra",
+        "qiskit-aer",
+        "qiskit-ignis",  # TODO: remove this?
+        "qiskit-aqua",  # TODO: remove this?
```
kt474, line 62: Yeah I think these can be removed since both of those repos have been deprecated for a while

kt474 (commented):
Mostly looks good to me, I assume this has already been tested on the server side.
Also the analytics team may be querying these qiskit versions so I'll let them know the format will be updated

qiskit_ibm_runtime/api/session.py
```diff
@@ -264,6 +282,26 @@ def request(  # type: ignore[override]
         headers = self.headers.copy()
         headers.update(kwargs.pop("headers", {}))
 
+        # Set default caller
+        headers.update(
+            {"X-Qx-Client-Application": "{}/qiskit".format(CLIENT_APPLICATION)}
+        )
```
jyu00, line 288: Wouldn't this overwrite any [custom headers](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/api/session.py#L215-L222)?

JakWai01: Good catch! Should be fixed in [6c0078d](https://github.com/Qiskit/qiskit-ibm-runtime/pull/733/commits/6c0078d289be112c7030d99792d17bf93820ea0f)

qiskit_ibm_runtime/api/session.py
JakWai01: Done!

qiskit_ibm_runtime/api/session.py
```diff
@@ -51,20 +54,33 @@
...
+        "qiskit-aer",
+        "qiskit-experiments",
+        "qiskit-nature",
+        "qiskit-machine-learning",
+        "qiskit-optimization",
+        "qiskit-finance",
+    ]
+
+    pkg_versions = {
+        "qiskit-ibm-runtime": "{}-{}".format("qiskit-ibm-runtime", ibm_runtime_version)
```
johker01, line 69:
```suggestion
        f"qiskit-ibm-runtime-{ibm_runtime_version}"
```

The above suggestion would be an alternative, which in my opinion is easier to read, however, I don't have any preference on whether to keep the line as it is or change it.

qiskit_ibm_runtime/api/session.py
```diff
@@ -51,20 +54,33 @@
...
+    }
     for pkg_name in qiskit_pkgs:
         try:
-            pkg_versions[pkg_name] = pkg_resources.get_distribution(pkg_name).version
+            version_info = "{}-{}".format(
```
johker01, line 73: you could also use an f-string here. I don't know which way is more pythonic, though.

qiskit_ibm_runtime/api/session.py
```diff
@@ -264,6 +272,35 @@ def request(  # type: ignore[override]
         headers = self.headers.copy()
         headers.update(kwargs.pop("headers", {}))
 
+        # Set default caller
+        headers.update(
+            {"X-Qx-Client-Application": "{}/qiskit".format(CLIENT_APPLICATION)}
+        )
+
+        # Use PurePath in order to support arbitrary path formats
+        caller_dict = {PurePath("qiskit/algorithms"): "qiskit-terra-algorithms"}
```
jyu00, line 281: This would only tells me that qiskit algorithms module was used. But I'd like to know a bit more. So instead of looking for just `qiskit/algorithms`, perhaps look for anything that starts with `qiskit_` in addition to `qiskit` itself. This would cover other packages like `qiskit_machine_learning`. In addition, perhaps just grab the short path of the last caller instead of using a hard coded value. 

For example, from VQE the calling path would be like 
```
/Users/jessieyu/.pyenv/versions/3.9.1/envs/sandbox-3.9/lib/python3.9/site-packages/qiskit/algorithms/minimum_eigensolvers/vqe.py
/Users/jessieyu/.pyenv/versions/3.9.1/envs/sandbox-3.9/lib/python3.9/site-packages/qiskit/algorithms/optimizers/spsa.py
/Users/jessieyu/.pyenv/versions/3.9.1/envs/sandbox-3.9/lib/python3.9/site-packages/qiskit/algorithms/optimizers/spsa.py
/Users/jessieyu/.pyenv/versions/3.9.1/envs/sandbox-3.9/lib/python3.9/site-packages/qiskit/algorithms/optimizers/spsa.py
/Users/jessieyu/.pyenv/versions/3.9.1/envs/sandbox-3.9/lib/python3.9/site-packages/qiskit/algorithms/minimum_eigensolvers/vqe.py
```

And this can grab and save the last one `qiskit/algorithms/minimum_eigensolvers/vqe.py`. Make sure you save only the Qiskit module path and not anything before that (so we don't accidentally include PII).

JakWai01: Do I understand it correctly that you want to store the whole path e.g. `qiskit/algorithms/minimum_eigensolvers/vqe.py` as the caller? This would require some sort of transformation/escaping since we use the `/` character as a delimiter in our code. That is why we hard coded the different "callers" in oder to give them aliases.

qiskit_ibm_runtime/api/session.py
```diff
@@ -264,6 +270,30 @@ def request(  # type: ignore[override]
...
+
+        # Use PurePath in order to support arbitrary path formats
+        callers = {PurePath("qiskit/"), "qiskit_"}
+
+        for frame in inspect.stack():
+            frame_path = str(PurePath(frame.filename))
+            for caller in callers:
+                if str(caller) in frame_path:
+                    caller_str = str(caller) + frame_path.split(str(caller), 1)[-1]
+                    sanitized_caller_str = caller_str.replace("/", "~")
```
jyu00, line 284: nit - can we use `.` instead of `~`? After looking at your example, `qiskit_nature.second_q.algorithms.ground_state_solvers.ground_state_eigensolver.py` would looks a lot nicer lol 

Also I think we can break out of both for loops once a matching str is found?

test/unit/test_account_client.py
```diff
@@ -46,24 +45,6 @@ def _get_client(self):
         )
         return AccountClient(params)
 
-    def test_custom_client_app_header(self):
```
jyu00, line 49: Why is the test removed instead of updated? We still accept custom header.

qiskit_ibm_runtime/api/session.py
JakWai01: Using `.` instead of `~` would create "collisions" in every single path due to file extensions including a `.` e.g. `ground_state_eigensolver.py`. It would be unnecessarily complicated to replace all the "right" `.` with `/` again when e.g. resubstituting the characters in the backend. 
Although we  could do this a split called on the string would create weird segments as well. One might expect that you get all the path segments when you call split with our chosen delimiter but with a `.`, you would get the `py` as an unexpected segment as well.

I don't think that we can break out of the loops. We are iterating through the stacktrace which consists out of many calls in `qiskit-ibm-runtime` and many of them also match the criteria. We want the last matching one since that one should be the one that is closest to the user.

test/unit/test_account_client.py
JakWai01: Yes, we still use custom headers but the way that they need to be set now can't be tested as easily as it was done before (since the custom headers are declared at a different point in the stack) and AFAIK we can't test there without massive amounts of mocking. Since I wasn't able to figure out how to write said test, I decided to just remove it. It might be smart to write one in the near future :)

qiskit_ibm_runtime/api/session.py
jyu00: > when e.g. resubstituting the characters in the backend.

I don't think we are doing any resubstitution in the backend? I think it's just stored in the DB and then be queried. But I don't feel too strongly if you don't want to change it.

> We want the last matching one since that one should be the one that is closest to the user.

But isn't the list returned by `inspect.stack()` already ordered by closest to the user first?

test/unit/test_account_client.py
jyu00: We already have a mock server [SimpleServer](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/unit/mock/http_server.py#L86) that's used in this test module. I think you can just send a request to that and have some code to inspect the header received by the server. You can also test the changes you made here that way.

qiskit_ibm_runtime/api/session.py
JakWai01: So far, we are not doing any resubstitution but I'd love to at least have the option which is why I think I will leave this part as is. 

If we take the first one of the stack, we would always get `qiskit-ibm-runtime/qiskit_ibm_runtime/api/session.py` which isn't the result we want.

JakWai01: Good point 😄 The latest commit implemented this change request.

test/unit/test_account_client.py
JakWai01: I managed to write just like the one before but I needed to call the `_set_custom_header` method manually this time. I think this should still be sufficient though.

jyu00 (approved):
Looks like there's some formatting issue but otherwise LGTM!

PR #739, opened 2023-03-06
Use Vale to lint docs writing

frankharkins:
### Summary

Uses [Vale](https://vale.sh) to lint writing in the docs. The most notable impact is the ability to catch spelling errors during CI, but this tool can also help enforce writing style rules.

To try it, [install Vale](https://vale.sh/docs/vale-cli/installation/) and run

```sh
make docs-test
```

(you may also need to `pip install docutils` if you don't have it already).

Most developers should be able to ignore this test (that is, no need to install Vale), and should only be impacted if a documentation change introduces an error, which is usually easily fixable.

### Details and comments

The rules used are from [IBM/ibm-quantum-style-guide](https://github.com/IBM/ibm-quantum-style-guide).

Similar to pylint, you can turn off rules for specific sections [using comments](https://vale.sh/docs/topics/config/#markup-based-configuration). Please [make an issue](https://github.com/IBM/ibm-quantum-style-guide/issues/new/choose) if a rule is not working correctly so we can improve it.

### How to read this PR

I've split this into three separate commits to make review easier:
- caab66aa417296515227123f7cd8c4cabf35df83 adds the ability to use Vale to lint the docs
- 0269ef6236cb26be738024b3e227188fdae87e83 edits the docs to adhere to the linter
- 10db466ec0d427de7857bd95a4b38dd637028b70 adds a linting test to the CI (will error on spelling mistakes and latin abbreviations)
- (Unrelated to this PR) 16752f30f0f64a1c572f3b5d688a645426fe43cc runs black on the notebooks so the CI no longer fails

***

In collaboration with @beckykd.


.github/workflows/ci.yml
```diff
@@ -55,8 +55,18 @@ jobs:
...
           pip install -e .
+
+          wget https://github.com/errata-ai/vale/releases/download/v2.23.0/vale_2.23.0_Linux_64-bit.tar.gz
+          mkdir $HOME/bin && tar -xf vale_2.23.0_Linux_64-bit.tar.gz -C $HOME/bin
+          export PATH=$HOME/bin:"$PATH"
+          echo "$HOME/bin" >> $GITHUB_PATH
+      - name: Lint documentation
+        run: |
+          echo "$PATH"
+          make docs-test
```
kt474, line 69: I've gone through the vale documentation but is this the only way to download/install? Also I think the echo statements can be removed

frankharkins: I'm installing like this to get the latest version. Once this is on the snap store, we can install via
```sh
sudo snap install vale --edge
```

.github/workflows/ci.yml
```diff
@@ -55,8 +55,18 @@ jobs:
...
           sudo apt install -y graphviz pandoc
           pip install -e .
+
+          wget https://github.com/errata-ai/vale/releases/download/v2.23.0/vale_2.23.0_Linux_64-bit.tar.gz
+          mkdir $HOME/bin && tar -xf vale_2.23.0_Linux_64-bit.tar.gz -C $HOME/bin
+          export PATH=$HOME/bin:"$PATH"
+          echo "$HOME/bin" >> $GITHUB_PATH
+      - name: Lint documentation
+        run: |
+          echo "$PATH"
```
frankharkins, line 68:
```suggestion
      - name: Lint documentation
        run: |
```

kt474 (approved):
Looks good on my end - If @beckykd is on board with these changes, we can merge

beckykd (approved):
Excited to add this process!

PR #744, opened 2023-03-09
Add note to session docs

### LINKED ISSUE #742: Add warning about mixing session and reservation in docs

Apparently you cannot start a session inside a reservation. For example, if you have

-- reservation starts --
job 1 inside session A
job 2 inside session A
-- reservation ends --
job 3 inside session A

Job 3 will actually be the one to start the session. That is, it will go through the queue as if it was the first job in the session. 

**Acceptance criteria**

Document this behavior to set expectations.

kt474:
### Details and comments
Fixes #742


beckykd (commented):
Sorry, I missed this earlier.  Looks fine to me!

PR #747, opened 2023-03-13
Update comparison table (`backend.run` vs primitives)

garrison:
### Summary

I updated the comparison table for `backend.run` vs. the primitives.

### Details and comments

With more and more examples of how to use Primitives, I believe that it makes sense to show "Yes" for both methods to "Working examples of code developed by Qiskit community"

I also added dynamic circuits to the table.  My current understanding is that the Sampler does not (currently) support dynamic circuits, so users should use `backend.run` if they need them.


kt474 (approved):
LGTM - Thanks @garrison!

PR #751, opened 2023-03-14
Fix error level default logic 

### LINKED ISSUE #750: resilience and optimization level 0 used by default
**Describe the bug**

It looks like #701 inadvertently changed the default resilience level to 0. The PR used

```
        if _options.resilience_level is None:
            if _options.simulator and (
                not hasattr(_options.simulator, "noise_model")
                or asdict(_options.simulator)["noise_model"] is None
            ):
                _options.resilience_level = 0
            else:
                _options.resilience_level = Options._DEFAULT_RESILIENCE_LEVEL
```

If no options are passed in to Estimator's or Sampler's constructor, [`Options()` is used](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/estimator.py#L167-L168), which sets `_options.simulator` to a new `SimulatorOptions` with  `noise_model=None`. This caused the code above to always set the resilience level (and optimization level in separate code) to 0.

**Steps to reproduce**

```
from qiskit_ibm_runtime import Sampler

backend = service.backend("ibmq_manila")
s = Sampler(session=backend)
print(s.options.resilience_level)
```

**Expected behavior**

Default should not be 0.

**Suggested solutions**

Check if the backend used is actually a simulator (e.g. via `backend.configuration().simulator` if an `IBMBackend` was passed) instead of using the simulator options. When in doubt, use the default value instead of 0. 

Also looks like more tests are needed if the existing ones didn't catch the problem. 

**Additional Information**

- **Python version**:
- **Operating system**:

kt474:
### Details and comments
Fixes #750

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -197,19 +197,35 @@ class Estimator(BaseEstimator):
             skip_transpilation
         )
 
+        if isinstance(session, Session):
+            self._session = session
+        else:
+            backend = session or backend
+            self._session = get_default_session(service, backend)
+
+        backend_obj = self._session._service._backends.get(self._session._backend)
+
         if _options.optimization_level is None:
-            if _options.simulator and (
-                not hasattr(_options.simulator, "noise_model")
-                or asdict(_options.simulator)["noise_model"] is None
+            if (
+                backend_obj
+                and backend_obj.configuration().simulator
+                and (
+                    not hasattr(_options.simulator, "noise_model")
+                    or asdict(_options.simulator)["noise_model"] is None
+                )
             ):
                 _options.optimization_level = 1
             else:
                 _options.optimization_level = Options._DEFAULT_OPTIMIZATION_LEVEL
 
         if _options.resilience_level is None:
-            if _options.simulator and (
-                not hasattr(_options.simulator, "noise_model")
-                or asdict(_options.simulator)["noise_model"] is None
+            if (
+                backend_obj
+                and backend_obj.configuration().simulator
+                and (
+                    not hasattr(_options.simulator, "noise_model")
+                    or asdict(_options.simulator)["noise_model"] is None
+                )
             ):
                 _options.resilience_level = 0
             else:
@@ -222,12 +238,6 @@ class Estimator(BaseEstimator):
             "observables": observables,
             "parameters": parameters,
         }
-
-        if isinstance(session, Session):
-            self._session = session
-        else:
-            backend = session or backend
-            self._session = get_default_session(service, backend)
 
         # self._first_run = True
         # self._circuits_map = {}
--- a/qiskit_ibm_runtime/sampler.py
+++ b/qiskit_ibm_runtime/sampler.py
@@ -172,19 +172,35 @@ class Sampler(BaseSampler):
             skip_transpilation
         )
 
+        if isinstance(session, Session):
+            self._session = session
+        else:
+            backend = session or backend
+            self._session = get_default_session(service, backend)
+
+        backend_obj = self._session._service._backends.get(self._session._backend)
+
         if _options.optimization_level is None:
-            if _options.simulator and (
-                not hasattr(_options.simulator, "noise_model")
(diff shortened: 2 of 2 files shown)


qiskit_ibm_runtime/estimator.py
```diff
@@ -197,19 +197,35 @@ def __init__(
             skip_transpilation
         )
 
+        if isinstance(session, Session):
+            self._session = session
+        else:
+            backend = session or backend
+            self._session = get_default_session(service, backend)
+
+        backend_obj = self._session._service._backends.get(self._session._backend)
```
jyu00, line 206:
```suggestion
        backend_obj = self._session.service.backend(self._session.backend())
```
Should avoid using private function/property when public ones are available. But it also means you'd need to check if the session actually has a backend.

qiskit_ibm_runtime/estimator.py
```diff
@@ -197,19 +197,35 @@ def __init__(
...
-            if _options.simulator and (
-                not hasattr(_options.simulator, "noise_model")
-                or asdict(_options.simulator)["noise_model"] is None
+            if (
+                backend_obj
+                and backend_obj.configuration().simulator
+                and (
+                    not hasattr(_options.simulator, "noise_model")
+                    or asdict(_options.simulator)["noise_model"] is None
+                )
```
jyu00, line 215: This logic really should be moved to `run()` because one can overwrite `noise_model` in `run()`. Also can have a utility function instead of repeating the same code.

test/integration/test_options.py
```diff
@@ -127,3 +131,19 @@ def test_optimization_and_resilience_levels(self, service):
...
+            primitive_simulator = primitive(session=backend)
+            self.assertEqual(primitive_simulator.options.resilience_level, 0)
+            self.assertEqual(primitive_simulator.options.optimization_level, 1)
+
+            if real_backend:
+                primitive_real_backend = primitive(session=real_backend)
+                self.assertEqual(
+                    primitive_real_backend.options.resilience_level,
+                    Options._DEFAULT_RESILIENCE_LEVEL,
+                )
```
jyu00, line 145: Since we have a [fake runtime client](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/unit/mock/fake_runtime_client.py), a better test would be to ensure the value sent to the "server" has the right level, and not just what's recorded locally. This would also avoid having to run on a real backend.

jyu00 (changes requested):

test/integration/test_options.py
merav-aharoni: You might want to have a look at the tests I wrote for this purpose - I split the test `test_optimization_and_resilience_levels` into two to cover more cases. See 
https://github.com/Qiskit/qiskit-ibm-runtime/pull/752/files#diff-9f1396626262723a7bef9a07fc75f30acccb1dd18cd8ea41498a7f050a265563

qiskit_ibm_runtime/estimator.py
```diff
@@ -268,6 +251,15 @@ def run(  # pylint: disable=arguments-differ
             ValueError: Invalid arguments are given.
         """
         # To bypass base class merging of options.
+        if "noise_model" not in kwargs:
```
jyu00, line 254: `kwargs` here actually accepts both structured and flat dictionaries, e.g. 

```
run(noise_model=foo)
```

and 

```
run(simulator={"noise_model": foo})
```

So it's probably easier to do this logic in `_run()` after `_merge_options()`, which combines everything into a "structured" dictionary.

It's done this way b/c we want to support both auto-complete with `Options` (structured) and flat dictionary as defined by the base class, _and_ `asdict(options)`.

qiskit_ibm_runtime/estimator.py
```diff
@@ -268,6 +251,15 @@ def run(  # pylint: disable=arguments-differ
             ValueError: Invalid arguments are given.
         """
         # To bypass base class merging of options.
```
jyu00, line 253: This comment is for line 263 and should stay with 263.

PR #752, opened 2023-03-15
[WIP] Fix default values for optimization_level and resilience_level

merav-aharoni:
… backend and not at the options

### Details and comments
Fixes #


merav-aharoni:
See #751

PR #756, opened 2023-03-20
Add support for OQ2 and OQ3 programs in primitives

mberna:
The main idea is that, in addition to QuantumCircuit objects, the primitives accept now a OpenQASM programs (as a string).

This implementation first uses a regex to determine the version of OpenQASMS in the program and then uses QuantumCircuit.from_qasm_str (version 2) or qiskit.qasm3.loads (version 3 and default version) to convert the input strings to QuantumCircuit objects. After the conversion is finished, the behaviour of the primitives is the usual one.


requirements.txt
```diff
@@ -1,4 +1,4 @@
-qiskit-terra>=0.23.1
+qiskit-terra[qasm3-import]>=0.23.1
 requests>=2.19
```
kt474, line 2: I think this will have to be added to `setup.py` as well

kt474 (commented):
Looks pretty good to me, will just need a release note

requirements.txt
mberna: Thanks! I added it in 246985e69bc72bc0210d226cac57cc411892910f

kt474:
@mberna what's the status on this? Does it need to be merged

kt474:
I'm going to close this for now since there have been no updates for over 3 months, feel free to reopen

PR #757, opened 2023-03-21
Issue warning when given an unsupported option

### LINKED ISSUE #754: Issue warning about unknown options

The `Options` class and its ` _merge_options()` method are designed to accept arbitrary keyword arguments. It should, however, issue a warning when an unrecognized option is used. This avoids a silence failure on syntax errors. 

**Acceptance criteria**

Issue a warning when an unrecognized option is specified.

merav-aharoni:
### Summary
Issue a user warning when given an unsupported option in Options.

### Details and comments
The warning is issued in `Options._get_program_inputs()`.
Fixes #754 .

### DIFF
--- a/qiskit_ibm_runtime/options/options.py
+++ b/qiskit_ibm_runtime/options/options.py
@@ -15,6 +15,7 @@
 from typing import Optional, Union, ClassVar
 from dataclasses import dataclass, fields, field
 import copy
+import warnings
 
 from .utils import _flexible, Dict
 from .environment_options import EnvironmentOptions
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/options/options.py
```diff
@@ -148,6 +149,9 @@ def _get_program_inputs(options: dict) -> dict:
         # Add additional unknown keys.
         for key in options.keys():
             if key not in known_keys:
+                warnings.warn(
+                    f"Key {key} not in available options. It will be ignored."
```
kt474, line 153:
```suggestion
                    f"Key '{key}' is an unrecognized option. It may be ignored."
```
Since arbitrary keyword arguments can be accepted by design (for testing) I think this should be reworded

kt474 (approved):
LGTM

PR #764, opened 2023-03-23
[WIP] Accept all hierarchical options if given flatly

### LINKED ISSUE #755: Accept all options in a dictionary as first level

The options one can pass to the runtime primitives are structured so that each falls under a category. For example, `shots` is under `execution`. This is true even when options are passed as a dictionary:

```
# You can do this
options = Options()
options.execution.shots = 1024

# Or this
options = {"execution": {"shots": 1024}}
```

It was designed this way to allow easier auto-complete when one simply does `options.<tab>`. However, this is inconsistent with the base primitive classes in qiskit-terra (which came after) that uses a flat structure:

```
options = {"shots": 1024}
```

This means users need to know to change their code when switching from terra (or aer) primitives to runtime primitives. Furthermore, if one does `options={"shots": 1024}`, the shot number will be silently ignored (see #754). 

**Acceptance criteria**

Allow users to specify all options as "first level" options, when passing it as a dictionary. For example, 

```
from qiskit_ibm_runtime import Estimator

estimator = Estimator(..., options={"shots": 1024})
```

The ` _merge_options()` can be used to do this.

merav-aharoni:
### Summary
Accept all options on given on level 1 and assign them to the appropriate hierarchical option type.

### Details and comments
For example, if the user defines `options = "shots": 1024}` as input to one of the primitives, we translate it to `options = {"execution": {"shots": 1024}}`.
Fixes #755.

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -180,11 +180,11 @@ class Estimator(BaseEstimator):
                     version="0.7",
                     remedy="Please pass the backend when opening a session.",
                 )
-            skip_transpilation = options.get("transpilation", {}).get(
-                "skip_transpilation", False
-            )
             default_options = asdict(Options())
             _options = Options._merge_options(default_options, options_copy)
+            skip_transpilation = _options.get("transpilation", {}).get(
+                "skip_transpilation", False
+            )
 
         _options["transpilation"][
             "skip_transpilation"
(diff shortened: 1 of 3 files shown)


releasenotes/notes/accept_level_1_options-06c09a67a9c23feb.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade:
```
jyu00, line 2: Personally I'd list this under `features` instead of `upgrade`, but it's not a big deal.

jyu00 (approved):
LGTM!

kt474 (approved):

PR #772, opened 2023-03-27
Set error levels to default when no backend is selected

kt474:
### Summary

As @merav-aharoni mentioned in https://github.com/Qiskit/qiskit-ibm-runtime/pull/748#issuecomment-1484110941, if `self._session.backend()` does not exist, the optimization and resilience levels will be set to `None` which would fail the job.

This could happen on cloud when no backend is selected, or when the sampler/estimator are used as sessions 

tests should be covered in https://github.com/Qiskit/qiskit-ibm-runtime/pull/748

### Details and comments
Fixes #


merav-aharoni (approved):
LGTM. I will test PR #748 after this one is merged.

PR #741, opened 2023-03-07
[WIP] Use qiskit-ibm-provider for duplicated code 

kt474:
### Details and comments
Fixes #548

PR #774, opened 2023-03-28
Retrieve backends with runtime api 

kt474:
### Summary

Related to #741 which is a much bigger effort to minimize duplicated code between provider repos.

Since #741 will likely not be ready this week, this PR only makes updates to use the runtime api instead of the iqx api to retrieve backends 

### Details and comments
Fixes #


qiskit_ibm_runtime/api/clients/runtime.py
```diff
@@ -342,13 +342,13 @@ def close_session(self, session_id: str) -> None:
 
     # IBM Cloud only functions
 
-    def list_backends(self) -> List[str]:
+    def list_backends(self, hgp: Optional[str] = None) -> List[str]:
         """Return IBM Cloud backends available for this service instance.
 
         Returns:
             IBM Cloud backends available for this service instance.
```
kt474, line 349: these docstring and comments should be updated since they are not cloud only

jyu00 (approved):
LGTM (other than the docstring you pointed out)! This should help with lazy loading as well.

Issue #641, opened 2022-12-06
resilience_levels 2 & 3 turns off mitigation for Sampler


nonhermitian:
Turning on `resilience_level` 2 or 3 in the `Sampler` disables readout mitigation, i.e. `resilience_level=1` functionality.  No other mitigation looks to be applied.  This is counter to the assumption that the higher the resilience level the more mitigation is performed.

I could not find any documentation that explicitly mentions this.

nonhermitian:
Actually I found it here: https://qiskit.org/documentation/partners/qiskit_ibm_runtime/how_to/error-mitigation.html#configure-error-mitigation (my screen just cut off the side of the table). 

Ideally, rather than no mitigation, these higher levels would keep M3 turned on.

king-p3nguin:
When I used the `Sampler` for the first time, I did not know that setting the `resilience_level` higher than one does not perform any error mitigation until I read the document. I would instead prefer raising an error when `resilience_level > 1`, just like when I set `resilience_level = 4` for `Estimator`.
```
--- Logging error ---
Traceback (most recent call last):
File "/provider/programruntime/program_starter_wrapper.py", line 88, in execute
  final_result = self.main(backend, self.messenger, **self.user_params)
File "/code/./program.py", line 1460, in main
  raise QiskitError(f"Resilience level {resilience_level} not supported.")
qiskit.exceptions.QiskitError: 'Resilience level 4 not supported.'
```

PR #781, opened 2023-03-29
Validate input to Options

### LINKED ISSUE #767: Input validation

Unlike the old IQX path, Qiskit Runtime doesn't do any input validation on the server side prior to adding jobs to the queue. This means a user can wait hours or even days just to have their job fail due to a typo. Some basic validation on the client side could reduce some of the pain. 

**Acceptance criteria**

Things to validate:
1. `resilience_level` and `optimization_level` should be within range for the primitive being called
1. `noise_amplifier` and `extrapolator` resilience options are one of those [supported values](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/options/resilience_options.py#L21-L32). 
2. `noise_factors` has at least 5 entries if `QuarticExtrapolator` is used; at least 4 if `CubicExtrapolator` is used
3. If `resilience_level=3` is specified and the backend is a simulator, a `coupling_map` must also be specified in simulator options.

A `ValueError` should be raise if the validation fails. However, there should be an env variable that one can set to bypass validation (e.g. for testing).

merav-aharoni:
### Summary
Added a few validations for inputs to Options.

### Details and comments
The validations are called from `Options._get_program_inputs`, i.e., they are called during `Sampler.run()/ Estimator.run()`.
All validations can be turned off by setting environment variable `QISKIT_RUNTIME_REMOVE_OPTIONS_VALIDATION`.
Validations performed in this PR:
- `optimization_level`
- `resilience_level`
- Specific options to `resilience`

Fixes #767, #641.


merav-aharoni:
In this PR, I only made specific validations for the values defined for `resilience`. I can do the same for the other option types as well. If I do the same for them, I will define the possible options of every category as `Literal`, like they are defined for `resilience`. @kt474 , @jyu00 - let me know what you think.

qiskit_ibm_runtime/estimator.py
```diff
@@ -317,7 +317,14 @@ def _run(  # pylint: disable=arguments-differ
...
         logger.info("Submitting job using options %s", combined)
-        inputs.update(Options._get_program_inputs(combined))
+        if self._session.backend():
+            inputs.update(
+                Options._get_program_inputs(
+                    combined, primitive="Estimator", backend=self._session.backend()
+                )
+            )
+        else:
+            inputs.update(Options._get_program_inputs(combined, primitive="Estimator"))
```
jyu00, line 327: `self._session.backend()` would return `None` if there is no backend, right? So there is no need for the if/else.

qiskit_ibm_runtime/options/options.py
```diff
@@ -106,15 +115,20 @@ class Options:
...
+    def _get_program_inputs(
+        options: dict, primitive: PrimitiveType, backend: IBMBackend = None
+    ) -> dict:
         """Convert the input options to program compatible inputs.
 
         Returns:
             Inputs acceptable by primitive programs.
         """
+        if not os.getenv("QISKIT_RUNTIME_REMOVE_OPTIONS_VALIDATION"):
+            Options._validate_program_inputs(options, primitive, backend)
```
jyu00, line 131: I'd recommend calling `_validate_program_inputs` separately from `_get_program_inputs` as to not complicate the functionality of `_get_program_inputs` which just converts options to program inputs.

qiskit_ibm_runtime/options/options.py
```diff
@@ -155,6 +169,56 @@ def _get_program_inputs(options: dict) -> dict:
                 inputs[key] = options[key]
         return inputs
 
+    @staticmethod
+    def _validate_program_inputs(
```
jyu00, line 173:
```suggestion
    def _validate_options(
```
nit - "options" is what users specify, and "program inputs" is what go to the runtime programs. They are different to keep backward compatibility. This function is validating the options, not the program inputs.

qiskit_ibm_runtime/options/options.py
```diff
@@ -155,6 +169,56 @@ def _get_program_inputs(options: dict) -> dict:
...
+        options: dict, primitive: PrimitiveType = "Sampler", backend: IBMBackend = None
```
jyu00, line 174: Instead of passing in the primitive name and have branches, perhaps it makes more sense for `Sampler` and `Estimator` classes to have their own `_validate_options` methods that call a general method (i.e. this one) and then do their own primitive-specific validation.

qiskit_ibm_runtime/options/options.py
```diff
@@ -155,6 +169,56 @@ def _get_program_inputs(options: dict) -> dict:
...
+            and options.get("resilience_level")
+            and not options.get("resilience_level") in [0, 1]
+        ):
+            raise ValueError(
+                f"resilience level can only take the values "
+                f"{list(range(Options._MAX_RESILIENCE_LEVEL_SAMPLER + 1))} in Sampler"
+            )
+        if (
+            options.get("resilience_level")
+            and options.get("resilience_level") == 3
```
jyu00, line 208: The first statement (line 207) seems unnecessary as the second one would fail if it's None.

qiskit_ibm_runtime/options/options.py
```diff
@@ -155,6 +169,56 @@ def _get_program_inputs(options: dict) -> dict:
...
+            and backend.simulator
+        ):
+            if (
+                not options.get("simulator").get("coupling_map")
+                or options.get("simulator").get("coupling_map") is None
```
jyu00, line 213:
```suggestion
                not options.get("simulator", {}).get("coupling_map")
```

jyu00:
> In this PR, I only made specific validations for the values defined for `resilience`. I can do the same for the other option types as well. If I do the same for them, I will define the possible options of every category as `Literal`, like they are defined for `resilience`. @kt474 , @jyu00 - let me know what you think.

Sure. I would have preferred an enum _and_ a literal (so the values are not hardcoded in the validation), but there doesn't seem to be a way to convert one to another.

qiskit_ibm_runtime/options/options.py
merav-aharoni: Good idea!

merav-aharoni:
> > In this PR, I only made specific validations for the values defined for `resilience`. I can do the same for the other option types as well. If I do the same for them, I will define the possible options of every category as `Literal`, like they are defined for `resilience`. @kt474 , @jyu00 - let me know what you think.
> 
> Sure. I would have preferred an enum _and_ a literal (so the values are not hardcoded in the validation), but there doesn't seem to be a way to convert one to another.

I will do this in a separate PR.

merav-aharoni:
@jyu00 - thanks for all your suggestions. I incorporated them all. Can you please do another review?

jyu00 (approved):
Some minor comments but otherwise LGTM

test/unit/test_estimator.py
```diff
@@ -82,3 +83,20 @@ def test_estimator_circuit_caching(self):
...
+        with Session(
+            service=FakeRuntimeService(channel="ibm_quantum", token="abc"),
+            backend="common_backend",
+        ) as session:
+            circuit = QuantumCircuit(1, 1)
+            obs = SparsePauliOp.from_list([("I", 1)])
+            for bad_opt in options_bad:
+                inst = Estimator(session=session)
+                with self.assertRaises(ValueError):
+                    _ = inst.run(circuit, observables=obs, **bad_opt)
```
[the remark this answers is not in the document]
jyu00: @merav-aharoni Can you also add this message verification to the other tests that have the same `self.assertRaises`?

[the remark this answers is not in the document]
merav-aharoni: I added this to all the new tests with the exception of `test_resilience_options` in `test_ibm_primitives.py`, because that has very diverse error messages. I will try to find a way to do that there as well. Was there any other place you think I missed?

kt474 (approved):

PR #791, opened 2023-04-04
Lazy load backend configurations 

### LINKED ISSUE #272: service initialization too slow
For credentials that are saved on disk, running this line takes 13 seconds for me.
`service = QiskitRuntimeService()`

I hope this can be improved with lazy loading. There is a similar issue open in `qiskit-ibm-provider`: https://github.com/Qiskit/qiskit-ibm-provider/issues/84

kt474:
### Summary

~~still need to test, not confident this works correctly yet - having trouble getting instances to work properly~~

need to fix unit tests 

### Details and comments
Fixes #272


test/unit/mock/fake_runtime_service.py
```diff
@@ -82,20 +100,28 @@ def _initialize_hgps(
...
+            for backend_name, backend in hgp.backends.items():
+                if not backend:
+                    self._set_backend_config(backend_name, hgp_name)
+                    backend_obj = ibm_backend.IBMBackend(
+                        instance=hgp_name,
+                        configuration=self._backend_configs[backend_name],
+                        service=self,
+                        api_client=hgp._runtime_client,
+                    )
+                    hgp.backends[backend_name] = backend_obj
```
kt474, line 123: Not sure what the best way to update these unit tests is without significant refactoring - thoughts? @jyu00

jyu00: From what I can tell this `_initialize_hgps()` method doesn't actually collect the backend configuration. It just initializes a fake account client, which will need to be changed to fake runtime client, since that's what is used now. The fake runtime client will also need to be updated to take the "spec", to know what backends to return.

Then in theory when the mainline code calls `self._api_client.backend_configuration`, it should go to the FakeRuntimeClient to get the backend config.

jyu00 (commented):
Did you skip lazy loading of cloud channel backends to keep things simple? If so, can you open a follow up issue so we don't forget?

test/unit/mock/fake_runtime_service.py
kt474: I believe the hgps on the `FakeRuntimeService` side will still need the backend configs somehow because when jobs are retrieved, the service will initialize without any backends 

then `self._api_client.backend_configuration` would result in in a backend cannot be found error

jyu00 (approved):
@kt474 if you are good with my changes, specifically the ones in `HubGroupProject`, then we are all set to go!

PR #794, opened 2023-04-07
Check for faulty qubits and edges

### LINKED ISSUE #792: Raise early if circuit uses faulty qubits or gates

Currently, if a user submits a circuit which uses gates or qubits which are marked as faulty, they will get an opaque `IBMJobFailureError` due to an error in transpilation.

It should be possible, when backends have faulty qubits or gates, to scan submitted circuits and raise early if there is an instruction on a faulty gate or qubit. This can help users avoid waiting in the queue for jobs which cannot run successfully, and gives an opportunity to inform the user to either use different physical qubits, or re-transpile their circuit with a different seed or settings .

N.B. https://github.com/Qiskit/qiskit-ibm-provider/issues/568 raises the same concern for `qiskit-ibm-provider`, but as far as I can tell, this would have to be addressed in both the `provider` and the `runtime`. Feel free to close one of them if that's not the case.

**Acceptance criteria**

jyu00:
### Summary

Closes #792

Checks if the input circuits use a faulty qubit or edge and raises if they do.

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -305,6 +305,8 @@ class Estimator(BaseEstimator):
         }
 
         combined = Options._merge_options(self._options, kwargs.get("_user_kwargs", {}))
+
+        backend_obj: Optional[IBMBackend] = None
         if self._session.backend():
             backend_obj = self._session.service.backend(self._session.backend())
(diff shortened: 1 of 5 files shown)


kt474 (approved):
LGTM

PR #818, opened 2023-05-02
Print nicer error messages

### LINKED ISSUE #811: No error message from provider but on IQP
**Describe the bug**

When my job `ch1ebkol1tcq36kffct0` failed, the IQP dashboard had a nice error message

image

But `job.result()` doesn't tell me anything:

```
RuntimeJobFailureError: 'Unable to retrieve job result. Job ch1ebkol1tcq36kffct0 has failed:\n'
```

**Steps to reproduce**

**Expected behavior**

I should get the same error message back.

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.9.3
- **Python version**:
- **Operating system**:

merav-aharoni:
### Summary
Print a `cleaner` and simpler error message to the user upon failure.

### Details and comments
Extract the `job._reason` field, which contains the reason for failure. Ignore the rest of the long error message.
Fixes #811.


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -220,7 +220,7 @@ def result(  # pylint: disable=arguments-differ
         if self._results is None or (_decoder != self._final_result_decoder):
             self.wait_for_final_state(timeout=timeout)
             if self._status == JobStatus.ERROR:
-                error_message = self.error_message()
+                error_message = self._reason if self._reason else self._error_message
                 if self._reason == "RAN TOO LONG":
```
kt474, line 224: Do we want to do something similar when `job.error_message()` is called? We could add the same logic to `_set_error_message()`

merav-aharoni: I am not sure. I don't think we should modify `self._error_message`. Do you think `job.error_message` should return the `_reason` rather than the entire `_error_message`?
In any case, the only place I saw this method called is in the tests.

kt474: Ah ok I see - yeah `result()` can return the reason and `error_message` can still return the full error message

kt474 (approved):

kt474 (approved):
Changes look good to me but I'm not sure if this fixes the original issue of the `Failed - RecursionError` not showing up on the client side. Were you able to verify this?

merav-aharoni:
> Changes look good to me but I'm not sure if this fixes the original issue of the `Failed - RecursionError` not showing up on the client side. Were you able to verify this?

I didn't actually produce a circuit that gives this particular error. My assumption was that this issue was not specific for this error, but general. I will try to build a circuit that fails with this message.

merav-aharoni:
> > Changes look good to me but I'm not sure if this fixes the original issue of the `Failed - RecursionError` not showing up on the client side. Were you able to verify this?
> 
> I didn't actually produce a circuit that gives this particular error. My assumption was that this issue was not specific for this error, but general. I will try to build a circuit that fails with this message.

I constructed a circuit as @jyu00 specified, but did not get an exception. Even went up to depth=40 and still no exception. So I didn't check this particular exception message.

jyu00:
Well I did say "before they fix it", and they did fix it 😄 so you won't get the `RecursionError` anymore. But yeah the issue was more general, and the `RecursionError` was just an example.

PR #823, opened 2023-05-04
Fix language toggle in docs

Eric-Arellano:
### Summary

The language toggle was broken for two reasons:

1. Sphinx 6 does not activate jquery by default anymore, so the interactivity was broken
2. The `versionutils.py` file should be setting `language_label`, not `version_label`

This PR also makes some other docs config improvements:

* Pin the sphinx theme version for more stability
* Move more docs dependencies from tox.ini to requirements-dev.txt to follow the convention of other Qiskit projects.
* Remove unnecessary pip install lines in tox.ini.
* Don't install sphinx-rtd-theme, which is superseded by qiskit-sphinx-theme
* Remove override of the `versions.html` template. We should use qiskit-sphinx-theme instead.


kt474 (approved):
LGTM

PR #824, opened 2023-05-04
Turn off CI for forks

Eric-Arellano:
### Summary

Most people agree that GitHub made a mistake by having forks run CI both on the fork and in the upstream repository. It wastes computing resources (which has a real environmental cost!).

### Details and comments

If people really want to run CI in their fork, they can always modify their fork's config to revert this change.


kt474 (approved):
LGTM!

PR #849, opened 2023-05-11
Remove unnecessary CSS files

Eric-Arellano:
* `gallery.css` is for `sphinx-gallery`, which is not used by this project.
* `style.css` is leftover from when we used the Read The Docs theme. It's no longer relevant to qiskit_sphinx_theme.


docs/conf.py
```diff
@@ -70,13 +70,11 @@
     'reno.sphinxext',
     'nbsphinx',
     'sphinx_design',
-    "sphinxcontrib.jquery",  # Remove when changing html_theme to qiskit_ecosystem.
+    "qiskit_sphinx_theme",
```
Eric-Arellano, line 73: `qiskit_sphinx_theme` already activates `sphinxcontrib.jquery`. This is more future proof for the upcoming Sphinx Theme 1.12, which will provide `versionutils.py` for you.

kt474 (approved):
LGTM!

Issue #851, opened 2023-05-12
The cost calculation fails with an expression


AzizNgoueya:
**Describe the bug**
When assigning a value containing an operation in the Options().max_execution_time, I got an error `{"code":1208,"message":"Invalid job request body: Key: \\\'ProgramRunParams.Cost\\\' Error:Field validation for \\\'Cost\\\' failed on the \\\'min\\\' tag.","solution":"Check the JSON schema.","more_info":"https://cloud.ibm.com/apidocs/quantum-computing#error-handling"}`

**Steps to reproduce**

```
options = Options()
options.max_execution_time = 4*60
sampler = Sampler(session = backend, options = options)
circuit = QuantumCircuit(1,1)
circuit.h(0)
circuit.measure_all()
job = sampler.run(circuit, shots = 1024) 
```

**Expected behavior**
It should work

**Suggested solutions**

`options.max_execution_time = 240` works

**Additional Information**

- **qiskit-ibm-runtime version**: 0.9.3
- **Python version**: 3.8.16
- **Operating system**: Mac OS

merav-aharoni:
`options.max_execution_time` must be at least 300 seconds. So actually both versions should fail. I suggest we improve the error message.

kt474:
fixed by #853

PR #853, opened 2023-05-14
Error message upon unsupported values for max_execution_time

merav-aharoni:
### Summary
Added error message when max_execution_time is outside the allowed range. Previously, this gave an error on the server side.

### Details and comments
Related to  #851.
Should actually have been part of #781.


kt474 (approved):
LGTM!

Issue #548, opened 2022-09-29
Use qiskit-ibm-provider


jyu00:
**What is the expected feature or enhancement?**

Now that `qiskit-ibm-provider` is released, we can include that as a dependency and reuse a lot of the same code, especially for API calls.

**Acceptance criteria**

Reuse same code in qiskit-ibm-provider.

jyu00:
@kt474 Why is this on hold?

kt474:
I thought it would be best to get https://github.com/Qiskit/qiskit-ibm-provider/issues/394 done first since there are still iqx api calls, but it's not a hard blocker. Removed the on hold label

kt474:
fixed by #871

PR #859, opened 2023-05-16
Use qiskit-ibm-provider

kt474:
### Summary

There are many pieces to remove/refactor so I will break it up into smaller PRs

Duplications removed

- [x] qpy
- [x] utils - `hgp`, `ibm_qubit_properties`, `backend_decoder` 
- [x] proxies 
- [x] `BackendReservation` - this file was never used, I think it was forgotten during the original migration  

I will take care of the api related code duplication in a separate PR - in #871 

### Details and comments
Fixes part of #548


merav-aharoni (approved):
Looks good! Will be nice to get rid of all these duplications.
You might as well remove the directories `qpy` and `proxies` while you're at it.

Issue #784, opened 2023-03-30
Remove deprecated features from >3 months ago and TODOs


jyu00:
**What is the expected feature or enhancement?**

Many features were deprecated in 0.7 release and can be removed now. In light of [this issue](https://github.com/Qiskit/qiskit-ibm-provider/issues/550), the TODOs in the code should be reviewed and actions taken accordingly. 

**Acceptance criteria**

- Remove features that had sufficient deprecation period per Qiskit's [deprecation policy](https://qiskit.org/documentation/deprecation_policy.html#:~:text=We%20consider%20a%20feature%20marked,merge%20new%20functionality%20to%20it.)
- Review/fix all TODOs

merav-aharoni:
@kt474 - deprecated code for which versions can be removed?

kt474:
Yes, it is a bit confusing - originally, I believe we wanted all options passed into `session.run()` to be honored, but in terms of `backend`, it never worked because sessions don't support cross multiple backends. This deprecation message was meant to just inform users not to pass in a backend in the options and the doc strings say the same thing. I'm ok with just leaving this deprecation message for now, as I don't believe any users are doing this anymore anyway. 

I agree that ideally, `backend` is not an option, but it would be a relatively big change unrelated to this original issue

jyu00:
We can just make a copy of the input `options` before adding `backend` to it. We really shouldn't be modifying a user's variable to begin with anyway. 

We can remove/privatize both `session.run()` and `service.run()` once everything becomes primitives only (i.e. no more custom programs).

kt474:
I believe #1093 handles the last deprecation warning we can remove, after it's in we can close this issue

jyu00:
closing this, but as part of the release process we should always check if there are deprecated features that can be removed.

PR #866, opened 2023-05-22
Remove deprecation warning  for unsupported parameters of transpilation settings

merav-aharoni:
### Summary
Removed 1 deprecation warning for unsupported parameters of `transpilation settings`.

### Details and comments
Part of #784 .
Question - regarding `translation_method` and `timing_constraints`, do we want to raise an error or simply ignore? Because for other unsupported options, we simply ignore.


kt474:
> Question - regarding translation_method and timing_constraints, do we want to raise an error or simply ignore? Because for other unsupported options, we simply ignore.

Since it's been more than 6 months, I think we can just ignore unsupported options

merav-aharoni:
When checking the removal of deprecation warning, I noticed that if an unsupported option is passed, the server will return an error (possibly after a long time). So I added validation checks for level 2 options. We already have checks for level 1 and level 3.

kt474 (approved):

jyu00:
We specifically allow kwargs in `Options` to allow people to test undocumented features. If you are going to raise, perhaps one can bypass that behavior using an env variable (I think there was one before, but maybe it's in qiskit-ibm-provider).

merav-aharoni:
@jyu00 - Indeed there is an environment variable that controls whether to run _validate_options or not. So the user can turn set `QISKIT_RUNTIME_SKIP_OPTIONS_VALIDATION=False` to turn off all the validations.

PR #871, opened 2023-05-24
Use qiskit-ibm-provider api code

kt474:
### Summary

Removing as much duplication as possible from the api code 

api/clients: auth.py, base.py, runtime_ws.py, version.py
api/rest: backend.py, base.py, program_job.py, root.py

### Details and comments
Fixes other half of #548


merav-aharoni (commented):
Overall looks fine. A few comments:

1. In `qiskit_ibm_runtime/api/clients/base.py`, `retries==8` and in `qiskit-ibm-provider/qiskit_ibm_provider/api/clients/base.py` it is `5`. Do you want to change it to 8 in provider?
2. There are quite a few differences between `qiskit_ibm_runtime/api/rest/backend.py` and `qiskit-ibm-provider/qiskit_ibm_provider/api/rest/backend.py`. Do we need to copy any of these to provider?
3. Same as (2) for `qiskit_ibm_runtime/api/rest/program_job.py` and `qiskit-ibm-provider/qiskit_ibm_provider/api/rest/program_job.py`

merav-aharoni (approved):

kt474:
> Overall looks fine. A few comments:
> 
1. Yes, we can also increase the retries in qiskit-ibm-provider
2. We don't need to copy any of the changes in `backend.py` because `job_limit` and `reservations are no longer used
3. I believe the same applies here, we don't need to copy the changes, because the `program_job.py` in qiskit-ibm-provider is more up to date 
4. There is no `reservations` endpoint in the runtime api, I don't believe it was ever used in this repo

PR #873, opened 2023-05-25
Overview updates

beckykd:
### Details and comments
Fixes #


docs/index.rst
```diff
@@ -5,11 +5,9 @@ Qiskit Runtime overview
 Overview
 ==============
 
-Qiskit Runtime is a quantum computing service and programming model that streamlines quantum computations. It is designed to use classical compute resources to execute quantum circuits with more efficiency on quantum processors. It improves efficiency by giving you access to *primitives*, which are designed to run in *sessions*.
+Qiskit Runtime is a cloud-based quantum computing service offered by IBM. It implements *primitives* to perform foundational quantum computing tasks and has built-in error suppression and mitigation. In addition, Qiskit Runtime has *sessions*, which allow you to run your circuits on a quantum computer without being interrupted by other users’ jobs. The combination of primitives, error suppression / mitigation, and sessions paves the way to efficiently build and execute scalable quantum applications.
```
blakejohnson, line 8: Perhaps instead of "it implements *primitives*" we could say "it exposes *primitives*", or "it exposes computational *primitives*".

docs/index.rst
```diff
@@ -19,26 +17,26 @@ Key concepts
 
 **Primitives**
 
-Primitives are core functions that provide a simplified interface for defining `near-time quantum-classical workloads <https://research.ibm.com/blog/near-real-time-quantum-compute>`_  required to efficiently build and customize applications. The initial release of Qiskit Runtime includes two primitives: Estimator and Sampler. They perform foundational quantum computing tasks and act as an entry point to the Qiskit Runtime service.
+Primitives are base level operations that serve as building blocks for many quantum algorithms and applications. The `base primitive interfaces <https://qiskit.org/documentation/apidoc/primitives.html>`__ are defined in Qiskit Terra, and many Qiskit algorithms use the primitives natively. This abstraction allows you to write the same code, using Qiskit algorithms or otherwise, that can run on different quantum hardware or simulators without having to explicitly manage some of the finer details. See `Introduction to primitives <primitives.html>`__ for further details.
```
blakejohnson, line 20: If we are going to introduce the term "base primitive" here, then we should follow that up by saying the primitves offered by the `qiskit_ibm_runtime` add additional options specific to IBM's service.

docs/index.rst
```diff
@@ -19,26 +17,26 @@ Key concepts
...
+
+**Error suppression / mitigation**
+
+Errors occur naturally in a computer, and building fault-tolerant quantum computation is our ultimate goal. While we continue to research how to build error-corrected qubits at scale, Qiskit Runtime offers a number of error suppression and mitigation techniques that alleviate the effect of noise.
 
 **Session**
 
-A session is a contract between the user and the Qiskit Runtime service that ensures that a collection of jobs can be grouped and jointly prioritized by the quantum computer’s job scheduler. This eliminates artificial delays caused by other users’ jobs running on the same quantum device during the session time.
+A session is a contract between the user and the Qiskit Runtime service that ensures that a collection of jobs can be grouped and jointly prioritized by the quantum computer’s job scheduler. This eliminates artificial delays caused by other users’ jobs running on the same quantum device during the session. See `Introduction to sessions <sessions.html>`__ for further details.
```
blakejohnson, line 39: Describing sessions as a "contract" isn't really true.  I think this would be clearer as:
> A session allows a collection of jobs to be grouped and jointly scheduled by the Qiskit Runtime service, facilitating iterative use of quantum computers without incurring queuing delays on each iteration.

blakejohnson (commented):
Certainly better. I've left some additional suggestions.

docs/index.rst
beckykd: I was chatting with @jyu00 and we decided on "offers" instead of "exposes"

jyu00 (approved):
LGTM!

PR #876, opened 2023-05-30
Add SimulatorOptions.set_backend method

ikkoham:
### Summary

This PR proposes the simple method to set the simulator options from backend.

Before
```python
from qiskit.providers.fake_provider import FakeManila
from qiskit_aer.noise import NoiseModel

# Make a noise model
fake_backend = FakeManila()
noise_model = NoiseModel.from_backend(fake_backend)

# Set options to include the noise model
options = Options()
options.simulator = {
    "noise_model": noise_model,
    "basis_gates": fake_backend.configuration().basis_gates,
    "coupling_map": fake_backend.configuration().coupling_map,
    "seed_simulator": 42
}
```

After
```python
from qiskit.providers.fake_provider import FakeManila
from qiskit_aer.noise import NoiseModel

# Make a noise model
fake_backend = FakeManila()

# Set options to include the noise model
options = Options()
options.simulator.set_backend(fake_backend)
options.simulator.seed_simulator = 42
```

I opened the PR to discuss. If this direction is ok, I'll finalize the PR.

### Details and comments
Fixes #


qiskit_ibm_runtime/options/simulator_options.py
```diff
@@ -47,3 +51,17 @@ class SimulatorOptions:
     seed_simulator: Optional[int] = None
     coupling_map: Optional[List[List[int]]] = None
     basis_gates: Optional[List[str]] = None
+
+    def from_backend(self, backend: Union[BackendV1, BackendV2]):
```
ikkoham, line 55:
```suggestion
    def set_backend(self, backend: Union[BackendV1, BackendV2]):
```

`from_backend` seems to be a class method. `set_backend` might be better.

veenaiyuri:
Thank you for this @ikkoham. As someone involved with teaching, I want to note that simplifying code implementation as proposed in this case may seem small but does make a real difference.

ikkoham:
Oh... There are no config file for `black`.

docs/how_to/noisy_simulators.rst
```diff
@@ -267,6 +267,26 @@ that the backend supports to have a more realistic noisy simulation.
     options.optimization_level = 0
     options.resilience_level = 0
 
+:meth:`Options.simulator.set_backend` is the syntax sugar of the above set options.
```
kt474, line 270:
```suggestion
`set_backend <https://qiskit.org/ecosystem/ibm-runtime/stubs/qiskit_ibm_runtime.options.SimulatorOptions.html>`_ is the syntactic sugar for setting options.
```

requirements-dev.txt
```diff
@@ -14,6 +14,7 @@ coverage>=6.3
 pylatexenc
 mthree
 scikit-learn
+ddt>=1.2.0,!=1.4.0,!=1.4.3
```
kt474, line 17: nitpicking a bit here but I don't think we need an extra library to test this when a for loop would do the same thing

kt474 (commented):
Thanks for adding this! Mostly looks good - can you also add a release note highlighting this new method?

requirements-dev.txt
ikkoham: I was making the same standard as Terra, but if you say so, I could write it with for-loop.

kt474: ah ok, never mind then, everything looks good to me

kt474 (approved):

ikkoham:
@kt474 Thank you very much. If you encounter any problems related to ddt in the future, please assign me.

PR #886, opened 2023-06-02
Remove non-runner programs

### LINKED ISSUE #880: Deprecate `hello-world`, `vqe`, and `qaoa` programs
These three programs have very low usage and will be deprecated and then removed (end of July)

If one of these programs is used, warn users of the deprecation and suggest the new alternative/replacement 

- VQE tutorial that uses primitives: https://qiskit.org/ecosystem/ibm-runtime/tutorials/vqe_with_estimator.html
- QAOA tutorial: https://qiskit.org/documentation/tutorials/algorithms/05_qaoa.html (will need to be copied into this repo)
- Update hello-world example to use the sampler 

We'll also have to update integration tests to use sampler instead of hello-world

kt474:
### Details and comments
Fixes #880


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -986,6 +986,13 @@ def run(
             RuntimeProgramNotFound: If the program cannot be found.
             IBMRuntimeError: An error occurred running the program.
         """
+        # TODO: Remove this after 3 months
+        if program_id in ["hello-world", "vqe", "qaoa"]:
+            raise IBMInputValueError(
+                "The hello-world, vqe, and qaoa programs have been retired in the "
+                "Qiskit Runtime service. "
+                "Please visit https://qiskit.org/ecosystem/ibm-runtime for more information."
```
jyu00, line 994: I think we can give a bit more information in the text. Something like 
```
Please visit https://qiskit.org/ecosystem/ibm-runtime for an introduction on Sessions and Primitives, and to access tutorials on how to execute VQE and QAOA using Qiskit Runtime Primitives.
```

jyu00 (commented):
We can take this opportunity to delete `program_source` entirely, since even `circuit-runner` has been outdated for a long time.

qiskit_ibm_runtime/qiskit_runtime_service.py
kt474: Sure, I'll leave out `hello-world` for now and add it back next sprint when it's actually removed

PR #890, opened 2023-06-05
Give warning when backend is not active

### LINKED ISSUE #884: Runtime should warn if a backend is not active
The provider will warn an user that they are submitting a job  to a device that is currently not active.  the runtime should do the same.  Ideally the warning would be when the backend is grabbed from `service.get_backend` because when the warning comes at job submission, it is too late, and one needs to go to the IQX to cancel the job.

**Acceptance criteria**

merav-aharoni:
### Summary
Give warning when the backend is not active in  QiskitRuntimeService.run.

### Details and comments
Fixes #884.


merav-aharoni:
@kt474 - do you think a test is necessary? I checked it locally on the real devices. But creating a test turned out to be complicated, because it involves calling `run`  on a backend whose status needs to be manipulated to be non-active.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1004,6 +1004,10 @@ def run(
                 instance=qrt_options.instance, backend_name=qrt_options.backend
             )
             hgp_name = hgp.name
+        backend = self.backend(name=qrt_options.backend, instance=hgp_name)
+        status = backend.status()
+        if status.operational is True and not status.status_msg == "active":
```
kt474, line 1009: nit: is this equivalent to 
```suggestion
        if status.operational is True and status.status_msg != "active":
```

I think no test for this would be ok

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1007,6 +1007,10 @@ def run(
                 instance=qrt_options.instance, backend_name=qrt_options.backend
             )
             hgp_name = hgp.name
+        backend = self.backend(name=qrt_options.backend, instance=hgp_name)
+        status = backend.status()
+        if status.operational is True and status.status_msg != "active":
+            warnings.warn(f"The backend {backend.name} is currently paused.")
```
kt474, line 1013: Just noticed that the status message can also be `dedicated` or `internal` so we should probably only give this warning if the message is `maintenance`
Screenshot 2023-06-06 at 10 35 46 AM

merav-aharoni: I think it would be good to give the warning for all these cases. Only instead of outputing 'no active' I should output the status message.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1007,6 +1007,12 @@ def run(
                 instance=qrt_options.instance, backend_name=qrt_options.backend
             )
             hgp_name = hgp.name
+        backend = self.backend(name=qrt_options.backend, instance=hgp_name)
+        status = backend.status()
+        if status.operational is True and status.status_msg != "active":
+            warnings.warn(
+                f"The backend {backend.name} is currently {status.status_msg}."
```
kt474, line 1014:
```suggestion
                f"The backend {backend.name} currently has a status of {status.status_msg}."
```

sorry, small nit: `The backend {} currently has a status of internal/maintenance/dedicated` sounds better than 
`The backend {} is currently internal/maintenance/dedicated`

kt474 (approved):

PR #891, opened 2023-06-05
Allow to override runtime url (for development purposes)

rathishcholarajan:
### Summary
This enables testing with experimental cluster.

### Details and comments
Fixes ws-qiskit-runtime-api issue 61

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -128,6 +128,7 @@ class QiskitRuntimeService(Provider):
         instance: Optional[str] = None,
         proxies: Optional[dict] = None,
         verify: Optional[bool] = None,
+        runtime_url: Optional[str] = None,
     ) -> None:
         """QiskitRuntimeService constructor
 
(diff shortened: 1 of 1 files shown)


jyu00:
I think having an "IBM developer only" option in one of the most public facing method is a bad idea. Surely we can find another way for the developers to achieve the same thing in a more subtle manner.

kt474:
Would it work if you just had this change in a fork? Is there a reason other than ease of use to have this in qiskit/main

daka1510:
Thanks for working on that, @rathishcholarajan. I agree it's important to have a good and consistent development experience and at the same time I also think we shouldn't include any special values in the public-facing APIs (even that would be the most straigh-forward solution).

I cannot think of a better solution than internal environment variables. An alternative option would be an internal (private) function that can be called after the initialization to override the runtime URL. I do think though this would require a bit more refactoring since the runtime URL is already used in calls triggered by the `QiskitRuntimeService` constructor.

jyu00:
> An alternative option would be an internal (private) function that can be called after the initialization to override the runtime URL. I do think though this would require a bit more refactoring since the runtime URL is already used in calls triggered by the `QiskitRuntimeService` constructor.

While a bit convoluted, one can create a child class of `QiskitRuntimeService`, overwrite `_authenticate_ibm_quantum_account()` to return an auth client that'd return a custom runtime url. 

But I'd be ok with an internal environment variable as well :)

kt474:
Closing this for now, if this is still necessary, we can create an issue to add an internal environment variable

PR #893, opened 2023-06-06
Disable default session

### LINKED ISSUE #842: Disable default session behavior

[Default session](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/session.py#L246-L248) was introduced as part of flexible session as an attempt to improve session usability. Unfortunately due to the current limitations, this is causing unexpected behaviors as outlined in #648. And since the session is managed implicitly, there is also no way for a user to close a session.

So until we can properly manage a default session, let's just disable this feature and always require an explicit session (or run in non-session mode if session is not used).

However, we can still keep some of default session's usability improvements. Namely.

1. Jobs running a session context manager should use that session, e.g.

```
with Session(...):
    job = Sampler().run()  # This uses the session from Session(), without requiring session being passed
    job = Estimator().run()  # This uses the session from Session(), without requiring session being passed
```

2. We can still use an explicitly created session without requiring it being passed. For example,

```
session = Session()
job = Sampler().run()  # This uses the session from Session(), without requiring session being passed
job = Estimator().run()  # This uses the session from Session(), without requiring session being passed
```

This will need some doc updates.

**Acceptance criteria**

Disable default session but preserve aforementioned usability

jyu00:
### Summary

Disable default session behavior. A primitive will instead run in non-session mode if a backend name or instance is passed. 

Since we are not opening a default session anymore, it's odd to do `session=backend`. I contemplated about using a new parameter like `target` but eventually decided `session=session` and `backend=backend` is the most natural. So I added a `backend` parameter to the primitives. 

I also broke down and wrote a base primitive class for all the duplicate code 😅 

### Details and comments
Fixes #842


jyu00:
> Can `backend` be the first kwarg here? That way one can do `Sampler(backend)` directly, rather than `Sampler(backend=backend)`

What about people who want to do `Sampler(session)` 😂 ? 

I'm just kidding. All our tutorials encourage people to use Session as a context manager, in which you don't have to explicitly pass the session. So it makes sense to have `backend` as the first kwarg.

nonhermitian:
> All our tutorials encourage people to use Session as a context manager

Hmm, I am explicitly changing that for the VQE (PR coming today).  Is this a requirement?

jyu00:
> > All our tutorials encourage people to use Session as a context manager
> 
> Hmm, I am explicitly changing that for the VQE (PR coming today). Is this a requirement?

I'd prefer if we stick with Session as a context manager, if Session is used. In 3Q we plan on making `session.close()` to mean "I'm done submitting jobs to the session", instead of "kill all my pending session jobs" lol. If we encourage people to use it as a context manager, then they will automatically get this new close behavior when it becomes available.

jyu00:
> Can `backend` be the first kwarg here? That way one can do `Sampler(backend)` directly, rather than `Sampler(backend=backend)`

@nonhermitian On a second thought, since `session` takes precedence over `backend`, it's kind of weird to have `backend` first. 

Another option is to use a generic kwarg like `target` that can take either a backend or a session.  Would you prefer that over `session` being first?

nonhermitian:
In my mind, the `backend` is the fundamental object.  It is required for the `Session` to be created in the first place.  And yes, I know that the cloud will defer the selection until later.

The session is really just a way to logically batch / box jobs together.  Could there not be a happy place where a call to create a primitive with no session takes the form:

```python

sampler = Sampler(backend)
```

and one with a session is:

```python
session = Session()
sampler = Sampler(backend, session=session)
```
where the session itself does not take the backend as an arg; it is just a reference object tracking the logical connection between jobs created by the primitive.

jyu00:
> and one with a session is:
> 
> ```python
> session = Session()
> sampler = Sampler(backend, session=session)
> ```
> 
> where the session itself does not take the backend as an arg; it is just a reference object tracking the logical connection between jobs created by the primitive.

Right now a session is tied to a backend, so it should take backend as an arg or people will like think cross-backend session is possible.

nonhermitian:
But that is a validation thing right rather than a fundamental blocker?  Internally it looks like the session is already separate in the `service.run` call, with backend in options and the session_id passed separately.  A completely general call could be 
```python

Sampler(session=session)
```
which would do the cloud-style backend selection for you and still allow for having a session that you could set.

But perhaps the smoke has gotten to my brain.

test/integration/test_estimator.py
```diff
@@ -297,3 +297,30 @@ def test_estimator_error_messages(self, service):
...
+
+        # pylint: disable=invalid-name
+        H1 = SparsePauliOp.from_list([("II", 1), ("IZ", 2), ("XI", 3)])
+
+        estimator = Estimator(backend=backend)
+        self.assertIsInstance(estimator, BaseEstimator)
+        self.assertIsNone(estimator.session)
+
+        theta1 = [0, 1, 1, 2, 3, 5]
+        circuits1 = [psi1]
```
merav-aharoni, line 317: It seems to me you are testing only a single circuit. So why call it circuits1`? and why check the length of the results?

merav-aharoni (approved):
Looks good!
I am particularly happy with the new `BasePrimitive` class that eliminates a lot of code duplication!
It was a bit difficult to review all the changes. I think it would have been better if you had divided this PR into two - one for `BasePrimitive` and the other for deprecating `default_session`. Anyway, I think I managed to review.

test/integration/test_sampler.py
```diff
@@ -251,3 +251,21 @@ def test_sampler_error_messages(self, service):
...
             self.assertFalse("python -m uvicorn server.main" in err.exception.message)
+
+    @run_integration_test
+    def test_sampler_no_session(self, service):
+        """Test sampler without session."""
+        backend = service.backend(self.backend)
+        sampler = Sampler(backend=backend)
+        self.assertIsInstance(sampler, BaseSampler)
+
+        circuits = [self.bell] * 3
```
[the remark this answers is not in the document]
jyu00: We had an issue a while back wherein the backend was returning results that don't match the input length. While it wasn't a provider related issue, it was easy to add a couple extra circuits and verify the result length. We did this for the integration tests that use session, so I figured I'd do the same here.

test/integration/test_estimator.py
jyu00: > why call it circuits1

Because copy & paste 😄. I updated the code to avoid confusion. 

> why check the length of the results

See above.

kt474 (approved):
Sorry, took me a while to review - happy with all the updates here 

LGTM!

PR #877, opened 2023-05-30
Support  CouplingMap in options

merav-aharoni:
### Summary
Support `CouplingMap` given as `simulator.coupling_map` option, in addition to a list of edges.

### Details and comments
Fixes #872.

PR #896, opened 2023-06-08
Fix CouplingMap test

merav-aharoni:
### Summary
`test_coupling_map_options` was failing because of different order of the couples in the list. 
Converted the lists to sets for the comparison.

### Details and comments
See also #877


kt474 (approved):
LGTM - Thanks!

PR #901, opened 2023-06-09
Replace QAOA tutorial

HwajungKang:
### Summary
QAOA tutorial using only Qiskit, i.e. without using qiskit_algorithm modules 

### Details and comments
Fixes #


jyu00:
> Looks great! Can you add this footer?
> 
> ```python
> import qiskit.tools.jupyter
> 
> %qiskit_version_table
> %qiskit_copyright
> ```

That and 

```
import qiskit_ibm_runtime

qiskit_ibm_runtime.version.get_version_info()
```

Since `%qiskit_version_table` doesn't include qiskit-ibm-runtime version.

docs/tutorials.rst
```diff
@@ -24,7 +24,7 @@ Sampler
    tutorials/user-transpiled-circuits
    tutorials/sea_with_sampler
    tutorials/qpe_with_sampler
-   tutorials/qaoa_with_sampler
+   tutorials/qaoa_with_primitives
```
blakejohnson, line 27: You need to make a similar change in index.rst to point to the renamed file.

blakejohnson (changes requested):
Just need to update index.rst and add the standard footer.

HwajungKang:
@blakejohnson Thanks! Just updated the index.rst and the qaoa_with_primitives for the footer.
@jyu00 Added the footer for qiskit_ibm_runtime as well.

docs/tutorials.rst
HwajungKang: Done!

docs/tutorials/qaoa_with_primitives.ipynb
```diff
@@ -0,0 +1,460 @@
+{
```
blakejohnson, line 1: <pre class="ql-syntax" spellcheck="false">Line #8.    ansatz.decompose(reps=3).draw("mpl")
</pre><p>I have been wondering if it would be better to show the circuit without <code>.decompose(reps=3)</code> because it makes the QAOA structure more evident. But, the way you have it now makes the representation in native gates clearer. So, there are pros and cons of each.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/qiskit-ibm-runtime/pull/901/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

HwajungKang: Hi @blakejohnson , I was actually debating which one to choose when I made it. I decided this way to show that the QAOAAnsatz from the circuit library implements the needed cost Hamiltonian correctly and easily. But I agree that displaying it without decompose method shows the QAOA structure more clearly. 

So, I just added an extra cell to show the circuit without decompose. If you think it is not necessary, then, I can delete it.

HwajungKang: For your convenience, this is the extra cell that I added. 
extra

blakejohnson: I guess the version that makes the structure clearest is a single `decompose()` call (i.e. `reps=1`). So, can we go with `ansatz.decompose().draw('mpl')`?

PR #902, opened 2023-06-11
Fixed bug in deepcopy of IBMBackend

### LINKED ISSUE #899: Unable to deepcopy an IBMBackend
**Describe the bug**

With the old IBMQBackends, deepcopying was allowed. With the new IBMBackends, it results in a recursion error. Loosely, this seems to be because deepcopy checks if `_properties` is an attribute of IBMBackend (when it is constructing the new, empty IBMBackend), and the `__getattr__` function of IBMBackend calls `self._get_properties()` internally, resulting in an infinite loop of trying to access the attribute `_properties` which doesn't exist yet. This is also true for `_defaults` and `_configuration`.

I'm open to the idea that deepcopying backends should be discouraged for some reason but feel like at least the error should be improved.

**Steps to reproduce**
```
import copy

backend = service.get_backend('ibm_perth')
backend_copy = copy.deepcopy(backend)
```

**Expected behavior**

A copy of the backend (preferable) or avoiding infinite recursion.

**Suggested solutions**

Currently, I modify `__getattr__` in `ibm_backend.py` to the following, which is very hacky and I'm worried there's some edge case behaviour I may affect

```
   def __getattr__(self, name: str) -> Any:
        """Gets attribute from self or configuration

        This magic method executes when user accesses an attribute that
        does not yet exist on IBMBackend class.
        """
        # Lazy load properties and pulse defaults and construct the target object.
        if name not in ['_properties', '_defaults','_configuration']:
            self._get_properties()
            self._get_defaults()
            self._convert_to_target()
        # Check if the attribute now is available on IBMBackend class due to above steps
        try:
            return super().__getattribute__(name)
        except AttributeError:
            pass
        # If attribute is still not available on IBMBackend class,
        # fallback to check if the attribute is available in configuration
        if name not in ['_configuration']:
            try:
                return self._configuration.__getattribute__(name)
            except AttributeError:
                raise AttributeError(
                    "'{}' object has no attribute '{}'".format(
                        self.__class__.__name__, name
                    )
                )
```

**Additional Information**

- **qiskit-ibm-runtime version**:  0.10.0
- **Python version**: 3.8.8
- **Operating system**: macOS Ventura

merav-aharoni:
### Summary
Fixed infinite recursion when attempting to deepcopy an IBMBackend.

### Details and comments
Fixes #899.
See also https://github.com/Qiskit/qiskit-ibm-provider/pull/658.


test/integration/test_backend.py
```diff
@@ -163,3 +163,16 @@ def test_backend_run(self):
         with self.subTest(backend=backend.name):
             with self.assertRaises(RuntimeError):
                 backend.run()
+
+    def test_backend_deepcopy(self):
+        """Test that deepcopy on IBMBackend works correctly"""
+        service = QiskitRuntimeService(channel="ibm_quantum")
+        sim_backend = service.get_backend("ibmq_qasm_simulator")
+        device_backend = service.get_backend("ibmq_lima")
```
kt474, line 171: I don't think we need to reinitialize the service here we can just use `self.dependencies.service`

Also some accounts may not have access to `ibmq_lima`, so we can just use `self.backend`

merav-aharoni: Done.

merav-aharoni:
When using the default `deepcopy`, we got a bug because `IBMBackend` contains a `QiskitRuntimeService`, and that contains a list of all the supported backends. So if the list of backends was already loaded in the method `QiskitRuntimeService.backends()`, we got an infinite recursion bug. 
In a conversation with @jyu00  and @kt474 , we decided to implement a `deepcopy` method for IBMBackend, to avoid this problem. In this implementation, the list of backends in `IBMBackend._service._backends` is copied and not deepcopied.

merav-aharoni:
@kt474 - can you review again?

kt474 (approved):
LGTM!

PR #916, opened 2023-06-14
Minor edits of Grover tutorial

beckykd:
@nonhermitian , should the session be manually closed?  I know it'll close on its own, but in other places we're closing it so it doesn't waste time waiting to time out.

### DIFF
--- a/docs/tutorials/grover_with_sampler.ipynb
+++ b/docs/tutorials/grover_with_sampler.ipynb
@@ -1,6 +1,7 @@
 {
  "cells": [
   {
+   "attachments": {},
    "cell_type": "markdown",
    "id": "db78a668",
    "metadata": {},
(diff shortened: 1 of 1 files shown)


nonhermitian:
There is no session here as it is a single job.  Thus we do not have to worry about the session logic in this notebook.

beckykd:
> There is no session here as it is a single job. Thus we do not have to worry about the session logic in this notebook.

Oh, I see.  Then the lines like "real_sampler = Sampler(session=backend)" aren't actually starting a session?  I didn't see the context manager, but thought it was just another way to instantiate a session :)

nonhermitian:
So to date it has made an implicit session, but also one the user does not have access to.   But that brings up a good point, in V0.11 I think one is allowed to just write `estimator = Estimator(backend)`, and an internal session is no longer created.  I will let @jyu00 validate that statement, but if true, we (me really) should go in and make that change.

kt474 (approved):
Since this PR is just grammar/wording improvements i'll merge it to get it in the release today.

Updates to the code can be done in a separate PR.

PR #917, opened 2023-06-14
Minor QAOA tutorial edits

beckykd:
@HwajungKang - Should the session be manually closed?  I know it doesn't have to be, but we've been closing it in other places so it doesn't have to wait the whole 5 min of idle time before timing out.


nonhermitian:
Here, the session is closed in `In[78]`

beckykd:
> Here, the session is closed in `In[78]`

Oh thanks - missed that

kt474 (approved):

PR #925, opened 2023-06-20
Remove QPE and SEA tutorials

nonhermitian:
Removes the QPE and SEA tutorials, as they are a bit too complex for addition here, and are more suitable for inclusion in other places if we decide to bring them back.


jyu00 (approved):
LGTM!

PR #930, opened 2023-06-27
Cancelled job should throw exception when trying to get result()

### LINKED ISSUE #885: job.result() should not return None if the job is cancelled
If a job is cancelled I should get an exception, like is done in the provider, when doing `job.result()`.  

**Acceptance criteria**

merav-aharoni:
### Summary
Cancelled job should throw exception when trying to get result(), instead of returning None.

### Details and comments
Fixes #885


kt474 (approved):
LGTM

PR #931, opened 2023-06-29
Fix/qr typos

raulotaolea:
### Summary
A couple of typos in the QR descriptions.


beckykd (approved):

raulotaolea:
@kt474 It seems that the deployment had errors. I don't see the message changed on production, so I guess we have to publish it again?

PR #943, opened 2023-07-11
Added error message for job failure

### LINKED ISSUE #841: Expose job status reason if there is no result on a failed job

With recent changes in the scheduler for the ibm_cloud and ibm_quantum channels, jobs in a session which didn't run before the session closed will be errored out and put into a `Failed` status.

On the client side, when you look for the error message on these jobs you just get something like this:

`Job chd744aceq17dip7od0g has failed:`

Note the colon without anything after it. I believe that comes from this code:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/63742c76c9dc93ca8cbadf82c17bf2e3db321047/qiskit_ibm_runtime/runtime_job.py#L451-L471

In this case the job failed before it ran, as part of the scheduler, so there are no results, but there is the reason, which in the DB is this:

`Session chd7442ceq17dip7octg has been closed`

It'd be nice if there are no results for the job, if we could expose the reason (if there is one) in the `RuntimeJob.error_message()` return value. There might be implications for not wanting to expose all reasons in case there are internal details that shouldn't be leaked out, but I'd think maybe the job status reason code could be used to determine that. In this case the reason code was 1217.

**Acceptance criteria**

Expose the job status reason if there are no results on a failed job.

merav-aharoni:
### Summary
Added reason for failure or full error message when invoking the method `job.error_message()`.

### Details and comments
Fixes #841.


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -453,14 +453,17 @@ def _error_msg_from_job_response(self, response: Dict) -> str:
...
         index = job_result_raw.rfind("Traceback")
         if index != -1:
             job_result_raw = job_result_raw[index:]
-
         error_msg = API_TO_JOB_ERROR_MESSAGE["FAILED"]
+        if not error_message is None:
+            error_msg += error_message + "\n"
         if status == "CANCELLED" and self._reason == "RAN TOO LONG":
             error_msg = API_TO_JOB_ERROR_MESSAGE["CANCELLED - RAN TOO LONG"]
         return error_msg.format(self.job_id(), job_result_raw)
```
kt474, line 469: In the original issue, we only wanted to add the reason or error message if the result isn't returned (`job_result_raw` is none) - otherwise, we'd just be appending the `error_message` to `error_msg` which would already contain the same, duplicated error message. 

I think something like this would work: 
```suggestion
        return error_msg.format(self.job_id(), job_result_raw or self._reason)
```

merav-aharoni: You are right. But I am thinking it may be better to leave `_error_msg_from_job_response` and instead change line 218 in `results()` as follows:
Instead of 
`error_message = self._reason if self._reason else self._error_message`
write
`error_message = self.error_message()`
What do you think?

kt474: Yeah that would work, but I think `_error_msg_from_job_response` would still have to be updated to handle the case if `job_result_raw` is none

merav-aharoni: I finally took your advice, with a little extra change. Can you have another look?

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -460,10 +460,12 @@ def _error_msg_from_job_response(self, response: Dict) -> str:
...
             job_result_raw = job_result_raw[index:]
 
-        error_msg = API_TO_JOB_ERROR_MESSAGE["FAILED"]
         if status == "CANCELLED" and self._reason == "RAN TOO LONG":
             error_msg = API_TO_JOB_ERROR_MESSAGE["CANCELLED - RAN TOO LONG"]
-        return error_msg.format(self.job_id(), job_result_raw)
+            return error_msg.format(self.job_id(), job_result_raw)
+        else:
+            error_msg = API_TO_JOB_ERROR_MESSAGE["FAILED"]
+            return error_msg.format(self.job_id(), self._reason or job_result_raw)
```
kt474, line 468: I think I would still prefer to have something like
```suggestion
            return error_msg.format(self.job_id(),  job_result_raw or self._reason)
```
Because the full `job_result_raw` could have more information useful to the user, and already contains the reason. If the job errors, the reason will also already be returned in `job.result()`.

This is more of a design nitpick so I'd be ok with doing it this way as well

kt474 (approved):

qiskit_ibm_runtime/runtime_job.py
merav-aharoni: My thought was that the user would probably prefer to see only the reason, because the `job_result_raw` contains a lot of information. But if you still think it should be the other way around, I will go with that. So waiting for your final input.

kt474: Thanks makes sense, I think the way it currently is works as well

PR #955, opened 2023-07-16
Use service defined by user as default

### LINKED ISSUE #923: Use service already initialized

Since `Sampler` and `Estimator` accepts a backend name string, a user can easily do something like

```
service = QiskitRuntimeService(..., instance="h1/g1/p1")
job = Sampler("ibmq_manila").run(...)
```

and expect the `service` object to be used (with h1/g1/p1). But since `service` is not passed to `Sampler`, it'd actually just use whatever the default is from `QiskitRuntimeService()`.  There is also no way to pass `service` to `Sampler` today. 

**Acceptance criteria**

Save and use the `QiskitRuntimeService` if one was previously initialized when backend name is passed to the primitive.

merav-aharoni:
### Summary
If the user defines a `QiskitRuntimeService`, then that service will be treated as the default service in the program.

### Details and comments
This is implemented by adding a static member `QiskitRuntimeService.global_service`. The primitive will use that last defined `service` as its service.

Fixes #923.


kt474 (approved):
LGTM

PR #961, opened 2023-07-18
Add an auto-try option if a job fails 

### LINKED ISSUE #682: Give the option to auto-retry failed jobs

Jobs sometimes fail because of temporary server / networking error, and a retry would work. Unfortunately for many iterative algorithms, this means the user would need to restart the entire workload from the beginning. For this reason `QuantumInstance` has [built-in retry](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/utils/run_circuits.py#L306). Since we are encouraging people to use Qiskit Runtime / primitives and _not_ `QuantumInstance`, it'd be nice to have the same capability.

**Acceptance criteria**

Have the options for users to specify max retry (0 being no retry) when running a QRT job.

kt474:
### Details and comments
Fixes #682


merav-aharoni:
I only did a overview. It generally looks fine. I have two questions/comments:

1. The support for retries is only if the job is run in a Session. Do we want to support retries also when no session?
2. I think the decision on whether to do a retry should depend on the reason for the error. For example, if the circuit is bad, there is no reason to do a retry.

merav-aharoni:
@kt474 - was this completed? should I review again?

kt474:
Going to close this for now, need to rethink a better way to handle this. I'll hopefully be able to work on this at the end of Q4

PR #976, opened 2023-07-27
Expose job cost estimation values

### LINKED ISSUE #972: Expose new values for job estimation
There will be 2 new parameters for jobs returned in `GET /jobs` and `GET /jobs/{id}`

- `estimated_running_time_seconds` - the value predicted by the service as it is
- `estimated_max_running_time_seconds` - the value with a safe multiplier based on the failure ratio known for the service.

We can implement this in `qiskit-ibm-provider` and then call the necessary methods from there. 

https://github.com/Qiskit/qiskit-ibm-provider/issues/681

kt474:
### Summary

Don't merge until change is live in prod on the server side 

Adding a new property, `job.usage_estimation` which returns 

```
{'quantum_seconds': 3.082247164179738}
```

### Details and comments
Fixes #972


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -655,3 +656,24 @@ def tags(self) -> List:
             Tags assigned to the job that can be used for filtering.
         """
         return self._tags
+
+    @property
+    def cost_estimation(self) -> Dict[str, Any]:
```
jyu00, line 661: Can we not call this `cost_estimation`? Since the returned values are estimated execution time, which is at best an indirect translation to cost. Maybe `estimated_execution_time`?

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -655,3 +656,24 @@ def tags(self) -> List:
...
+        """Return cost estimation infromation for this job.
+
+        Returns:
+            Cost estimation information in dictionary format, with the estimated
+            running time and estimated max running time in seconds.
```
jyu00, line 666: What is estimated max running time?

qiskit_ibm_runtime/runtime_job.py
kt474: Sure, @francabrera fyi

qiskit_ibm_runtime/runtime_job.py
kt474: I will update the docstrings & release note to be more clear - my understanding is:

`estimated_running_time_seconds`: the value returned by estimation service, estimating how long the job will run
`estimated_max_running_time_seconds`: the estimated running time multiplied by some multiplier based on failure ratio that estimates the absolute maximum the job could run (before it should be canceled?)

jyu00: > before it should be canceled?

Yeah that was my question. Is that the new max execution time? Presumably only used if user didn't specify one?

qiskit_ibm_runtime/runtime_job.py
francabrera: since this is an estimation of `usage.quantum_seconds` let's make it `usage_estimation.quantum_seconds` the value being `estimated_running_time_seconds` cc @jyu00

qiskit_ibm_runtime/runtime_job.py
francabrera: we only need to worry about `estimated_running_time_seconds` as mentioned in my other comment

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -655,3 +656,19 @@ def tags(self) -> List:
...
+
+        Returns:
+            ``quantum_seconds`` which is the estimated running
+            time of the job in seconds.
+        """
+        if not self._usage_estimation:
+            response = self._api_client.job_get(job_id=self.job_id())
+            self._usage_estimation = {
+                "quantum_seconds": response.pop("estimated_running_time_seconds", None),
+            }
```
jyu00, line 672: Is this data always available upon first query? If not, it'd be set to `"quantum_seconds": None` and never updated again (since `self._usage_estimation` won't be False).

kt474: Yes, can confirm it's always available on first query

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -655,3 +656,19 @@ def tags(self) -> List:
...
         """
         return self._tags
+
+    @property
+    def usage_estimation(self) -> Dict[str, Any]:
+        """Return the usage estimation infromation for this job.
+
+        Returns:
+            ``quantum_seconds`` which is the estimated running
+            time of the job in seconds.
```
jyu00, line 666: Perhaps we can be more explicit here, since this is different from the unit used in `max_execution_time`. Perhaps something like

```suggestion
            ``quantum_seconds`` which is the estimated quantum time 
            of the job in seconds. Quantum time represents the time that 
            the QPU complex is occupied exclusively by the job.
```

kt474: sure, will copy these changes to qiskit-ibm-provider as well

PR #979, opened 2023-07-31
Filter backends by instance when given in QiskitRuntimeService constructor

### LINKED ISSUE #938: Specifying instance in QiskitRuntimeService does not filter available systems
**Describe the bug**
I would expect that passing an instance to the service instance filters the available devices to only those available from that instance, e.g.

```python
service = QiskitRuntimeService(instance='ibm-q-skku/2023-quantum-lea/2023-quantum-lea')
```

but it does not.

**Steps to reproduce**
Try to filter based on instance.

**Expected behavior**
Passing an instance to the service gives me access to only those services accessible to the instance given.

**Suggested solutions**

**Additional Information**

- **Python version**:
- **Operating system**:

merav-aharoni:
### Summary
When `QiskitRuntimeService()` is initialised with `instance`, filter the backends according to the `instance`.

### Details and comments
Previously, the backends were filtered by `instance` only when it was given as parameter to `QiskitRuntimeService.backends()`.
In addition, we now raise an exception if the `instance` parameter to `QiskitRuntimeService` is not supported. Previously, we only  gave a warning.
Fixes #938.

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -358,6 +358,7 @@ class QiskitRuntimeService(Provider):
         Raises:
             IBMInputValueError: If the URL specified is not a valid IBM Quantum authentication URL.
             IBMAccountError: If no hub/group/project could be found for this account.
+            IBMInputValueError: If instance parameter is not found in hgps.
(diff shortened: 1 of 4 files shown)


kt474 (approved):
LGTM!

PR #997, opened 2023-08-06
Correct backend given by RuntimeJob.backend()

### LINKED ISSUE #625: RuntimeJob.backend() does not show the correct backend a job was run on
**Describe the bug**

I created a test where I submit 50 cloud channel jobs without specifying a backend, so the scheduler picks a backend. I wait for all jobs to complete - they al completed successfully. The jobs are spread across two possible backends. However, when I get the backend counts for the jobs it says they are all on the same backend:

```python
from collections import defaultdict

backend_counts = defaultdict(int)
for job in jobs:
    backend_counts[job.backend()] += 1

print(f'backend distribution of {len(jobs)} jobs')
for backend, count in backend_counts.items():
    print(f'{backend}: {count}')
```

Output:

```
backend distribution of 50 jobs
<IBMBackend('alt_algiers')>: 50
```

In the database though, this is the distribution:

```
ibmclouddb=> select count(*),backend_name from jobs where tags @> '{concurrent-jobs}' and created_at > now() - interval '20 minutes' group by backend_name order by count;
 count | backend_name 
-------+--------------
     2 | alt_algiers
    48 | alt_canberra
(2 rows)
```

**Steps to reproduce**

This is my script that generates the jobs and waits for them to complete:

```python
from concurrent import futures
import time

options = {
    # Be aware of https://github.com/Qiskit/qiskit-ibm-runtime/issues/622 if
    # not specifying a backend.
    "backend": "",
    "job_tags": ["proxy_mcs", "use_runtime_workers", "concurrent-jobs"]
}

def submit_job(index):
    # copy job tags so we can add one for the index per job
    job_tags = list(options['job_tags'])
    job_tags.append(f'job{index}')
    my_options = dict(options)
    my_options['job_tags'] = job_tags
    return service.run(program_id="hello-world",
                       options=my_options,
                       inputs=program_inputs)

# submit jobs concurrently (runtime API rate limits to 50 jobs per minute per user)
count = 50
print(f'starting to submit {count} jobs concurrently')
start_time = time.time()
jobs = []
with futures.ThreadPoolExecutor(thread_name_prefix='mcs-integration-test') as executor:
    # submit the jobs asynchronously
    f = []
    for index in range(count):
        f.append(executor.submit(submit_job, index))
    # wait for the jobs to be created
    for future in futures.as_completed(f):
        try:
            job = future.result()
            jobs.append(job)
            print(f'submitted job: {job.job_id()}')
        except Exception as e:
            print(f'error submitting job: {e}')

# wait for the jobs to complete
print('waiting for jobs to complete')
for index, job in enumerate(jobs):
    job_id = job.job_id()
    print(f'{index+1}) waiting for job {job_id} to complete; status: {job.status()}')
    print(f'job {job_id} result: {job.result()}')
end_time = time.time()

print(f'time from jobs submitted to all jobs completed: {end_time - start_time}')

# how many failed?
print('seeing how many jobs failed')
errored_jobs = {}
for job in jobs:
    if job.error_message():
        errored_jobs[job.job_id()] = job.error_message()

if errored_jobs:
    print(f'{len(errored_jobs)} jobs failed')
elif not jobs:
    print('failed to submit all jobs')
else:
    print('all jobs passed')
```

**Expected behavior**

I would expect that once the jobs are complete, Qiskit would show me the actual backend they ran on.

**Suggested solutions**

Use the `backend` field from the actual `GET /jobs/{id}` response from the runtime API for the cloud channel:

https://us-east.quantum-computing.cloud.ibm.com/openapi/#/Jobs/get_job_details_jid

**Additional Information**

It seems that maybe the code is getting here:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/5ef5711716b52a95b8a0211385f3b4450da82f74/qiskit_ibm_runtime/qiskit_runtime_service.py#L701

And just taking the first backend available, which between `alt_algiers` and `alt_canberra` is `alt_algiers`, just based on sort order, even though 48 of the 50 jobs ran on `alt_canberra`. However, I'm not sure how `RuntimeJob.backend()` gets to that `QiskitRuntimeService.backends()` method.

This may also be related to #622 in some way, i.e. if the jobs don't have a backend specified when they are created maybe that influences how this works.

- **Python version**: 3.8

merav-aharoni:
### Summary
Fixed bug where the `RuntimeJob.backend()` did not return the correct backend.

### Details and comments
When running on channel `ibm_cloud`, it is possible not to specify
the `backend`. In this case, the system selects one of the available
backends for this service. 
As reported in the issue below, the the backend returned by `job.backend()` was not always the actual backend on which the job was run. This PR fixes
this bug.
In the previous version, the `RuntimeJob._backend` was assigned when the job was created. If the backend was not given as a parameter, the backend selected was simply the first in the list of `QiskitRuntimeService.backends()`. But the server later selects an actual backend to run the job on, and this may be a different one.
The fix is to not assign a backend if not passed as a parameter. And to assign the backend after the job has completed.
I did not add a test, because these test may take a very long time to run, and the results are not predictable.
Fixes #625 .


merav-aharoni:
@mriedem -  - Can you please try this on your failing test case?

kt474 (approved):
I tested on a cloud premium account, LGTM

PR #1010, opened 2023-08-09
Don't store circuit params automatically 

### LINKED ISSUE #987: Circuit runner shouldn't store automatically store circuits in run jobs
**Describe the bug**

When using the circuit runner, the returned job includes a list of all circuit objects for ran circuits. If these circuits are large this massively increases the size of the job and how long it takes to load. 

**Steps to reproduce**

Generate a circuit with a large number of gates (ie roughly the number of 1-qubit gates for square PEC like circuits)

```python

# Large file size circuit
size = 10
qc = QuantumCircuit(1)
for i in range(size * size):
    qc.rz(0.1, 0)
    qc.sx(0)
    qc.rz(0.2, 0)
    qc.sx(0)
    qc.rz(-0.3, 0)
qc.measure_all()
```

For generate batches of these circuits to run, where we don't care about keeping the circuit in memory for longer than it takes to spawn the job

```python
from qiskit_ibm_runtime import QiskitRuntimeService, Session

job_size = 10
num_batches = 10
jobs = []    
    for i in range(num_batches):
        # Simulate sampling circuits via making copies that use memory
        circuits = [qc.copy() for _ in range(job_size)]
        jobs.append(session.run("circuit-runner", inputs={"circuits": circuits}))
```

Each job contains the full list of circuit objects for run circuits in `job._params["circuits"]` and will lead to running out of memory if the number of circuits and batches becomes large enough.

```python
jobs[0]._params
{'circuits': [<qiskit.circuit.quantumcircuit.QuantumCircuit at 0x28db02370>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x28db02b50>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x28db02b80>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x290207400>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x28dbb43a0>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x29106ceb0>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x29103bc40>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x28dbcb820>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x17156e820>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x28dca96d0>]}
```

**Expected behavior**

When a job is loaded from the service, the `_params` field appears to be empty unless `job.inputs` is called.
```python
loaded_job = [service.job(jobs[0].job_id())
loaded_job._params
# Returns
{}

# If inputs is called it gets populated
loaded_job.inputs
{'circuits': [<qiskit.circuit.quantumcircuit.QuantumCircuit at 0x2918a6d90>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x2918a6e20>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x2918d8c70>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x29190aca0>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x29193dcd0>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x291970d00>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x2919a5d30>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x2919d8d60>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x291a0bd90>,
  <qiskit.circuit.quantumcircuit.QuantumCircuit at 0x2914f2ca0>]}
```
**Suggested solutions**

Can we purge the circuits from the job returned by circuit-runner, and have the be lazily loaded like if the job was itself loaded from the service?

**Additional Information**

- **qiskit-ibm-runtime version**: 0.11.2
- **Python version**:3.9
- **Operating system**:

kt474:
### Details and comments
Fixes #987


releasenotes/notes/dont-store-circuit-params-9d87ffefc2a7d257.yaml
```diff
@@ -0,0 +1,5 @@
+---
+upgrade:
+  - |
+    Circuits will no longer be automatically stored in runtime jobs. They can still be retrieved
+    normally with :meth:`qiskit_ibm_runtime.RuntimeJob.inputs`.
```
merav-aharoni, line 5:
```suggestion
    Circuits and other input parameters will no longer be automatically stored in runtime jobs. They can still be retrieved
    with :meth:`qiskit_ibm_runtime.RuntimeJob.inputs`.
```

merav-aharoni (approved):
LGTM.
Two questions just for my understanding - 

1. Why did you add the `kwargs` parameter and not keep `params` or simply remove both?
And why not remove the `params` in the creation of the `RuntimeJob` in `QiskitRuntimeService`?
2. Why did you modify the `SerializableClass.__eq__` method?

kt474:
> 1. Why did you add the `kwargs` parameter and not keep `params` or simply remove both?
>    And why not remove the `params` in the creation of the `RuntimeJob` in `QiskitRuntimeService`?
> 2. Why did you modify the `SerializableClass.__eq__` method?

You are right, removing `params` in the creation of the runtime job here is a much better approach - I was a bit backwards here (fixed). Since the params are no longer automatically a part of the runtime job, this custom serialization class returns `{'value': 'foo'}` which needs to be deserialized

PR #1012, opened 2023-08-10
Add note for docs of error mitigation

ikkoham:
### Details and comments
Fixes #


frankharkins (changes requested):
I'm not familiar with resiliance level so can't review accuracy, maybe @pedrorrivero can help?

Writing-wise, it would be good to explain the acronyms (TREX etc.) or link to the docs as earlier in the page.

Also, not really a problem with this PR, but if
> Each method has its strengths and weaknesses. For example, TREX is good for shallow circuits since it's readout error mitigation whereas ZNE for deeper circuits. PEC can mitigate arbitrary errors but may not work in practice because of its large overhead.

Maybe "resilience level" is a misnomer? As a user I would expect `2` to be more resilient to errors than `1`.

ikkoham:
@frankharkins Thank you for your comment. Yes. @pedrorrivero can help the review.

> Writing-wise, it would be good to explain the acronyms (TREX etc.) or link to the docs as earlier in the page.

Thanks :+1: I'll update.

> Maybe "resilience level" is a misnomer? As a user I would expect 2 to be more resilient to errors than 1.

hmm... Agree. But I cannot modify the name here.

docs/how_to/error-mitigation.rst
```diff
@@ -144,7 +144,8 @@ The Estimator interface lets users seamlessly work with the variety of error mit
       session.close() 
 
 .. note::
-    As you increase the resilience level, you will be able to use additional methods to improve the accuracy of your result. However, because the methods become more advanced with each level, they require additional sampling overhead (time) to generate more accurate expectation values.     
+    As you increase the resilience level, you will be able to use additional methods to improve the accuracy of your result. However, because the methods become more advanced with each level, they require additional sampling overhead (time) to generate more accurate expectation values.
+    Note that higher resilience levels do not guarantee better quality. Higher levels only mean greater overhead. Each method has its strengths and weaknesses. For example, TREX (Twirled Readout Error eXtinction) is good for shallow circuits since it's readout error mitigation whereas ZNE (Zero Noise Extrapolation) for deeper circuits. PEC can mitigate arbitrary errors but may not work in practice because of its large overhead.
```
kt474, line 148:
```suggestion
    Note that higher resilience levels do not guarantee better quality. Higher levels only mean greater overhead. Each method has its strengths and weaknesses. For example, TREX (Twirled Readout Error eXtinction) is good for shallow circuits because of its readout error mitigation whereas ZNE (Zero Noise Extrapolation) is good for deeper circuits. PEC can mitigate arbitrary errors but may not work in practice because of its large overhead.
```

kt474 (approved):
Small comment but LGTM

docs/how_to/error-mitigation.rst
ikkoham: Thank you very much!

PR #1013, opened 2023-08-10
Fix broken integration/E2E tests

### LINKED ISSUE #1009: Integration test and e2e test are both failing
**Describe the bug**

[Integration test](https://github.com/Qiskit/qiskit-ibm-runtime/actions/workflows/integration-tests.yml) has been failing for >3 weeks now, and [e2e](https://github.com/Qiskit/qiskit-ibm-runtime/actions/workflows/e2e-tests.yml) for 1 week. 

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

**Additional Information**

- **Python version**:
- **Operating system**:

kt474:
### Summary

- Updated `test_device_status` because the CI account has hgps that don't have any real backends
- Refactor `_run_program` because some inputs no longer work after the replacement of hello-world with sampler
- ~~Refactored `cancel()` method - comment below~~

### Details and comments
Fixes part of #1009


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -249,7 +249,11 @@ def cancel(self) -> None:
                 raise RuntimeInvalidStateError(f"Job cannot be cancelled: {ex}") from None
             raise IBMRuntimeError(f"Failed to cancel job: {ex}") from None
         self.cancel_result_streaming()
-        self._status = JobStatus.CANCELLED
+        self.status()
+        if self._status != JobStatus.CANCELLED:
+            raise RuntimeInvalidStateError(
+                f"Failed to cancel job: {self.job_id()} has a status of {self._status}"
+            )
```
kt474, line 256: I think this is a bug - it appears that it's possible to call the cancel [endpoint](https://runtime-us-east.quantum-computing.ibm.com/openapi#/Jobs/cancel_job_jid) and no error is raised but the job is actually completed. 

Either way I don't think we should manually set the status to cancelled and instead actually confirm the status.

PR #1018, opened 2023-08-15
Allow user to define a default account as an environment variable

### LINKED ISSUE #954: Allow users to specify a default channel

Currently, if a user has both a cloud and quantum channel account saved, calling `QiskitRuntimeService()` without any parameters will always default to the cloud channel. It would be useful to specify a default channel, maybe with `os.environ["QISKIT_IBM_CHANNEL"]`, which already exists but is only used when the auth and url are also saved as env variables. 

I believe there may have been a specific reason why the cloud channel is the default (but I don't remember)

merav-aharoni:
### Summary
Allow the user to define the default channel either by an environment variable, QISKIT_IBM_CHANNEL, or by saving the default channel to the account. 

### Details and comments
Order of preference for channel selection:
1. Parameter to `QiskitRuntimeService`
2. Definition in `json` file (defined in `save_account()`)
3. Environment variable (`QISKIT_IBM_CHANNEL`)
4. Default in the code (currently "ibm_cloud")

Fixes #954


jyu00:
can we make it an option as well? see my comment [here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/954#issuecomment-1667794814)

merav-aharoni:
> can we make it an option as well? see my comment [here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/954#issuecomment-1667794814)

@jyu00 - we cannot easily set an environment variable permanently from the Python code. Setting `os.environ[]` will only last for the duration of the program. We could manipulate the `qiskit-ibm.json` file, but I am not sure we want to do that. Any suggestions?

jyu00:
> > can we make it an option as well? see my comment [here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/954#issuecomment-1667794814)
> 
> We could manipulate the `qiskit-ibm.json` file, but I am not sure we want to do that.

Why not?

merav-aharoni:
> > > can we make it an option as well? see my comment [here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/954#issuecomment-1667794814)
> > 
> > 
> > We could manipulate the `qiskit-ibm.json` file, but I am not sure we want to do that.
> 
> Why not?

I started modifying the `qiskit-ibm.json` to contain the default channel. I just want to say that in this case, we won't have an environment variable for the default channel, as was originally suggested, because we can't have both (can cause conflicts).

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -700,6 +702,7 @@ def save_account(
         verify: Optional[bool] = None,
         overwrite: Optional[bool] = False,
         channel_strategy: Optional[str] = None,
+        set_default_channel: Optional[bool] = True,
```
kt474, line 705: Regarding the comment [here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/954#issuecomment-1667794814) I think @jyu00 meant that there should be a way to set an entire account as the default account, not just a default channel. Then this default account would be used if no parameters were passed in to `QiskitRuntimeService()`. Users can still use the env var to set a default channel(original issue), but a default account would take precedence. 

Because it would be a little strange to have a parameter like `set_default_channel` passed in to `save_account` when the value isn't actually saved with the account, it's being saved as an extra value inside `qiskit-ibm.json`. There also wouldn't be a way to remove the `default_channel` value from the config json.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -271,7 +272,7 @@ def _discover_account(
             if channel and channel not in ["ibm_cloud", "ibm_quantum"]:
                 raise ValueError("'channel' can only be 'ibm_cloud' or 'ibm_quantum'")
             channel = channel or self._get_channel_for_auth(auth=auth)
-            if token:
+            if channel and token:
```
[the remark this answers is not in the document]
merav-aharoni: Not always - only if `channel` or `auth` were passed as parameters.

qiskit_ibm_runtime/qiskit_runtime_service.py
merav-aharoni: You are probably right and my misunderstanding. I will wait for input from @jyu00  to make sure, before I make modifications.

qiskit_ibm_runtime/qiskit_runtime_service.py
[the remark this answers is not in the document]
kt474: but it's inside `elif auth or channel:` so one of them will always be defined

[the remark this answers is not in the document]
merav-aharoni: Okay, I see your point now. `_get_channel_for_auth` always returns a channel and not None.

qiskit_ibm_runtime/qiskit_runtime_service.py
merav-aharoni: > Users can still use the env var to set a default channel(original issue), but a default account would take precedence.

@kt474 - you wrote here that the default account should take precedence over the the env var account. It seemed in the meeting today, you both thought the env var should take precedence. Any final opinion?

merav-aharoni: I am copying the existing documentation here:

     An account is selected in the following order:

            - Account with the input `name`, if specified.
            - Default account for the `channel` type, if `channel` is specified but `token` is not.
            - Account defined by the input `channel` and `token`, if specified.
            - Account defined by the environment variables, if defined.
            - Default account for the ``ibm_cloud`` account, if one is available.
            - Default account for the ``ibm_quantum`` account, if one is available.
So I guess this is consistent with what @kt474 wrote above and with the order I specified in my PR.

merav-aharoni: But now I am even more confused: there is a default account defined for `ibm_cloud` and a default for `ibm_quantum`. So I think it would make sense to simply define the `default_channel`, and then we would select the account according to the `default_channel`. What do you think?
The other option would be to have an additional account, name "default" which would probably be a copy of one of the accounts above. Or else to add a field to the account, something like "is_default_account", but I think this is more confusing.

kt474: Yes, I think that's a good solution - we can have a `default_channel` and then it would just select the appropriate default channel account. So setting the env variable `QISKIT_IBM_CHANNEL` will allow users to to set their default channel to the quantum channel. 

If we really wanted to, we could add another "default" flag but that would overcomplicate the already very complicated account management. This wasn't part of the original issue so I'm ok with not having this

merav-aharoni: So I think I will go with adding a "is_default_account" field to the account that is default. It will require going over all the accounts to check which one is default, but that should be so bad, because normally people will only keep at most a handful of accounts.

merav-aharoni:
I changed this PR to save the default account. I suggest you look at the documentation of `AccountManager.get()` to see the precedence order of selecting an account. It is a bit complicated, so I am not so sure this is the right thing to do.

kt474 (approved):
LGTM!

Issue #1019, opened 2023-08-15
Investigate/fix `test_cancel_running_job`


kt474:
This test has been failing for a while now - https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/5863436233/job/15896900505

It appears that when attempting to cancel a running job, calling `job.cancel()` can return a 204 and raise no errors while the job is actually completed. 

This snippet shows that the status of `job` is `CANCELLED` but retrieving the job and status again shows `DONE`, which is the correct status. 

I'm guessing this is some sort of race condition where the job runs too quickly and if this is the case, it should be handled appropriately on the client side. 

```python
def test_cancel_job_running(self, service):
        """Test canceling a running job."""
        job = self._run_program(service, iterations=5)
        if not cancel_job_safe(job, self.log):
            return
        time.sleep(10)  # Wait a bit for DB to update.
        rjob = service.job(job.job_id())
        print("job==rjob is " + str(job == rjob))
        print("job.job_id() = " + str(job.job_id()))
        print("rjob.job_id() = " + str(rjob.job_id()))
        print("job.status() = " + str(job.status()))
        print("rjob.status()) = " + str(rjob.status()))
        self.assertEqual(job.status(), JobStatus.CANCELLED)
```

kt474:
For more context, here in the cancel method we are manually setting the job status to CANCELLED on line 252 even though in some cases the job is actually completed. 

https://github.com/Qiskit/qiskit-ibm-runtime/blob/3d8b9c9f5a81b0c1850cbfe44625216908c41977/qiskit_ibm_runtime/runtime_job.py#L238-L252

robotAstray:
@kt474 I believe there might be an issue both in the `cancel()` function race condition and `test_cancel_job_running`. 
It appears that the job status is not updated correctly, which could be causing the problem. The Job is completing before the `CANCELLATION` requests goes through. 

I think the following might resolve the problem:

1.  **Modify test function:** Check the server-side status after using the `cancel()` method and wait for the state to be updated before proceeding with any assertions in `test_cancel_job_running` case. 
2. **Modify `cancel()` method:**  make sure `cancel()` method  waits until the server-side status is updated before marking the job as "CANCELLED" on the client side.
Let me know what you think and perhaps I can try to implement this.

merav-aharoni:
I think the test can be fixed quite simply by replacing the lines:
```
if not cancel_job_safe(job, self.log):
        return
[...]

PR #1022, opened 2023-08-19
Investigate/fix `test_cancel_running_job()` in `test_job.py`

robotAstray:
### Summary

This PR addresses the issue described in [Issue #1019](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1019),

### Changes Made

~- Modified the `cancel` method in `runtime_job.py` to handle status codes 409 and 204 as cases where job cancellation cannot be performed due to specific conditions. This ensures that the client-side status accurately reflects the server-side status.~

- Updated the `test_cancel_job_running` function to wait for the server-side status update after cancellation before performing assertions. This change ensures that the test accurately reflects the behaviour of cancelling a running job.

Fixes:
  - |
    Fixes a race condition in the function `test_cancel_running_job()` in `test_job()`where job cancellation 
  cannot be performed due to specific conditions.  Refer to #1019 for more details.


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -245,7 +245,7 @@ def cancel(self) -> None:
         try:
             self._api_client.job_cancel(self.job_id())
         except RequestsApiError as ex:
-            if ex.status_code == 409:
+            if ex.status_code in (409, 204):
```
kt474, line 248: Thanks for taking care of this @robotAstray! I think this change is unnecessary because if the status code was 204 there would be no exception caught.

robotAstray: @kt474 Ok Thank you! that makes sense. I will add the changes

robotAstray:
@kt474 I wil fix the release notes as well

test/integration/test_job.py
```diff
@@ -161,6 +161,7 @@ def test_cancel_job_running(self, service):
             service,
             circuits=[ReferenceCircuits.bell()] * 10,
         )
+        wait_for_status(job, JobStatus.CANCELLED)
```
merav-aharoni, line 164: I believe this addition does not really fix the problem, because it causes the test to be skipped if the `job.status() != CANCELLED`, and does not really check if the job was cancelled. See inside `wait_for_status`. We need some additional investigation.

robotAstray: @merav-aharoni apologies, I now realise that the initial solution was not ideal. Furthermore, I was checking the status before calling the cancel function. 

As a possible solution, I was thinking about repeatedly querying the server to check the status of the job and waiting for the status to change to CANCELLED on the server side to provide a more accurate validation.

merav-aharoni: @robotAstray - please see my comment in #1019 .

robotAstray: @merav-aharoni thank you, I have made the changes and tested it locally. It was successful. I will push the changes.

kt474 (approved):
Thanks for taking care of this @robotAstray! I think there is still an issue with the `cancel()` method itself but that can be taken care of in a separate PR 

https://github.com/Qiskit/qiskit-ibm-runtime/pull/1038

PR #1060, opened 2023-09-07
Making account code more generic 

### LINKED ISSUE #1052: Make channels handling code more generic

Today we have two channels - `ibm_quantum` and `ibm_cloud`. The URLs are hard coded, and there is channel specific logic in various places. 

Since there are plans to onboard additional channels, we should make channel account management more generic. That is, to have a configuration file that lists the channels, their default urls, and any channel strategies they support. 

It is unlikely the code can be completely channel agnostic, but we can try to isolate such code to separate classes/methods. If there are actions that can be done on the server side, we can also bring that back to them. 

**Acceptance criteria**

Ideally, when adding a new channel, all one has to do it to update the config file. 

**Suggested fix**

Instead of hard coding the channels, we should have a config file that lists what channels are support. For example, in a format of:
```
channels:
    - name: ibm_quantum
      default_url: api.quantum-computing.ibm.com
      token_required: True
      instance_required: True
      channel_strategy: []
    - name: ibm_cloud 
      ...
```

Then a new channel can just update this file with a new section. `create_account` can return a `GenericAccount` (or use the base `Account` since it's not actually declared as an abstract class despite the comment).

merav-aharoni:
### Summary
Made the `Account` class an abstract class. The specific channel accounts inherit from it: `QuantumAccount` and `CloudAccount`. All the specific-channel behavior is in the subclasses.

### Details and comments
Fixes #1052.


merav-aharoni:
@jyu00 - I would appreciate it if you could take a quick look to tell me whether this is more or less what you had in mind.

PR #1062, opened 2023-09-08
Update default resilience options

### LINKED ISSUE #1048: Set default resilience options based on level

Today the default resilience options (e.g. `extrapolator=LinearExtrapolator`) are always set, regardless which resilience level is used, even though those options only apply to resilience level 2. This could cause confusion to users. 

**Acceptance criteria**

Only options relevant to the resilience level should be set. While only level 2 has options today, the code should be structured to accommodate options for other levels in the future.

kt474:
### Summary

Options that have the value `None` should be removed and not sent to the server. 

### Details and comments
Fixes #1048

### DIFF
--- a/qiskit_ibm_runtime/base_primitive.py
+++ b/qiskit_ibm_runtime/base_primitive.py
@@ -183,11 +183,9 @@ class BasePrimitive(ABC):
             combined["resilience_level"] = Options._DEFAULT_RESILIENCE_LEVEL
 
         self._validate_options(combined)
-        if combined["resilience_level"] == 2:
-            if not combined["resilience"]["noise_factors"]:
-                combined["resilience"]["noise_factors"] = (1, 3, 5)
-            if not combined["resilience"]["extrapolator"]:
-                combined["resilience"]["extrapolator"] = "LinearExtrapolator"
+
+        combined = Options._set_default_resilience_options(combined)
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/base_primitive.py
```diff
@@ -183,6 +183,12 @@ def _run_primitive(self, primitive_inputs: Dict, user_kwargs: Dict) -> RuntimeJo
             combined["resilience_level"] = Options._DEFAULT_RESILIENCE_LEVEL
 
         self._validate_options(combined)
+        if combined["resilience_level"] == 2:
+            if not combined["resilience"]["noise_factors"]:
+                combined["resilience"]["noise_factors"] = (1, 3, 5)
+            if not combined["resilience"]["extrapolator"]:
+                combined["resilience"]["extrapolator"] = "LinearExtrapolator"
+
```
kt474, line 191: Maybe it's better to have this in a helper method? Not sure if this is the best base to check for this

merav-aharoni: I agree with your comment here. I suggest moving all the default setting into a separate method (`set_default_options()`).

merav-aharoni (approved):
Except for that one suggestion of yours, looks fine.

PR #1065, opened 2023-09-08
Remove min execution time check

### LINKED ISSUE #1064: Remove client side enforced min execution time

The server side API for POST /jobs no longer enforces a min value of 300 for `max_execution_time` but the client still does:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/7752b68c749cc434feeacb5caf72d02de91aed46/qiskit_ibm_runtime/options/options.py#L173

The client should leave the validation to the server side and just handle errors that arise from that validation rather than impose it's own limits, unless there is some good reason, but here it just seems redundant and in this case no longer applicable.

**Acceptance criteria**

The min execution time validation is removed.

cc @riley2012 @jyu00

kt474:
### Details and comments
Fixes #1064


merav-aharoni (approved):
LGTM!

Issue #1057, opened 2023-09-06
qiskit.opflow is deprecated


1ucian0:
**Describe the bug**

The module `qiskit.opflow` is used by [`qiskit.opflow` was deprecated in Qiskit 0.43](https://qiskit.org/documentation/release_notes.html#quantuminstance-opflow-and-algorithms-usage-deprecation) (released on 2023-05-04) and will be removed eventually. The migration guide can be found in https://qisk.it/opflow_migration.

**Steps to reproduce**

The grep results in two instances of qiskit.opflow usage:
```sh
grep -ri '\<opflow\>' qiskit_ibm_runtime test
```
```
qiskit_ibm_runtime/estimator.py:from qiskit.opflow import PauliSumOp
test/unit/test_data_serialization.py:from qiskit.opflow import (
```

**Expected behavior**

The usage of `qiskit.opflow` needs to be migrated to [`quantum_info`](https://qiskit.org/documentation/apidoc/quantum_info.html#module-qiskit.quantum_info) or removed in order to continue running with coming Qiskit releases.

**Suggested solutions**

With the help of  https://qisk.it/opflow_migration, migrate to `qiskit.quantum_info`.

kt474:
fixed by #1085, feel free to reopen if something else needs to be done

PR #1079, opened 2023-09-13
Remove importing PauliSumOp, which is deprecated

1ucian0:
Partial fix to https://github.com/Qiskit/qiskit-ibm-runtime/issues/1057

Skip importing `PauliSumOp`, which is only used for type hinting.

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -16,9 +16,9 @@ from __future__ import annotations
 import os
 from typing import Optional, Dict, Sequence, Any, Union
 import logging
+import typing
 
 from qiskit.circuit import QuantumCircuit
-from qiskit.opflow import PauliSumOp
 from qiskit.quantum_info.operators.base_operator import BaseOperator
 from qiskit.primitives import BaseEstimator
 
@@ -31,6 +31,9 @@ from .utils.qctrl import validate as qctrl_validate
 
 # pylint: disable=unused-import,cyclic-import
 from .session import Session
+
+if typing.TYPE_CHECKING:
+    from qiskit.opflow import PauliSumOp
 
 logger = logging.getLogger(__name__)
 


ElePT (commented):
A couple of comments:
1. As far as I know, this typehint was left to maintain consistency with the base classes, which still preserve the `PauliSumOp` input type. Shouldn't this change be coordinated with the base primitives?? (for which, I believe, the plan is to keep the opflow typehint until opflow is removed)
2. There are still a few uses of `PauliSumOp` in the code (lines 125, 161)

1ucian0:
> 1. As far as I know, this typehint was left to maintain consistency with the base classes, which still preserve the `PauliSumOp` input type. Shouldn't this change be coordinated with the base primitives?? (for which, I believe, the plan is to keep the opflow typehint until opflow is removed)

You believe is correct. As a typehint still works, but does not import it at runtime.

> 2. There are still a few uses of `PauliSumOp` in the code (lines 125, 161)

Also typehints, right?

ElePT:
aha, yes, I don't know how I read the changes the other way round (as removing `if typing.TYPE_CHECKING:` instead of adding). No comments then.

kt474:
Can `PauliSumOp` be removed completely then?

1ucian0:
> Can PauliSumOp be removed completely then?

If you do, I think mypy will fail.

kt474:
> > Can PauliSumOp be removed completely then?
> 
> If you do, I think mypy will fail.

I meant if it's being deprecated, will `PauliSumOp` still be an acceptable type for observables in `Estimator`? If it is, what would the appropriate import be?

ElePT:
> I meant if it's being deprecated, will PauliSumOp still be an acceptable type for observables in Estimator? If it is, what would the appropriate import be?

@kt474 Maybe a bit more background: `PauliSumOp` was deprecated in qiskit 0.43 (not removed, but deprecated). When we deprecated it, we removed its internal use within the primitives, but we kept the user-facing typehint to respect the deprecation period of opflow (we wanted to give users some time to update their code) and for consistency with the base primitives (which respect the deprecation period of opflow). In other words, functionally, the typehint can be removed now and the code will work just fine, but interface-wise, we shouldn't remove it before the base primitives do it.  Why can we remove it easily?? Because its almost-direct replacement is `SparsePauliOp`, which:

1. is of type `BaseOperator` (already supported in the primitives)
2. is actually what is currently used in all of the code examples and internally maintained libraries that use the primitives

Does this make sense?

kt474:
> Does this make sense?

Makes sense to me! Thanks for explaining

kt474 (approved):

PR #1085, opened 2023-09-15
Remove opflow and algorithms from serialization tests

### LINKED ISSUE #1057: qiskit.opflow is deprecated

ElePT:
### Summary
I started looking into the unit tests which trigger `qiskit.opflow` and `qiskit.algorithms` deprecation warnings (see https://github.com/Qiskit/qiskit-ibm-runtime/issues/1057) to see what is the best way to address these, and I think that:

1. the optimizer tests can directly be removed
2. the operator tests don't need to test any opflow class beyond PauliSumOp (which is the only one documented as being supported)

Let me know if I missed anything, but from what I see, these tests are not really actionable unit tests. They are an integration test of sorts to check that common inputs/outputs to the encoder/decoder “work”. However, all the encoder does is call whatever object’s `.to_json` method, which is not runtime’s “responsibility” to fix. I guess that a selection of “common” inputs was picked when the tests were designed, and this selection included, for example, the optimizers in `qiskit.algorithms` (which I imagine lead to some edge cases that you wanted to cover). But I think that there is no strict reason why runtime should be checking the optimizers ability to serialize to json, especially now that the classes are deprecated and **not used by the primitives**. At first, I even removed the `PauliSumOp` test, but decided to add it back to be on the safe side, as this is still a documented input type.

### Details and comments
Could be a fix for #1057 @1ucian0 

Note: I also don't think these tests should be moved to, for example, `qiskit_algorithms`, because we never mention that these are json-serializable.

### DIFF
--- a/test/unit/test_data_serialization.py
+++ b/test/unit/test_data_serialization.py
@@ -101,12 +101,12 @@ class TestDataSerialization(IBMTestCase):
             warnings.filterwarnings(
                 "ignore", category=DeprecationWarning, module=r"qiskit\.opflow\."
             )
-            from qiskit.opflow import PauliSumOp
+            from qiskit.opflow import PauliSumOp # pylint: disable=import-outside-toplevel
 
         # catch warnings triggered by opflow use
-        with warnings.catch_warnings(record=True) as w:
+        with warnings.catch_warnings(record=True) as w_log:
             deprecated_op = PauliSumOp(SparsePauliOp(Pauli("XYZX"), coeffs=[2]))
-            self.assertTrue(len(w) > 0)
+            self.assertTrue(len(w_log) > 0)
 
         coeff_x = Parameter("x")
         coeff_y = coeff_x + 1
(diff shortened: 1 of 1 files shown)


kt474 (approved):
LGTM

PR #1092, opened 2023-09-19
Allow the user to define the session_id

### LINKED ISSUE #581: Instantiate a Session with an existing session id
One should be able to instantiate a `Session` object with an existing session id `str`.

**Acceptance criteria**
Users should have the ability to instantiate a `Session` object with its private `_session_id` attribute set.

One implementation might look like:
```
session_id: str = ...
session = Session.from_id(session_id)
```

merav-aharoni:
### Summary
Allow the user to define the `session_id` when creating a new session, rather than always getting the value of the `job_id` of the first job in the session.

### Details and comments
Fixes #581 .

### DIFF
--- a/qiskit_ibm_runtime/session.py
+++ b/qiskit_ibm_runtime/session.py
@@ -75,6 +75,7 @@ class Session:
         service: Optional[QiskitRuntimeService] = None,
         backend: Optional[Union[str, IBMBackend]] = None,
         max_time: Optional[Union[int, str]] = None,
+        session_id: Optional[str] = None
     ):  # pylint: disable=line-too-long
         """Session constructor.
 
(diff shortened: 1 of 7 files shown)


qiskit_ibm_runtime/session.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         service: Optional[QiskitRuntimeService] = None,
         backend: Optional[Union[str, IBMBackend]] = None,
         max_time: Optional[Union[int, str]] = None,
+        session_id: Optional[str] = None,
```
jyu00, line 78: This should not be a parameter to the Session constructor, because it implies you can also change other parameters (e.g. `backend`) of an existing session. Instead it should be a class method like `Session.from_id()`

merav-aharoni: My misunderstanding of the intent of this issue. I thought we want the user to be able to supply the `session_id`. But from what you wrote, the intent is to get a particular session using the `session_id`. I will change it.

merav-aharoni:
I misunderstood the intent of this issue, so this fix is not really relevant. I will open a new PR.

PR #1093, opened 2023-09-19
Support only LocalFoldingAmplifier as noise_amplifier option

merav-aharoni:
### Summary
Removed support for all noise_amplifier options other than `LocalFoldingAmplifier`. Removed deprecation warning. Was deprecated 3 weeks ago, in 0.12.0.

### Details and comments
Fixes part of #784.


docs/how_to/error-mitigation.rst
```diff
@@ -228,7 +219,7 @@ Example of adding ``resilience_options`` into your estimator session
     options.optimization_level = 3
     options.resilience_level = 2
     options.resilience.noise_factors = (1, 2, 3, 4)
-    options.resilience.noise_amplifier = 'CxAmplifier'
+    options.resilience.noise_amplifier = 'LocalFoldingAmplifier'
```
kt474, line 222: We should double check if we want the default value to be `LocalFoldingAmplifier` or just `None`

kt474 (approved):
LGTM

qiskit_ibm_runtime/options/resilience_options.py
```diff
@@ -48,10 +44,9 @@ class ResilienceOptions:
             Only applicable for ``resilience_level=2``.
             Default: ``None``, and (1, 3, 5) if resilience level is 2.
 
-        noise_amplifier (DEPRECATED): A noise amplification strategy. One of ``"TwoQubitAmplifier"``,
-            ``"GlobalFoldingAmplifier"``, ``"LocalFoldingAmplifier"``, ``"CxAmplifier"``.
-            Only applicable for ``resilience_level=2``.
-            Default: "TwoQubitAmplifier".
+        noise_amplifier (DEPRECATED): A noise amplification strategy. Currently only
+        ``"LocalFoldingAmplifier"`` is supported Only applicable for ``resilience_level=2``.
+            Default: "LocalFoldingAmplifier".
```
jyu00, line 49: We don't really need an option if there is only 1 choice. And it shouldn't be marked as `DEPRECATED` without a deprecation warning.

merav-aharoni: Indeed the `DEPRECATED` should have been removed.
At the time, we left `noise_amplifier` as an option because it can be None, for resilience level lower than 2. Do you think we should remove it entirely, assuming the server knows to set it correctly? Or else keep as an option with just one choice that is hard coded?

jyu00: @merav-aharoni Well from UX perspective, we should remove it. However, you are correct that it looks like the server side primitive would blow up if it's not passed in 😞. We can remove it after it's fixed on the server side.

PR #1095, opened 2023-09-19
Only return channel strategy supported backends 

### LINKED ISSUE #1058: Support filtering by Q-CTRL backends
If `q-ctrl` is selected as the channel strategy, then only `q-ctrl` supported backends should be returned. There should be an option to filter backends by channel strategy. 

```python
service = QiskitRuntimeService(..., channel_strategy="q-ctrl")
backends = service.backends()
```

On hold until server side changes are ready

kt474:
### Summary

If a channel strategy is passed in during initialization, only backends supporting that channel strategy will be returned. 
We can add a test when we can test the full Q-CTRL path.

### Details and comments
Fixes #1058


merav-aharoni (approved):
LGTM!

PR #1100, opened 2023-09-21
Warn users if job submitted will exceed quota

### LINKED ISSUE #1088: Issue warning if job will exceed capacity limit
Open users have a per-month quota on quantum time and we should raise a warning if the job submitted has a higher estimated quantum time than the remaining capacity 

The message is returned from the `POST /Jobs` API

kt474:
### Summary

~~Don't merge until at least 9/26, I need to test this on production~~

Warning returned from /jobs in this format:
```python

{'id': 'cm65ft7e49b0008szmdg', 'backend': 'fake_backend1', 'messages': [{'level': 'warn', 'data': 'Your current pending jobs are estimated to consume 7.329993678195504 quantum seconds, but you only have 1 quantum seconds left in your monthly quota; therefore, it is likely this job will be canceled'}]}

```

### Details and comments
Fixes #1088

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -1022,6 +1022,11 @@ class QiskitRuntimeService(Provider):
                 if self._channel_strategy == "default"
                 else self._channel_strategy,
             )
+            messages = response.get("messages")
+            if messages:
+                warning_message = messages[0].get("data")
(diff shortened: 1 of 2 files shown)


kt474:
Thoughts on how to test this? The message will be only returned if the user's 10 minutes of job execution time per month is going to be exceeded

I only tested on staging where my account was temporarily modified to exceeded the limit

jyu00 (approved):

PR #1106, opened 2023-09-25
Remove old deprecations

kt474:
### Summary

Removing bss.seconds deprecation warning, it's been 3 months
Removing passing in a backend into options, this was deprecated 9 months ago 

Also replaced `assertDictContainsSubset` because it was removed in python 3.11

### Details and comments
Fixes part of #784


merav-aharoni (approved):
LGTM.

PR #1108, opened 2023-09-26
Disallow unsupported options

### LINKED ISSUE #1066: Don't allow arbitrary kwargs in options

Each of the classes in `options` is decorated with `@_flexible`, which "is used to dynamically create a new dataclass with the arbitrary kwargs input". This was done because the options were still in flux at the time. 

However, validation was added to _disallow_ arbitrary kwargs input ([for example](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/options/resilience_options.py#L89-L91)), which contradicts with `_flexible`. 

Since options are more settled, we don't really need to force option classes to accept arbitrary kwargs. If we remove that, we also don't need this additional validation, since dataclass would take care of that for us.

We can still allow arbitrary kwargs in `run()` that will be merged into the final dictionary. 

We still need the part of `_flexible` that converts input dictionary to objects. 

**Acceptance criteria**

Options classes no longer need to accept arbitrary kwargs, which can still be specified in `run()`.

merav-aharoni:
### Summary
Raise an exception `Options` is initialized with unsupported options. We still allow arbitrary options when calling `run()`.

### Details and comments
Fixes #1066.


merav-aharoni:
I finally decided to simply raise an exception when an illegal option is given in `__init__`. I didn't touch the `kwargs` parameter of `Options`. I am not sure whether this was the intent or not. Feedback is welcome @kt474 , @jyu00 .

jyu00:
So Python dataclass already enforces that only documented args and kwargs are allowed. For example, if you do

```
from dataclasses import dataclass

@dataclass
class Foo:
    foo: bool

Foo(bar=1)
```

This would raise an exception `TypeError: __init__() got an unexpected keyword argument 'bar'`.

The `Options` class _bypasses_ this by using the `_flexible` decorator, which then allows arbitrary kwargs. But new code was added to invalidate this bypass ([example](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/options/resilience_options.py#L89-L91)). So basically the situation right now is

1. Python dataclass disallows arbitrary kwargs
2. `_flexible` overwrites (1)
3. Code in options classes overwrite (2)

What #1066 is trying to say is that we should just remove (2) and (3), and let Python handles the enforcement. This is more than just removing the `_flexible` decorator though, since it does more than allowing kwargs.

merav-aharoni:
> So Python dataclass already enforces that only documented args and kwargs are allowed. For example, if you do
> 
> ```
> from dataclasses import dataclass
> 
> @dataclass
> class Foo:
>     foo: bool
> 
> Foo(bar=1)
> ```
> 
> This would raise an exception `TypeError: __init__() got an unexpected keyword argument 'bar'`.
> 
> The `Options` class _bypasses_ this by using the `_flexible` decorator, which then allows arbitrary kwargs. But new code was added to invalidate this bypass ([example](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/options/resilience_options.py#L89-L91)). So basically the situation right now is
> 
> 1. Python dataclass disallows arbitrary kwargs
> 2. `_flexible` overwrites (1)
> 3. Code in options classes overwrite (2)
> 
> What #1066 is trying to say is that we should just remove (2) and (3), and let Python handles the enforcement. This is more than just removing the `_flexible` decorator though, since it does more than allowing kwargs.

If I understand you correctly, you want the exception to come from the python `dataclass` code, rather than from the `Primitive.__init__`. I made the change. Please let me know what you think.

qiskit_ibm_runtime/options/utils.py
```diff
@@ -92,8 +92,8 @@ def __new__(cls, *_, **kwargs):  # type: ignore
             orig_field_names.add(fld.name)
 
         for key, val in kwargs.items():
-            if key not in orig_field_names:
-                all_fields.append((key, type(val), field(default=None)))
+            if key in dir(cls):
+                setattr(cls, key, val)
```
jyu00, line 96: While this accomplishes the desired output, creating a new dataclass (line 98) is unnecessary if we are not allowing arbitrary kwargs. I think it'd be worth understanding what this decorator does to fix the issue properly.

jyu00 (changes requested):
> If I understand you correctly, you want the exception to come from the python `dataclass` code, rather than from the `Primitive.__init__`. I made the change. Please let me know what you think.

Correct, since there is no need to write custom code when python does it for you.

kt474 (approved):
LGTM

PR #1119, opened 2023-09-30
Expose new session details 

### LINKED ISSUE #887: Expose session details

An user should be able to query for the remaining session time from a session instance. In addition, it'd also be useful to know about session status, and what the various TTLs are. 

**Acceptance criteria**

- [ ] Expose session details, including
    - backend name
    - created/updated/started/closed timestamps
    - last job start/complete
    - interactive ttl (map to `interactive_timeout`)
    - max ttl (map to `max_time` to match session parameter)
    - state (should map to `open`, `closed` (not accepting job, see [this issue](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1008)), and `cancelled`)
- [ ] Doc update

kt474:
### Summary

Adding a new `session.details()` method that returns the following:

Both channels will return these 11 fields:
```python
{
'id': 'cki5d18m3kt305s4pndg',
 'backend_name': 'ibm_algiers',
 'interactive_timeout': 300,
 'max_time': 28800,
 'active_ttl': 28800,
 'state': 'closed',
 'accepting_jobs': True,
 'last_job_started': '2023-10-09T19:37:42.004Z',
 'last_job_completed': '2023-10-09T19:38:10.064Z',
 'started_at': '2023-10-09T19:37:42.004Z',
 'closed_at': '2023-10-09T19:38:39.406Z'
}

```

Also added `runtime_session.py` and `base.py` back to this repo since we are eventually merging qiskit-ibm-provider back here anyway.

### Details and comments
Fixes #887 

~~Also, open to different options on how we want these new details exposed. Do we want just a method that returns all the information, or maybe specific properties added to the `Session` class?~~

### DIFF
--- a/qiskit_ibm_runtime/api/rest/runtime_session.py
+++ b/qiskit_ibm_runtime/api/rest/runtime_session.py
@@ -13,7 +13,7 @@
 """Runtime Session REST adapter."""
 
 from typing import Dict, Any
-from qiskit_ibm_provider.api.rest.base import RestAdapterBase
+from .base import RestAdapterBase
 from ..session import RetrySession
 
 
@@ -43,7 +43,11 @@ class RuntimeSession(RestAdapterBase):
     def details(self) -> Dict[str, Any]:
         """Return the details of this session."""
         try:
-            return self.session.get(self.get_url("self")).json()
+            if "cloud" in self.session.base_url:
+                return self.session.get(self.get_url("self")).json()
+            else:
+                # TODO: remove this once "v2" is removed from the url path
+                return self.session.get(self.get_prefixed_url("/v2", "self")).json()
         # return None if API is not supported
         except:  # pylint: disable=bare-except
             return None
--- a/qiskit_ibm_runtime/api/rest/runtime_session.py
+++ b/qiskit_ibm_runtime/api/rest/runtime_session.py
@@ -1,0 +1,45 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2022.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Runtime Session REST adapter."""
+
+from typing import Dict, Any
+from qiskit_ibm_provider.api.rest.base import RestAdapterBase
+from ..session import RetrySession
+
+
+class RuntimeSession(RestAdapterBase):
+    """Rest adapter for session related endpoints."""
+
+    URL_MAP = {
+        "self": "",
+        "close": "/close",
+    }
+
+    def __init__(self, session: RetrySession, session_id: str, url_prefix: str = "") -> None:
+        """Job constructor.
+
+        Args:
+            session: RetrySession to be used in the adapter.
+            session_id: Job ID of the first job in a runtime session.
+            url_prefix: Prefix to use in the URL.
+        """
+        super().__init__(session, "{}/sessions/{}".format(url_prefix, session_id))
+
+    def close(self) -> None:
+        """Close this session."""
+        url = self.get_url("close")
+        self.session.delete(url)
+
+    def details(self) -> Dict[str, Any]:
+        """Return the details of this session."""
+        return self.session.get(self.get_url("self")).json()
--- a/qiskit_ibm_runtime/api/rest/runtime_session.py
+++ b/qiskit_ibm_runtime/api/rest/runtime_session.py
@@ -42,4 +42,8 @@ class RuntimeSession(RestAdapterBase):
 
     def details(self) -> Dict[str, Any]:
         """Return the details of this session."""
-        return self.session.get(self.get_url("self")).json()
+        try:
+            return self.session.get(self.get_url("self")).json()
+        # return None if API is not supported
+        except:  # pylint: disable=bare-except
+            return None
--- a/qiskit_ibm_runtime/api/rest/runtime_session.py
+++ b/qiskit_ibm_runtime/api/rest/runtime_session.py
@@ -43,11 +43,7 @@ class RuntimeSession(RestAdapterBase):
     def details(self) -> Dict[str, Any]:
         """Return the details of this session."""
         try:
-            if "cloud" in self.session.base_url:
-                return self.session.get(self.get_url("self")).json()
-            else:
-                # TODO: remove this once "v2" is removed from the url path
-                return self.session.get(self.get_prefixed_url("/v2", "self")).json()
+            return self.session.get(self.get_url("self")).json()
         # return None if API is not supported
         except:  # pylint: disable=bare-except
             return None
--- a/qiskit_ibm_runtime/session.py
+++ b/qiskit_ibm_runtime/session.py
@@ -197,7 +197,7 @@ class Session:
         """Return current session status.
 
         Returns:
-            The current status of the session, including:
+            The current status of the session, including the following.
             Pending: Session is created but not active. It will become active when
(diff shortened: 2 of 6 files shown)


qiskit_ibm_runtime/api/rest/runtime_session.py
```diff
@@ -0,0 +1,53 @@
...
+        """Return the details of this session."""
+        try:
+            if "cloud" in self.session.base_url:
+                return self.session.get(self.get_url("self")).json()
+            else:
+                # TODO: remove this once "v2" is removed from the url path
+                return self.session.get(self.get_prefixed_url("/v2", "self")).json()
+        # return None if API is not supported
+        except:  # pylint: disable=bare-except
+            return None
```
kt474, line 53: > Something to keep in mind with this, and I'm not sure how Qiskit works with this, but the GET /sessions/:id API is only being implemented in the ntc and sw-api-iqp (new job service) repos, not the ntc-iqp-channel fork. The ntc repo is deployed for cloud pay-go and IQP DE (Fraunhofer). The sw-api-iqp service is only going to be deployed in the US.

Copying these concerns from the original issue, I think this logic handles everything but if/when `/v2` is removed from the path, that would be a breaking change?

jyu00:
> Also, open to different options on how we want these new details exposed. Do we want just a method that returns all the information, or maybe specific properties added to the Session class?

Since there are many fields, adding each as a property is inconvenient for both us and the users. However, session status should be a method, and it should return an enum that also indicates whether it's accepting new jobs, e.g. 
`in progress, accepting new jobs`. Please check with the IQP team to see what statuses are used for the UI so we can match.

As for the `details()` method, 
1. the returned dictionary should have the same keys regardless which channel they are using (i.e. don't return `backend_name` for paygo and `backend` for IQP). 
2. some keys should have the same names as parameters/documents we already have. For example, `maxSessionTTL` should become `max_time`, since that's the parameter name we have. #887 has other examples
3. keys should use snake case
4. don't expose `queueName` since it means nothing to the users.

qiskit_ibm_runtime/api/rest/runtime_session.py
kt474: Update, the `/v2` path will be removed, so we should wait for that

qiskit_ibm_runtime/session.py
```diff
@@ -193,6 +193,38 @@ def backend(self) -> Optional[str]:
         """
         return self._backend
 
+    def status(self) -> Optional[str]:
+        """Return current session status."""
+        # TODO implement this when API changes are done
```
mriedem, line 198: What does this mean? The session status is just the `state` field.

qiskit_ibm_runtime/session.py
```diff
@@ -193,6 +193,38 @@ def backend(self) -> Optional[str]:
...
+                    "root_job": response.get("rootJob"),
+                    "root_job_started": response.get("rootJobStarted"),
+                    "last_job": response.get("lastJob"),
+                    "last_job_started": response.get("lastJobStarted")
+                    or response.get("last_job_started"),
+                    "last_job_completed": response.get("lastJobCompleted")
+                    or response.get("last_job_completed"),
+                    "activated_at": response.get("activatedAt") or response.get("started_at"),
+                    "closed_at": response.get("closedAt") or response.get("closed_at"),
+                }
```
mriedem, line 225: Need to work with the IQP API team on this because Jorge has a change up which is going to severely limit the number of fields they return from their API, like as written it doesn't include any of the ttl fields which I think is a mistake.

mriedem: At some point (hopefully in the near future) the cloud API should return an `activated_at` timestamp field and the quantum API should return an `activatedAt` field. Those will indicate when the session last went into the `active` state and can be used with the `active_ttl` to calculate how much time is remaining in the active window for the session (assuming the session has more jobs to run and doesn't go inactive after the interactive_ttl expires).

qiskit_ibm_runtime/session.py
kt474: Originally I thought we would have to map the state `'active' | 'open' | 'inactive' | 'closed'`, to some status like `in progress, accepting new jobs` - but I believe that's going to be handled on the server side now so I can update this when ready.

rachaelhouse:
@kt474 FYI the changes for IQP api are in staging now https://api-dev.quantum-computing.ibm.com/runtime/api#/sessions/GetSessionDetailsController_getSessionDetails

qiskit_ibm_runtime/session.py
mriedem: >but I believe that's going to be handled on the server side now so I can update this when ready.

That's news to me...

qiskit_ibm_runtime/session.py
```diff
@@ -193,6 +193,50 @@ def backend(self) -> Optional[str]:
...
+            state = details["state"]
+            accepting_jobs = details["accepting_jobs"]
+            if state == "open":
+                return "Pending"
+            if state == "active" and accepting_jobs:
+                return "In progress, accepting new jobs"
+            if state == "active" and not accepting_jobs:
+                return "In progress, not accepting new jobs"
+            return state.capitalize()
+
```
kt474, line 213: @jyu00 is something like this what you had in mind?

jyu00: I think `inactive` should also be mapped to `Pending`. 

In addition, the docstring should list out the possible statuses and explain what they mean, including `None`.

qiskit_ibm_runtime/session.py
jyu00: The docstring should explain what these fields mean, or at least point to the docs that do.

kt474: Updated docstrings in [5cbdd16](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1119/commits/5cbdd16519cae64278adf5e5249f789891362495)

qiskit_ibm_runtime/session.py
```diff
@@ -193,6 +193,70 @@ def backend(self) -> Optional[str]:
         """
         return self._backend
 
+    def status(self) -> Optional[str]:
+        """Return current session status.
+
+        Returns:
+            The current status of the session, including:
+                Pending: Session is created but no jobs have been dequeued, or
+                    the ``interactive_timeout`` expired before more jobs were available
```
jyu00:
```suggestion
                Pending: Session is created but not active. It will become active when the next job of this session is dequeued. 
```

qiskit_ibm_runtime/session.py
```diff
@@ -193,6 +193,70 @@ def backend(self) -> Optional[str]:
...
+
+    def details(self) -> Optional[Dict[str, Any]]:
+        """Return session details.
+
+        Returns:
+            A dictionary with session details, including:
+            id: id of the session
+            backned_name: backend used for the session
+            interactive_timeout: The max time (in seconds) between jobs that
+                is allowed to occur to keep the session active
```
jyu00, line 230:
```suggestion
            interactive_timeout: The maximum idle time (in seconds) between jobs that
                is allowed to occur to before the session is deactivated.
```

qiskit_ibm_runtime/session.py
```diff
@@ -193,6 +193,70 @@ def backend(self) -> Optional[str]:
...
+            max_time: Max allowed time for the session to run in seconds, subject to plan limits
```
jyu00, line 231:
```suggestion
            max_time: Maximum allowed time (in seconds) for the session, subject to plan limits
```

qiskit_ibm_runtime/session.py
```diff
@@ -193,6 +193,70 @@ def backend(self) -> Optional[str]:
...
+            active_timeout: The remaining time (in seconds) for the session to
+                be in the active state whilst jobs are running
```
jyu00, line 233:
```suggestion
            active_timeout: The maximum time (in seconds) a session can stay active. 
```

jyu00 (approved):
Some minor changes in wording, but otherwise LGTM!

PR #1123, opened 2023-10-03
Allow users to indicate they are done submitting jobs to a session

### LINKED ISSUE #1008: Allow users to indicate they are done submitting jobs to a session

Session will soon support a new flag `accepting_jobs` that indicates if the session should continue accepting jobs (but not terminating existing queued/running session jobs). This should be exposed to the end users.

**Acceptance criteria**
- [ ] Re-purpose `session.close()` to indicate one is done submitting to the session. This sets `accepting_jobs=false`. 
- [ ] If `session.close()` is invoked manually (i.e. not through context manager), issue a one-time warning about the change. 
- [ ] Add a new method `session.cancel()` for users to terminate all jobs inside a session. 
- [ ] Call `session.close()` upon exiting session in context manager
- [ ] docs like [these](https://qiskit.org/ecosystem/ibm-runtime/how_to/run_session.html#close-a-session) should then be updated to reflect the above changes

Adding in more info from Matt:

The change with the `accepting_jobs` flag, as it's going to be implemented in qiskit-ibm-runtime, is that the `Session` context manager behavior will mark the session as no longer accepting jobs when it exits the context manager. Trying to submit jobs against the session after that will be rejected. The session won't accept more jobs but it will continue to run any queued jobs until they are done or the max time expires (and then the same closed rules apply as above). When a session that is not accepting jobs has run out of jobs to run, it's immediately closed, freeing up the backend to run more jobs rather than wait for the [interactive timeout](https://docs.quantum-computing.ibm.com/run/sessions#interactive-timeout-value).

The Session context manager behavior is like this code snippet:

https://docs.quantum-computing.ibm.com/run/run-sessions#close-a-session

With the accepting jobs change, that snippet can change to this:

```py
    estimator = Estimator()
    job = estimator.run(...)

# The session is no longer accepting jobs but the submitted job will run to completion
result = job.result()
```

The user doesn't need to explicitly close the session and they don't want to wait for `job.result()` (a blocking call) to return before exiting the context manager (the `with` block).

We'll have to be clear when it comes to docs because different versions of qiskit-ibm-runtime will have different behaviors after this is implemented

kt474:
### Summary

WIP

### Details and comments
Fixes #1008

### DIFF
--- a/qiskit_ibm_runtime/api/clients/runtime.py
+++ b/qiskit_ibm_runtime/api/clients/runtime.py
@@ -345,6 +345,17 @@ class RuntimeClient(BaseBackendClient):
         """
         self._api.runtime_session(session_id=session_id).close()
 
+    def session_details(self, session_id: str) -> Dict[str, Any]:
+        """Get session details.
+
+        Args:
+            session_id: Session ID.
+
+        Returns:
+            Session details.
+        """
+        return self._api.runtime_session(session_id=session_id).details()
+
     def list_backends(
         self, hgp: Optional[str] = None, channel_strategy: Optional[str] = None
     ) -> List[str]:
--- a/qiskit_ibm_runtime/api/rest/base.py
+++ b/qiskit_ibm_runtime/api/rest/base.py
@@ -1,0 +1,57 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Base REST adapter."""
+
+from ..session import RetrySession
+
+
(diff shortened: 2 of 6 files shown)


docs/how_to/run_session.rst
```diff
@@ -88,20 +88,17 @@ There is also an interactive timeout value (5 minutes), which is not configurabl
 Close a session
 ---------------
 
-When jobs are all done, it is recommended that you use `session.close()` to close the session. This allows the scheduler to run the next job without waiting for the session timeout,  therefore making it easier for everyone.  You cannot submit jobs to a closed session.
+When the context manager exits or when `session.close()` is explicity called, the session won't accept any more jobs. Any queued jobs will run until completion or until the max time expires.
 
-.. warning::
-  Close a session only after all session jobs **complete**, rather than immediately after they have all been submitted. Session jobs that are not completed will fail.
+When a session that is not accepting jobs has run out of jobs to run, it is immediately closed.
```
kt474, line 94: @beckykd should docs updates be in a separate PR? I'll probably need help with the wording on a lot of these - it will also be a bit tricky since once this is released, older versions of `qiskit-ibm-runtime` will behave differently

kt474:
closing this and reopening on a fresh branch since this no longer depends on a different PR

PR #1137, opened 2023-10-08
Update docstring & remove max_time in `from_id()`

### LINKED ISSUE #1134: Session.from_id() shouldn't accept max_time

#1101 added a `from_id()` method to `Session`. It takes in a `max_time` parameter, which is misleading because you cannot change a session max_time after it's created (and it has to have been created to have an id). 

In addition, this method doesn't have docstring on the parameters. 

**Acceptance criteria**

- [ ] Remove `max_time` from `from_id()`
- [ ] Explain what parameters are in the docstring

kt474:
### Details and comments
Fixes #1134


merav-aharoni (approved):
LGTM

PR #1139, opened 2023-10-09
Allow users to indicate they are done submitting jobs to a session

### LINKED ISSUE #1008: Allow users to indicate they are done submitting jobs to a session

kt474:
### Summary

 - Re-purposed `session.close()` to indicate one is done submitting to the session by using `accepting_jobs = false`.
 - New method `session.cancel()` terminates all jobs inside a session (this is the old close method, still calls `/close`).
 - `session.close()` is called upon exiting session in context manager
 - Removed the `session.close()` in docs and docstrings since it is no longer necessary/recommended 
 - ~~Added `rest/base.py` and `rest/runtime_session.py` back to this repo since we are merging qiskit-ibm-provider into here soon anyway~~ (done in separate PR) 
 - Removed `qiskit_runtime_service/runtime_session.py`, unused file 

### Details and comments
Fixes #1008

I'd like to test this a bit more and add tests, it just takes a while since I need to run on real devices


docs/how_to/run_session.rst
```diff
@@ -88,20 +88,17 @@ There is also an interactive timeout value (5 minutes), which is not configurabl
 Close a session
 ---------------
 
-When jobs are all done, it is recommended that you use `session.close()` to close the session. This allows the scheduler to run the next job without waiting for the session timeout,  therefore making it easier for everyone.  You cannot submit jobs to a closed session.
+When the context manager exits or when `session.close()` is explicitly called, the session won't accept any more jobs. Any queued jobs will run until completion or until the max time expires.
 
-.. warning::
-  Close a session only after all session jobs **complete**, rather than immediately after they have all been submitted. Session jobs that are not completed will fail.
+When a session that is not accepting jobs has run out of jobs to run, it is immediately closed.
```
kt474, line 93: @beckykd - sorry for tagging you twice (closed the first pr) 

should docs updates be in a separate PR? I'll probably need help with the wording on a lot of these - it will also be a bit tricky since once this is released, older versions of qiskit-ibm-runtime will behave differently

qiskit_ibm_runtime/session.py
```diff
@@ -179,10 +176,19 @@ def run(
...
         if self._session_id:
+            self._service._api_client.cancel_session(self._session_id)
+
+    def close(self) -> None:
+        """Update the session so new jobs will not be accepted, but existing
+        queued or running jobs will run to completion. The session will be closed when there
+        are no more jobs to run."""
+        backend = self._service.backend(self._backend)
+        # There is a 500 internal error on IQP if the backend is a simulator
+        if not backend.simulator and self._session_id:
```
kt474, line 191: I'll have to test this again and ask the IQP team, I don't this should error when using simulators

docs/how_to/backends.rst
```diff
@@ -124,9 +124,7 @@ If you are using a runtime session, add the ``backend`` option when starting you
   with Session(service=service, backend="ibmq_qasm_simulator") as session:
        estimator = Estimator(session=session, options=options)
        job = estimator.run(circuit, observable)
-       result = job.result()
-       # Close the session only if all jobs are finished, and you don't need to run more in the session
-       session.close() # Closes the session
+  result = job.result()
```
jyu00, line 127: The indentation looks wrong.

docs/how_to/run_session.rst
```diff
@@ -88,20 +88,17 @@ There is also an interactive timeout value (5 minutes), which is not configurabl
...
 
 .. code-block:: python
 
   with Session(service=service, backend=backend) as session:
```
jyu00, line 97: We don't need `as session` if it's not used.

qiskit_ibm_runtime/session.py
```diff
@@ -179,10 +176,19 @@ def run(
...
+    def cancel(self) -> None:
+        """Cancel all jobs in a session."""
         self._active = False
         if self._session_id:
+            self._service._api_client.cancel_session(self._session_id)
+
+    def close(self) -> None:
+        """Update the session so new jobs will not be accepted, but existing
+        queued or running jobs will run to completion. The session will be closed when there
+        are no more jobs to run."""
```
jyu00, line 188:
```suggestion
        """Close the session so new jobs will no longer be accepted, but existing
        queued or running jobs will run to completion. The session will be terminated once there
        are no more pending jobs."""
```

docs/how_to/run_session.rst
jyu00: We should add `session.cancel()` behavior as well

qiskit_ibm_runtime/session.py
```diff
@@ -179,10 +176,19 @@ def run(
 
         return job
 
-    def close(self) -> None:
-        """Close the session."""
+    def cancel(self) -> None:
+        """Cancel all jobs in a session."""
```
jyu00, line 180:
```suggestion
        """Cancel all pending jobs in a session."""
```

merav-aharoni (commented):
It generally looks okay to me, but I am not sure I understand it well enough to approve. Better wait for other reviewers.

test/unit/test_estimator.py
```diff
@@ -38,15 +37,12 @@ def test_unsupported_values_for_estimator_options(self):
             {"resilience_level": 4, "optimization_level": 3},
             {"optimization_level": 4, "resilience_level": 2},
         ]
-        with Session(
-            service=FakeRuntimeService(channel="ibm_quantum", token="abc"),
-            backend="common_backend",
-        ) as session:
```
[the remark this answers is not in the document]
kt474: I will add these back - removed them originally because I added a check to see if the backend was a simulator. The IQP API will be updated so this check is no longer necessary

qiskit_ibm_runtime/api/rest/runtime_session.py
```diff
@@ -35,11 +37,23 @@ def __init__(self, session: RetrySession, session_id: str, url_prefix: str = "")
...
 
+    def close(self) -> None:
+        """Set accepting_jobs flag to false, so no more jobs can be submitted."""
+        payload = {"accepting_jobs": False}
+        url = self.get_url("self")
+        try:
+            self.session.patch(url, json=payload)
+        except RequestsApiError as ex:
+            # TODO change this to 404
+            if ex.status_code == 500:
```
kt474, line 53: PR should be good to go, just need to change this to `404` once the IQP API changes are in production

qiskit_ibm_runtime/session.py
```diff
@@ -179,9 +176,16 @@ def run(
...
+    def cancel(self) -> None:
+        """Cancel all pending jobs in a session."""
         self._active = False
+        if self._session_id:
+            self._service._api_client.cancel_session(self._session_id)
+
+    def close(self) -> None:
+        """Close the session so new jobs will no longer be accepted, but existing
+        queued or running jobs will run to completion. The session will be terminated once there
+        are no more pending."""
```
jyu00, line 188:
```suggestion
        are no more pending jobs."""
```

qiskit_ibm_runtime/session.py
```diff
@@ -179,9 +176,16 @@ def run(
...
 
-    def close(self) -> None:
-        """Close the session."""
+    def cancel(self) -> None:
+        """Cancel all pending jobs in a session."""
         self._active = False
+        if self._session_id:
+            self._service._api_client.cancel_session(self._session_id)
+
+    def close(self) -> None:
```
jyu00, line 185: This function should set `self._active=False` as well (maybe rename `_active` since it's now confusing). The `_active` flag is only used to check whether you can run a job in a session, and you cannot once `session.close()` is called.

qiskit_ibm_runtime/session.py
```diff
@@ -308,6 +312,7 @@ def __exit__(
         exc_tb: Optional[TracebackType],
     ) -> None:
         set_cm_session(None)
+        self.close()
```
mriedem, line 315: Can there be a way to control this behavior in case I want to continue accepting jobs on the session even after I've exited the context manager?

My use case is I have some test code using pytest. I define a fixture which creates a session:

```py
@pytest.fixture
def session(service):
    """Creates and yields a Session on the configured backend"""
    session = Session(service=service, backend=os.getenv('IBM_CLOUD_BACKEND'))
    yield session
    # Make sure that we always close the session so there isn't idle time on the backend.
    LOG.info(f'closing session {session.session_id}')
    session.close()
```

I have a utility function to submit jobs, optionally against a session (from the fixture):

```py
def run_job(service, session=None, wait=True):
    """Runs a Sampler job

    :param service: The QiskitRuntimeService instance to use to run the job
    :param session: Optional session to use for the job. If provided the
    session is left open.
    :param wait: If true, waits for the job to complete.
    :returns: RuntimeJob
    """
    options = Options(
        resilience_level=0,
        execution=dict(shots=1)
    )
    if session:
        with session:
            sampler = Sampler(options=options)
            job = sampler.run(ReferenceCircuits.bell())
            LOG.info(f'created job: {job.job_id()}; session: {job.session_id}; backend: {get_job_backend(job)}')
    else:
        # Only get the backend if one is specified otherwise Qiskit returns
        # the first available which is not what we want if we're trying to
        # let the scheduler pick the backend from the default resource set.
        if 'IBM_CLOUD_BACKEND' in os.environ:
            backend = service.backend(os.getenv('IBM_CLOUD_BACKEND'))
        else:
            backend = None
        sampler = Sampler(backend=backend, options=options)
        job = sampler.run(ReferenceCircuits.bell())
        LOG.info(f'created job: {job.job_id()}; backend: {get_job_backend(job)}')

    if wait:
        wait_for_final_state(job)
    return job
```

And then I have a test which submits jobs against the session before explicitly closing it:

```py
def test_close_session_with_queued_jobs(service, session):
    with session:
        jobs = []
        # concurrent_jobs=3 (for AoT) so create 4 jobs so we're sure to have one queued
        for _ in range(4):
            job = run_job(service, session=session, wait=False)
            jobs.append(job)

        # Immediately explicitly close the session before at least the last job can run.
        LOG.info(f'closing session {session.session_id}')
        session.close()
```

With the change proposed here, when the `with session:` exits in `run_job`, trying to submit another job against the same session will fail.

It'd be nice if there were a way to control that behavior to opt out of it if you know what you're doing, so make setting `accepting_jobs=False` on exit the default, but be able to override that if you have a case for it.

kt474: @jyu00 thoughts on this? We could maybe have another method like `session.open` that sets `accepting_jobs` to true.

mriedem: That feels a bit clunky because I'd have to call it every time I submit a job to the session. I was thinking more like a flag you could set on the session to control the behavior when the context manager exits.

drew-distefano: @mriedem can you help me better understand the reason for the change you're asking for? Does the use case also exist outside of the tests you're describing?

jyu00: > in case I want to continue accepting jobs on the session even after I've exited the context manager?

The point of a context manager is to clean up when the context ends. In your case, you should not use a context manager and just pass `session` directly to the primitive, e.g. `Sampler(session=session, options=options)`.

mriedem: > In your case, you should not use a context manager and just pass `session` directly to the primitive, e.g. `Sampler(session=session, options=options)`.

This is what I was wondering, if I should be using `session.run` or something else - sounds like I should be using `Sampler`. I figured there were multiple options. So I'll change to do that.

qiskit_ibm_runtime/session.py
```diff
@@ -179,8 +176,16 @@ def run(
 
         return job
 
+    def cancel(self) -> None:
+        """Cancel all pending jobs in a session."""
+        self._active = False
+        if self._session_id:
+            self._service._api_client.cancel_session(self._session_id)
+
     def close(self) -> None:
```
mriedem, line 185: Since this is a behavior change will this trigger a minor version bump when this is released? I'm thinking about how my code explicitly calls `close()` today when my session jobs are done running. Now it won't close the session, but the end result is similar - it should close the session indirectly this way because if there are no more jobs to run when accepting_jobs gets set to false the session will close.

But if I had a case where I wanted to kill a session _before_ jobs run and was relying on `close()` to do that, then I'd need to change to use `cancel()`. And that feels like a minor version bump to me. Assuming people define requirements like `qiskit-ibm-runtime~=0.12` then they could opt into the change when ready.

mriedem (commented):
Just a question inline about versioning when this is released.

mriedem:
> I'd like to test this a bit more and add tests, it just takes a while since I need to run on real devices

@kt474 ping me next week and we can get you into a hub that has access to `fake_backend1` which is essentially a simulator but not a simulator, it just doesn't run jobs on a real backend but the API flow is all the same.

qiskit_ibm_runtime/session.py
kt474: Yes, this will trigger a minor version bump - the next release will be `0.13` (hopefully on monday)

jyu00:
> New method session.cancel() terminates all jobs inside a session (this is the old close method, still calls /close).

@mriedem If a session job is already running, does this fail the running job?

releasenotes/notes/session-accepting-jobs-d7ef6b60c0f5527b.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - |
+    :meth:`qiskit_ibm_runtime.Session.close` has been updated to mark a ``Session`` as no longer
+    accepting new jobs. The session won't accept more jobs but it will continue to run any 
+    queued jobs until they are done or the max time expires. This will also happen
+    automatically when the session context manager is exited. When a session that is not accepting 
+    jobs has run out of jobs to run, it's immediately closed, freeing up the backend to run more jobs rather 
+    than wait for the interactive timeout.
+
+    The old close method behavior has been moved to a new method, 
+    :meth:`qiskit_ibm_runtime.Session.cancel`, where all jobs within a session are
```
jyu00, line 12:
```suggestion
    :meth:`qiskit_ibm_runtime.Session.cancel`, where all queued jobs within a session are 
```

jyu00:
> > New method session.cancel() terminates all jobs inside a session (this is the old close method, still calls /close).
> 
> @mriedem If a session job is already running, does this fail the running job?

ok Matt confirmed the running job wouldn't fail. Nothing in the docstring says otherwise, so I think we are good.

kt474:
> ok Matt confirmed the running job wouldn't fail. Nothing in the docstring says otherwise, so I think we are good.

Still blocked by IQP, closing sessions doesn't seem to work at all after their most recent prod deploy 
Opened issue 693 in sw-api-iqp

PR #1140, opened 2023-10-10
Migrate fake backends from `qiskit.providers.fake_provider`

ElePT:
### Summary
This PR replaces https://github.com/Qiskit/qiskit-ibm-provider/pull/740 (see original PR for context on the move) in anticipation of the deprecation of the provider repository. This move was previously consulted with @kt474, and it would allow us to directly point users to the final location of the module in the Qiskit deprecation warning (planned for 0.45). An alternative to this PR would be merging https://github.com/Qiskit/qiskit-ibm-provider/pull/740 and having the fake backends migrated in a second step. Tagging @1ucian0 for reference. 

### Details and comments
- Fixes https://github.com/Qiskit/qiskit-ibm-provider/issues/511
- This time I have not performed a history-preserving copy, because I was not sure of how needed it would be and it was quite noisy, but I am open to re-doing it if necessary.


kt474 (approved):

kt474:
@ElePT If everything still looks good on your end, I can get this in the next release

ElePT (commented):
Thanks @kt474, including this PR in the next release would be great. I have given it some thought and decided to make a few changes, but these should hopefully not take long to review, they make the PR simpler. I was not too happy duplicating the base classes, I think this increases the maintenance load in the `qiskit-ibm-runtime` repository and doesn't really provide any advantage, as these classes are still actively developed in the `qiskit` repo (bug fixes, and even new features).  So I have decided to reduce the migrated files to the minimal necessary to enable the use of fake backends here, that is, the `backends` directory and the `fake_provider.py` file (the only duplication). I have also **not** included `FakeQasmBackend` in the list of migrated backends (I think the focus here are real device snapshots).

Until the code diverges (if we want the fake backends to be closer to `BackendV2`), I think that this is the easiest path, and we can always introduce new base classes in the future. What do you think?

qiskit_ibm_runtime/fake_provider/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+
+        from qiskit_ibm_runtime import IBMProvider
+        from qiskit_aer import AerSimulator
+
+        # get a real backend from a real provider
+        provider = IBMProvider()
+        backend = provider.get_backend('ibmq_manila')
+
+        # generate a simulator that mimics the real quantum system with the latest calibration results
+        backend_sim = AerSimulator.from_backend(backend)
```
ElePT, line 81: one more comment, could you let me know what the new import path of the `IBMProvider` will be and whether this snippet is correct?

kt474: Do you need the the provider here? Can you just retrieve the backend from the runtime service? 

```python
backend = service.get_backend('ibmq_manila')
```

ElePT: That's a good point, this was just old code that got updated automatically. Fixed in [9b4ad12](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1140/commits/9b4ad12ad4a26669a170448393d3e4d9f12d3fdf).

ElePT: From my side, I don't have further comments on the PR, it's good to go.

PR #1150, opened 2023-10-16
Log instance on initialization and when running a job

### LINKED ISSUE #1126: Print the instance used to connect or run on the service

When connecting to the quantum channel, when the user doesn't specifying an instance or they have setup the environment variable `QISKIT_IBM_INSTANCE`, it's hard to tell which instance the provider chose so I have no clue which instance I'm using.

**Acceptance criteria**

When connecting to the service, if one instance is selected (including by the env variable or when the user specifies one programatically), print it. When running a job print which instance it's using.

kt474:
### Summary

At initialization, if not passed in directly, the default `instance` selected by the provider 
will be logged at the "INFO" level. When running a job, if the backend selected is not in 
the default instance but in a different instance the user also has access to, that instance 
will also be logged.

### Details and comments
Fixes #1126

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -219,8 +219,10 @@ class QiskitRuntimeService(Provider):
                 for backend_name in hgp.backends:
                     if backend_name not in self._backends:
                         self._backends[backend_name] = None
-            if not self._account.instance:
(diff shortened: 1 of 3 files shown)


merav-aharoni (approved):
It looks fine. The only thing I am wondering about is if `_default_instance` is a representative name. Because it is not really the default, but rather the currently used instance.

PR #1153, opened 2023-10-17
Add deprecation warnings for everything related to custom programs

### LINKED ISSUE #1130: Deprecate custom programs
Add a deprecation warning when custom programs are used. After a 1 month deprecation period we can then fully remove custom program support in the provider

Docs will also have to be updated, all mentions of private programs should be removed.

kt474:
### Summary

- Added deprecation messages to all methods related to custom programs 
- Removed `test_programs.py` (I think it's ok to remove this now since it's not worth fixing the tests if we're going to completely remove custom programs anyway) 

Don't merge/release until Nov 1st

### Details and comments
Fixes #1130 

We can update documentation in a separate PR when we completely remove custom programs


merav-aharoni (approved):
LGTM.
Question (probably for the next stage): how do we continue allowing internal users to use custom programs while disabling them for external users?

kt474:
> LGTM. Question (probably for the next stage): how do we continue allowing internal users to use custom programs while disabling them for external users?

We are going to create an internal fork of this repo where custom programs are still enabled

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -884,6 +897,12 @@ def program(self, program_id: str, refresh: bool = False) -> RuntimeProgram:
             RuntimeProgramNotFound: If the program does not exist.
             IBMRuntimeError: If the request failed.
         """
+        issue_deprecation_msg(
+            msg="Custom programs are being deprecated",
+            version="0.13.1",
```
jyu00, line 902:
```suggestion
            version="0.14.0",
```

We need a minor version bump for a deprecation.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -884,6 +897,12 @@ def program(self, program_id: str, refresh: bool = False) -> RuntimeProgram:
...
+            remedy="Please use the provided primitives instead.",
+            period="1 month",
```
jyu00, line 904: We should point people to the migration guide. Also the deprecation period is technically slightly less than a month. So the message should look something like

`
Custom programs are being deprecated with qiskit-ibm-runtime 0.14.0 and will be removed on November 27, 2023. You can instead convert your custom programs to use Qiskit Runtime primitives with Quantum Serverless. Refer to the migration guide for instructions: https://qiskit-extensions.github.io/quantum-serverless/migration/migration_from_qiskit_runtime_programs.html
`

PR #1162, opened 2023-10-19
Fix templates documentation

arnaucasau:
### Summary
This PR fixes the problem of having duplicated methods when a class has inherited_members.

In the original `class.rst`, we have two consecutive for, one for all methods, and another one for only the inherited_members. With this two for, we are duplicating the entries of the inherited methods (they appear two times in the page).


Eric-Arellano (approved):
Thanks. This looks good

arnaucasau:
Thank you for the review @Eric-Arellano! @kt474 I don't have write access to merge the PR, if you think it's ready to be merged and have time, can you do it for me, please?

PR #1163, opened 2023-10-19
Update session.from_id()

kt474:
### Summary

`backend` should not be a parameter in `from_id()` because sessions do not support multiple backends. Also, the passed in `session_id` must be an existing `session_id` . 

Although I'm not sure how useful this will be because a new session created from an existing session id will likely not be accepting jobs (because this happens automatically when the original session's context manager exits). 

### Details and comments
Addresses this comment https://github.com/Qiskit/qiskit-ibm-runtime/issues/1134#issuecomment-1753198038


kt474:
@merav-aharoni @jyu00 thoughts on the `from_id` method? I know we briefly talked about this before but I don't see how this method would be useful after the `accepting_jobs` change. The original use case ([here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/581#issuecomment-1664665575)) was for a way to create a session from a previous session id and then run jobs again - this isn't possible unless we set `accepting_jobs` back to `True`

jyu00:
> @merav-aharoni @jyu00 thoughts on the `from_id` method? I know we briefly talked about this before but I don't see how this method would be useful after the `accepting_jobs` change. The original use case ([here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/581#issuecomment-1664665575)) was for a way to create a session from a previous session id and then run jobs again - this isn't possible unless we set `accepting_jobs` back to `True`

There were 2 use cases from the original issue: 
1. (This was an offline discussion) A user wants to run multiple jobs in parallel using different processor / container, but within the same session. `from_id` is still useful in this case as long as they don't use a context manager.
2. A user is submitting consecutive jobs to the backend within a session, and their local system crashes. This is also still a valid  use case since the crash may happen before the session is closed.

qiskit_ibm_runtime/session.py
```diff
@@ -279,28 +280,28 @@ def service(self) -> QiskitRuntimeService:
         return self._service
 
     @classmethod
-    def from_id(
-        cls,
-        session_id: str,
-        service: Optional[QiskitRuntimeService] = None,
-        backend: Optional[Union[str, IBMBackend]] = None,
-    ) -> "Session":
+    def from_id(cls, session_id: str, service: QiskitRuntimeService) -> "Session":
```
jyu00, line 283: You need to deprecate the parameter before removing it. So during the deprecation period it needs to check if the backends match.

merav-aharoni (approved):
Looks okay to me.

PR #1170, opened 2023-10-24
Don't block for the first job in a session

### LINKED ISSUE #1165: Sampler.run blocks for first job in session to complete if no backend was specified
**Describe the bug**

Using the ibm_cloud channel, I create a Session without a backend specified to let the server pick the backend. 

I run a Sampler job using:

```py
        channel="ibm_cloud",
        token=os.getenv('IBM_CLOUD_API_KEY'),
        instance=os.getenv('IBM_CLOUD_CRN'),
        url=os.getenv('IBM_CLOUD_URL', "https://test.cloud.ibm.com"))
session = Session(service=service)
options = Options(
        resilience_level=0,
        execution=dict(shots=1)
    )
sampler = Sampler(options=options, session=session)
job = sampler.run(ReferenceCircuits.bell())
```

The `run()` call blocks until the job is complete. This blocks me from being able to submit more jobs in the same session in batch mode.

**Steps to reproduce**

See above.

**Expected behavior**

`Sampler.run()` should return once I've submitted the job and not block waiting for the backend to be determined once the first job in the session is complete.

**Suggested solutions**

I believe the relevant code is here:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.12.2/qiskit_ibm_runtime/session.py#L178

That `job.backend()` call is blocking when no backend is specified when creating the session:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.12.2/qiskit_ibm_runtime/runtime_job.py#L258-L259

Maybe pass a `timeout` value of 0 and handle `RuntimeJobTimeoutError` and just leave backend unset in the Session instance?

**Additional Information**

- **qiskit-ibm-runtime version**: 0.12.2
- **Python version**: 3.8

kt474:
### Details and comments
Fixes #1165


merav-aharoni:
I am not quite sure how this will work and how you can check it. One thing that maybe should be changed is the method Session.backend(). Perhaps it should be as following:
```
    def backend(self) -> Optional[str]:
        self._backend = job.backend().name
        return self._backend
```
because calling `run` may not set the backend.

kt474:
> I am not quite sure how this will work and how you can check it. One thing that maybe should be changed is the method Session.backend(). Perhaps it should be as following:
> 
> ```
>     def backend(self) -> Optional[str]:
>         self._backend = job.backend().name
>         return self._backend
> ```
> 
> because calling `run` may not set the backend.

`Session.backend()` wouldn't have access to "job" so I'm not sure how that could work

merav-aharoni:
> > I am not quite sure how this will work and how you can check it. One thing that maybe should be changed is the method Session.backend(). Perhaps it should be as following:
> > ```
> >     def backend(self) -> Optional[str]:
> >         self._backend = job.backend().name
> >         return self._backend
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > because calling `run` may not set the backend.
> 
> `Session.backend()` wouldn't have access to "job" so I'm not sure how that could work

You are right of course. My concern is that if that `self._backend` is not set in `Session.run()`, it won't be set at all, and the user can't know what backend their job ran on. Do you have any idea how to resolve this?

drew-distefano:
@kt474 to retest

kt474:
> You are right of course. My concern is that if that self._backend is not set in Session.run(), it won't be set at all, and the user can't know what backend their job ran on. Do you have any idea how to resolve this?

Even if there is some error and `session._backend` is `None` the job will always have a backend because this value is retrieved from the server. 

On the cloud channel when there is no backend passed in to the session, the backend is set in the scheduler - I don't think we have to worry about not being able to see which backend a job ran on.

merav-aharoni (approved):
Looks fine.

PR #1184, opened 2023-10-31
Cleaned code involving noise_amplifier options

merav-aharoni:
### Summary
Removed the code involving noise_amplifier documentation and checking, since currently only one option is supported.

### Details and comments
See comment in  #1093.

### DIFF
--- a/qiskit_ibm_runtime/options/resilience_options.py
+++ b/qiskit_ibm_runtime/options/resilience_options.py
@@ -41,10 +41,6 @@ class ResilienceOptions:
             Only applicable for ``resilience_level=2``.
             Default: ``None``, and (1, 3, 5) if resilience level is 2.
 
-        noise_amplifier (DEPRECATED): A noise amplification strategy. Currently only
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/options/resilience_options.py
```diff
@@ -53,7 +49,7 @@ class ResilienceOptions:
             Default: ``None``, and ``LinearExtrapolator`` if resilience level is 2.
     """
 
-    noise_amplifier: NoiseAmplifierType = None
+    noise_amplifier: NoiseAmplifierType = "LocalFoldingAmplifier"
```
kt474, line 52: We can't merge this until there is an update on the server side right? Because this would add `noise_amplifier` to every job, including those without `resilience_level` 2

merav-aharoni: As far as I understand, this will have no effect on other resilience levels, because ZNE won't be run. @jyu00 - any comment on this?

PR #1185, opened 2023-10-31
Pydantic options

merav-aharoni:
### Summary
Changed the `Options` classes to be of type `pydantic.dataclass`, and use automatic validation.

### Details and comments
Fixes #


merav-aharoni:
@jyu00 - here is a first version of using `pydantic` for options. It works partially. A few problems so far:

1. In `test_options.py`, changed the exception types to `Exception` . Need to check if there is a more specific exception type.
2. There is a problem with importing `qiskit_aer`. I didn't look into it, did a temporary hack to be able to run some tests.
3. We had an environment variable to control whether to perform validations or not. I don't see how this can be implemented now.
4. Didn't change `RuntimeOptions` to use `pydantic` yet.

jyu00:
Thanks @merav-aharoni! I'm in the process of updating the experimental branch to use pydantic for the new options, so I'll likely copy some of your code 😁

merav-aharoni:
So for now, I am stopping work on this. Let me know if you need anything more here.

drew-distefano:
Closing as Jessie has incorporated these changes in another PR

PR #1196, opened 2023-11-05
Make sessions thread safe

### LINKED ISSUE #1195: Make session setup thread-safe
Using sessions in a concurrent environment can lead to send multiple jobs starting the session. Race conditions can appear when reading the attribute `_session_id` in which multiple threads can read it as `None`, making the [starter condition `self._session_id is None`](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/session.py#L166) equals `True` and, consequently, sending multiple "session starter jobs". 

Motivation arise from the need of storming a session with multiple jobs in order to make the most of the parallelization built into IBM quantum computers. And so, sending multiple interactive experiments is a good way of exploiting parallelism.

**Acceptance criteria**
When using a session, the session must ensure it sends one, and only one starter job.

delapuente:
### Summary

Locks code accessing the session underlying service to ensure the session sends only one job starting the session.

### Details and comments
Fixes #1195 

The lock prevents other threads from sending new jobs to the service only if the session has not been setup yet. I.e.: when sending the first "session starter job", or root job. After this happens, the code is left unlocked to allow threads to access the service concurrently.

Error handling is needed to ensure the lock is always released.


qiskit_ibm_runtime/session.py
```diff
@@ -25,6 +26,8 @@
 from .program.result_decoder import ResultDecoder
 from .ibm_backend import IBMBackend
 
+_session_setup_lock = Lock()
```
jyu00, line 29: why making this a global variable instead of part of `Session`? If I create 2 `Session` objects for 2 different backends, one shouldn't have to wait for the other.

test/unit/test_session.py
```diff
@@ -113,6 +117,28 @@ def test_run(self):
...
+        service.run = Mock(side_effect=_wait_a_bit)
+
+        session = Session(service=service, backend="ibm_gotham")
+        with ThreadPoolExecutor(max_workers=5) as executor:
+            results = list(map(lambda _: executor.submit(session.run, "", {}), range(5)))
+            wait(results)
+
+        calls = service.run.call_args_list
+        session_starters = list(filter(lambda c: c.kwargs["start_session"] is True, calls))
+        self.assertEqual(len(session_starters), 1)
```
jyu00, line 140: Since both threads would encounter the delay in `_wait_a_bit`, it's not a definitive test on whether the lock is working. A better test would be that only the first one has to wait. So in the old path, the second one who doesn't have to wait would have seen the empty job id and proceed to set it before the first is done waiting.

qiskit_ibm_runtime/session.py
delapuente: You're right. Actually, if you create two separated sessions, even for the same backend, they should not wait for each other. I will modify.

test/unit/test_session.py
delapuente: That is intended although I see it as a definitive test so let me try to explain what this test is trying to do with the hope you can catch the flaw in my reasoning, and not in my implementation. The test tries to demonstrate that, without the lock, several threads will evaluate the following parameter expression as `True`, starting the session more than once, regardless setting the session id twice.

```python
job = self._service.run(
                ...
                start_session=self._session_id is None,
                ...
            )
```

Making all the threads wait, gives an opportunity to the last ones to reach the parameter evaluation point. Actually, after reading your comments, I realize making them wait for more than [`sys.getswitchinterval()`](https://docs.python.org/3/library/sys.html#sys.getswitchinterval), makes sure more than one will hit the offending line.

I can change the test to wait double of that interval, instead of a random value. I can also reduce the number of threads to 2 which will make the test more deterministic and easy to debug.

Does it make sense?

jyu00 (approved):

PR #1203, opened 2023-11-08
Make sure backend.run and Sampler.run can run in a single session

### LINKED ISSUE #1180: Enable Session to contain both primitive.run and backend.run
After the implementation of `backend.run` is complete, we would like to be able 
to run both primitive.run and backend.run in a single session. See #1138.

**Acceptance criteria**
Be able to run something like:
```
backend = service.backend('ibmq_qasm_simulator')

with Session(service, backend=backend) as session:
    
    job1 = sampler.run(bell)
    job2 = backend.run(bell) # this would not run in a session 
```

merav-aharoni:
### Summary
Added a test to make sure they run properly.

### Details and comments
Fixes #1180


merav-aharoni:
There was a proposal to issue a warning to let the user know that if `backend` is run in `qiskit_ibm_runtime.Session`, then it will actually not be executed in the session (and vice versa). I couldn't find a way to check whether this occurs. Any suggestions?

kt474:
> There was a proposal to issue a warning to let the user know that if `backend` is run in `qiskit_ibm_runtime.Session`, then it will actually not be executed in the session (and vice versa). I couldn't find a way to check whether this occurs. Any suggestions?

That's true, I don't see a good way to do this either - maybe we can just write in the docs that this type of behaviour is not supported?

merav-aharoni:
I added a warning for the case that a `Primitive` is run within a `IBMBackend` session. The opposite is not possible, because it creates a circular import.This occurs because `Session` includes an `IBMBackend`, so `IBMBackend` cannot include a `Session`.

merav-aharoni:
@kt474 - I think this can be reviewed now. I think a release note is not needed, because it is really part of the `backend.run()` PR.

kt474 (approved):
LGTM

PR #1205, opened 2023-11-08
Remove caching of results in RuntimeJob

### LINKED ISSUE #1159: Support `refresh` functionality
Followup from #1138:
@jyu00 wrote:
> I guess we can keep refresh since it was a [feature request] (https://github.com/Qiskit/qiskit-ibmq-provider/issues/466), and it should be fairly easy to do (e.g. just call the API again).

>On a second thought, instead of adding a refresh, we should just not cache the results and call server again if someone calls result() again. Results can be large and caching them in jobs means there is no easy way for a user to delete them to free up memory.

**Acceptance criteria**

merav-aharoni:
### Summary
Removed the `RuntimeJob._results` member, and don't store the results every time they are retrieved. Instead, retrieve the results from the server every time.
Also fixed several related tests.
Noticed `job_tags` was not passed as a parameter when creating a new `RuntimeJob`, so added it.

### Details and comments
Fixes #1159

### DIFF
--- a/qiskit_ibm_runtime/ibm_backend.py
+++ b/qiskit_ibm_runtime/ibm_backend.py
@@ -787,6 +787,7 @@ class IBMBackend(Backend):
                 program_id=program_id,
                 session_id=session_id,
                 service=self.service,
+                tags=job_tags,
             )
             logger.debug("Job %s was successfully submitted.", job.job_id())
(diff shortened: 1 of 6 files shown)


kt474 (approved):
LGTM

Issue #1023, opened 2023-08-21
Remove `qiskitrc` related code 


kt474:
In https://github.com/Qiskit/qiskit-ibm-runtime/pull/370 we added a way for users to automatically copy/use their old credentials from `qiskit-ibmq-provider` in this repo. `qiskit-ibmq-provider` is now fully deprecated so we no longer need this functionality. 

The `qiskitrc` file credentials were copied over to the new `qiskit-ibm.json` credentials file.

mtreinish:
Is there actively a reason we **need** to do this? I feel like leaving the backwards compatibility code in place makes sense since just because we've archived the ibmq project doesn't mean people aren't using it still. Especially since it's still installable and will work.

merav-aharoni:
I find it confusing from a developer's perspective. I actually used `_QISKITRC_CONFIG_FILE` in an issue I am working on now, and it took me a while to realize my mistake.
I think we should at least add a comment in the code not to use this. And I think we should deprecate it at some point in the future.

jyu00:
> Is there actively a reason we **need** to do this? I feel like leaving the backwards compatibility code in place makes sense since just because we've archived the ibmq project doesn't mean people aren't using it still. Especially since it's still installable and will work.

People using ibmq-provider can still use `qiskitrc`. I think the proposal here is to just stop using `qiskitrc` in `qiskit-ibm-runtime` (not deleting it), since it moved to the new file over a year ago. 

> And I think we should deprecate it at some point in the future.

I thought it was deprecated (with a deprecation warning) when we moved to the json file? 😅

merav-aharoni:
There was a notification on this in https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/releasenotes/notes/0.1/account-management-b9908e0b62dc1d81.yaml. However, I didn't see a relevant deprecation message in the code. Much of the `qiskitrc` related code was removed as far back as 2021. It seems we only have some leftovers now. 
Do you think we still need to issue a deprecation warning on this, or is it safe to simply remove the related code? @kt474 , @jyu00 .

PR #1219, opened 2023-11-15
Added deprecation warning when using qiskitrc file

merav-aharoni:
### Summary
Deprecation of qiskitrc file before complete removal.

### Details and comments
 See review comment in #1023


merav-aharoni:
When I tried to use a `DeprecationWarning` rather than a simple `warning`, I get a failure on `test_enable_account_by_qiskitrc`. As far as I understand, the reason is that in the tests, deprecation messages are treated as errors. I realize this should only hold for deprecation messages from `qiskit`. However, when I comment out the line 
```
        warnings.filterwarnings("error", category=DeprecationWarning, module="qiskit")

```
in `ibm_test_case.py`, the deprecation message does not cause a failure.
Any suggestions?

kt474:
https://github.com/Qiskit/qiskit-ibm-runtime/blob/d23fa543ef60c194ddce223b50740894a287dd22/test/ibm_test_case.py#L51

I see, I shouldn't have used `module="qiskit"` here since that would match all modules starting with qiskit, which includes qiskit_ibm_runtime. Instead, regex should be used and we can replace it with `module=r'^qiskit$'`

merav-aharoni:
Is this the correct version number? period for removal?

merav-aharoni:
@kt474 - when I run the test `test_enable_account_by_qiskitrc` with the deprecation warning, the test passes, but I get the following messages, about a file not being closed:
```
OK
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmppaexvbyw' mode='w+' encoding='UTF-8'>
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmpek7vt_qi' mode='w+' encoding='UTF-8'>
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmp3vt623pj' mode='w+' encoding='UTF-8'>
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmpz0t4dfjo' mode='w+' encoding='UTF-8'>
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmprz2cq7y4' mode='w+' encoding='UTF-8'>
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmp35mtw27q' mode='w+' encoding='UTF-8'>
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmpez_a_afy' mode='w+' encoding='UTF-8'>
```
Any idea what is causing this?

kt474:
> Any idea what is causing this?

Is it coming from that specific test? I don't see those errors when I run the tests locally and I can't find them in the github actions logs

merav-aharoni:
> > Any idea what is causing this?
> 
> Is it coming from that specific test? I don't see those errors when I run the tests locally and I can't find them in the github actions logs

I get these messages when I run `test_enable_account_by_qiskitrc` locally. If you aren't getting them, maybe it is a problem in my environment, and then we are okay.

kt474 (approved):

PR #1220, opened 2023-11-16
Fix BYOB job results bytes typing issue 

kt474:
### Summary

Currently in staging, after BYOB was enabled, if a job is not successful, a bytes object is returned from COS. This causes a type error in `_error_msg_from_job_response` 

The same fix will have to be done in `qiskit-ibm-provider` 

### Details and comments
Fixes #


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -485,9 +485,13 @@ def _error_msg_from_job_response(self, response: Dict) -> str:
...
-        index = job_result_raw.rfind("Traceback")
-        if index != -1:
-            job_result_raw = job_result_raw[index:]
+
+        try:
+            index = job_result_raw.rfind("Traceback")
+            if index != -1:
+                job_result_raw = job_result_raw[index:]
+        except TypeError:
+            pass
```
kt474, line 494: @cpkurotori What do you think about this? Should we add some sort of try/except 

Or check the type with something like:

```python
if isinstance(job_result_raw, bytes):
     job_result_raw = job_result_raw.decode('utf-8')
```

cpkurotori: honestly i think it would be better to change `_download_external_result` (https://github.com/Qiskit/qiskit-ibm-runtime/pull/1220/files#diff-f14680ef8724b0c65249d7b84df515bd3db9ce534b509e373580eacc6400609bR169)

```diff
- return result_response.content
+ return result_response.text
```

kt474: Wasn't aware of the difference, thanks for the suggestion

cpkurotori (approved):

PR #1193, opened 2023-11-03
Add channel strategy instance validation

kt474:
### Summary

If `q-ctrl` is passed in as the channel_strategy but the instance does not have it enabled, the following error will be raised: 

```
IBMNotAuthorizedError: 'This account is not authorized to use ``q-ctrl`` as a channel strategy.'
```

### Details and comments
Fixes #1176

PR #1233, opened 2023-11-17
Instance & channel_strategy validation 

kt474:
### Summary

In https://github.com/Qiskit/qiskit-ibm-runtime/pull/1193 we raised an error if the channel_strategy  `q-ctrl` was passed in but the instance used did not support `q-ctrl`. We also want to raise an error if the opposite happens - if the instance is `q-ctrl` enabled but `q-ctrl` is not passed in as the channel_strategy. 

### Details and comments
Fixes #


merav-aharoni (approved):
Looks good!

PR #1243, opened 2023-11-29
Support local execution on simulators

merav-aharoni:
### Summary
When the `backend` is a `FakeBackend` or `AerSimulator`, run the program locally, rather than through the IBM Cloud.

### Details and comments
Fixes #


merav-aharoni:
I wrote the basic flow, including a simple test. However, I encountered quite a few problems/questions:

- [ ] What is the correct way to transfer options? Which options should be transferred?
- [ ] Currently working only when `skip_transpilation=True`
- [ ] Should we support `FakeBackends V1` in addition to `V2`?
- [ ] Still need to add support for `circuit-runner`. Is QASM3 supported on simulators?
- [ ] Should `ibmq_qasm_simulator` also run locally?
- [ ] Does `Session` have any meaning when running on simulator?
- [ ] Fix when `QiskitRuntimeService` is not defined

@kt474 , @jyu00 - any insights on the above are welcome.

kt474:
> What is the correct way to transfer options? Which options should be transferred?
I'm not sure about how options would work in these fake backends, i'll have to take a look. @jyu00 thoughts here? 

>Should we support `FakeBackends V1` in addition to `V2`?

I don't believe so, we can just support V2 for now

>Still need to add support for `circuit-runner`. Is QASM3 supported on simulators?

No, QASM3 is not currently supported on simulators 
>Should `ibmq_qasm_simulator` also run locally?

No, I don't believe so - `ibmq_qasm_simulator` is one of the simulators that will be sunset. 

>Does `Session` have any meaning when running on simulator?

No, sessions don't have any meaning when using simulators - jobs on simulators run the same way, independent of whether 
or not they are in a session
>Must the user have an account to run on a simulator? Currently, we need to initialize a `QiskitRuntimeService`, and this fails when no account.

Yes, the user must have an account

merav-aharoni:
A few comments/questions for the reviewers:

1. I prefer to leave support for `circuit_runner` for a separate PR
2. Do we want `FakeProviderForBackendV2.get_backend()` to throw when backend not found? For me it was more convenient to return None as a check for whether this name is a fake backend.
3. `FakeRuntimeJob` – there are several members that I think are not needed. Do you agree?
4. See questions I wrote inside the code.
5. Ideas for additional tests?

qiskit_ibm_runtime/fake_provider/fake_provider.py
```diff
@@ -16,8 +16,11 @@
 Fake provider class that provides access to fake backends.
 """
 
+from typing import List
+
 from qiskit.providers.provider import ProviderV1
 from qiskit.providers.exceptions import QiskitBackendNotFoundError
+from qiskit.providers.fake_provider.fake_backend import FakeBackendV2
```
ElePT, line 23: I have recently opened a PR that migrates the fake provider base classes from qiskit (https://github.com/Qiskit/qiskit-ibm-runtime/pull/1270), should it be merged before this PR, this import path would change to `from qiskit_ibm_runtime.fake_provider.fake_backend import FakeBackendV2`. I'm happy to monitor it on my side, but thought it would make sense to bring your attention to it too.

qiskit_ibm_runtime/fake_runtime_job.py
```diff
@@ -0,0 +1,80 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Qiskit fake runtime job."""
+
+from typing import Optional, Dict, Any
+
+from qiskit.primitives.primitive_job import PrimitiveJob
+from qiskit.providers import JobStatus, JobV1
+from qiskit.providers.fake_provider import FakeBackendV2 as FakeBackend
```
ElePT, line 19: same comment as above regarding the import path

qiskit_ibm_runtime/options/utils.py
```diff
@@ -12,6 +12,7 @@
 
 """Utility functions for options."""
```
ElePT, line 14: similar comment here too.

qiskit_ibm_runtime/options/utils.py
```diff
@@ -12,6 +12,7 @@
...
 
+from qiskit.providers.fake_provider import fake_backend
```
ElePT, line 15: Here too

qiskit_ibm_runtime/base_primitive.py
```diff
@@ -154,16 +169,30 @@ def _run_primitive(self, primitive_inputs: Dict, user_kwargs: Dict) -> RuntimeJo
...
+            primitive_inputs["optimization_level"] = combined["optimization_level"]
+            runtime_options["backend"] = self._backend
+
+            return self._service.run(
+                program_id=self._program_id(),
+                options=runtime_options,
+                inputs=primitive_inputs,
+                # are the following relevant for simulators?
+                # callback=combined.get("environment", {}).get("callback", None),
+                # result_decoder=DEFAULT_DECODERS.get(self._program_id()),
```
kt474, line 194: I think we can leave these out for simulators

qiskit_ibm_runtime/fake_provider/fake_provider.py
```diff
@@ -69,18 +72,19 @@ class FakeProviderForBackendV2(ProviderV1):
...
         if name:
-            filtered_backends = [backend for backend in self._backends if backend.name() == name]
-            if not filtered_backends:
-                raise QiskitBackendNotFoundError()
-
-            backend = filtered_backends[0]
+            filtered_backends = [backend for backend in self.backends() if backend.name == name]
+            # if not filtered_backends:
+            #     raise QiskitBackendNotFoundError()
+            # I prefer returning None rather than throwing an error. Any objection?
```
kt474, line 82: I think returning `None` is fine

qiskit_ibm_runtime/fake_provider/fake_provider.py
```diff
@@ -69,18 +72,19 @@ class FakeProviderForBackendV2(ProviderV1):
     available in the :mod:`qiskit_ibm_runtime.fake_provider`.
     """
 
-    def get_backend(self, name=None, **kwargs):  # type: ignore
-        backend = self._backends[0]
+    def get_backend(self, name=None, **kwargs) -> FakeBackendV2:  # type: ignore
+        # backend = self._backends[0]
+        # when name=None, this will simply return Almaden. Is that what we want?
```
kt474, line 77: Do we need to return a default backend here? I think users should just pass in the fake backend they want here

kt474: Also, this will be renamed to just `backend` in https://github.com/Qiskit/qiskit-ibm-runtime/pull/1270

qiskit_ibm_runtime/base_primitive.py
kt474: In terms of the other options, I think it'd be nice to have the run options like:

`{'shots': 400, 'init_qubits': True, 'noise_model': None, 'seed_simulator': None}`
and 
`{'max_execution_time': None, 'log_level': 'WARNING', 'job_tags': []}`

qiskit_ibm_runtime/base_primitive.py
```diff
@@ -98,6 +101,13 @@ def __init__(
         if isinstance(backend, IBMBackend):
             self._service = backend.service
             self._backend = backend
+        elif isinstance(backend, AerSimulator):
```
jyu00, line 104: This should check for both Aer and FakeBackend

qiskit_ibm_runtime/base_primitive.py
```diff
@@ -20,10 +20,13 @@
 from dataclasses import asdict
 import warnings
 
+from qiskit_aer import AerSimulator
```
jyu00, line 23: Aer is not a required package, so this import be wrapped with try/except

qiskit_ibm_runtime/base_primitive.py
```diff
@@ -154,16 +169,30 @@ def _run_primitive(self, primitive_inputs: Dict, user_kwargs: Dict) -> RuntimeJo
 
         combined = Options._set_default_resilience_options(combined)
         combined = Options._remove_none_values(combined)
-
         primitive_inputs.update(Options._get_program_inputs(combined))
 
-        if self._backend and combined["transpilation"]["skip_transpilation"]:
+        if self._backend and combined["transpilation"]["skip_transpilation"] and not run_simulator:
```
jyu00, line 174: `check_faulty` should apply to fake backends as well, if we ever get one that mimics a real backend with faulty edge/qubits

qiskit_ibm_runtime/fake_runtime_job.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def result(self) -> Any:
+        """Return the results of the job."""
+        return self._primitive_job.result()
+
+    def cancel(self) -> None:
+        self._primitive_job.cancel()
+
+    def status(self) -> JobStatus:
+        return self._primitive_job.status()
```
jyu00, line 57: Why not just inherit `PrimitiveJob`, then you only need to add methods not already supported.

qiskit_ibm_runtime/options/utils.py
```diff
@@ -32,20 +33,17 @@ def set_default_error_levels(
     Returns:
         options with correct error level defaults.
     """
+    is_simulator = (
+        isinstance(backend, fake_backend.FakeBackendV2) or backend.configuration().simulator
+    )
```
jyu00, line 38: There is no need to change optimization_level and resilience_level for fake backends. Fake backends should always have noise model (from the real backend it mimics), so one doesn't have to pass it in. And none supports resilience_level.

qiskit_ibm_runtime/base_primitive.py
```diff
@@ -98,6 +101,13 @@ def __init__(
         if isinstance(backend, IBMBackend):
             self._service = backend.service
             self._backend = backend
+        elif isinstance(backend, AerSimulator):
+            self._backend = backend
+            self._service = (
+                QiskitRuntimeService()
+                if QiskitRuntimeService.global_service is None
+                else QiskitRuntimeService.global_service
+            )
```
jyu00, line 110: Instead of adding all the local logic into `QiskitRuntimeService`, it'd be much cleaner to just create a new `LocalRuntimeService` to "mock" any server requests.

jyu00 (changes requested):

qiskit_ibm_runtime/fake_runtime_job.py
```diff
@@ -0,0 +1,80 @@
...
+        # service: "qiskit_runtime_service.QiskitRuntimeService",
+        params: Optional[Dict] = None,
+        # creation_date: Optional[str] = None,
+        # user_callback: Optional[Callable] = None,
+        # result_decoder: Optional[Union[Type[ResultDecoder], Sequence[Type[ResultDecoder]]]] = None,
+        # tags: Optional[List] = None,
+    ) -> None:
+        """FakeRuntimeJob constructor."""
+        super().__init__(backend=backend, job_id=job_id)
+        # self._service = service  # do we need this?
```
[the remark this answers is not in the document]
merav-aharoni: I added `creation_date` and `tags`. 
Removed `user_callback` as you wrote below and also because it refers to interim results. 
Removed `result_decoder` as you wrote below, because results will not be returned encoded.

qiskit_ibm_runtime/base_primitive.py
merav-aharoni: - `shots` and `seed_simulator` are already supported.
- I couldn't fine `init_qubits` or anything similar in the `AerSimulator`, so I don't think there is any point of passing it on. Let me know if you know otherwise. I did find it as a parameter of `assemble_circuits` in Terra, but I am not sure how and if this is called.
- Same as above for `max_execution_time`.
- `noise_model` - I think this is only relevant when calling `AerSimulator` and not for fake backends, because the latter define their own noise models. For Aer, I pass along all the options that are `AerSimulator._default_options()`.
- Regarding `log_level`, in `Terra` and in `Aer`, I only found setting the `log_level` globally as an environment variable, so I am not sure what we can do here.
- `job_tags` - I added to `FakeRuntimeJob`.

qiskit_ibm_runtime/fake_runtime_job.py
merav-aharoni: I tried, but couldn't manage to do that, because I am receiving `primitive_job` as an object and couldn't find a way to construct the `FakeRuntimeObject` with a pre-constructed `super()` object. I would need something like a copy constructor.
I could simply use `PrimitiveJob` directly, but then I wouldn't be able to add any fields/methods.
Can you advise how to do this with inheritance?

qiskit_ibm_runtime/base_primitive.py
merav-aharoni: Added `try` and raise exception when attempting to run with a simulator and not aer.
Also added `skip` for relevant tests.
Please have a look if it is done properly.

qiskit_ibm_runtime/base_primitive.py
merav-aharoni: There is currently no `check_faulty` method in `FakeBackendV2` nor in `AerSimulator`. To support this, we would need to add such a method to `FakeBackendV2` after it is transferred to Runtime. So I suggest we defer this to a separate issue.

qiskit_ibm_runtime/options/utils.py
merav-aharoni: Changed so we don't apply this to a FakeBackend.

qiskit_ibm_runtime/base_primitive.py
merav-aharoni: I am not sure I understand your intention in this comment. Do you mean that when running on a simulator, we don't need a real service, and don't need to have a real account?

jyu00: Yes, so instead of modifying `QiskitRuntimeService`, `base_primitive` can set `self._service` to a mock (local) service. And its `run()` method would handle the local operation.

merav-aharoni: I will try this. Are you also saying that the local run will not need an account? because in a comment above, @kt474 thought it should have an account.

kt474: We can do it this way, without an account

merav-aharoni: I didn't get a chance to complete this item. I think it might be best to do in a separate PR. In any case, it is up to you @kt474 or @jyu00 to continue from here.

kt474:
Closing, this was done in #1495

PR #1245, opened 2023-11-29
Add options for summit PF

chriseclectic:
### Summary

Add necessary options for summit practitioners forums demo and tutorial notebooks

### Details and comments


chriseclectic:
Im not sure what to do about test failures here, they seem to have nothing to do with this PR so I suspect are failing on the branch already.

PR #1249, opened 2023-11-30
Raise error if backend retrieved not in current instance 

### LINKED ISSUE #1246: backend doesn't grab instance from service
**Describe the bug**

It is possible for the `instance` attached to the service to not get passed to the `backend` object. Only occurs when a user has access to a backend via multiple instances.

**Steps to reproduce**

```python
service = QiskitRuntimeService(channel="ibm_quantum", instance="hgp1")
backend = service.get_backend("ibm_algiers")
# ... returns a non-matching hgp
print(backend._instance)
```

**Suggested solutions**

I suspect that somewhere [around here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/qiskit_runtime_service.py#L588) there are a few instances of `instance` that should be `instance_filter`.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.16.1
- **Python version**: 3.11
- **Operating system**: macOS

kt474:
### Summary

In the following scenario where an instance is passed in at initialization and a backend that is not in that instance is retrieved (but the user has access to in a different instance), we should raise an error instead of silently switching to a different instance. 

```python
service = QiskitRuntimeService(channel="ibm_quantum", instance="hgp1")
backend = service.get_backend("ibm_algiers") 
# algiers is not in hgp1 but in a different instance the user has access to 
```

We should also return a more informative error message - for example:

```python
QiskitBackendNotFoundError: "Backend ibm_algiers is not in {hgp}. 
Please try a different instance. ibm_algiers is in the following instances you have access to: 
['hgp', 'hgp', 'hgp']"
```

### Details and comments
Fixes #1246


jyu00:
Does this really fix #1246? My interpretation of #1246 is that the retrieve backend doesn't have the right h/g/p, when it's accessible via multiple instances.

kt474:
> Does this really fix #1246? My interpretation of #1246 is that the retrieve backend doesn't have the right h/g/p, when it's accessible via multiple instances.

Discussed with @blakejohnson on slack - in his scenario, the backend (ibm_algiers) was not actually in hgp1 so the backend was retrieved from a different instance. This was confusing because the provider just silently chooses the different instance that has access to the backend. When this happens, I think we should raise an error. 

If an instance is passed in at initialization, retrieving a backend will always use that instance if the backend is actually in it. 

So yes, this doesn't actually fix #1246 because it's not actually a bug - this adds a helpful error message.

kt474:
> I am a bit confused about what this fix will do. If the user has several hgp accounts, then I think that for each of them that does not have the specified backend, an error will be issued. But perhaps I am mistaken. Perhaps a test would be helpful to demonstrate this. Also, the phrasing of the release note is cumbersome.

This behavior already existed - if a user did something like this, we raise an error since even though the user has access to `ibm_torino`, it is not in `ibm-q/open/main`. 

```python
service.backend('ibm_torino', instance='ibm-q/open/main')
```
This PR makes it so the same error will be raised if an instance is passed in at initialization:

```python
service = QiskitRuntimeService(channel="ibm_quantum", instance="ibm-q/open/main")
service.backend('ibm_torino') 
```

I'll try to add a test and update the release note to make it more clear

test/integration/test_backend.py
```diff
@@ -45,6 +45,25 @@ def test_backends(self, service):
...
+
+        hgp_1 = hgps[0]
+        hgp_2 = hgps[1]
+        hgp_1_backends = service.backends(instance=hgp_1)
+        hgp_2_backends = service.backends(instance=hgp_2)
+        unique_hgp = list(
+            set(hgp_2_backends) - set(hgp_1_backends)
+        )  # get differences between the two lists
+        if unique_hgp:
+            unqiue_backned = unique_hgp[0]
```
merav-aharoni, line 63:
```suggestion
            unique_backend = unique_hgp[0]
```

test/integration/test_backend.py
```diff
@@ -45,6 +45,25 @@ def test_backends(self, service):
...
+            with self.assertRaises(QiskitBackendNotFoundError):
+                service.backend(unqiue_backned.name, instance=hgp_1)
```
merav-aharoni, line 65:
```suggestion
                service.backend(unique_backend.name, instance=hgp_1)
```

merav-aharoni (approved):
LGTM

PR #1256, opened 2023-12-04
Remove non-API docs and translations

Eric-Arellano:
This is the same as https://github.com/Qiskit/qiskit/pull/11352. Now that qiskit.org/ecosystem/ibm-runtime redirects to docs.quantum.ibm.com, we can simplify the docs. 

See the PR description in https://github.com/Qiskit/qiskit/pull/11352 for details on the motivation for each change. The only difference from Qiskit is that we still keep the tutorials around because learning.quantum.ibm.com uses them.

I used a script with `ripgrep` to confirm that none of the images were in use.


kt474 (approved):
LGTM 🎉

PR #1270, opened 2023-12-08
Migrate fake provider base classes from Qiskit

ElePT:
### Summary
This PR migrates the fake backend base classes from Qiskit:

- `FakeBackend`
- `FakeBackendV2`
- `FakeQasmBackend`
- `FakePulseBackend`

and updates the backend snapshot to inherit from the new `qiskit-ibm-runtime`-based classes.
The motivations behind this move are:

- decoupling the fake provider in `qiskit_ibm_runtime` from the deprecations taking place in the `qiskit` fake provider
- setting the stage for a future change of base classes of the fake backends in `qiskit_ibm_runtime`, as indicated in https://github.com/Qiskit/qiskit/issues/9553

### Details and comments

- This PR performs a history-preserving copy of the files. To preserve the history, the commits shouldn't be squashed before merging. If preserving the history is not a priority then they can just be squashed as it's usually done.
- This PR additionally changes the V2 fake provider's `get_backend` method to `backend` as indicated in #85, but I can revert it if you prefer it to be in a separate PR.
- I am not sure how much of this change should be documented in a release note, in principle users should not be interacting with these classes directly, the initial idea was to even make them private.

### DIFF
--- a/qiskit_ibm_runtime/fake_provider/backends/almaden/fake_almaden.py
+++ b/qiskit_ibm_runtime/fake_provider/backends/almaden/fake_almaden.py
@@ -15,7 +15,7 @@ Fake Almaden device (20 qubit).
 """
 
 import os
-from qiskit.providers.fake_provider import fake_qasm_backend, fake_backend
+from qiskit_ibm_runtime.fake_provider import fake_qasm_backend, fake_backend
 
 
 class FakeAlmadenV2(fake_backend.FakeBackendV2):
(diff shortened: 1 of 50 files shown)


kt474 (approved):

ElePT (commented):
It looks like the CLA signature is not working for some of the transferred commits, let me know if you'd still be able to merge the PR. Else, I guess we could not do the history-preserving copy.

kt474:
> It looks like the CLA signature is not working for some of the transferred commits, let me know if you'd still be able to merge the PR. Else, I guess we could not do the history-preserving copy.

Is preserving the git history important here? It's not currently enabled in this repo (I can enable it) but I would still prefer to squash and merge.

ElePT:
I always try to preserve git history, but if the PR is going to be squashed anyway, the origin of the files is traceable, and preserving history becomes problematic (with unsigned CLAs), I don't think it's a big issue not to do it. I just pushed the diff as a single commit.

PR #1273, opened 2023-12-11
Deprecated QiskitRuntimeService.get_backend()

### LINKED ISSUE #1261: Replace get_backend with backend in the README
The method `get_backend` has been renamed to `backend`, but the README was not updated.

Grepping `get_backend`, it still appears in additional places in the code (fake provider and integration tests); these possibly should be updated too.

merav-aharoni:
### Summary
Users should use ``QiskitRuntimeService.backend()`` instead.

### Details and comments
Fixes #1261

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ bell.cx(0, 1)
 bell.measure_all()
 
 # 3. Execute using the Sampler primitive
-backend = service.get_backend('ibmq_qasm_simulator')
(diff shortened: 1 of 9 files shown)


kt474:
I think we can close this for now, I'll take care of it when we stop inheriting from `Provider` 

https://github.com/Qiskit/qiskit-ibm-runtime/issues/1261#issuecomment-1861621220

PR #1280, opened 2023-12-12
Unite `Session` from `qiskit_ibm_provider` with `Session` in this repository

merav-aharoni:
### Summary
`IBMBackend.run()` now uses the same `Session` as used by the primitives.

### Details and comments

I made changes in the imports because of cyclic imports.
`IBMBackend` can receive a Session in 3 ways:

1. As a context manager (as for primitives)
2. As a parameter to `IBMBackend.run()`. This makes more sense than setting it in `__init__`, because often the user will define the backend as `backend = service.get_backend()`, so the backend will keep the same `Session` object even if the session is closed, causing failures. Even if we call `backend = service.get_backend()` again, backend will still be the same object as before.
3. Using `IBMBackend.open_session()`. I believe this should eventually be deprecated.


qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -752,13 +753,13 @@ def _runtime_run(
...
 
-        # Check if initialized within a Primitive session. If so, issue a warning.
-        if get_cm_primitive_session():
-            warnings.warn(
-                "A Primitive session is open but Backend.run() jobs will not be run within this session"
-            )
+        if session:
+            self._session = session
+        elif get_cm_session():
+            self._session = get_cm_session()
```
kt474, line 759:
```python
backend = self.service.get_backend("ibmq_qasm_simulator")
with Session(backend=backend) as session:
    job1 = sampler.run(circuits=ReferenceCircuits.bell())
    job2 = backend.run(circuits=ReferenceCircuits.bell())
```
With this approach, we wouldn't even have to pass in `session` to `backend.run()` right? Something like this would work fine too

test/integration/test_session.py
```diff
@@ -129,38 +128,66 @@ def test_backend_run_with_session(self):
...
+            job2 = sampler.run(circuits=ReferenceCircuits.bell())
+            session_id = session.session_id
+            self.assertEqual(session_id, job1.job_id())
+            self.assertEqual(job2.session_id, session_id)
+
+    def test_backend_open_session_with_context_manager(self):
+        """Test Sampler.run and backend.run in the same session.
+        Session is created by open_session. Sampler gets
+        Session from  backend without parameter"""
+        backend = self.service.get_backend("ibmq_qasm_simulator")
```
kt474, line 169: I think we should add another test for this scenario:

```python
backend = self.service.get_backend("ibmq_qasm_simulator")
with Session(backend=backend) as session:
    job1 = sampler.run(circuits=ReferenceCircuits.bell())
    job2 = backend.run(circuits=ReferenceCircuits.bell())

job3 = backend.run(...)
```

This currently wouldn't work because the session is still tied to the backend, and it's closed

qiskit_ibm_runtime/ibm_backend.py
merav-aharoni: Yes, this should work.

test/integration/test_session.py
merav-aharoni: I added a test for this case. I hope this is what you meant. You will have to take over from here...

kt474:
With the upcoming deprecation of backend.run() I will close this for now

PR #1281, opened 2023-12-12
Replace programs with primitives in api docs 

kt474:
### Details and comments
Fixes #


drew-distefano (commented):
@jyu00 would it make sense for this to read as:

Some runtime primitives provide interim results that inform you about **primitive** progress

beckykd (approved):

jyu00:
> @jyu00 would it make sense for this to read as:
> 
> Some runtime primitives provide interim results that inform you about **primitive** progress

How about

`Some runtime primitives provide interim results that inform you about the progress of your job.`

PR #1293, opened 2023-12-28
Update grover tutorial equation

### LINKED ISSUE #1222: Sub-optimal number of iterations in Grover's algorithm tutorial
In the tutorial [Grover's algorithm using the Sampler primitive](https://learning.quantum-computing.ibm.com/tutorial/grovers-algorithm-using-the-sampler-primitive) the number of iterations claimed to be optimal for Grover's algorithm is this number:

$$
\left\lfloor \frac{\pi}{4} \sqrt{\frac{N}{s}} \right\rfloor
$$

where $N = 2^n$ and $s$ is the number of marked items.

Screen Shot 2023-11-16 at 11 56 12 AM

This number is usually optimal but not always — the number should be this:

$$
\left\lfloor \frac{\pi}{4 \sin^{-1}\left(\sqrt{s/N}\right)} \right\rfloor
$$

For example, if there are $n=7$ bits (so $N=2^7 = 128$) and there are $s = 19$ marked items, then the two expressions give a number of iterations that differ by one, and the second number of iterations (which is optimal) succeeds with a probability slightly larger than the first: about 85.9% versus 84.3%.

**Suggested solutions**

Use the second formula.

kt474:
### Details and comments
Fixes #1222

### DIFF
--- a/docs/tutorials/grover_with_sampler.ipynb
+++ b/docs/tutorials/grover_with_sampler.ipynb
@@ -212,7 +212,7 @@
    "outputs": [],
    "source": [
     "optimal_num_iterations = math.floor(\n",
-    "    math.pi / 4 * math.sqrt(2**grover_op.num_qubits / len(marked_states))\n",
+    "    math.pi / (4 * math.asin(math.sqrt(len(marked_states) / 2**grover_op.num_qubits)))\n",
     ")"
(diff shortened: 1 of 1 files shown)


kt474:
cc: @jhwatrous

PR #1301, opened 2024-01-02
switch docs from mathjax to katex

kevinsung:
### Summary

Updates the sphinx documentation build to use KaTeX instead of MathJax because the final docs build uses KaTeX. The same change was made in Qiskit in https://github.com/Qiskit/qiskit/pull/11435.


Eric-Arellano (approved):
Thank you!

PR #1303, opened 2024-01-03
Add PyPi version checker

nonhermitian:
### Summary

Requires: https://github.com/Qiskit/qiskit-ibm-provider/pull/781

Often often runs into incompatibility issues with the Runtime because something has changed without much warning.  Time and time again, it is the case that a patch has been released that fixes the issue, but the user has no idea that a newer version is available.  This fixes this by adding a version checker that looks for an update on PyPi and gives the users a warning if a newer version is found.

### Details and comments
The mechanism for doing this is generic, and technically works for any package. There should be no issues with doing this as both Provider and Runtime need internet access to function.

As an extension, this can be made to be disabled with a config var or made to compare only major, minor, or micro version differences.

### DIFF
--- a/qiskit_ibm_runtime/__init__.py
+++ b/qiskit_ibm_runtime/__init__.py
@@ -162,6 +162,8 @@ Classes
 
 import logging
 
+from qiskit_ibm_provider.utils.version_check import pypi_version_check, update_warning
+
 from .qiskit_runtime_service import QiskitRuntimeService
 from .ibm_backend import IBMBackend
 from .runtime_job import RuntimeJob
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/__init__.py
```diff
@@ -162,6 +162,8 @@ def result_callback(job_id, result):
 
 import logging
 
+from qiskit_ibm_provider.utils.version_check import pypi_version_check, update_warning
+
```
kt474, line 166: Since there are plans to deprecate `qiskit-ibm-provider`, I think this file should live in this repo

kt474:
Going to close this for now, if you think this is still necessary, we can add the code from https://github.com/Qiskit/qiskit-ibm-provider/pull/781 into this repo since `qiskit-ibm-provider` is deprecated.

PR #1311, opened 2024-01-10
Update from_id()

### LINKED ISSUE #1309: backend data missing when using Session.from_id()
**Describe the bug**

If I do

```
session = Session.from_id(id, sevice)
backend = session.backend()
```

I'd expect the backend used for the session to be returned. But it's not even though this information can be query from the server. 

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

Get backend information from the server is not already known. 

**Additional Information**

- **Python version**:
- **Operating system**:

### LINKED ISSUE #1369: Verify execution mode in from_id()

Today `Batch` and `Session` both use the same `from_id()` method, and it doesn't verify if the execution mode matches the class. For example I can do 

```
session = Session(...)
batch = Batch.from_id(session.session_id)
```

while this is technically supported by the server and a user error, it'd be nice to let the user know especially since we (will) already query session details upon initialization.

Note that this needs to wait for execution mode being returned by the server. 

**Acceptance criteria**

An error is raise if one tries to use `Batch.from_id()` to initialize a session, and vice versa.

kt474:
### Details and comments
Fixes #1309
Fixes #1369


qiskit_ibm_runtime/session.py
```diff
@@ -301,15 +302,36 @@ def from_id(
...
+         Raises:
+            IBMInputValueError: If given `session_id` does not exist. or the backend passed in does
+                not match the original session backend.
 
         Returns:
             A new Session with the given ``session_id``
 
         """
         if backend:
             deprecate_arguments("backend", "0.15.0", "Sessions do not support multiple backends.")
```
jyu00, line 317: Not part of this PR, but I found the `Sessions do not support multiple backends` message very confusing, since I can be passing in the correct backend. The message should be something more like `The backend used to open the session will be used`.

qiskit_ibm_runtime/session.py
```diff
@@ -301,15 +302,36 @@ def from_id(
...
+            response = service._api_client.session_details(session_id)
+            if response:
+                session_backend = response.get("backend_name")
+                if backend and backend != session_backend:
+                    raise IBMInputValueError(
+                        f"The session_id {session_id} was created with backend {session_backend}, "
+                        f"but backend {backend} was given."
+                    )
+                backend = session_backend
+            else:
```
jyu00, line 331: Does the server just return an empty response if session is not found? I'd think it'd raise an 404 error.

qiskit_ibm_runtime/session.py
```diff
@@ -301,15 +302,36 @@ def from_id(
...
+                raise IBMInputValueError(f"The session_id {session_id} does not exist.")
+        if not service and not backend:
+            raise IBMInputValueError("Either service or backend must be given.")
```
jyu00, line 334: Since `backend` is deprecated, we should just make `service` a required keyword.

jyu00 (changes requested):

qiskit_ibm_runtime/session.py
kt474: I'll update the message

qiskit_ibm_runtime/session.py
kt474: When `session.details` was first implemented, we decided to use a `try/except` because of the differences between the US and DE environments 

https://github.com/kt474/qiskit-ibm-runtime/blob/8723ccf827fd7b43bf653d76e9c254ad00f9be4b/qiskit_ibm_runtime/api/rest/runtime_session.py#L57-L63

So `None` will be returned if the session is not found

qiskit_ibm_runtime/session.py
kt474: I agree that `service` should be a required argument, but is there some sort of deprecation policy when it comes to changing parameters from optional to required? Or can we just put in the error message that `service` is now required?

qiskit_ibm_runtime/session.py
mriedem: DE is dead so does that change any minds on how to work this now?

kt474: Yeah that logic should be updated, we can also remove the check [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/qiskit_runtime_service.py#L220)

qiskit_ibm_runtime/session.py
```diff
@@ -314,15 +318,47 @@ def from_id(
...
             A new Session with the given ``session_id``
 
         """
         if backend:
-            deprecate_arguments("backend", "0.15.0", "Sessions do not support multiple backends.")
+            deprecate_arguments(
+                "backend", "0.15.0", "The backend used to open the session will be used."
+            )
+            if isinstance(backend, IBMBackend):
+                backend = backend.name
```
jyu00, line 337: 0.15 was released >4 months ago, so we can remove this now.

qiskit_ibm_runtime/session.py
```diff
@@ -314,15 +318,47 @@ def from_id(
...
+
+        if service:
+            response = service._api_client.session_details(session_id)
+            if response:
+                session_backend = response.get("backend_name")
+                if backend and backend != session_backend:
+                    raise IBMInputValueError(
+                        f"The session_id {session_id} was created with backend {session_backend}, "
+                        f"but backend {backend} was given."
+                    )
```
jyu00, line 347: We don't need this anymore if `backend` is not accepted.

qiskit_ibm_runtime/session.py
jyu00: But this statement `if not service and not backend:` is also a breaking change since the old code didn't say anything about service being required. 

If service is not given, we should give a deprecation warning like you have, and just initialize one (i.e. `QiskitRuntimeService()`) since that's what `Session()` does if you don't pass a service. Then we can use that to query session details. This now should also check for `mode`.

kt474: If service is not given and we initialize a new service, `QiskitRuntimeService()`, it appears that it does not have access to previous sessions. Querying session details results in a session not found error.

qiskit_ibm_runtime/session.py
```diff
@@ -332,23 +335,40 @@ def from_id(
...
         session_id: str,
         service: Optional[QiskitRuntimeService] = None,
-        backend: Optional[Union[str, IBMBackend]] = None,
     ) -> "Session":
         """Construct a Session object with a given session_id
 
         Args:
             session_id: the id of the session to be created. This must be an already
                 existing session id.
             service: instance of the ``QiskitRuntimeService`` class.
```
jyu00, line 344:
```suggestion
            service: instance of the ``QiskitRuntimeService`` class. If ``None``, ``QiskitRuntimeService()`` is used to initialize your default saved account.
```

qiskit_ibm_runtime/session.py
```diff
@@ -332,23 +335,40 @@ def from_id(
...
-            backend: instance of :class:`qiskit_ibm_runtime.IBMBackend` class or
-                string name of backend.
+
+         Raises:
+            IBMInputValueError: If given `session_id` does not exist. or the backend passed in does
+                not match the original session backend.
```
jyu00, line 348:
```suggestion
            IBMInputValueError: If given `session_id` does not exist.
```

qiskit_ibm_runtime/session.py
```diff
@@ -332,23 +335,40 @@ def from_id(
...
+                stacklevel=2,
+            )
+            service = QiskitRuntimeService()
+
+        response = service._api_client.session_details(session_id)
+        backend = response.get("backend_name")
+        mode = response.get("mode")
+        if mode != cls.__name__.lower():
+            raise IBMInputValueError(
+                f"Session ID {session_id} has session mode {mode} instead of {cls.__name__}."
```
jyu00, line 370:
```suggestion
                f"Input ID {session_id} has execution mode {mode} instead of {cls.__name__}."
```

jyu00 (approved):
Some comments on docstring but otherwise looks good.

PR #1292, opened 2023-12-27
Remove `qiskit.test`

kt474:
### Details and comments
Fixes #1244

PR #1321, opened 2024-01-17
EstimatorV2 updates

### LINKED ISSUE #1314: Primitive interface updates for EstimatorV2
Part of https://github.com/Qiskit/qiskit-ibm-runtime/issues/1131 

Support EstimatorV2 input/output:
- [Input RFC](https://github.com/Qiskit/RFCs/pull/51) [Output RFC](https://github.com/Qiskit/RFCs/pull/53) 
- Input would be one or more PUBs (Primitive Unified Bloc). Each is a tuple of (circuit, observables, parameter values). 
- Output would be PubResults

Based on the following Qiskit PRs [Estimator 11227](https://github.com/Qiskit/qiskit/pull/11227):
- EstimatorV2 Container https://github.com/Qiskit/qiskit/pull/11524
- BaseEstimator V2 https://github.com/Qiskit/qiskit/pull/11527

Some of the work has already been done in https://github.com/Qiskit/qiskit-ibm-runtime/pull/1224

kt474:
### Summary

~~There are a lot of files copied over from [qiskit](https://github.com/Qiskit/qiskit/tree/main/qiskit/primitives), they should be replaced after the next qiskit release.~~ 

Simple job example: 

Screenshot 2024-01-22 at 12 12 05 PM

### Details and comments
Fixes #1314


ihincks:
BaseEstimatorV2 (and all supportive container classes) has now been merged into qiskit's main branch

qiskit_ibm_runtime/qiskit/primitives/pub_result.py
```diff
@@ -20,7 +21,7 @@
 
 
 @dataclass(frozen=True)
-class PubsResult:
+class PubResult:
```
kt474, line 24: I think this was a typo in #1247? It should be PubResult

kt474:
@jyu00 @gadial 
Should be ready for initial review, I left a couple questions in about things I wasn't sure about

qiskit_ibm_runtime/utils/json.py
```diff
@@ -254,8 +253,14 @@ def default(self, obj: Any) -> Any:  # pylint: disable=arguments-differ
                 serializer=lambda buff, data: dump(data, buff),  # type: ignore[no-untyped-call]
             )
             return {"__type__": "Instruction", "__value__": value}
-        if isinstance(obj, BasePub):
-            return asdict(obj)
+        # TODO proper way to do this?
```
gadial, line 256: I suggest we treat this similarly to other complex data types, e.g. pass a the following dict:

```
        if isinstance(obj, EstimatorPub):
            return {
                "__type__": "EstimatorPub",
                "__value__": {
                    "circuit": obj.circuit,
                    "observables": obj.observables,
                    "parameter_values": obj.parameter_values,
                    "precision": obj.precision,
                },
            }
```

However, I think we should delay that until we start using the `EstimatorPub` from qiskit itself.

test/integration/test_estimator_v2.py
```diff
@@ -0,0 +1,60 @@
...
+from qiskit_ibm_runtime.qiskit.primitives.containers.primitive_result import PrimitiveResult
+from qiskit_ibm_runtime.qiskit.primitives.containers.pub_result import PubResult
+from qiskit_ibm_runtime.qiskit.primitives.containers.data_bin import DataBin
+
+from qiskit_ibm_runtime import EstimatorV2, Session
+from ..decorators import run_integration_test
+from ..ibm_test_case import IBMIntegrationTestCase
+
+
+class TestEstimatorV2(IBMIntegrationTestCase):
```
gadial, line 28: The test suite is currently relatively small; perhaps we can use the test suite in qiskit:

https://github.com/Qiskit/qiskit/blob/main/test/python/primitives/test_estimator.py

See the SamplerV2 PR where I do something similar for Sampler:

https://github.com/gadial/qiskit-ibm-runtime/blob/8102831d23fba61fbe294770accbe093f32929a6/test/integration/test_sampler_v2.py

gadial (approved):
Generally LGTM. I believe we should have more tests, and my main concern is the large amount of copy-pasting of container data from qiskit, even if we believe it's not subject to change. However, if we have no way around it I have no objections to merging this PR.

jyu00:
Now that BaseEstimatorV2 and container classes are all merged into Qiskit main, I think it'd make more sense to just include qiskit main (then rc1 when it's released) in this branch

kt474:
> Now that BaseEstimatorV2 and container classes are all merged into Qiskit main, I think it'd make more sense to just include qiskit main (then rc1 when it's released) in this branch

Removed all of the copied files and using the main branch of qiskit now. I believe the build is failing because the CI is not using the main branch and the qiskit version is being overwritten by another dependency.

kt474:
Lint is now failing because qiskit.test was removed from qiskit 1.0 (Have a separate PR to take care of that https://github.com/Qiskit/qiskit-ibm-runtime/pull/1292, too many moving parts 🫠) 

Ran unit tests locally - tests are passing except `test_parameters_vals_kwvals` (which I think is from a bug in `BindingsArray.coerce()`

jyu00:
This can be a separate PR, but v2 primitives `run()` method now returns an instance of `BasePrimitiveJob`, which does _not_ inherit from `JobV1`. `RuntimeJob` needs to be updated to inherit this (plus `JobV1` to support `backend.run` still, if the same class is used).

jyu00:
@kt474 Now that the 2 blockers have been merged, can you update this PR so we can get it merged? We'll need more tests, but that can go into a separate PR.

PR #1326, opened 2024-01-21
[WIP] SamplerV2

### LINKED ISSUE #1313: Primitive interface updates for SamplerV2
Part of https://github.com/Qiskit/qiskit-ibm-runtime/issues/1131 

Support SamplerV2 input/output:
- [RFC](https://github.com/Qiskit/RFCs/pull/56)
- Input/output would also be PUB based
- Output is per shot measurement, with convenience functions to returns counts/probabilities

Based on the following Qiskit PRs https://github.com/Qiskit/qiskit/pull/11566 (replaced https://github.com/Qiskit/qiskit/pull/11264)
- SamplerV2 Container https://github.com/Qiskit/qiskit/pull/11542
- BaseSamplerV2 https://github.com/Qiskit/qiskit/pull/11529

Work should be done off the experimental [branch](https://github.com/Qiskit/qiskit-ibm-runtime/tree/experimental-0.2) where some of the work is already done (https://github.com/Qiskit/qiskit-ibm-runtime/pull/1224)

gadial:
### Summary
Fixes #1313


qiskit_ibm_runtime/utils/json.py
```diff
@@ -68,7 +69,7 @@
 )
 
 # TODO: Remove when they are in terra
-from ..qiskit.primitives import ObservablesArray, BindingsArray
+from ..qiskit.primitives import ObservablesArray
```
kt474, line 72: I think we can import `ObservablesArray` from qiskit too

kt474 (commented):
Mostly looks good to me, I believe we're just waiting on pec-runtime to finalize the sampler result format

qiskit_ibm_runtime/sampler.py
```diff
@@ -90,11 +91,36 @@ def __init__(
         Sampler.__init__(self)
         BasePrimitiveV2.__init__(self, backend=backend, session=session, options=options)
 
-        raise NotImplementedError("SamplerV2 is not currently supported.")
-
         # if self._service._channel_strategy == "q-ctrl":
         #     raise NotImplementedError("SamplerV2 is not supported with q-ctrl channel strategy.")
 
+    def run(self, pubs: SamplerPubLike | Iterable[SamplerPubLike]) -> RuntimeJob:
```
ihincks, line 97: The parent signature is 
```python
def run(
        self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None
    ) -> BasePrimitiveJob[PrimitiveResult[PubResult]]:
```

So this signature is extending it to allow `SamplerPubLike` in addition to `Iterable[SamplerPubLike]`. It might be confusing to have this extension.

But more importantly, we need to handle the shots argument. The order of precedence is 1) pub.shots if not None, 2) run shots kwarg if not None, 3) whatever behaviour the sampler chooses if both (1) and (2) are None. I'll leave another code suggestion in the body of this function.

```suggestion
    def run(self, pubs: SamplerPubLike | Iterable[SamplerPubLike], *, shots: int | None = None) -> RuntimeJob:
```

qiskit_ibm_runtime/sampler.py
```diff
@@ -90,11 +91,36 @@ def __init__(
...
+        pubs_to_run: List[SamplerPub]
+        if isinstance(pubs, SamplerPub):
+            pubs_to_run = [pubs]
+        elif isinstance(pubs, tuple) and isinstance(pubs[0], QuantumCircuit):
+            pubs_to_run = [SamplerPub.coerce(pubs)]
+        elif pubs is not SamplerPub:
+            pubs_to_run = [SamplerPub.coerce(pub) for pub in pubs]
+
+        for pub in pubs_to_run:
+            pub.validate()
```
ihincks, line 120: I removed the validation block because `coerce` does validation (except when the input is already a SamplerPub, but in this case the user would have had to turn off validation manually because the constructor does it, so that's on them).

Note that we _don`t_ need to pass the `shots` kwarg to `_run` because coersion has resolved the values and placed them into each pub, which takes precedence.

```suggestion
        pubs_to_run: List[SamplerPub]
        if isinstance(pubs, SamplerPub):
            pubs_to_run = [pubs]
        elif isinstance(pubs, tuple) and isinstance(pubs[0], QuantumCircuit):
            pubs_to_run = [SamplerPub.coerce(pubs, shots)]
        elif pubs is not SamplerPub:
            pubs_to_run = [SamplerPub.coerce(pub, shots) for pub in pubs]
```

ihincks (commented):
Thanks @gadial !

qiskit_ibm_runtime/sampler.py
gadial: Thanks. I'll make sure it conforms to the parent signature.

gadial: This was done by Kevin and was now merged to the current version.

gadial:
Replaced by the newer #1420

PR #1328, opened 2024-01-22
Use IBM-Provider-vendored pub/sub mechanism

jakelishman:
### Summary

This is being removed from Qiskit 1.0, and the IBM Provider is (until deprecated) taking over brokering the IBM-Runtime-internal events.

### Details and comments

See Qiskit/qiskit-ibm-provider#799 for the base, though this PR should apply cleanly even before that one, because of the backwards-compatibilty check.

### DIFF
--- a/qiskit_ibm_runtime/ibm_backend.py
+++ b/qiskit_ibm_runtime/ibm_backend.py
@@ -21,7 +21,6 @@ import warnings
 
 from qiskit import QuantumCircuit
 from qiskit.qobj.utils import MeasLevel, MeasReturnType
-from qiskit.tools.events.pubsub import Publisher
 
 from qiskit.providers.backend import BackendV2 as Backend
 from qiskit.providers.options import Options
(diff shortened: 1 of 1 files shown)


kt474:
Since qiskit-ibm-provider is being deprecated soon we'll have to copy the pub sub code into this repo as well

kt474 (approved):

jakelishman:
The way the pub/sub scheme is brokered by a singleton, as long as `qiskit-ibm-runtime` depends on `qiskit-ibm-provider`, it should pull the code from that package to ensure that the data streams are both using the same singleton.  The backwards-compatibility imports I stuck in here means that this will fallback to using Terra's broker if it's still in existence, for the same reason.

Once this package is deprecated, removed and `qiskit-ibm-runtime` no longer depends on it, (or more specifically, as long as it's not supposed to interoperate with _anything_ that `qiskit-ibm-runtime` is doing, include parallel job submission), then yeah, I guess this file can just be shipped over wholesale.

I think that the only place that this is actually _used_ (as a subscriber) is in the Jupyter monitor.  With all available configurations of packages right now, that can detect jobs submitted via either `qiskit-ibm-runtime` or `qiskit-ibm-provider`, but if we didn't have the backwards compatibility shims here, or `qiskit-ibm-runtime` immediately vendored the pub/sub code without the same sort of eager fall-back mechanism I added here, a running Jupyter monitor would only detect new jobs submitted from this package, rather than Runtime.  I strongly suspect that's not a blocker for anybody, but it's more the principle of stability and backwards compatibility of Qiskit!

PR #1329, opened 2024-01-22
Fix pip environment resolution

jakelishman:
### Summary

Calling `pip install` multiple times can (deeply unfortunately) allow an environment to become out-of-sync; `pip` doesn't "remember" the constraints of previous installation commands.

One of the largest effects here is that doing `pip install -e .` _followed_ by `pip install git+<qiskit>@main` installs `qiskit-terra` from the initial installation (via `qiskit==0.45.2`), then installs `qiskit==1.0.0.dev0` from Git, which is an incompatible environment. Doing the whole environment resolution in a single `pip install` command fixes this, as the `qiskit` dependency is correctly resolved to be _only_ the `1.0.0.dev0` version.

### Details and comments

See Qiskit/qiskit-ibm-provider#798 for the same PR over there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e .
-          pip install -c constraints.txt -r requirements-dev.txt
+          pip install -c constraints.txt -r requirements-dev.txt -e .
       - name: Run black
         run: make style
       - name: Run lint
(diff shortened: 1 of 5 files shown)


jakelishman:
Thanks!  For what it's worth, the top comment about "pip not remembering the constraints" isn't actually the problem in the example I gave - the problem in that example is actually that pip doesn't let us specify "conflicts" constraints, and the packaging-structure changes of Qiskit 1.0 make it conflict with Qiskit 0.45 in a way that we can't communicate to pip, or have it automatically resolve unless we poison the `qiskit-terra` package for older installations (also not great).

PR #1337, opened 2024-01-25
Support rep_delay

jyu00:
### Summary

Many users have asked for the ability to specify a `rep_delay`, which is already supported on the server side. This PR officially adds that support. 

### Details and comments
Fixes #


kt474:
Is this something we still want to add? Don't want this PR to get too stale/forgotten

jyu00:
It's already stale. We are deprecating V1 primitive in about a week, so there is really no point making updates.

PR #1353, opened 2024-01-31
Use BasePrimitiveJob

### LINKED ISSUE #1352: Switch to using BasePrimitiveJob for v2 primitives

V2 primitives `run()` method now returns an instance of `BasePrimitiveJob`, which does _not_ inherit from `JobV1`. `RuntimeJob` needs to be updated to inherit this (plus `JobV1` to support `backend.run` still, if the same class is used). 

**Acceptance criteria**

kt474:
### Summary

Can't be merged until Qiskit 1.0 release, because `BasePrimitiveJob` is a new class. Or we can merge this into `experimental-0.2` first. 

### Details and comments
Fixes #1352


jyu00:
Let's just merge into `experimental-0.2`. 

Also we should try to make v2 primitives to return a job that does _not_ inherit from `V1`, so we don't need to go through a deprecation period for methods we want to change (such as `job.status()`, see #1366)

kt474:
Closing this PR, i'll reopen in a new branch

PR #1356, opened 2024-01-31
Fix for unit tests

dieris:
Fix a couple of import on unit tests


kt474 (approved):
Thanks for taking care of this

PR #1363, opened 2024-02-01
Handle Qiskit 1.0 removals

kt474:
### Summary

Removing/replacing `qiskit.opflow`, `qiskit.quantum_info.synthesis`, `qiskit.extensions`, and `qiskit.test._canonical` because they will be removed in Qiskit 1.0 

I think these changes make `qiskit-ibm-runtime` compatible with Qiskit 0.45 and 1.0 

### Details and comments
Fixes #


qiskit_ibm_runtime/transpiler/passes/scheduling/dynamical_decoupling.py
```diff
@@ -23,12 +23,16 @@
...
 from qiskit.transpiler.exceptions import TranspilerError
 from qiskit.transpiler.instruction_durations import InstructionDurations
 from qiskit.transpiler.passes.optimization import Optimize1qGates
 from qiskit.transpiler import CouplingMap
 
+try:
+    from qiskit.quantum_info.synthesis import OneQubitEulerDecomposer
+except ImportError:
+    from qiskit.synthesis import OneQubitEulerDecomposer
+
```
kt474, line 35: Making sure this import works on both Qiskit 0.45 and 1.0 - since `OneQubitEulerDecomposer` is moved in Qiskit 1.0

kt474:
Lint and all integration tests passing with both Qiskit 0.46 and 1.0

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -39,6 +38,11 @@
     decode_pulse_defaults,
 )
 
+try:
+    from qiskit.providers import basicaer
+except ImportError:
+    from qiskit.providers import basic_provider as basicaer
```
wshanks, line 44: This is not enough because there is no QasmSimulatorPy in Qiskit 1.0. You need to use BasicSimulator from basic_provider with Qiskit 1.0 instead of QasmSimulatorPy from basicaer. Maybe @ElePT could tell us if those are totally interchangeable (the deprecation warning says to make the change but the classes don't look identical possibly just due to refactoring).

These are the changes I had to make to get the qiskit-experiments tests to pass with qiskit-ibm-runtime. I think you addressed these changes properly (I was just patching things out that we didn't need but were causing import errors) here and in #1280 and #1365 other than QasmSimulatorPy usage. #1280 or a follow up also needs to remove the qiskit-ibm-provider dependency because just having it installed in the environment causes errors with Qiskit 1.0 because Qiskit 1.0 errors trying to load qiskit-ibm-provider's plugins even if you are not using them.

ElePT, line 44: @wshanks is right, the refactoring affects both the module name (`basicaer `-> `basic_provider`) and simulator name (`QasmSimulatorPy` -> `BasicSimulator`). They are interchangeable with a caveat, that is that `QasmSimulatorPy` is a `BackendV1` subclass and `BasicSimulator` is `BackendV2`, but the basic init/run capabilities are the same. I'd import the simulator class directly to avoid having to handle two changes:

```suggestion
try:
    from qiskit.providers import QasmSimulatorPy as BasicSimulator
except ImportError:
    from qiskit.providers import BasicSimulator
```

ElePT: Thanks for the feedback on the deprecation message, I will take a second look to see how we can improve it to make the migration path clearer. We are working on a 1.0 migration guide, so linking it in the message would probably make sense.

ElePT: And I cannot write an inline comment but to finish my suggestion, on line 135 I'd write: `self.sim = BasicSimulator()` and on line 212, `return basicaer.BasicSimulator._default_options()`, and then update the docstrings (there are a few references to BasicAer that I guess cannot be replaced yet, but I'd say `BasicAer simulator/ BasicSimulator` instead).

ElePT (commented):
The changes look reasonable, I just left a few comments on the `basicaer -> basic_provider` transition.

ElePT:
Have you considered adding a CI run that tests the 1.0 integration? I feel like that would help catch a lot of these small bugs. We do it in [`qiskit-algorithms`](https://github.com/qiskit-community/qiskit-algorithms/pull/141) (you can see there are 2 checks called `Monitor_Qiskit_Main`). What we did in algorithms is not to add it to the branch protection rules, so that the test failing would not break PRs from being merged (this was maybe more relevant in the pre-1.0 period), but we could still check the 1.0 compatibility. I think that it was very helpful for the transition period, and we'll probably keep doing it for future majors.

kt474:
> @kt474 Have you considered adding a CI run that tests the integration with the qiskit main branch (in this case, 1.0)? 

Yes, I think that'd be helpful - I'll try to get that setup today

ElePT:
> > @kt474 Have you considered adding a CI run that tests the integration with the qiskit main branch (in this case, 1.0)?
> 
> Yes, I think that'd be helpful - I'll try to get that setup today

Nice! Feel free to use the [algorithms setup](https://github.com/qiskit-community/qiskit-algorithms/blob/main/.github/workflows/main.yml#L168) for reference.

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -39,6 +38,11 @@
     decode_pulse_defaults,
 )
 
+try:
+    from qiskit.providers import QasmSimulatorPy as BasicSimulator
+except ImportError:
+    from qiskit.providers import BasicSimulator
+
```
ElePT, line 45: Sorry, my bad, didn't include the full import path:
```suggestion
try:
    from qiskit.providers.basicaer import QasmSimulatorPy as BasicSimulator
except ImportError:
    from qiskit.providers.basic_provider import BasicSimulator

```

kt474: No worries, just noticed this too!

kt474:
Thanks for taking the time to review @ElePT @wshanks!

PR #1365, opened 2024-02-01
QPY Qiskit 1.0 updates

### LINKED ISSUE #1338: Update QPY Module for Qiskit 1.0
There were some updates made to QPY in qiskit for the 1.0 release, including removing qiskit.extensions https://github.com/Qiskit/qiskit/pull/11488

We'll have to update the QPY module in this repo as well

kt474:
### Summary

Credit goes to:

https://github.com/Qiskit/qiskit/pull/10902
https://github.com/Qiskit/qiskit/pull/11260
https://github.com/Qiskit/qiskit/pull/11522
https://github.com/Qiskit/qiskit/pull/11403
https://github.com/Qiskit/qiskit/pull/11495
https://github.com/Qiskit/qiskit/pull/11488

https://github.com/Qiskit/qiskit/pull/11647
https://github.com/Qiskit/qiskit/pull/11646
https://github.com/Qiskit/qiskit/pull/11644
https://github.com/Qiskit/qiskit/pull/11505

### Details and comments
Fixes #1338


qiskit_ibm_runtime/qpy/exceptions.py
```diff
@@ -12,7 +12,7 @@
 
 """Exception for errors raised by the pulse module."""
 from typing import Any
-from qiskit.qpy.exceptions import QpyError
+from qiskit.qpy.exceptions import QpyError, QiskitWarning
```
ElePT, line 15:
```suggestion
from qiskit.exceptions import QiskitWarning
from qiskit.qpy.exceptions import QpyError
```

ElePT (commented):
Hi @kt474, I see a few 1.0 QPY changes not reflected in this PR, they were last-minute changes so I'm bringing them to your attention in case they had slipped under your radar (if not, feel free to ignore):

- https://github.com/Qiskit/qiskit/pull/11647 (affects qpy)
- https://github.com/Qiskit/qiskit/pull/11646
- https://github.com/Qiskit/qiskit/pull/11644

qiskit_ibm_runtime/qpy/exceptions.py
kt474: Thanks for brining this up, i'll cherry pick those PRs

qiskit_ibm_runtime/qpy/exceptions.py
```diff
@@ -12,6 +12,7 @@
 
 """Exception for errors raised by the pulse module."""
 from typing import Any
+from qiskit.exceptions import QiskitWarning
```
ElePT, line 15: This feels ugly, but a workaround to make this import work in both 0.45 and 1.0 could be aliasing `UserWarning`:

```suggestion
try:
    from qiskit.exceptions import QiskitWarning
except ImportError:
    QiskitWarning = UserWarning
```

mtreinish:
There's also: https://github.com/Qiskit/qiskit/pull/11505

mtreinish:
FWIW, I just opened https://github.com/Qiskit/qiskit-ibm-runtime/issues/1375 to track deleting the fork of the qpy module here. In qiskit 1.0.0 we added the necessary functionality to cover emitting older versions of the QPY format from `qiskit.qpy.dump()`. However, we obviously can't remove the fork until we're ready to set the qiskit dependency to `>=1.0.0`

jyu00:
Would this work with pre-1.0 QPY? Because the server is still at 0.*

mtreinish:
> Would this work with pre-1.0 QPY? Because the server is still at 0.*

Yes, there is specifically overlap in the versions supported by `dump()` to enable this. The minimum qpy compatibility version is 10 in the 1.x series, which is the version `dump()` emits by qiskit 0.45.x and 0.46.x and is the maximum version `load()` supports in 0.45.x and 0.46.x. I outlined a few of the details in #1375 on how ideally we can do version negotiation if there is an API we can query to either figure out what the maximum qpy version is supported by the server (or alternatively which qiskit version is deployed)

qiskit_ibm_runtime/qpy/binary_io/value.py
```diff
@@ -44,7 +48,7 @@ def _write_parameter_vec(file_obj, obj):  # type: ignore[no-untyped-def]
             formats.PARAMETER_VECTOR_ELEMENT_PACK,
             len(name_bytes),
             obj._vector._size,
-            obj._uuid.bytes,
+            obj.uuid.bytes,
```
kt474, line 51: @mtreinish to be compatible with `0.45`, this is causing an error

```AttributeError: 'ParameterVectorElement' object has no attribute 'uuid'```

Because uuid is added as an attribute to Parameter [here](https://github.com/Qiskit/qiskit/pull/11647/files#diff-3b25b84ba5e294ae8700df4ca202b9964f05194850cca0bde8e15c2be8104c3eR117-R126), only in 1.0 

How should this be handled here?

mtreinish: Fully porting the qpy changes from qiskit 1.0.0rc1 will inherently be unsound for use with qiskit 0.45.0 or 0.46.0. The internals of the qpy module are updated in lock-step with the `QuantumCircuit` class. This is basically the same underlying issue we hit that needed: qiskit/qiskit-ibm-provider#757

This will also start emitting qpy version 11 which will cause a version mismatch with what the server side can parse until it is upgraded to qiskit 1.0.0. I pushed up: #1377 as an alternative to this which will let you do the upgrade in a manner which is compatible with both versions so you can publish a release that is both 0.45.x/0.46.x and 1.0.0 compatible. It also removes the vendored fork completely as it is not needed anymore (and luckily has never been released either, so there is no backwards compatibility issue).

kt474: Got it, thank you so much for taking care of this! 

I'll close this PR

kt474: I'll create a PR to do the same in  `qiskit-ibm-provider` (but we can't remove the fork completely yet)

1ucian0:
@mtreinish wrote https://github.com/Qiskit/qiskit-ibm-runtime/pull/1377 as an alternative to this PR. 1377 handles QPY in a way that supports qiskit 0.45/0.46 and 1.0.

PR #1372, opened 2024-02-02
Use POST /sessions endpoint 

### LINKED ISSUE #1327: Batch mode migration
The plan is for all sessions to be switched to dedicated sessions but since we introduced `Batch` a couple versions ago, the server can not differentiate between `Session` and `Batch`. 

In the next version of `qiskit-ibm-runtime`, we will start passing the `mode` field to indicate which execution mode is being used. 

Depends on NTC issue # 3751

kt474:
### Details and comments
Fixes #1327


qiskit_ibm_runtime/session.py
```diff
@@ -78,6 +78,7 @@ def __init__(
         service: Optional[QiskitRuntimeService] = None,
         backend: Optional[Union[str, IBMBackend]] = None,
         max_time: Optional[Union[int, str]] = None,
+        mode: Optional[str] = None,
```
jyu00, line 81: `mode` shouldn't be a externally visible parameter. I.e. we should not allow `Session(mode="batch")`.

qiskit_ibm_runtime/api/clients/runtime.py
```diff
@@ -230,6 +230,14 @@ def job_metadata(self, job_id: str) -> Dict[str, Any]:
         """
         return self._api.program_job(job_id).metadata()
 
+    def create_session(self, session_id: str = "", mode: str = None) -> Dict[str, Any]:
```
jyu00, line 233: Is there really a code path where `session_id` is passed?

qiskit_ibm_runtime/session.py
```diff
@@ -124,7 +127,6 @@ def __init__(
         self._backend = backend
 
         self._setup_lock = Lock()
```
jyu00, line 129: This lock was added by https://github.com/Qiskit/qiskit-ibm-runtime/pull/1196 to make session thread safe, e.g. it enabled the following case

```
with Session(backend=backend), ThreadPoolExecutor() as executor:
    estimator1 = Estimator(options={"job_tags": ["cobyla"]})
    estimator2 = Estimator(options={"job_tags": ["nelder-mead"]})

    cobyla_result = executor.submit(minimize_thread, estimator1, "cobyla").result()
    nelder_mead_result = executor.submit(minimize_thread, estimator2, "nelder-mead").result()
```

I don't think it's needed anymore since session is now created prior to the threads start.

kt474:
The /sessions endpoint returns:

```
{'id': 'fa7726cb-1280-4182-83b1-cfc62240fd68', 'backend_name': '', 'interactive_ttl': 60, 'max_ttl': 28800, 'active_ttl': 28800, 'state': 'open', 'accepting_jobs': True, 'mode': 'legacy'}
```

When the id returned is used for calling session details, I get a 404 error

```
404 Client Error: Not Found for url: https://us-east.quantum-computing.test.cloud.ibm.com/sessions/ef09dc0d-0581-488e-abfa-81fda878ec4f. 
```

mriedem:
> When the id returned is used for calling session details, I get a 404 error

Known issue: https://github.ibm.com/IBM-Q-Software/ntc/issues/3819

qiskit_ibm_runtime/session.py
```diff
@@ -120,18 +118,36 @@ def __init__(
...
+            sim_backend = backend_obj.simulator
         self._backend = backend
 
-        self._setup_lock = Lock()
-        self._session_id: Optional[str] = None
+        if not sim_backend:
+            self._session_id = self._create_session(self._backend, self._instance)
+        else:
+            self._session_id = None
+
```
kt474, line 133: We need to add this logic because on IQP, POST `/sessions` requires `backend` and `instance` and will result in an error if a simulator is used

qiskit_ibm_runtime/session.py
```diff
@@ -162,29 +178,23 @@ def run(
...
-                self._session_id = job.job_id()
-        finally:
-            if self._setup_lock.locked():
-                self._setup_lock.release()
+        job = self._service.run(
+            program_id=program_id,
+            options=options,
+            inputs=inputs,
+            session_id=self._session_id,
+            start_session=self._session_id is None,
```
jyu00, line 191: I think the plan is to reject jobs that have `start_session=True` once dedicated session is rolled out. I don't know if that applies to simulator jobs, but probably safer to always set this to False.

qiskit_ibm_runtime/session.py
```diff
@@ -162,29 +178,23 @@ def run(
...
+            callback=callback,
+            result_decoder=result_decoder,
+        )
+
+        if self._session_id is None:
+            self._session_id = job.job_id()
```
jyu00, line 197: For simulators we should just keep `_session_id` as `None`, to prevent people from trying to use it in `from_id()` (which presumably would fail as job ids are no longer session ids).

releasenotes/notes/session-modes-5c22b68620f8d690.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    Sessions will now be started with a new ``/sessions`` endpoint that allows for different 
+    execution modes. Batch mode is now supported through :class:`~qiskit_ibm_runtime.Batch`, and 
+    :class:`~qiskit_ibm_runtime.Session` will work the same as way as before. 
+    Please see https://docs.quantum.ibm.com/run/sessions for more information.
+
+    Note that starting a session using the ``/jobs`` endpoint will no longer be 
+    supported after March 31, 2024. Please update your code as soon as possible before this date.
```
jyu00, line 10:
```suggestion
    Note that ``Session`` and ``Batch`` created from ``qiskit-ibm-runtime`` prior to this release will no longer be 
    supported after March 31, 2024. Please update your ``qiskit-ibm-runtime`` version as soon as possible before this date. 
```

jyu00 (commented):
The signature/docstring for `from_id()` should also be updated to

```
    def session_id(self) -> Optional[str]:
        """Return the session ID.

        Returns:
            Session ID. None the backend is a simulator.
        """
```

qiskit_ibm_runtime/session.py
mriedem: I wouldn't assume that as it's be a breaking api change. POST /jobs should still be possible to create a session. The server side default might change to dedicated mode but that's different.

qiskit_ibm_runtime/session.py
mriedem: As mentioned above POST /jobs can still create a session implicitly otherwise it's a breaking api change.

kt474:
Should be ready for final review

jyu00 (changes requested):

qiskit_ibm_runtime/api/rest/runtime_session.py
```diff
@@ -35,7 +35,34 @@ def __init__(self, session: RetrySession, session_id: str, url_prefix: str = "")
...
+            payload["mode"] = mode
+        if backend:
+            payload["backend"] = backend
+        if instance:
+            payload["instance"] = instance
+        if max_time:
+            if channel == "ibm_quantum":
+                payload["max_session_ttl"] = max_time  # type: ignore[assignment]
+            else:
+                payload["max_ttl"] = max_time  # type: ignore[assignment]
```
kt474, line 64: These fields are named differently in the two APIs 🫠

jyu00 (approved):

PR #1385, opened 2024-02-07
Cast use_symengine input to a bool

mtreinish:
### Summary

This commit works around a bug in Qiskit 0.45.x, 0.46.0, and 1.0.0rc1 with the `use_symengine` flag on `qpy.dump()`. The dump function has a bug when it receives a truthy value instead of a bool literal that it will generate a corrupt qpy because of a mismatch between how the encoding was processed (the encoding is incorrectly set to sympy in the file header but uses symengine encoding in the actual body of the circuit.  This is being fixed in Qiskit/qiskit#11730 for 1.0.0, and will be backported to 0.46.1. But to ensure compatibility with 0.45.x, 0.46.0, and 1.0.0rc1 while waiting for those releases we can workaround this by just casting the value to a boolean.

### Details and comments

### DIFF
--- a/qiskit_ibm_runtime/utils/json.py
+++ b/qiskit_ibm_runtime/utils/json.py
@@ -245,7 +245,7 @@ class RuntimeEncoder(json.JSONEncoder):
         if isinstance(obj, ParameterView):
             return obj.data
         if isinstance(obj, Instruction):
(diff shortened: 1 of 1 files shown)


kt474 (approved):

kt474:
LGTM, thanks - was made aware of this bug earlier today when running the `vqe_with_estimator` tutorial 

We'll have to do the same in qiskit-ibm-provider

PR #1392, opened 2024-02-08
Update qiskit version

gadial:
### Summary
Updates the qiskit version to 1.0


test/unit/transpiler/passes/scheduling/control_flow_test_case.py
```diff
@@ -15,8 +15,8 @@
 from typing import Any, Optional
 
 from qiskit import QuantumCircuit
-from qiskit.test._canonical import canonicalize_control_flow
 
+from .....utils import canonicalize_control_flow
```
kt474, line 19: We can actually remove this file, it was removed in `main` but accidentally added again when we updated `experimental-0.2`

kt474 (approved):

PR #1401, opened 2024-02-13
Unpack Pubs after validation before sending to server

chriseclectic:
### Summary

The current experimental branch is incompatible with server code since it is sending classes not supported for round trip JSON serialization. This PR unpacks the input pubs into pubs-like lists of circuits and numpy arrays for json encoding so the server doesn't need to do any special decoding.

### DIFF
--- a/qiskit_ibm_runtime/base_primitive.py
+++ b/qiskit_ibm_runtime/base_primitive.py
@@ -136,7 +136,7 @@ class BasePrimitiveV2(ABC, Generic[OptionsT]):
 
         if self._backend and options_dict["transpilation"]["skip_transpilation"]:
             for pub in pubs:
-                self._backend.check_faulty(pub[0])
+                self._backend.check_faulty(pub.circuit)
 
         logger.info("Submitting job using options %s", options_dict)
 
--- a/qiskit_ibm_runtime/base_primitive.py
+++ b/qiskit_ibm_runtime/base_primitive.py
@@ -136,7 +136,7 @@ class BasePrimitiveV2(ABC, Generic[OptionsT]):
 
         if self._backend and options_dict["transpilation"]["skip_transpilation"]:
             for pub in pubs:
-                self._backend.check_faulty(pub.circuit)
+                self._backend.check_faulty(pub[0])
 
         logger.info("Submitting job using options %s", options_dict)
 
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -17,6 +17,7 @@ import os
 from typing import Optional, Dict, Sequence, Any, Union, Iterable
 import logging
 
+import numpy as np
 from qiskit.circuit import QuantumCircuit
 from qiskit.quantum_info.operators.base_operator import BaseOperator
 from qiskit.quantum_info.operators import SparsePauliOp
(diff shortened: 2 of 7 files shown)


mberna (approved):

qiskit_ibm_runtime/base_primitive.py
```diff
@@ -136,7 +136,7 @@ def _run(self, pubs: Union[list[EstimatorPub], list[SamplerPub]]) -> RuntimeJob:
 
         if self._backend and options_dict["transpilation"]["skip_transpilation"]:
             for pub in pubs:
-                self._backend.check_faulty(pub.circuit)
+                self._backend.check_faulty(pub[0])
```
jyu00, line 139: ~What's wrong with `pub.circuit`? If the input is of type Pub~ 

If the input type is no longer a Pub, the type hint should be updated.

qiskit_ibm_runtime/estimator.py
```diff
@@ -146,8 +147,19 @@ def run(
...
+        # Unpack pubs back to Pub-like tuples for serialization
+        unpacked = [
+            (
+                pub.circuit,
+                np.asarray(pub.observables),
+                pub.parameter_values.as_array(pub.circuit.parameters),
+                pub.precision,
+            )
+            for pub in coerced_pubs
+        ]
```
jyu00, line 161: Why not do this in the `RuntimeEncoder` to keep the code here clean?

qiskit_ibm_runtime/sampler.py
```diff
@@ -117,8 +118,12 @@ def run(self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None) -> Ru
                 "will be empty. Did you mean to add measurement instructions?",
                 UserWarning,
             )
-
-        return self._run(coerced_pubs)  # type: ignore[arg-type]
+        # Unpack pubs back to Pub-like tuples for serialization
+        unpacked_pubs = [
+            (pub.circuit, pub.parameter_values.as_array(pub.circuit.parameters), pub.shots)
```
jyu00, line 123: Does the simple `as_array` work? Since parameter values can be a dictionary, the values are not necessarily in the same order as the circuit parameters.

qiskit_ibm_runtime/estimator.py
chriseclectic: It's not round-trip able if you put it there, and the point of this was to avoid having special encoding just for the program inputs that you know are pubs.

qiskit_ibm_runtime/sampler.py
chriseclectic: Yes, this is the intended API for `BindingsArray` to convert to numpy array on correctly ordered parameters

qiskit_ibm_runtime/base_primitive.py
chriseclectic: Should i put it as union PubLike's or just `tuple`? (its not as general as pub like). I didn't put the unpacking here in the base class because I didn't want to have to do an if /else for if its an estimator or sampler pub.

qiskit_ibm_runtime/estimator.py
chriseclectic: This also allows us to later add proper round-trippable JSON encoding/decoding into the encoder which I think we want eventually once everything else is working properly

jyu00: I don't understand why not. `RuntimeEncoder` can convert an `EstimatorPub` to a tuple like this. And `RuntimeDecoder` continues to do nothing (since it wouldn't know if a tuple is a pub). Then server gets a list of tuples under the `pubs` keyword and knows they are pubs and can convert them back to pubs.

chriseclectic: Ok, i added unpacking to RuntimeEncoder and removed them from the deceoder

jyu00 (approved):

PR #1404, opened 2024-02-13
Add deprecation warning for abstract circuits

### LINKED ISSUE #1403: Issue warning about abstract circuit change

We are going to drop support for abstract circuits in 2 weeks. Therefore we need to inform the users about this change asap.

**Acceptance criteria**

If a user passes abstract circuits to the primitives, both v1 and v2, issue a deprecation warning: 

`
Circuits that do not match the target hardware definition will no longer be supported after March 1, 2024. 
See the transpilation documentation (https://docs.quantum.ibm.com/transpile) for instructions to transform circuits and the primitive examples (https://docs.quantum.ibm.com/run/primitives-examples) to see this coupled with operator transformations.
`

kt474:
### Details and comments
Fixes #1403


jyu00 (approved):
Thanks for making the update so quickly!

Issue #1354, opened 2024-01-31
Encoding/decoding does not work for `np.array(0, dtype=object)`


mberna:
**Describe the bug**
An error occurs when encoding the object mentioned in the title and then decoding the resulting string.
`TypeError: argument should be a bytes-like object or ASCII string, not 'int'`

**Steps to reproduce**

```python
obj = {"ndarray": np.array(0, dtype=object)}
encoded = json.dumps(obj, cls=RuntimeEncoder)
self.assertIsInstance(encoded, str)
decoded = json.loads(encoded, cls=RuntimeDecoder)
```

**Expected behavior**
Object is properly enconded/decoded.

**Additional Information**
Might be a good idea to add a (sub) test case here: https://github.com/Qiskit/qiskit-ibm-runtime/blob/6c2cf0e71b1273923086845d4383dc00cfa559f6/test/unit/test_data_serialization.py#L159-L163

- **qiskit-ibm-runtime version**: 0.18.0
- **Python version**: 3.11
- **Operating system**: MacOS

PR #1407, opened 2024-02-15
Convert the observables to a list when encoding

mberna:
### Summary
Avoid converting observables to numpy arrays because https://github.com/Qiskit/qiskit-ibm-runtime/issues/1354 hasn't been fixed yet.

Using `.tolist` as a temporary replacement.


chriseclectic (approved):
Since its always an object array this probably makes little difference to the serialized file size.

It might be good to add a couple of Encoder/Decoder tests for these edge cases  like shape () observable, where you check you can round trip via something like
```
encoded = Encoder.dumps(input_pub)
decoded = Decoder.loads(encoded)
decoded_pub = EstimatorPub.coerce(decoded)
# check input_pub equiv to decoded_pub
```

PR #1414, opened 2024-02-19
Update ISA validation

jyu00:
### Summary

Update ISA circuit validation:
- Allow circuits without `layout`. This applies to manually built ISA circuits and backends that  don't have connectivity constraints (e.g. simulators)
- Verify the circuit has the same number of qubits as the target system
- Allow pulse gates


kt474 (approved):

qiskit_ibm_runtime/utils/utils.py
```diff
@@ -29,38 +29,40 @@
...
     Args:
         circuit: A single QuantumCircuit
-        target: A Qiskit Target
+        backend: The target backend.
+
     Returns:
         Boolean True if the circuit is an ISA circuit
     """
-    if circuit.layout is None:
+    if backend.coupling_map and circuit.layout is None:
```
mtreinish, line 47: I would probably just remove this check. Having the `layout` attribute set isn't hard requirement to run the circuit. Like you could manually build a hardware aware circuit and directly submit it and it would otherwise be completely valid to run right?

jyu00:
Putting this on hold while we discuss if we want to force users to call `transpile()` to explicitly state their intent, or just assume virtual indexes match physical indexes if layout is not given.

jyu00:
Updated to remove check of `layout` and add check of `num_qubits`. Also made the message more explicit.

blakejohnson (approved):

PR #1422, opened 2024-02-23
Add initial qiskit-neko CI job

mtreinish:
### Summary

In Qiskit/qiskit-neko#41 we're updating the default simulator backend plugin used by qiskit-neko to rely on the fake backends provider in the qiskit-ibm-runtime project. A key aspect of qiskit-neko is the reciprocity of testing to maintain compatibility in the qiskit ecosystem (see:
https://github.com/Qiskit/qiskit-neko?tab=readme-ov-file#downstream-usage-in-testing) so to ensure this is maintainable and the interface remains compatible for every project using neko this commit adds a corresponding job to exercise the qiskit-ibm-runtime usage in the tests. However, this is blocked until after Qiskit/qiskit-neko#41 merges as that is needed to fix the test suite. A followup will be needed to add a backend selection to the action configuration so that we can actually use a fake backend.

A longer term step we should investigate is adding a qiskit-neko plugin to the repo which would enable using qiskit-ibm-runtime's backends and primitives directly for integration testing with real backends. But that is much longer term and isn't part of this commit.

### Details and comments


kt474 (approved):

ElePT:
The current neko failure is dues to an incompatibility in `qiskit_algorithms` with `scipy>=1.14`, the bug was fixed in main (https://github.com/qiskit-community/qiskit-algorithms/commit/3fb69b3051f39602b1e4c3f0229a849a2ff6b8e4), but neko is pulling the latest released version which doesn't include the fix. I can do a patch release of algorithms, but given the current level of maintenance of the repo, it might make sense considering installing the package from source when running neko.

ElePT (commented):
The current neko failure is dues to an incompatibility in `qiskit_algorithms` with `scipy>=1.14`, the bug was fixed in main (https://github.com/qiskit-community/qiskit-algorithms/commit/3fb69b3051f39602b1e4c3f0229a849a2ff6b8e4), but neko is pulling the latest released version which doesn't include the fix. I can do a patch release of algorithms, but given the current level of maintenance of the repo, it might make sense considering installing the package from source when running neko.

PR #1423, opened 2024-02-23
Upgrade to qiskit 1.0

mberna:
### Summary
Move away from the release candidate and require `qiskit>=1.0.0`.


kt474:
Version should be updated here https://github.com/Qiskit/qiskit-ibm-runtime/blob/experimental-0.2/setup.py#L30 as well

Issue #1416, opened 2024-02-20
Make backend required for cloud channel


jyu00:
**What is the expected feature or enhancement?**

Soon primitives will no longer support abstract circuits, and as a result of that `backend` can no longer be optional when calling primitives. 

**Acceptance criteria**

- [x] Issue a warning that `backend` will soon be required when calling Sampler, Estimator, Session, and Batch (#1425). 
- [ ] Make `backend` required after the 3 months deprecation period

PR #1425, opened 2024-02-23
Add warning if no backend provided on cloud channel

kt474:
### Details and comments
Fixes part of #1416

### DIFF
--- a/qiskit_ibm_runtime/session.py
+++ b/qiskit_ibm_runtime/session.py
@@ -12,6 +12,7 @@
 
 """Qiskit Runtime flexible session."""
 
+import warnings
 from typing import Dict, Optional, Type, Union, Callable, Any
 from types import TracebackType
 from functools import wraps
@@ -112,8 +113,15 @@ class Session:
         else:
             self._service = service
 
-        if self._service.channel == "ibm_quantum" and not backend:
-            raise ValueError('"backend" is required for ``ibm_quantum`` channel.')
+        if not backend:
+            if self._service.channel == "ibm_quantum":
+                raise ValueError('"backend" is required for ``ibm_quantum`` channel.')
+            warnings.warn(
+                "In a future release no sooner than 3 months after the release date of "
+                "qiskit-ibm-runtime version 0.21.0, "
+                "``backend`` will no longer be optional when using primitives. "
+                "Please provide a backend."
+            )
 
         self._instance = None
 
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/session.py
```diff
@@ -112,8 +113,15 @@ def __init__(
...
-            raise ValueError('"backend" is required for ``ibm_quantum`` channel.')
+        if not backend:
+            if self._service.channel == "ibm_quantum":
+                raise ValueError('"backend" is required for ``ibm_quantum`` channel.')
+            warnings.warn(
+                "In a future release no sooner than 3 months after the release date of "
+                "qiskit-ibm-runtime version 0.21.0, "
+                "``backend`` will no longer be optional when using primitives. "
+                "Please provide a backend."
+            )
```
jyu00, line 124: Why not use the `issue_deprecation_msg()` so the format is consistent and sets the right type.

jyu00 (changes requested):
The same check needs to apply to `BasePrimitive` as well.

qiskit_ibm_runtime/session.py
kt474: I didn't think the wording in `issue_deprecation_msg()` fit well here 

```f"and will be removed no sooner than {period} after the release date. {remedy}"``` 

I can update the message to more closely match that format and also set it as `DeprecationWarning`

File: releasenotes/notes/require-cloud-backend-38a74144347c735d.yaml [hunk 2]
```diff
@@ -0,0 +1,6 @@
+---
+deprecations:
+  - |
+    In a future release, ``backend`` will be a required parameter for :class:`qiskit_ibm_runtime.Sampler`, 
+    :class:`qiskit_ibm_runtime.Estimator`, :class:`qiskit_ibm_runtime.Session` and 
+    :class:`qiskit_ibm_runtime.Batch` when using the ``ibm_cloud`` channel.
```
jyu00, line 6: We should be more explicit that `backend` is only required if `session` is not specified. 

```suggestion
    In a future release, ``backend`` will be a required parameter for :class:`qiskit_ibm_runtime.Sampler`, and
    :class:`qiskit_ibm_runtime.Estimator` if ``session`` is not specified,
    even when using the ``ibm_cloud`` channel. 
    
    It will also be a required parameter for 
    :class:`qiskit_ibm_runtime.Session` and 
    :class:`qiskit_ibm_runtime.Batch`.
```

jyu00 (approved):
1 comment but otherwise looks good

PR #1427, opened 2024-02-24
Remove unused classes from encoder

mberna:
### Details and comments
Fixes #


kt474 (approved):
Looks like `test_bit_array` and `test_data_bin` also need to be removed

jyu00:
But if we don't put the encoder and decoder together, how do we make sure they are aligned?

PR #1431, opened 2024-02-27
Use sphinx.ext.linkcode for more precise source code links

melechlapson:
### Summary
This addresses https://github.com/Qiskit/documentation/issues/517 by switching out the sphinx.ext.viewcode for the sphinx.ext.linkcode extension which allows our GitHub links in the documentation to be linked to the specific lines of code, not just the file. This was tested successfully in https://github.com/Qiskit/qiskit_sphinx_theme/pull/589 and in the qiskit repo in https://github.com/Qiskit/qiskit/pull/11851 and now we want to implement it in the runtime repo.

The API generation script at Qiskit/documentation already knows how to handle sphinx.ext.linkcode.

Example links from this PR:

https://github.com/Qiskit/qiskit-ibm-runtime/tree/main/qiskit_ibm_runtime/fake_provider/backends/armonk/fake_armonk.py#L21-L33

https://github.com/Qiskit/qiskit-ibm-runtime/tree/main/qiskit_ibm_runtime/fake_provider/fake_backend.py#L269-L282

PR #1439, opened 2024-02-28
Update deprecated actions in workflows 

melechlapson:
### Summary
In recent PRs including my PR https://github.com/Qiskit/qiskit-ibm-runtime/pull/1431, the CI has been throwing several errors due to old versions of actions that are using node12 or node16 instead of node20. I updated the actions to use the latest versions which should eliminate the CI warnings.


Eric-Arellano (approved):
Thank you 🙌

PR #1443, opened 2024-02-28
Handle control flow instructions in Target generation

mtreinish:
### Summary

This commit updates the convert_to_target() function that is used to convert the API response objects used to describe a backend and generate a Qiskit Target object from that to properly handle control flow instructions. The supported control flow instructions are listed in the `supported_instructions` field of the configuration payload. [1] This commit updates the logic to look for the presence of known control flow instruction names in that field and add them as appropriate to the target as global variadic instructions to indicate to Qiskit's transpiler and other Target object users that the backend supports the instructions.

[1] https://github.com/Qiskit/ibm-quantum-schemas/blob/0231221082ec722cc31db09c0b41a25f441ac338/schemas/backend_configuration_schema.json#L165-L170

### Details and comments


jyu00 (approved):
Thanks @mtreinish!

PR #1347, opened 2024-01-29
Support Python 3.12

kt474:
### Details and comments
Fixes #1345

PR #1449, opened 2024-02-29
Migrate from `pkg_resources` to `importlib.metadata`

garrison:
### Summary

Migrate from `pkg_resources` to `importlib.metadata` to improve Python 3.12 support

### Details and comments
PR #1347 added `setuptools` to `requirements-dev.txt` in an effort to make `qiskit-ibm-runtime` work with Python 3.12.  However, `qiskit_ibm_runtime/api/session.py` actually imports `pkg_resources`, which is provided by setuptools, so really setuptools should have been added to [`REQUIREMENTS`](https://github.com/Qiskit/qiskit-ibm-runtime/blob/e152d122782959a04ebf2330f776d4c4e309ad14/setup.py#L21) in `setup.py` so that it is installed automatically for any downstream users of Qiskit Runtime, too.

Better yet, however, is to drop this explicit dependency on setuptools, since [`pkg_resources` is deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html) in favor of `importlib`.  Hence, this PR migrates from `pkg_resources` to `importlib.metadata` (there are a few more words about this [here](https://github.com/googleapis/python-api-core/issues/27#issuecomment-676563754)).

My investigation into this was prompted by us trying to add Python 3.12 support to the circuit-knitting-toolbox (https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/429), and me then seeing CI errors prompted by `qiskit-ibm-runtime`'s importing of `pkg_resources`.


kt474 (approved):
Thank you @garrison!

PR #1450, opened 2024-02-29
[WIP] Add RuntimeJobV2 class

### LINKED ISSUE #1352: Switch to using BasePrimitiveJob for v2 primitives

kt474:
### Details and comments
Fixes #1352

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -24,8 +24,8 @@ from qiskit.primitives import BaseEstimator
 from qiskit.primitives.base import BaseEstimatorV2
 from qiskit.primitives.containers import EstimatorPubLike
 from qiskit.primitives.containers.estimator_pub import EstimatorPub
-
 from .runtime_job import RuntimeJob
+from .runtime_jobV2 import RuntimeJobV2
 from .ibm_backend import IBMBackend
 from .options import Options
 from .options.estimator_options import EstimatorOptions
@@ -132,7 +132,7 @@ class EstimatorV2(BasePrimitiveV2[EstimatorOptions], Estimator, BaseEstimatorV2)
 
     def run(
         self, pubs: Iterable[EstimatorPubLike], *, precision: float | None = None
-    ) -> RuntimeJob:
+    ) -> RuntimeJobV2:
         """Submit a request to the estimator primitive.
 
         Args:
--- a/qiskit_ibm_runtime/__init__.py
+++ b/qiskit_ibm_runtime/__init__.py
@@ -180,6 +180,7 @@ import warnings
 from .qiskit_runtime_service import QiskitRuntimeService
 from .ibm_backend import IBMBackend
 from .runtime_job import RuntimeJob
+from .runtime_jobV2 import RuntimeJobV2
 from .runtime_options import RuntimeOptions
 from .utils.json import RuntimeEncoder, RuntimeDecoder
 from .session import Session  # pylint: disable=cyclic-import
--- a/qiskit_ibm_runtime/base_primitive.py
+++ b/qiskit_ibm_runtime/base_primitive.py
@@ -30,6 +30,7 @@ from .options import Options
 from .options.options import BaseOptions, OptionsV2
 from .options.utils import merge_options, set_default_error_levels
 from .runtime_job import RuntimeJob
+from .runtime_jobV2 import RuntimeJobV2
 from .ibm_backend import IBMBackend
 from .utils.default_session import get_cm_session
 from .utils.utils import validate_isa_circuits
@@ -120,7 +121,7 @@ class BasePrimitiveV2(ABC, Generic[OptionsT]):
                     "A backend or session must be specified when not using ibm_cloud channel."
                 )
 
-    def _run(self, pubs: Union[list[EstimatorPub], list[SamplerPub]]) -> RuntimeJob:
+    def _run(self, pubs: Union[list[EstimatorPub], list[SamplerPub]]) -> RuntimeJobV2:
         """Run the primitive.
 
         Args:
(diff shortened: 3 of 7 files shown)


qiskit_ibm_runtime/job_status.py
```diff
@@ -0,0 +1,37 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""This module defines an enumerated type for the state of primitive jobs"""
+
+import enum
+
+
+class JobStatus(enum.Enum):
```
jyu00, line 18: If you use an enum for job status, and we introduce a new status, wouldn't that break downlevel providers?

kt474: Should strings be returned instead? As mentioned in https://github.com/Qiskit/qiskit/issues/11541

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -0,0 +1,788 @@
...
+
+        result_raw = self._download_external_result(
+            self._api_client.job_results(job_id=self.job_id())
+        )
+        # TODO: Remove getting/setting version once it's in result metadata
+        if _decoder.__name__ == EstimatorResultDecoder.__name__:
+            if not self._version:
+                self._version = self.inputs.get("version", 1)
+
+        return _decoder.decode(result_raw) if result_raw else None  # type: ignore
```
kt474, line 253: I don't see the version in the result metadata, but removed the `version_param` object because it's not accepted by `estimator_result_decoder` anymore

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -0,0 +1,788 @@
...
+API_TO_JOB_STATUS: Dict[str, JobStatus] = {
+    "QUEUED": "QUEUED",
+    "RUNNING": "RUNNING",
+    "COMPLETED": "DONE",
+    "FAILED": "ERROR",
+    "CANCELLED": "CANCELLED",
+}
+
+
+class RuntimeJobV2(BasePrimitiveJob):
```
jyu00, line 67: Instead of duplicating all the same methods, why not just create a `BaseRuntimeJob` that has all the common method. Then `RuntimeJobV2` can inherit from both `BaseRuntimeJob` and `BasePrimitiveJob`.

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -0,0 +1,408 @@
...
+API_TO_JOB_STATUS: Dict[str, JobStatus] = {
+    "QUEUED": "QUEUED",
+    "RUNNING": "RUNNING",
+    "COMPLETED": "DONE",
+    "FAILED": "ERROR",
+    "CANCELLED": "CANCELLED",
+}
+
+
+class RuntimeJobV2(BasePrimitiveJob, BaseRuntimeJob):
```
jyu00, line 54: Similar to my previous comment, I think this should be 

```
class RuntimeJobV2(BasePrimitiveJob[PrimitiveResult][JobStatus], BaseRuntimeJob):
```

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -0,0 +1,408 @@
...
+
+        Returns:
+            Error message string or ``None``.
+        """
+        self._set_status_and_error_message()
+        return self._error_message
+
+    def _set_status_and_error_message(self) -> None:
+        """Fetch and set status and error message."""
+        if self._status not in JOB_FINAL_STATES:
```
jyu00, line 191: I don't think these methods need to be duplicated, especially since there are methods like `in_final_state()`

qiskit_ibm_runtime/estimator.py
```diff
@@ -132,7 +132,7 @@ def __init__(
 
     def run(
         self, pubs: Iterable[EstimatorPubLike], *, precision: float | None = None
-    ) -> RuntimeJob:
+    ) -> RuntimeJobV2:
```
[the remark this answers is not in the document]
kt474: mypy complains about this but I believe it is correct

`"RuntimeJobV2" expects no type arguments, but 1 given  [type-arg]`

qiskit_ibm_runtime/runtime_job_v2.py
kt474: If these status related methods are in `BaseRuntimeJob` instead, it makes things messier because `JOB_FINAL_STATES` is different, and all the statuses are strings instead of enums.

jyu00: But if the only difference is `JOB_FINAL_STATES`, and you're just checking if a status is in final state, can't you just call `if not self.in_final_state()` and have each child implement its own `in_final_state()` ?

qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -0,0 +1,463 @@
...
+        """
+        if self.errored():
+            self._error_message = self._error_msg_from_job_response(job_response)
+        else:
+            self._error_message = None
+
+    @abstractmethod
+    def _status_from_job_response(self, response: Dict) -> str:
+        """Returns the job status from an API response."""
+        return response["state"]["status"].upper()
```
kt474, line 266: Wasn't sure what the best way to handle this `_status_from_job_response` method was - I don't think it's best practice to make a private method abstract but it makes sense for the child classes to have to implement this method since the `API_TO_JOB_STATUS` is different

jyu00 (approved):
I think the 2 job classes can be consolidated a bit more, but we can do that later.

PR #1460, opened 2024-03-03
Update primitive options

jyu00:
### Summary

Update primitive options to match server side support 

### Details and comments
Fixes #


qiskit_ibm_runtime/options/options.py
```diff
@@ -123,48 +124,78 @@ def _set_if_exists(name: str, _inputs: dict, _options: dict) -> None:
...
             "twirling",
-            "dynamical_decoupling",
-            "seed_estimator",
         ]:
-            _set_if_exists(fld, inputs, options_copy)
+            _set_if_exists(fld, output_options, options_copy)
+
+        resilience_options = output_options.get("resilience", {})
+        # Convert zne/pec settings to layer mitigation
+        layer_mitigation = []
```
ihincks, line 150: Please check if this should be a string or a list, because I think it needs to match Literal["zne", "pec"] | None.

(I also left another comment wrt this discrepancy)

qiskit_ibm_runtime/options/options.py
```diff
@@ -123,48 +124,78 @@ def _set_if_exists(name: str, _inputs: dict, _options: dict) -> None:
...
+            if new_keys:
+                output_options["experimental"] = new_keys
 
         # Remove image
-        inputs.pop("image", None)
+        output_options.pop("image", None)
 
-        inputs["version"] = OptionsV2._VERSION
-        _remove_dict_unset_values(inputs)
+        output_options["version"] = OptionsV2._VERSION
```
ihincks, line 192: Should the version go in "inputs" instead? I know the options get flattened elsewhere, so maybe it doesn't make a difference, but if we switch that, I believe the version will be in the wrong spot.

test/unit/test_estimator.py
```diff
@@ -129,32 +129,32 @@ def test_run_default_options(self):
...
             ),
             (
                 EstimatorOptions(optimization_level=1),  # pylint: disable=unexpected-keyword-arg
                 {"transpilation": {"optimization_level": 1}},
             ),
             (
                 {
-                    "optimization_level": 1,
-                    "execution": {"shots": 100},
+                    "default_precision": 0.5,
```
ihincks, line 141: fwiw this would correspond to like 4 shots. I assume this doesn't matter in this test. Set to 0.1 for ~100 shots.

ihincks (commented):
Thanks @jyu00 ! This is looking great.

qiskit_ibm_runtime/base_primitive.py
```diff
@@ -132,17 +132,18 @@ def _run(self, pubs: Union[list[EstimatorPub], list[SamplerPub]]) -> RuntimeJob:
         primitive_inputs = {"pubs": pubs}
         options_dict = asdict(self.options)
         self._validate_options(options_dict)
-        primitive_inputs.update(self._options_class._get_program_inputs(options_dict))
+        primitive_options = self._options_class._get_program_inputs(options_dict)
+        primitive_inputs.update(primitive_options)
```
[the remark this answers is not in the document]
jyu00: I'm not sure what you meant. This will output something like

```
{"pubs": ..., "options": {...}, "resilience_level": 1}
```

to match what you have in your PR.

qiskit_ibm_runtime/options/dynamical_decoupling_options.py
```diff
@@ -0,0 +1,53 @@
...
+            ``pulse_alignment`` is specified. This sometimes results in the duration of the
+            created sequence being shorter than the idle time that you want to fill with the
+            sequence, i.e. `extra slack`. This option takes following values.
+
+            * ``"middle"``: Put the extra slack to the interval at the middle of the sequence.
+            * ``"edges"``: Divide the extra slack as evenly as possible into intervals at
+              beginning and end of the sequence.
+
+        scheduling_method: Whether to schedule gates as soon as ("asap") or
+            as late as ("alap") possible.
```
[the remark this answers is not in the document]
jyu00: I was debating whether to document the defaults here, or in the [IQP doc](https://docs.quantum.ibm.com/run), or both. Having them here has the disadvantage that down level provider will have the outdated information if the default changes on the server side. But maybe we can just add a disclaimer to that 🤔

qiskit_ibm_runtime/options/estimator_options.py
```diff
@@ -58,55 +79,36 @@ class EstimatorOptions(OptionsV2):
...
     """
 
     # Sadly we cannot use pydantic's built in validation because it won't work on Unset.
+    default_precision: Union[UnsetType, float] = Unset
+    default_shots: Union[UnsetType, int] = Unset
     optimization_level: Union[UnsetType, int] = Unset
     resilience_level: Union[UnsetType, int] = Unset
-    dynamical_decoupling: Union[UnsetType, DDSequenceType] = Unset
     seed_estimator: Union[UnsetType, int] = Unset
+    dynamical_decoupling: Union[DynamicalDecouplingOptions, Dict] = Field(
```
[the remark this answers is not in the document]
jyu00: It is exactly to signal to users that it's a valid coercion type, and that they don't need to import/pass  `DynamicalDecouplingOptions`.

qiskit_ibm_runtime/options/resilience_options.py
```diff
@@ -34,115 +34,60 @@
...
-    # TREX
-    measure_noise_mitigation: Union[UnsetType, bool] = Unset
-    # TODO: measure_noise_local_model
-
-    # ZNE
+    measure_mitigation: Union[UnsetType, bool] = Unset
+    measure_noise_learning: Union[MeasureNoiseLearningOptions, Dict] = Field(
+        default_factory=MeasureNoiseLearningOptions
+    )
     zne_mitigation: Union[UnsetType, bool] = Unset
```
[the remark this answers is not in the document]
jyu00: Yes, I think it's more intuitive to have them as bools, which also matches `measure_mitigation`.

qiskit_ibm_runtime/options/dynamical_decoupling_options.py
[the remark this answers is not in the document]
ihincks: I'm not sure what would be best. There's also possible defaulting at the endpoint schema level.

qiskit_ibm_runtime/base_primitive.py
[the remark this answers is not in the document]
ihincks: Oh great, I just misunderstood your code.

qiskit_ibm_runtime/options/options.py
ihincks: Am I also wrong here? Please resolve if so.

qiskit_ibm_runtime/options/utils.py
```diff
@@ -61,13 +64,23 @@ def set_default_error_levels(
...
+
+
+def reomve_empty_dict(in_dict: dict) -> None:
+    """Remove empty dictionaries."""
+    for key, val in list(in_dict.items()):
+        if isinstance(val, dict):
+            if val:
+                reomve_empty_dict(val)
+            if not val:
+                del in_dict[key]
```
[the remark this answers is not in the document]
jyu00: fixed in [7e49516](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1460/commits/7e49516094b89c40f7cb69ca8541a6e9b51c7e5f)

qiskit_ibm_runtime/options/options.py
jyu00: You were right. It's fixed in [7e49516](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1460/commits/7e49516094b89c40f7cb69ca8541a6e9b51c7e5f)

ihincks:
This PR LGTM. Holding off on approval until we agree on how to synchronize server-side efforts.

qiskit_ibm_runtime/options/twirling_options.py
```diff
@@ -30,29 +30,34 @@ class TwirlingOptions:
     """Twirling options.
 
     Args:
-        gates: Whether to apply 2-qubit gate twirling.
-            By default, gate twirling is enabled for resilience level >0.
+        enable_gates: Whether to apply 2-qubit gate twirling. Default: False.
 
-        measure: Whether to apply measurement twirling.
-            By default, measurement twirling is enabled for resilience level >0.
+        enable_measure: Whether to apply measurement twirling. Default: True.
```
chriseclectic, line 35: This should be more clear that it only applys to expectation value measurements, not any measurements in the pub circuit.

qiskit_ibm_runtime/options/resilience_options.py
```diff
@@ -34,115 +34,63 @@
...
-        zne_noise_factors: An list of real valued noise factors that determine by what amount the
-            circuits' noise is amplified.
-            Only applicable if ZNE is enabled.
+        zne: Additional zero noise extrapolation mitigation options.
+            See :class:`ZneOptions` for all options.
 
-        zne_extrapolator: An extrapolation strategy. One or more of ``"exponential"``,
-            ``"double_exponential"``, and ``"linear"``.
-            Only applicable if ZNE is enabled.
+        pec_mitigation: Whether to turn on Probabilistic Error Cancellation error mitigation method.
```
chriseclectic, line 57: Since pec and zne mitigation are mutually exclusive, why not have a single option like on the server side that has to be set to PEC or ZNE?

qiskit_ibm_runtime/options/layer_noise_learning_options.py
```diff
@@ -0,0 +1,75 @@
...
+        These options are only used when the resilience level or options specify a
+        technique that requires layer noise learning.
+
+    Args:
+        max_layers_to_learn: The max number of unique layers to learn.
+            A ``None`` value indicates that there is no limit.
+            If there are more unique layers present, then some layers will not be learned or
+            mitigated. The learned layers are prioritized based on the number of times they
+            occur in a set of run estimator PUBs, and for equally occurring layers are
+            further sorted by the number of two-qubit gates in the layer. Default: 20.
```
chriseclectic, line 42: The default of 20 is way to high, this could easily timeout an 8 hour job just with learning. In V1 the default was 3 or 4 and we should use the same here unless a user explicitly changes it.

qiskit_ibm_runtime/options/measure_noise_learning_options.py
```diff
@@ -0,0 +1,41 @@
...
+
+    .. note::
+        These options are only used when the resilience level or options specify a
+        technique that requires measurement noise learning.
+
+    Args:
+        meas_num_randomizations: The number of random circuits to draw for the measurement
+            learning experiment. Default: 32.
+
+        meas_shots_per_randomization: The number of shots to use for the learning experiment
```
chriseclectic, line 32: The `"auto"` is the default value here, we should also document the default heuristic which scales the total number of shots with the pub precision and number of basis observable being measureed

qiskit_ibm_runtime/options/measure_noise_learning_options.py
```diff
@@ -0,0 +1,41 @@
...
+            per random circuit.
+    """
+
+    meas_num_randomizations: Union[UnsetType, int] = Unset
```
chriseclectic, line 36:
```suggestion
    num_randomizations: Union[UnsetType, int] = Unset
```
With nesting, we can keep consistent names for all the sampling options

qiskit_ibm_runtime/options/measure_noise_learning_options.py
```diff
@@ -0,0 +1,41 @@
...
+    meas_shots_per_randomization: Union[UnsetType, int] = Unset
```
chriseclectic, line 37:
```suggestion
    shots_per_randomization: Union[UnsetType, int, Literal["auto"]] = Unset
```

qiskit_ibm_runtime/options/twirling_options.py
```diff
@@ -30,29 +30,34 @@ class TwirlingOptions:
...
+
+        num_randomizations: The number of random samples to use when twirling or
+            peforming sampled mitigation.
```
chriseclectic, line 38: A default value of `"auto"` should be documented here.

qiskit_ibm_runtime/options/twirling_options.py
```diff
@@ -30,29 +30,34 @@ class TwirlingOptions:
...
+
+        shots_per_randomization: The number of shots to run for each random sample.
```
chriseclectic, line 40: Same here (default "auto")

qiskit_ibm_runtime/options/twirling_options.py
```diff
@@ -30,29 +30,34 @@ class TwirlingOptions:
...
-            Default: ``"active-accum"`` for resilience levels 0, 1, 2. ``"active"`` for
-                     resilience level 3.
+            Default: "active-accum".
     """
 
-    gates: Union[UnsetType, bool] = Unset
-    measure: Union[UnsetType, bool] = Unset
+    enable_gates: Union[UnsetType, bool] = Unset
+    enable_measure: Union[UnsetType, bool] = Unset
+    num_randomizations: Union[UnsetType, int] = Unset
```
chriseclectic, line 59:
```suggestion
    num_randomizations: Union[UnsetType, int, Literal["auto"]] = Unset
```

qiskit_ibm_runtime/options/twirling_options.py
```diff
@@ -30,29 +30,34 @@ class TwirlingOptions:
...
+    shots_per_randomization: Union[UnsetType, int] = Unset
```
chriseclectic, line 60:
```suggestion
    shots_per_randomization: Union[UnsetType, int, Literal["auto"]] = Unset
```

qiskit_ibm_runtime/options/sampler_options.py
```diff
@@ -14,62 +14,38 @@
...
-from .utils import Dict, Unset, UnsetType, skip_unset_validation
+from .utils import Dict, Unset, UnsetType
 from .execution_options import ExecutionOptionsV2
-from .twirling_options import TwirlingOptions
 from .options import OptionsV2
 from .utils import primitive_dataclass
+from .dynamical_decoupling_options import DynamicalDecouplingOptions
 
 DDSequenceType = Literal["XX", "XpXm", "XY4"]
 MAX_OPTIMIZATION_LEVEL: int = 1
```
chriseclectic, line 26:
```suggestion
```
Don't think these are needed here anymore

qiskit_ibm_runtime/options/estimator_options.py
```diff
@@ -12,101 +12,108 @@
...
+            collection of qubit-wise commuting observables for some specific circuit/parameter
+            value set to create a single measurement with basis rotations that is inserted into
+            hardware executions.
+
+            If twirling is enabled, the value of this option will be divided over circuit,
+            randomizations, with a smaller number of shots per randomization. See the
+            :attr:`~twirling` options.
+
+
         optimization_level: How much optimization to perform on the circuits.
```
chriseclectic, line 62: Is optimization_level current used for anything? If not I suggest we remove it and only add it back later when it is used for something. If it is used for something it should be documented as something more specific than "light optimization"

qiskit_ibm_runtime/options/estimator_options.py
```diff
@@ -12,101 +12,108 @@
...
             * 1: light optimization
 
+            Default: 1.
+
         resilience_level: How much resilience to build against errors.
             Higher levels generate more accurate results,
             at the expense of longer processing times.
 
             * 0: No mitigation.
             * 1: Minimal mitigation costs. Mitigate error associated with readout errors.
```
chriseclectic, line 76: Should we add a bit more of a summary of what these do here? 

Maybe we should use `measure` error terminology like in the resilience options
```suggestion
            * 1: Minimal mitigation costs. Apply measure twirling and mitigate error associated with measure errors.
```

qiskit_ibm_runtime/options/estimator_options.py
```diff
@@ -12,101 +12,108 @@
...
             * 2: Medium mitigation costs. Typically reduces bias in estimators but
```
chriseclectic, line 77: If you wanted to add more detail here it could be something like level 1 methods supplemented with gate twirling and zero noise extrapolation layer noise mitigation.

qiskit_ibm_runtime/options/estimator_options.py
```diff
@@ -12,101 +12,108 @@
...
-
-        simulator: Simulator options. See
-            :class:`SimulatorOptions` for all available options.
-
+        experimental: Experimental options.
     """
 
     # Sadly we cannot use pydantic's built in validation because it won't work on Unset.
+    default_precision: Union[UnsetType, float] = Unset
+    default_shots: Union[UnsetType, int] = Unset
```
chriseclectic, line 104: This one None is an allowed value, since it is only used to override default_precision (which always has to be float)
```suggestion
    default_shots: Union[UnsetType, int, None] = Unset
```

chriseclectic (changes requested):

qiskit_ibm_runtime/options/estimator_options.py
jyu00: There is a link to the docs that goes into more details, including what's done for each level, and what each method is about (at least there will be once docs are updated for V2)  .

qiskit_ibm_runtime/options/estimator_options.py
jyu00: There is a link to the docs that goes into more details, including what's done for each level, and what each method is about (at least there will be once docs are updated for V2) .

qiskit_ibm_runtime/options/estimator_options.py
jyu00: How would `None` be different from `Unset`?

qiskit_ibm_runtime/options/measure_noise_learning_options.py
jyu00: `"auto"` is essentially saying "use whatever the server decides", which is the same as `Unset`.

qiskit_ibm_runtime/options/measure_noise_learning_options.py
jyu00: So `qiskit-ibm-runtime` allows flat kwargs when updating an option. For example, I can do 
```
estimator.options.update(meas_num_randomizations=16)
```
instead of
```
estimator.options.update({"resilience": {"layer_noise_learning": {"meas_num_randomizations": 16} }})
```

The caveat is the options cannot have the same names, hence the renaming.

qiskit_ibm_runtime/options/resilience_options.py
jyu00: Two reason: (1) it's easier to code, like you don't need to know which methods are available, and (2) it matches the other switches, like `measure_mitigation` and `dynamical_decouplig.enable`. 

There is a check to verify they're not both specified.

qiskit_ibm_runtime/options/twirling_options.py
jyu00: Same as my other comment, isn't `"auto"` the same as `Unset`? i.e. use server's auto detection

qiskit_ibm_runtime/options/measure_noise_learning_options.py
chriseclectic: Unset might mean let server decide, but `Literal["auto"]` is only an allowed value for certain options (like this one), and if that is the server default it means the server will choose a value based on some heuristics, not just the server will choose a fixed value.

In absence of having an exact description of the heuristic here, which would probably be better in API docs like you say, we could just say something like: if `"auto"` this value will be chosen automatically basedo n the input pubs (Default `"auto"`), or something along those lines

qiskit_ibm_runtime/options/measure_noise_learning_options.py
chriseclectic: You already have this name twice in `layer_noise_learning` and `twirling`

ihincks: It seems like the update method is really limiting our ability to use nested options effectively. It's unusual to have restrictions on names for things in different namespaces. I'd rather see the update method changed than have this restriction.  I'm also not convinced that update should be thought of as a good way to set a single option, more of as a way to update a whole bunch of options in a script.

You get nice tab-completion with this:
```
estimator.options.resilience.layer_noise_learning.num_randomizations = 50
```

jyu00: oh I missed that! I guess I'll remove it for now even though I really hate having to do 3-layers of nested dictionary 😞

qiskit_ibm_runtime/options/measure_noise_learning_options.py
ihincks: Off the top of my head, all of the places where we have "auto", the default value is "auto". To me, this sort of means that Unset is fine until and if that changes. Doesn't mean we can't describe it. In some places, we do distinguish between None and "auto", where "auto" means "it will have a value, but we'll choose it for you later" and None means "this thing won't ever have a value".

jyu00: Updated in [db05269](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1460/commits/db052691a0591c4cff71c6c7156e4b5c9f7dad1b)

qiskit_ibm_runtime/options/estimator_options.py
```diff
@@ -12,101 +12,114 @@
...
 class EstimatorOptions(OptionsV2):
-    """Options for EstimatorV2.
+    """Options for V2 Estimator.
 
     Args:
+        default_precision: The default precision to use for any PUB or ``run()``
+            call that does not specify one.
+            Each estimator pub can specify its own precision. If the ``run()`` method
+            is given a precision, then that value is used for all PUBs in the ``run()``
+            call that do not specify their own. Default: 0.02.
```
ihincks, line 44: The default is actually  0.015625, which is a weird number, because its 1 / sqrt(4096), since 4096 was the previous default shots.

mberna (approved):

qiskit_ibm_runtime/options/layer_noise_learning_options.py
chriseclectic: Maybe just remove the default for now then and we can add something back in later
```suggestion
            further sorted by the number of two-qubit gates in the layer.
```

chriseclectic (approved):
one minor comment, but not critical

PR #1477, opened 2024-03-06
Allow importing SamplerV1 and EstimatorV1

Eric-Arellano:
Right now, users can't import `qiskit_ibm_runtime.SamplerV1` and `qiskit_ibm_runtime.EstimatorV1` because we only expose the aliases `Sampler` and `Estimator`. That gets in the way of users being explicit about which primitive version they're using.

This also is resulting in the API docs not including the pages for both the V1 and V2 primitives in https://github.com/Qiskit/documentation/pull/973 because we didn't tell autosummary about them.

### DIFF
--- a/qiskit_ibm_runtime/__init__.py
+++ b/qiskit_ibm_runtime/__init__.py
@@ -173,15 +173,15 @@ from .exceptions import *
 from .utils.utils import setup_logger
 from .version import __version__
 
-from .estimator import (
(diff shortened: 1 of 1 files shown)


Eric-Arellano:
@kt474 @jyu00 note that this is an actual user API change. You may want to do a patch release today with this fix.

jyu00 (approved):

PR #1478, opened 2024-03-06
Allow importing SamplerV1 and EstimatorV1 (Cherry-pick of #1477)

Eric-Arellano:
Right now, users can't import `qiskit_ibm_runtime.SamplerV1` and `qiskit_ibm_runtime.EstimatorV1` because we only expose the aliases `Sampler` and `Estimator`. That gets in the way of users being explicit about which primitive version they're using.

This also is resulting in the API docs not including the pages for both the V1 and V2 primitives in https://github.com/Qiskit/documentation/pull/973 because we didn't tell autosummary about them.

### DIFF
--- a/qiskit_ibm_runtime/__init__.py
+++ b/qiskit_ibm_runtime/__init__.py
@@ -144,7 +144,11 @@ Classes
 
    QiskitRuntimeService
    Estimator
+   EstimatorV1
+   EstimatorV2
    Sampler
+   SamplerV1
+   SamplerV2
    Session
    IBMBackend
    RuntimeJob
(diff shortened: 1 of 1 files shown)


jyu00:
Isn't this a duplicate of #1477?

Eric-Arellano:
Oops, I forgot to change the branch. Fixed. Thanks!

jyu00:
I think this needs to go into 0.21.1 instead of just cherry picked into `stable/0.21`. cc @kt474

Eric-Arellano:
@jyu00 we need to cherry-pick to stable/0.21, then you/Kevin will need to make a 0.21.1 release from the stable/0.21 branch after. We can't cherry-pick a PR into a tag since a tag is a single snapshot in time.

jyu00:
Yes you're right. Sorry it's been too long since I did a patch (or any) release.

Eric-Arellano:
All good! @jyu00 can you please approve this so we can merge?

kt474 (approved):

kt474:
Thanks @Eric-Arellano! I'll do a patch release today

Issue #989, opened 2023-08-02
Use status code instead of `Ran too long`


kt474:
In a couple of places in `RuntimeJob` we are checking for the error reason `RAN TOO LONG` ([example](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/runtime_job.py#L221)) - this is being changed on the server side so instead we should check for the error code, 1305.

Wait until ntc issue 2790 is complete

kt474:
> Unblocked by ntc/3494

I think we'll have to look into this again on the server side I'm not getting any error reason code when the error is `RAN TOO LONG` 

```python
'state': {'status': 'Cancelled', 'reason': 'Ran too long', 'reasonCode': None}, 'status': 'Cancelled - Ran too long'
```

PR #1494, opened 2024-03-07
Allow new job status' in RuntimeJobV2

### LINKED ISSUE #1476: New RuntimeJobV2 doesn't allow new statuses
**Describe the bug**

The new `RuntimeJobV2` was meant to return job status in string format, so that even down level provider can tolerate new statues in the future (see #1366). But it's using a [hard coded status mapping](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/runtime_job_v2.py#L188), meaning it'd blow up if a new status is introduced by the server today. 

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

Return raw server status response if it can't be found in the mapping. 

**Additional Information**

- **Python version**:
- **Operating system**:

kt474:
### Details and comments
Fixes #1476


qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -181,10 +181,13 @@ def _status_from_job_response(self, response: Dict) -> JobStatus:
...
         """
-        mapped_job_status = API_TO_JOB_STATUS[response["state"]["status"].upper()]
-        if mapped_job_status == "CANCELLED" and self._reason == "RAN TOO LONG":
-            mapped_job_status = "ERROR"
-        return mapped_job_status
+        api_status = response["state"]["status"].upper()
+        if api_status in API_TO_JOB_STATUS:
+            mapped_job_status = API_TO_JOB_STATUS[api_status]
+            if mapped_job_status == "CANCELLED" and self._reason == "RAN TOO LONG":
+                mapped_job_status = "ERROR"
```
ptristan3, line 188: Maybe we could define constants

ptristan3 (approved):
I left a minor comment/suggestion

qiskit_ibm_runtime/runtime_job_v2.py
kt474: Yeah we have an old issue - https://github.com/Qiskit/qiskit-ibm-runtime/issues/989 to avoid hard coding these job statuses. I'll follow up separately, I believe the server side is not returning an error code

PR #1503, opened 2024-03-11
Fix test_estimator_v2

kt474:
### Summary

Integration tests have been failing for a couple days - https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/8227453042/job/22495383437

I'm not sure if something on the server side changed to cause this test to start to fail? Was this expected? 

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_estimator_v2.py
+++ b/test/integration/test_estimator_v2.py
@@ -60,11 +60,11 @@ class TestEstimatorV2(IBMIntegrationTestCase):
 
             job2 = estimator.run([(psi1, [H1, H3], [theta1, theta3]), (psi2, H2, theta2)])
             result2 = job2.result()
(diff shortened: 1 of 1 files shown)


test/integration/test_backend_serialization.py
```diff
@@ -76,7 +76,8 @@ def test_backend_properties(self, service):
         for backend in backends:
             with self.subTest(backend=backend):
                 properties = backend.properties()
-                self._verify_data(properties.to_dict(), good_keys)
+                if properties:
+                    self._verify_data(properties.to_dict(), good_keys)
```
kt474, line 80: Adding this check here because `fake_backend1` is no longer a simulator and causes this line to error

PR #1524, opened 2024-03-16
Return correct job version 

### LINKED ISSUE #1471: service.job() doesn't know about job version
**Describe the bug**

If I submit a primitive V2 job then do `service.job()`, it will return a V1 `RuntimeJob` instead of `RuntimeJobV2`. This applies to `service.jobs()` as well. 

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

Unfortunately there is no good solution since the version number is only in inputs and results, which means potentially download the entire circuit inputs. 

**Additional Information**

- **qiskit-ibm-runtime version**: main
- **Python version**:
- **Operating system**:

kt474:
### Summary

~~WIP, needs ntc issue 3938, iqp-api 1184~~

~~Need to confirm if we want to retrieve `version` from the job params, doing so will make retrieving jobs significantly more slow~~

### Details and comments
Fixes #1471


kt474:
This is ready for review - I tested on cloud but the IQP `jobs()` [endpoint](https://runtime-us-east.quantum-computing.ibm.com/openapi#/Jobs/list_jobs) does not return job params, even if `exclude_params` is set to false. 

Following up on server side

ptristan3 (approved):
LGTM

PR #1532, opened 2024-03-18
Cherry pick 0.21.2 release notes back to main

kt474:
### Details and comments
Fixes #


Eric-Arellano (approved):
Nice. This is expected btw that 1527.bug.rst is deleted. That's so that the release note doesn't show up in both 0.21.2 and 0.22.0

PR #1538, opened 2024-03-20
Pretty print options

### LINKED ISSUE #1490: Implement an `_html_repr_` for the option classes

The printout of an options class is hard to read. If we implemented `_html_repr_` to return an HTML table, it would render nicely in environments that support this method, like jupyter notebooks. When I looked into it briefly, it seemed easier than trying to modify the `__str__` or the `__repr__` because dataclass creation seems to eagerly overwrite such definitions from a base.

**Acceptance criteria**

kt474:
### Summary

Added a simple `_repr_html_` method to format `Options` and `OptionsV2` if they are printed in a jupyter/vscode cell. 

Screenshot 2024-03-26 at 2 13 28 PM

### Details and comments
Fixes #1490


ptristan3 (approved):
LGTM.

PR #1558, opened 2024-03-25
Fix `BackendSamplerV2` when used with an empty classical register

garrison:
### Summary

This is a port of https://github.com/Qiskit/qiskit/pull/12055 to this repository, since the current repository has its own fork of `BackendSamplerV2` until Qiskit 1.1 is released (#1523).

### Details and comments
I would appreciate if this fix could be backported to the stable branch as well.


kt474 (approved):
Thanks! I can cherry pick this into the stable branch

PR #1562, opened 2024-03-26
Update v2 primitive unit tests

### LINKED ISSUE #1540: Incomplete unit tests

Some unit tests call `np.allclose` without `self.assertTrue`. But it does not raise anything even if args are not equal.
We should use `np.testing.assert_allclose` instead.

E.g., https://github.com/Qiskit/qiskit-ibm-runtime/blob/5625e98c4c7a6e215c252aca72c2f871fed9bc17/test/unit/test_estimator.py#L94

It is also better to replace `np.assertTrue(np.allclose(...))` with `np.testing.assert_allclose` because the latter shows the details of the error.

E.g., https://github.com/Qiskit/qiskit-ibm-runtime/blob/5625e98c4c7a6e215c252aca72c2f871fed9bc17/test/integration/test_estimator.py#L189

**Acceptance criteria**

### LINKED ISSUE #1543: Check skipped tests like TestSamplerV2
While working on a PR, I noticed `TestSamplerV2` tests are [skipped](https://github.com/Qiskit/qiskit-ibm-runtime/blob/5625e98c4c7a6e215c252aca72c2f871fed9bc17/test/unit/test_sampler.py#L54) "until SamplerV2 is supported". SamplerV2 is now supported so the tests should be re-enabled (and like need some work).

It'd also be good to check all the skipped tests to see if some can be re-enabled or deleted.

kt474:
### Details and comments
Fixes #1540 
Fixes #1543

### DIFF
--- a/test/unit/test_sampler.py
+++ b/test/unit/test_sampler.py
@@ -80,7 +80,7 @@ class TestSamplerV2(IBMTestCase):
             self.assertEqual(a_pub_param.circuit, an_in_taks[0])
             # Check parameter values
             an_input_params = an_in_taks[1] if len(an_in_taks) == 2 else []
-            np.allclose(a_pub_param.parameter_values.vals, an_input_params)
+            np.testing.assert_allclose(a_pub_param.parameter_values.vals, an_input_params)
 
     @data(
         {"optimization_level": 4}, {"resilience_level": 1}, {"resilience": {"zne_mitigation": True}}
--- a/test/unit/test_sampler.py
+++ b/test/unit/test_sampler.py
@@ -23,7 +23,7 @@ from qiskit.circuit.library import RealAmplitudes
 from qiskit_ibm_runtime import Sampler, Session, SamplerV2, SamplerOptions
 
 from ..ibm_test_case import IBMTestCase
-from ..utils import bell, MockSession, dict_paritally_equal, get_mocked_backend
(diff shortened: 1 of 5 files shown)


test/unit/test_sampler.py
```diff
@@ -98,32 +98,32 @@ def test_unsupported_values_for_sampler_options(self, opt):
...
+                SamplerOptions(  # pylint: disable=unexpected-keyword-arg
+                    dynamical_decoupling={"sequence_type": "XX"}
+                ),
+                {"dynamical_decoupling": {"sequence_type": "XX"}},
             ),
             (
-                SamplerOptions(optimization_level=3),  # pylint: disable=unexpected-keyword-arg
-                {"transpilation": {"optimization_level": 3}},
+                SamplerOptions(default_shots=1000),  # pylint: disable=unexpected-keyword-arg
+                {"default_shots": 1000},
```
ptristan3, line 111: The pylint disabling is necessary? `dynamical_decoupling` and `default_shots` are valid options.

ptristan3 (approved):
Left a minor comment/doubt. LGTM

test/unit/test_sampler.py
kt474: I believe there's an incompatibility between pylint and pydantic causing this lint error.

PR #1565, opened 2024-03-26
Add delay instruction without pulse defaults

### LINKED ISSUE #1564: Delay instruction is not added to backends without pulse defaults
If the backend has no `PulseDefaults`, the target is built without adding a `Delay` instruction here:
https://github.com/Qiskit/qiskit-ibm-runtime/blob/ba5bd69e4e4856f28fd6a0beb129bd1c7eeb1dc5/qiskit_ibm_runtime/utils/backend_converter.py#L164-L168

dieris:
Fixes #1564

### DIFF
--- a/qiskit_ibm_runtime/utils/backend_converter.py
+++ b/qiskit_ibm_runtime/utils/backend_converter.py
@@ -65,8 +65,10 @@ def convert_to_target(
     }
     custom_gates = {}
     target = None
+    faulty_qubits = set()
     # Parse from properties if it exists
     if properties is not None:
+        faulty_qubits = set(properties.faulty_qubits())
         qubit_properties = qubit_props_list_from_props(properties=properties)
         target = Target(num_qubits=configuration.n_qubits, qubit_properties=qubit_properties)
         # Parse instructions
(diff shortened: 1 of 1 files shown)


kt474 (approved):

qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -161,11 +160,11 @@ def convert_to_target(
                         if any(qubit in faulty_qubits for qubit in qarg):
                             continue
                         target[inst][qarg].calibration = sched
-        if "delay" not in target:
-            target.add_instruction(
-                Delay(Parameter("t")),
-                {(bit,): None for bit in range(target.num_qubits) if bit not in faulty_qubits},
-            )
+    if "delay" not in target:
```
kt474, line 163: If `target` is still `None` here won't there be an error?

dieris: `target` should be constructed in either case, whether `properties` are `None` or not. Do you see a case where `target` is still undefined? We may need to add a couple of tests here

kt474: Ah I see, you're right I got confused with the identations - originally the indentation was added because `faulty_qubits` could be undefined.

This LGTM

kt474 (approved):

PR #1568, opened 2024-03-28
Use status() for related methods

### LINKED ISSUE #1547: job.in_final_state() does not fetch status
**Describe the bug**

`RuntimeJob.in_final_state()` does not report correctly until you force the job to fetch its status through another method such as `.status()`.

**Steps to reproduce**

```python
>>> from qiskit_ibm_runtime import QiskitRuntimeService
>>> # Choose completed job
>>> job = QiskitRuntimeService().job("cqy1mxvhepxg0081acqg")
>>> job.in_final_state()
False

>>> job.status()
>>> job.in_final_state()
True
```

**Expected behavior**

Calling `in_final_state` should fetch the most recent status.

**Suggested solutions**

Replace `self._status` with `self.status()` in the method:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/503e76ed8f4805d74bef6be33aef0c63165149e0/qiskit_ibm_runtime/runtime_job_v2.py#L201-L203

**Additional Information**

- **qiskit-ibm-runtime version**: 0.21.2
- **Python version**: 3.11
- **Operating system**: OSX

kt474:
### Details and comments
Fixes #1547

### DIFF
--- a/qiskit_ibm_runtime/base_runtime_job.py
+++ b/qiskit_ibm_runtime/base_runtime_job.py
@@ -13,7 +13,7 @@
 """Base runtime job class."""
 
 from abc import ABC, abstractmethod
-from typing import Any, Optional, Callable, Dict, Type, Union, Sequence, List
+from typing import Any, Optional, Callable, Dict, Type, Union, Sequence, List, Tuple
 import json
 import logging
 from concurrent import futures
@@ -54,6 +54,8 @@ class BaseRuntimeJob(ABC):
     """Used to inform streaming to stop."""
 
     _executor = futures.ThreadPoolExecutor(thread_name_prefix="runtime_job")
+
+    JOB_FINAL_STATES: Tuple[Any, ...] = ()
 
     def __init__(
         self,
@@ -220,7 +222,7 @@ class BaseRuntimeJob(ABC):
 
     def _set_status_and_error_message(self) -> None:
         """Fetch and set status and error message."""
-        if not self.in_final_state():
+        if self._status not in self.JOB_FINAL_STATES:
             response = self._api_client.job_get(job_id=self.job_id())
             self._set_status(response)
             self._set_error_message(response)
--- a/qiskit_ibm_runtime/base_runtime_job.py
+++ b/qiskit_ibm_runtime/base_runtime_job.py
@@ -56,6 +56,7 @@ class BaseRuntimeJob(ABC):
     _executor = futures.ThreadPoolExecutor(thread_name_prefix="runtime_job")
 
     JOB_FINAL_STATES: Tuple[Any, ...] = ()
+    ERROR: Union[str, RuntimeJobStatus] = None
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -196,11 +199,11 @@ def status(self) -> JobStatus:
 
     def in_final_state(self) -> bool:
         """Return whether the job is in a final job state such as ``DONE`` or ``ERROR``."""
-        return self._status in JOB_FINAL_STATES
+        return self.status() in JOB_FINAL_STATES
```
gadial, line 202: Should it be `self.JOB_FINAL_STATES`?

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -196,11 +199,11 @@ def status(self) -> JobStatus:
...
 
     def errored(self) -> bool:
         """Return whether the job has failed."""
-        return self._status == JobStatus.ERROR
+        return self.status() == JobStatus.ERROR
```
gadial, line 206: Should it be `self.JobStatus.ERROR`?

qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -220,7 +223,7 @@ def error_message(self) -> Optional[str]:
 
     def _set_status_and_error_message(self) -> None:
         """Fetch and set status and error message."""
-        if not self.in_final_state():
+        if self._status not in self.JOB_FINAL_STATES:
```
gadial, line 226: not sure why this change is needed, although I assume there is a good reason.

qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -255,7 +258,7 @@ def _set_error_message(self, job_response: Dict) -> None:
         Args:
             job_response: Job response from runtime API.
         """
-        if self.errored():
+        if self._status == self.ERROR:
```
gadial, line 261: Same as above, I'm not sure why this change is required.

gadial (commented):
Generally looks good, left some minor comments.

qiskit_ibm_runtime/base_runtime_job.py
kt474: It has to be done this way because `_set_stauts_and_error_message` is called in `status` so calling `in_final_state` would result in an infinite loop

qiskit_ibm_runtime/base_runtime_job.py
kt474: Same reasoning as above, it would result in an infinite loop

qiskit_ibm_runtime/runtime_job.py
```diff
@@ -196,11 +199,11 @@ def status(self) -> JobStatus:
 
     def in_final_state(self) -> bool:
         """Return whether the job is in a final job state such as ``DONE`` or ``ERROR``."""
-        return self._status in JOB_FINAL_STATES
+        return self.status() in self.JOB_FINAL_STATES
```
frankharkins, line 202: Would it be better in these cases to reuse `job.in_final_state()`?

frankharkins (approved):
Thanks! Tested and it works correctly 🙂

qiskit_ibm_runtime/runtime_job.py
kt474: We can't here because `in_final_state()` calls `status()` which would result in an infinite loop

PR #1575, opened 2024-04-01
Fix incorrect specification of noise_gain option

chriseclectic:
### Summary

The current description and validation of the `PecOption.noise_gain` option is incorrect.

### Details and comments

Values of 0 should be allowed, and correspond to full PEC mitigation of the learned noise model.
A Value of 1 corresponds to no mitigation (ie running the circuit with its native noise model)


ihincks (approved):
Thanks @chriseclectic

PR #1576, opened 2024-04-02
Fix wording in error raised when validating options

mberna:
Fix wording in error raised when using a simulator, pec_mitigation is enabled but a coupling map has not been provided.

### Details and comments
Fixes #


kt474 (approved):
Remember to create branches off of your fork instead of directly from qiskit/main

Issue #1554, opened 2024-03-25
Dynamic circuit & meas level 1 support


jyu00:
**What is the expected feature or enhancement?**

Qiskit Runtime primitives will soon support dynamic circuits and measurement level 1 in SamplerV2. qiskit-ibm-runtime will need to make corresponding changes for that. 

**Acceptance criteria**

- [x]  Document that dynamic circuits are supported in SamplerV2 without DD
- [x]  Raise an error if dynamic circuits are passed with DD
- [x]  Separate execution options for SamplerV2 that includes measurement level and return (under `meas_type`)

ihincks:
The two options in the third bullet will actually be lumped into one option called `meas_type` as follows:

```python
class SamplerExecutionOptions(ExecutionOptions):
    r"""Extension of :class:~.ExecutionOptions` for the sampler primitive."""

    meas_type: Literal["classified", "kerneled", "avg_kerneled"] = "classified"
    r"""How to process and return measurement results.
    
    This option sets the return type of all classical registers in all :class:`~.PubResult`\s.
    If a sampler pub with shape ``pub_shape`` has a circuit that contains a classical register 
    with size ``creg_size``, then the returned data associated with this register will have one of
    the following formats depending on the value of this option.

    * ``"classified"``: A :class:`~.BitArray` of shape ``pub_shape`` over ``num_shots`` with a 
      number of bits equal to ``creg_size``.

    * ``"kerneled"``: A complex NumPy array of shape ``(*pub_shape, num_shots, creg_size)``, where
      each entry represents an IQ data point (resulting from kerneling the measurement trace) in 
      arbitrary units.

    * ``"avg_kerneled"``: A complex NumPy array of shape ``(*pub_shape, creg_size)``, where
      each entry represents an IQ data point (resulting from kerneling the measurement trace and 
      averaging over shots) in arbitrary units. This option is equivalent to selecting
      ``"kerneled"`` and then averaging over the shots axis, but requires less data bandwidth.

    See https://pubs.aip.org/aip/rsi/article/88/10/104703/836456 for a description of kerneling.
    """
```

This can be further word-smithed. We could also add a note about how "kerneled" is essentially (meas_level=1, meas_return=avg), "avg_kerneled" is (meas_level=1, meas=single), and "classified" is meas_level 2.

jyu00:
Closing since both PRs are merged.

PR #1594, opened 2024-04-09
Level 1 measurement support in Sampler options

gadial:
### Summary
This PR adds support for level 1 measurement options in `SamplerV2`

### Details and comments
This PR extends `ExecutionOptionsV2` to a new class, `SamplerExecutionOptionsV2` having an additional field `meas_type` which can accept the values `"classified", "kerneled", "avg_kerneled"`, where "classified" represents the usual measurement level 2 output, and the other options represent measurement level 1 output.

Fixes part of #1554

### DIFF
--- a/qiskit_ibm_runtime/options/sampler_execution_options.py
+++ b/qiskit_ibm_runtime/options/sampler_execution_options.py
@@ -12,7 +12,7 @@
 
 """Sampler Execution options."""
 from typing import Literal, Union
-from . import ExecutionOptionsV2
+from .execution_options import ExecutionOptionsV2
 from .utils import primitive_dataclass, Unset, UnsetType
 
 
@@ -40,5 +40,6 @@ class SamplerExecutionOptionsV2(ExecutionOptionsV2):
       averaging over shots) in arbitrary units. This option is equivalent to selecting
       ``"kerneled"`` and then averaging over the shots axis, but requires less data bandwidth.
 
-    See `Here <https://pubs.aip.org/aip/rsi/article/88/10/104703/836456>`_ for a description of kerneling.
+    See `Here <https://pubs.aip.org/aip/rsi/article/88/10/104703/836456>`_ for 
+    a description of kerneling.
     """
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/options/sampler_execution_options.py
```diff
@@ -0,0 +1,44 @@
...
+    * ``"kerneled"``: A complex NumPy array of shape ``(*pub_shape, num_shots, creg_size)``, where
+      each entry represents an IQ data point (resulting from kerneling the measurement trace) in 
+      arbitrary units.
+
+    * ``"avg_kerneled"``: A complex NumPy array of shape ``(*pub_shape, creg_size)``, where
+      each entry represents an IQ data point (resulting from kerneling the measurement trace and 
+      averaging over shots) in arbitrary units. This option is equivalent to selecting
+      ``"kerneled"`` and then averaging over the shots axis, but requires less data bandwidth.
+
+    See https://pubs.aip.org/aip/rsi/article/88/10/104703/836456 for a description of kerneling.
```
ihincks, line 43: please turn this into an ReST link https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages

ihincks (commented):
Thanks @gadial , one minore comment.

More generally, is the testing here in line with the standards of this library? I have an `IQBackendWrapper < Backend` class implement elsewhere if we want to use it.

gadial:
> More generally, is the testing here in line with the standards of this library? I have an `IQBackendWrapper < Backend` class implement elsewhere if we want to use it.

I added relevant tests for this PR to the existing ones; there are no extensive tests as far as I can see. However, Such a wrapper sounds like a good idea and I can add more exhaustive tests around it.

qiskit_ibm_runtime/options/sampler_execution_options.py
gadial: Done

ihincks (approved):

Issue #1570, opened 2024-03-29
FakeSherbrooke is missing reset in its properties


jyu00:
**Describe the bug**

FakeSherbrooke properties dictionary is missing `reset` in its list of gates. Consequently, a circuit with a `reset` will not pass the ISA validation. The real ibm_sherbrooke has `reset`.

We should also check all the other fake backends (that still have a real counterpart) to see if their `target.operation_names` match the real ones.

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**:
- **Python version**:
- **Operating system**:

gregbyrd:
Yes, this appears to be a problem with all of the V2 backends. I added a comment to #975 yesterday rather than opening up a new issue. I saw the problem with FakeAlmadenV2 and with FakeCusco.

nonhermitian:
It would be nice if this can be prioritized a bit as the lack of reset support is breaking a lot of our internal benchmarking codes

jyu00:
@nonhermitian Yes this is planned for this sprint (next 2 weeks)

kt474:
Fixed by #1624 #1600

PR #1600, opened 2024-04-11
Replace `convert_to_target` with the Qiskit version

### LINKED ISSUE #1590: Measurement instruction calibration bug
**Describe the bug**

IBM real backend reports wrong calibration data for measure $0. Also calibration data immediately is parsed into the Qiskit `Schedule` form with heavy computational cost although user doesn't need calibration at all. Lazy data loading mechanism was implemented but likely accidentally disabled.

**Steps to reproduce**

```python

service = QiskitRuntimeService(channel="ibm_quantum")
backend = service.get_backend("ibm_sherbrooke")
type(backend.target["measure"][(0,)]._calibration)  # qiskit.pulse.calibration_entries.ScheduleDef
len(backend.target["measure"][(0,)].calibration.instructions)  # 381 -- this is not correct
len(backend.target["measure"][(1,)].calibration.instructions)  # 129

from qiskit_ibm_runtime.fake_provider import FakeSherbrooke

backend = FakeSherbrooke()
type(backend.target["measure"][(0,)]._calibration)  # qiskit.pulse.calibration_entries.PulseQobjDef
len(backend.target["measure"][(0,)].calibration.instructions)  # 129
len(backend.target["measure"][(1,)].calibration.instructions)  # 129
```

**Expected behavior**

Do not report wrong calibration. Use `PulseQobjDef` format.

**Suggested solutions**

Fake backends work normally because it uses the Qiskit-builtin backend model converter `qiskit.providers.backend_compat.convert_to_target`.

https://github.com/Qiskit/qiskit-ibm-runtime/blob/833b9d244663aaa5129638ab3ffbf9185bc9c7ad/qiskit_ibm_runtime/fake_provider/fake_backend.py#L181-L183

Real backend should use the same function unless they need special edge case handling.

**Additional Information**

Related issue https://github.com/Qiskit/qiskit/issues/12040#issuecomment-2042417686

- **qiskit-ibm-runtime version**: 0.22.0
- **Python version**: doesn't matter
- **Operating system**: doesn't matter

t-imamichi:
### Summary

This PR  copies `qiskit.providers.backend_compat.convert_to_target` to `utils.backend_converter.convert_to_target` with some modifications for mypy and lint. Particularly, I added `reset` to `required` to ensure `reset` is included in target.

Fixes #1590

### Details and comments

The script #1590 works fine as follows

```python

service = QiskitRuntimeService(channel="ibm_quantum")
backend = service.get_backend("ibm_sherbrooke")
print('ibm_sherbrooke')
print(type(backend.target["measure"][(0,)]._calibration))  # qiskit.pulse.calibration_entries.ScheduleDef
print(len(backend.target["measure"][(0,)].calibration.instructions))  # 381 -- this is not correct
print(len(backend.target["measure"][(1,)].calibration.instructions))  # 129
print()

from qiskit_ibm_runtime.fake_provider import FakeSherbrooke

backend = FakeSherbrooke()
print('fake_sherbrooke')
print(type(backend.target["measure"][(0,)]._calibration))  # qiskit.pulse.calibration_entries.PulseQobjDef
print(len(backend.target["measure"][(0,)].calibration.instructions))  # 129
print(len(backend.target["measure"][(1,)].calibration.instructions))  # 129
```

main
```
ibm_sherbrooke
<class 'qiskit.pulse.calibration_entries.ScheduleDef'>
381
129

fake_sherbrooke
<class 'qiskit.pulse.calibration_entries.PulseQobjDef'>
129
129
```

this PR
```
ibm_sherbrooke
<class 'qiskit.pulse.calibration_entries.PulseQobjDef'>
129
129

fake_sherbrooke
<class 'qiskit.pulse.calibration_entries.PulseQobjDef'>
129
129
```


jyu00:
@t-imamichi we recently had an issue that the `convert_to_target` from Qiskit wasn't picking up `reset`:

```python
from qiskit_ibm_runtime.utils.backend_converter import convert_to_target as qr_c2t
from qiskit.providers.backend_compat import convert_to_target as qk_c2t

qr_target = qr_c2t(backend.configuration(), backend.properties(), backend.defaults())
qk_target = qk_c2t(backend.configuration(), backend.properties(), backend.defaults())

# then qr_target.operation_names gives:
dict_keys(['id', 'rz', 'sx', 'x', 'cx', 'reset', 'measure', 'switch_case', 'if_else', 'for_loop', 'delay'])

# but qk_target.operation_names gives:
dict_keys(['id', 'rz', 'switch_case', 'if_else', 'cx', 'x', 'measure', 'for_loop', 'delay', 'sx'])
```

Presumably there was a reason for qiskit-ibm-runtime to have its own `convert_to_target`, so I'd also consult with @mtreinish before removing it.

mtreinish:
We can look at fixing the issue in Qiskit's converter function; is there an issue open about it already? I assume `reset` is being put in a field in the configuration that's not being looked at for the reset instruction by the qiskit version, or there are conflicting fields and qiskit is looking at one that's missing the reset instruction.

Besides that the other consideration is that we're planning to deprecate the `BackendV1` interface in 1.2.x which will include removing the `BackendConfiguration`, `BackendProperties`, `PulseDefaults`, and by extension this will include dropping `convert_to_target` because there is nothing to convert from. The removal won't be until Qiskit 2.0, but it might be better to just leave this in place since we'll need to add it back before too long.

t-imamichi:
I copied Qiskit `convert_to_target` to qiskit-ibm-runtime with some modifications.
Particularly, I added `reset` to `required` to ensure `reset` is included in target.
https://github.com/Qiskit/qiskit-ibm-runtime/blob/50bfff0cc087a40a989ce4f2b7efbcddb5877765/qiskit_ibm_runtime/utils/backend_converter.py#L63

jyu00:
> We can look at fixing the issue in Qiskit's converter function; is there an issue open about it already?

@mtreinish I'm guessing no if you're not aware. Somehow I thought you were added to the thread where this was discussed. Is it worth opening an issue in Qiskit at this point if it's going to be deprecated in 1.2? 

> Besides that the other consideration is that we're planning to deprecate the BackendV1 interface in 1.2.x which will include removing the BackendConfiguration, BackendProperties, PulseDefaults

Sounds like we'll need to copy those over to qiskit-ibm-runtime, since that's still how we get data from the server.

jyu00:
> I copied Qiskit convert_to_target to qiskit-ibm-runtime with some modifications.

@t-imamichi can you also add more tests? Like maybe copy the ones on faulty qubits from Qiskit (we already have one on control flow). They can just be unit tests using fake backends.

mtreinish:
> > We can look at fixing the issue in Qiskit's converter function; is there an issue open about it already?
> 
> @mtreinish I'm guessing no if you're not aware. Somehow I thought you were added to the thread where this was discussed. Is it worth opening an issue in Qiskit at this point if it's going to be deprecated in 1.2?

I vaguely remember the issue, but I don't remember the specifics. But yeah we still support the converter function for > 1 year so it's good to ensure we get this fixed.

> 
> > Besides that the other consideration is that we're planning to deprecate the BackendV1 interface in 1.2.x which will include removing the BackendConfiguration, BackendProperties, PulseDefaults
> 
> Sounds like we'll need to copy those over to qiskit-ibm-runtime, since that's still how we get data from the server.

Yeah, I would think so. There's no rush until ~August 2025 (that's when the 1.x series goes EoL) but planning for that and taking ownership of the container objects for the api payloads in the medium term is probably prudent.

t-imamichi:
I copied unit tests on faulty qubits from https://github.com/Qiskit/qiskit/blob/main/test/python/providers/test_faulty_backend.py

t-imamichi:
I notice some error occurs due to random seed issue. Need to merge #1616

t-imamichi:
Puala suggested replacing `convert_to_target` for fake backends too https://github.com/t-imamichi/qiskit-ibm-runtime/pull/2. Do you anyone have any suggestion or comment? If no, I will merge it.

kt474:
> Puala suggested replacing `convert_to_target` for fake backends too [t-imamichi#2](https://github.com/t-imamichi/qiskit-ibm-runtime/pull/2). Do you anyone have any suggestion or comment? If no, I will merge it.

No objections, after that, this PR LGTM

qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -12,192 +12,307 @@
...
+from qiskit.circuit.library.standard_gates import get_standard_gate_name_mapping
+from qiskit.circuit.parameter import Parameter
+from qiskit.providers.backend import QubitProperties
+from qiskit.providers.exceptions import BackendPropertyError
+from qiskit.providers.models.backendconfiguration import BackendConfiguration
+from qiskit.providers.models.backendproperties import BackendProperties
+from qiskit.providers.models.pulsedefaults import PulseDefaults
+from qiskit.transpiler.target import InstructionProperties, Target
 
-from ..ibm_qubit_properties import IBMQubitProperties
```
t-imamichi, line 41: qiskit-ibm-runtime has its own qubit property class. I'm wondering whether I can replace it with Qiskit's qubit properties or not.

kt474: I believe this is fine, `qubit_props_list_from_props` is a helper function only used in `convert_to_target`. It's not a breaking change

kt474 (approved):

wshanks:
This is labeled as new feature. Can it not be backported to 0.23.1 if there will be one? Besides fixing #1570 and #1590, it also fixes another issue I am running into (where a backend defines pulse calibrations for a qubit pair that does not have an entry in the gate properties; the old code in 0.23.0 raises a KeyError while the version here skips the missing keys without an exception).

PR #1617, opened 2024-04-15
Allow QASM string as input to V2 primitives

jyu00:
### Summary

As mentioned in this [comment](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1593#discussion_r1564782561), QASM is currently a valid input to `backend.run`. This PR adds QASM string as an input to V2 primitives, so `SamplerV2` can properly replace `backend.run`. 

### Details and comments
Fixes #


ihincks (commented):
If this PR is being introduced to gain feature parity with `backend.run`, I wonder if we could instead replace this PR with a documentation explanation of how to convert qasm strings into `QuantumCircuit`s and pass them to the `SamplerV2`?

It's not clear to me that this feature is so demanded that changing the input type is warranted.

jyu00:
> If this PR is being introduced to gain feature parity with `backend.run`, I wonder if we could instead replace this PR with a documentation explanation of how to convert qasm strings into `QuantumCircuit`s and pass them to the `SamplerV2`?
> 
> It's not clear to me that this feature is so demanded that changing the input type is warranted.

That's a great idea! We'd also not have to rush this PR. I'll close this and work with docs team to update.

PR #1628, opened 2024-04-18
Remove reference to the same file in `options/__init__.py`

arnaucasau:
The PR removes one link from `qiskit_ibm_runtime/options/__init__.py` that was pointing to itself. The link had a strange behavior when used in historical versions.

The PR will need backport to `stable/0.23`.


kt474 (approved):
LGTM, Thanks for taking care of this

PR #1630, opened 2024-04-19
Fix multiple cycles of staggered DD

dieris:
### Summary

Update the `insert_multiple_cycles` option to support staggered DD pulses.


kt474 (approved):
LGTM, but probably best for someone else to review as well

PR #1653, opened 2024-05-02
Update service outage message

### LINKED ISSUE #1647: Return more helpful exception if Quantum Service is in maintenance

In the rare times when the IBM Quantum service is in maintenance, there is no longer any API available to the user, and the site is fully replaced with a maintenance page - html, returns a 200. 

The runtime client should catch these items as a specific exception when expecting JSON, and convert to an exception that the service is in maintenance mode with a link to the page.

**Acceptance criteria**

kt474:
### Details and comments
Fixes #1647


ihincks (approved):
LGTM

PR #1674, opened 2024-05-13
`_write_parameter_expression` takes version parameter 

kt474:
### Summary

@jakelishman `version` was added as a new argument [here](https://github.com/Qiskit/qiskit/pull/12310) which I believe caused our unit [tests](https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/9057469723/job/24881530623#step:5:3952) against Qiskit/main to start failing. Is this the correct way to handle `version` from the provider?

### DIFF
--- a/qiskit_ibm_runtime/utils/json.py
+++ b/qiskit_ibm_runtime/utils/json.py
@@ -61,6 +61,7 @@ from qiskit.qpy import (
     _read_parameter_expression_v3,
     load,
     dump,
+    QPY_VERSION,
 )
 from qiskit.primitives.containers.estimator_pub import EstimatorPub
 from qiskit.primitives.containers.sampler_pub import SamplerPub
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/utils/json.py
```diff
@@ -243,12 +244,21 @@ def default(self, obj: Any) -> Any:  # pylint: disable=arguments-differ
...
-                compress=False,
-                use_symengine=bool(optionals.HAS_SYMENGINE),
-            )
+            if QPY_VERSION >= 12:
+                value = _serialize_and_encode(
+                    data=obj,
+                    serializer=_write_parameter_expression,
+                    compress=False,
+                    use_symengine=bool(optionals.HAS_SYMENGINE),
+                    version=QPY_VERSION,
```
jakelishman, line 253: This version wants to be compatible (less than) whatever version of QPY you'll be using on the backend to deserialise - `QPY_VERSION` will always be the max the local version of Qiskit can handle.

jakelishman (commented):
That's probably the calling convention for it now, yeah, but that's very much not a public API function, and we cannot make guarantees that it'll even continue to exist in the future, given that we may well need to refactor the QPY export to be faster.

qiskit_ibm_runtime/utils/json.py
kt474: Is the use of `QPY_VERSION` incorrect here then? I don't believe the server side has updated to qiskit 1.1 yet

mtreinish: You should repeat the logic in other places around the version kwarg and keep the version there the same as in other places. It might makes sense to extract that to a constant like `RUNTIME_QPY_VERSION = 11` or something.

```
if (_TERRA_VERSION[0] == 1 and _TERRA_VERSION[1] >= 1) or _TERRA_VERSION[0] > 1:
    kwargs["version"] = 11
```

jyu00:
From the Slack discussion I think we can just drop the support for `ParameterExpression` type, since it's not an input type to primitives anymore.

kt474:
> From the Slack discussion I think we can just drop the support for `ParameterExpression` type, since it's not an input type to primitives anymore.

Closing this PR, i'll open a separate one

PR #1679, opened 2024-05-14
Add `resilience.zne.amplifier` to EstimatorV2 options

### LINKED ISSUE #1670: Add ZNE gate folding methods (`random`, `from_front`, `from_back`, etc.) as options
Allow users to choose from different ZNE gate folding methods.
For fractional noise factors, there are a several folding strategies available in the literature. Suppose, we have circuit with a total of $n$ (2-qubit) gates and need to fold $k$ gates to achieve a fractional noise amplification factor. Then, we can fold using following methods:

- **random**: randomly fold any $k$ gates.
- **from first/front**: fold first $k$ gates.
- **from last/back**: fold last $k$ gates.

**Acceptance criteria**
Maybe, a new field in [`ZneOptions`](https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.ZneOptions) to allow users select their choice of folding method.

ihincks:
### Summary

This PR adds the `ZneOptions` attribute `amplifier` which can be one of these strings:

* `"gate_folding"` (default) uses 2-qubit gate folding to amplify noise. If the noise
  factor requires amplifying only a subset of the gates, then these gates are chosen
  randomly.
* `"gate_folding_front"` uses 2-qubit gate folding to amplify noise. If the noise
  factor requires amplifying only a subset of the gates, then these gates are selected
  from the front of the topologically ordered DAG circuit.
* `"gate_folding_back"` uses 2-qubit gate folding to amplify noise. If the noise
  factor requires amplifying only a subset of the gates, then these gates are selected
  from the back of the topologically ordered DAG circuit.

### Details and comments
Fixes #1670

### DIFF
--- a/qiskit_ibm_runtime/options/zne_options.py
+++ b/qiskit_ibm_runtime/options/zne_options.py
@@ -37,6 +37,18 @@ class ZneOptions:
     """Zero noise extrapolation mitigation options.
 
     Args:
+        amplifier: Which technique to use for amplifying noise. One of:
+
+            * `"gate_folding"` (default) uses 2-qubit gate folding to amplify noise. If the noise
+              factor requires amplifying only a subset of the gates, then these gates are chosen
+              randomly.
+            * `"gate_folding_front"` uses 2-qubit gate folding to amplify noise. If the noise
+              factor requires amplifying only a subset of the gates, then these gates are selected
+              from the front of the topologically ordered DAG circuit.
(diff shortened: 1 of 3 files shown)


kt474 (approved):

jyu00:
We'll need to add this to the schema as well, cc @gadial

PR #1682, opened 2024-05-15
Fix 0-d zne options encoding bug

kt474:
### Summary

Need to wait for server side changes to fully test. I'm still a bit confused about this bug - 

```python 

if isinstance(obj, np.ndarray):
    if obj.dtype == object:
        ...
    elif obj.ndim: # this value is 1? Do we need to check for obj.size instead
        ...
```

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/utils/json.py
+++ b/qiskit_ibm_runtime/utils/json.py
@@ -214,6 +214,8 @@ class RuntimeEncoder(json.JSONEncoder):
         if isinstance(obj, np.ndarray):
             if obj.dtype == object:
                 return {"__type__": "ndarray", "__value__": obj.tolist()}
+            elif obj.size == 0:
+                return obj.tolist()
             value = _serialize_and_encode(obj, np.save, allow_pickle=False)
             return {"__type__": "ndarray", "__value__": value}
         if isinstance(obj, np.int64):


qiskit_ibm_runtime/utils/json.py
```diff
@@ -214,6 +214,8 @@ def default(self, obj: Any) -> Any:  # pylint: disable=arguments-differ
         if isinstance(obj, np.ndarray):
             if obj.dtype == object:
                 return {"__type__": "ndarray", "__value__": obj.tolist()}
+            elif obj.size == 0:
```
ihincks, line 217: The problem with using size here is that `np.array(3.14)` and `np.array([3.14])` and `np.array([[3.14]])` all have size 1. But they have respective `.shape`s `()`, `(1,)`, and `(1,1)`. I believe it's only the case `()` that's causing us problems, which is why I've been checking `obj.ndim`, which is just `len(obj.shape)`.

What's happening is that `(x:=np.array("abc", dtype=object)).tolist()` is equal to "abc", so that `x` ends up getting serialized to `{"__type__": "ndarray", "__value__": "abc"}` which the decoder doesn't know how to handle. (it tries to b64 unpack "abc", just like all the other ndarrays).

kt474: Do we have to wait for the server side changes to be deployed? I was confused because the test failing [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/integration/test_estimator_v2.py#L70) seems to have a  `obj.ndim` value of 1 for the extrapolator value.

kt474: I think the issue here is still the metadata serialization on the server side (from pec-runtime 991). The client side does not handle the serialization of the metadata. I've modified this PR to handle this case from the client side but I think the encoder on the server side needs to be patched again.

ihincks: You're right. @ptristan3 discovered that the patch on the server side was put on the wrong encoder class. We'll be issuing a "real" fix.

I think we still want this PR, though, right? Because we don't want to implement this as a server-side patch except for backwards compat.

kt474: Yeah we should still get this PR in

ptristan3 (approved):

PR #1689, opened 2024-05-17
Remove Provider as parent class from QiskitRuntimeService

### LINKED ISSUE #1261: Replace get_backend with backend in the README

### LINKED ISSUE #1589: Remove Provider as a parent class

Qiskit is deprecating the `Provider` interface (see https://github.com/Qiskit/qiskit/pull/12145). `QiskitRuntimeService` still inherits from `Provider`, and it wouldn't be a very good experience if everyone starts getting a deprecation warning every time they use Qiskit Runtime until the release of 2.0, especially since there is nothing they can do about it. 

`Provider` today only has 3 methods:
- `get_backend()` - we can deprecate this in favor of `backend()`
- `backends()` - we already overwrite this
- `__eq__()` - we should define this method with a proper comparison (e.g. including channel, credential, instance, and channel_strategy) 

**Acceptance criteria**

- Stop inheriting from `Provider`
- Update the Provider related methods as described above
- Update docstrings as needed, related to https://github.com/Qiskit/qiskit-ibm-runtime/issues/1261
- If possible, issue a deprecation warning if `isinstance(QiskitRuntimeService, Provider)` is used. 

cc @1ucian0

kt474:
### Details and comments
Fixes #1589 
Fixes #1261

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -21,7 +21,6 @@ from collections import OrderedDict
 from typing import Dict, Callable, Optional, Union, List, Any, Type, Sequence
 
 from qiskit.providers.backend import BackendV2 as Backend
-from qiskit.providers.provider import ProviderV1 as Provider
 from qiskit.providers.exceptions import QiskitBackendNotFoundError
 from qiskit.providers.providerutils import filter_backends
 from qiskit.providers.models import (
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/transpiler/passes/scheduling/__init__.py
```diff
@@ -56,12 +56,12 @@
...
     from qiskit_ibm_runtime.transpiler.passes.scheduling import PadDelay
-    from qiskit_ibm_runtime.fake_provider import FakeJakarta
+    from qiskit_ibm_runtime.fake_provider import FakeJakartaV2
 
-    backend = FakeJakarta()
+    backend = FakeJakartaV2()
 
     # Temporary workaround for mock backends. For real backends this is not required.
-    backend.configuration().basis_gates.append("if_else")
+    backend._conf_dict['basis_gates'].append('if_else')
```
ptristan3, line 64: This hack could be deleted when merging the [PR](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1710).

PR #1705, opened 2024-05-22
Add `private_endpoint` param when saving account

### LINKED ISSUE #1699: When "url" param is provided in QiskitRuntimeService do not modify it.
When I save an account information like this for `ibm_cloud` channel:

```python
QiskitRuntimeService.save_account(
    channel="ibm_cloud",
    token="<TOKEN>",
    instance="<INSTANCE>",
    name="<NAME>",
    url="https://a.domain.example/",
    overwrite=True
)
```
the final url used is the addition of a prefix based on the region of the CRN and the IBM subdomain, plus the `url` passed as parameter. This limits the use of alternative urls to send jobs.

**Acceptance criteria**
If the `url` param is provided in the creation of a `QiskitRuntimeService` it should connect to that url unchanged. If the url param is not provided, it should build the final url based on the CRN region and IBM domain as it does currently.

kt474:
### Details and comments
Fixes #1699


qiskit_ibm_runtime/accounts/account.py
```diff
@@ -278,9 +282,13 @@ def __init__(
             proxies: Proxy configuration.
             verify: Whether to verify server's TLS certificate.
             channel_strategy: Error mitigation strategy.
+            private_endpoint: Connect to private API URL.
         """
         super().__init__(token, instance, proxies, verify, channel_strategy)
-        resolved_url = url or IBM_CLOUD_API_URL
+        if private_endpoint:
+            resolved_url = PRIVATE_EU_API_URL
```
raulotaolea, line 289: Private endpoint should only add the `private` prefix without the region. All regions can have private endpoints, so it should not be attached to any region by default.

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -30,6 +30,7 @@
 
 IBM_QUANTUM_API_URL = "https://auth.quantum-computing.ibm.com/api"
 IBM_CLOUD_API_URL = "https://cloud.ibm.com"
+PRIVATE_EU_API_URL = "https://private.eu-de.quantum-computing.cloud.ibm.com/"
```
raulotaolea, line 33: Not sure if it is a good idea to have a constant per region. I think it would work better the current approach, where you build the url depending on the region coming in the CRN. After that, if the `private_endpoint` param is provided add `private` subdomain at the very beginning of the url. In summary, the pattern is `https://<private>.<region>.<subdomain>.cloud.ibm.com` where:
   - `<private>` is only added if `private_endpoint` is `True`.
   - `<region>` is extracted from the CRN.

raulotaolea (changes requested):

raulotaolea (commented):
This covers the case when a user provides `private_endpoint` param. But if the user provides an `url` param, it will be changed, right? I mean that this doesn't cover that case described in the [issue](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1699#issuecomment-2127586329)

kt474:
> This covers the case when a user provides `private_endpoint` param. But if the user provides an `url` param, it will be changed, right? I mean that this doesn't cover that case described in the [issue](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1699#issuecomment-2127586329). Or do you prefer to do it in a different PR?

Wouldn't this be a breaking change? Then, even if the correct URL is entered (https://cloud.ibm.com/) then the region and sub domain would not be added

raulotaolea:
> > This covers the case when a user provides `private_endpoint` param. But if the user provides an `url` param, it will be changed, right? I mean that this doesn't cover that case described in the [issue](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1699#issuecomment-2127586329). Or do you prefer to do it in a different PR?
> 
> Wouldn't this be a breaking change? When the correct default URL is passed in (https://cloud.ibm.com/) in `save_account` the region and sub domain would not be added.
> 
> For example, when users save a staging account, the url needs to be passed in (https://test.cloud.ibm.com).

I guess that `url` is only used by internal teams, so we can change the behavior of it from '_the subdomain that will be completed with the region and private endpoints subdomain_' to '_use the url as provided without changing it_'. That said, if this affects to e2e tests of other teams, or it needs more discussion, we can treat it independently. With the `private_endpoint` we already fulfill the VPEs use case.

qiskit_ibm_runtime/accounts/management.py
```diff
@@ -61,6 +62,7 @@ def save(
             proxies=proxies,
             verify=verify,
             channel_strategy=channel_strategy,
+            private_endpoint=private_endpoint if not url else False,
```
raulotaolea, line 65: Ignoring `private_endpoint` when `url` param is provided only makes sense if the `url` is kept unchanged. If we change the `url` with the location, then we have to always take into account the `private_endpoint` param. So the condition is not necessary.

test/unit/test_account.py
```diff
@@ -405,6 +405,40 @@ def test_default_env_channel(self):
...
+            instance=_TEST_IBM_CLOUD_ACCOUNT.instance,
+            url=_TEST_IBM_CLOUD_ACCOUNT.url,
+            channel="ibm_cloud",
+            overwrite=True,
+            set_as_default=True,
+            private_endpoint=True,
+        )
+
+        account = AccountManager.get(filename=_TEST_FILENAME)
+        self.assertFalse(account.private_endpoint)
```
raulotaolea, line 440: If we are changing the `url` with the location, the `private_endpoint` param must be taken into account, so you can remove this test.

raulotaolea (changes requested):

kt474:
> That said, if this affects to e2e tests of other teams, or it needs more discussion, we can treat it independently. With the `private_endpoint` we already fulfill the VPEs use case.

Yeah lets handle that independently, in a separate PR.

raulotaolea (approved):
LGTM 👍🏼

raulotaolea (approved):
LGTM 👍🏼

Issue #1604, opened 2024-04-11
Deprecate V1 fake backends


jyu00:
**What is the expected feature or enhancement?**

The fake backends in this repo are meant to represent IBM backends and be similar to `IBMBackend`. Since `IBMBackend` is inherits from `BackendV2`, there isn't much point supporting V1. 

**Acceptance criteria**

Deprecate V1 fake backends.

PR #1715, opened 2024-05-30
Fractional gate opt-in

nkanazawa1989:
### Summary

This PR adds the fractional gate opt-in feature. In IBM backend architecture, dynamic circuits and fractional gates will be exclusively supported. However, Qiskit Target model doesn't express such a constraint. This means one must create two target instances, namely, one only with the fractional gate instructions and the other one only with the control flow instructions. The choice of target is the responsibility of the end users. 

This opt-in might be removed without announce when IBM backend architecture changes.

### Details and comments

This PR is draft to conclude the final design.

- API (fixed)

Both `QiskitRuntimeService.backends()` and `QiskitRuntimeService.backend()` gains `use_fractional_gate` argument. User needs to set True to include fractional gates in the backend target.

- Mutability (need help)

In the current implementation, the service instance pre caches all backends associated with the account. This has been reasonable design. However, with the new flag, the same backend will have two instances with/without fractional gates/control flow instructions in their target. Change of the flag will silently mutate the cached backend. For example,

```python
backend_dc = service.backends("some_backend", use_control_flow=False)
backend_fg = service.backends("some_backend", use_control_flow=True)

qc = QuantumCircuit(...)
qc.if_else(...)

isa_qc = transpile(qc, backend_dc)  # This fails because id(backend_dc) == id(backend_fg)
```

The easiest solution is to deepcopy the instance before returning it to users (target is created when the end user calls any attribute from the target for the first time). We don't like deepcopy though. Alternatively we can remove this cache and always call `_create_backend_obj` when the backend is retrieved. This maybe increase REST GET call overhead to get configuration data from server, but makes internal code simpler (and might reduce memory footprint if retrieved backend is GC'd). If the user doesn't retrieve the same backend multiple times there is no overhead with this change.

- Breaking API change (need help)

We set default of `use_fractional_gates` to True because dynamic circuit seems not being heavily used. However this is indeed a breaking API change for end users because it silently mutates basis gates and changes the resulting isa circuits, and maybe fidelity too. We can also start with False, then turn it on with deprecation warning if we want. From the API break viewpoint, another concern is a user can still use qiskit-ibm-runtime before this commit even after our backends start to report fractional basis gates. This will create a target with mixture of fractional and control flow instructions, which will create invalid isa circuits when it includes any control flow. In the current version there is no warning mechanism in the client software, which might frustrate end users using outdated runtime.

### TODO

- [x] Write unittest
- [x] Write release note


nkanazawa1989:
(clue for review)

I added unit tests in three steps:

- [b9ac43b](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1715/commits/b9ac43bfbb74bdad784fd0d3f898a415b0c3182c): Test for convert_to_target function is added. This is the core feature of the PR. I also modified the interface so that we can individually select dynamic and fractional feature. When our backends support both, we can just set True in both arguments without any API change. Note that FakeBackends already do this, because the simulator backend doesn't have constraints. In addition, if we disable dynamic feature, this could break other unit tests (including community package) for dynamic circuits relying on our fake backends.

- [736e060](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1715/commits/736e060ae9072c4663ff3a939f0258364e9becb7): Test for `backends(..., use_fractional_gates=True)` is added. This requires modification for the test backend, so that the fake runtime client can refer to the actual model data of fake backends by backend names (before this commit every backend uses `FakeLima`).

- [7453ce3](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1715/commits/7453ce3b488c867a8228da185a1690cd749358cd): Test to verify Sampler.run can reject un-executable pubs on our backends.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -506,6 +508,15 @@ def backends(
                         filters=lambda b: b.max_shots > 50000)
                     QiskitRuntimeService.backends(
                         filters=lambda x: ("rz" in x.basis_gates )
+            use_fractional_gates: Set True to allow for the backends to include
+                fractional gates in target. Currently this feature cannot be used
+                simulataneously with the dynamic circuits.
```
jyu00, line 513: We should also point out it doesn't work with PEA/PEC. So if a user plans to enable those features, they should transpile with a target that doesn't include fractional gates. 

```suggestion
            use_fractional_gates: Set True to allow for the backends to include
                fractional gates in target. Currently this feature cannot be used
                simulataneously with the dynamic circuits, PEC, or PEA.
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -489,6 +489,8 @@ def backends(
         instance: Optional[str] = None,
         dynamic_circuits: Optional[bool] = None,
         filters: Optional[Callable[[List["ibm_backend.IBMBackend"]], bool]] = None,
+        *,
+        use_fractional_gates: Optional[bool] = True,
```
jyu00, line 493: Can you add this parameter to the `backend()` method as well? I think many people use that instead of `backends()` when they know the name of the backend.

release-notes/unreleased/1715.feat.rst
```diff
@@ -0,0 +1,26 @@
...
+        "name_of_your_backend", 
+        use_fractional_gates=True,
+    )[0]
+
+When the fractional feature is enabled, transpiled circuits may have
+shorter depth compared with the conventional IBM basis gates, e.g. [sx, rz, ecr].
+
+When you use control flow instructions, e.g. ``if_else``, in your circuit,
+you must disable the fractional gate feature to get executable ISA circuits.
+The choice of the instruction set is now responsibility of users.
```
jyu00, line 23: We should also point out that PEA and PEC are also not supported with fractional gates, and that people should use `use_fractional_gates=False` if they plan to use those error mitigation methods in the primitives.

qiskit_ibm_runtime/fake_provider/__init__.py
```diff
@@ -135,6 +135,7 @@
     FakeKyoto
     FakeLagosV2
     FakeLimaV2
+    FakeAltLimaV2
```
jyu00, line 138: Looks like `FakeAltLimaV2` is only created to include fractional gates + control flow. So why not use a more meaningful name, like `FakeFractionalBackend` ?

qiskit_ibm_runtime/fake_provider/__init__.py
```diff
@@ -196,6 +197,7 @@
     FakeKolkata
     FakeLagos
     FakeLima
+    FakeAltLima
```
jyu00, line 200: We've [deprecated](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1604) V1 fake backends, so there is no need for V1.

jyu00 (commented):
Some relatively minor comments, but overall looks good!

nkanazawa1989:
Thanks @jyu00 I updated the PR. Do you have any comment on the rest of two topics in the PR comment?: Mutability and Breaking API change

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -506,6 +508,15 @@ def backends(
...
                         filters=lambda x: ("rz" in x.basis_gates )
+            use_fractional_gates: Set True to allow for the backends to include
+                fractional gates in target. Currently this feature cannot be used
+                simulataneously with the dynamic circuits, PEC, or PEA.
+                When this flag is set, control flow instructions are automatically
+                removed from the backend target.
+                When you use the dynamic circuits feature (e.g. if_else) in your
+                algorithm, you must disable this flag to create executable ISA circuits.
+                This flag might be modified or removed when our backend
+                supports dynamic circuits and fractioanl gates simulataneously.
```
yaelbh, line 519:
```suggestion
                supports dynamic circuits and fractional gates simultaneously.
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -767,6 +788,15 @@ def backend(
...
                 For users without access to a premium provider, the default open provider will be used.
+            use_fractional_gates: Set True to allow for the backends to include
+                fractional gates in target. Currently this feature cannot be used
+                simulataneously with the dynamic circuits, PEC, or PEA.
+                When this flag is set, control flow instructions are automatically
+                removed from the backend target.
+                When you use the dynamic circuits feature (e.g. if_else) in your
+                algorithm, you must disable this flag to create executable ISA circuits.
+                This flag might be modified or removed when our backend
+                supports dynamic circuits and fractioanl gates simulataneously.
```
yaelbh, line 799:
```suggestion
                supports dynamic circuits and fractional gates simultaneously.
```

release-notes/unreleased/1715.feat.rst
```diff
@@ -0,0 +1,28 @@
+New opt-in feature to support fractioanl gates is added to :class:`.IBMBackend`.
```
yaelbh, line 1:
```suggestion
New opt-in feature to support fractional gates is added to :class:`.IBMBackend`.
```

release-notes/unreleased/1715.feat.rst
```diff
@@ -0,0 +1,28 @@
...
+IBM backends currently support dynamic circuits and fractional gates exclusively and
+two features cannot be used in the same primitive job.
```
yaelbh, line 3:
```suggestion
the two features cannot be used in the same primitive job.
```

yaelbh:
Do you want to add checks at the client side that PEC and PEA are not requested along with fractional gates?

jyu00:
> Do you want to add checks at the client side that PEC and PEA are not requested along with fractional gates?

Doing it on the client side has the disadvantage that down level client code would become stale when updates are made on the server side (e.g. if/when we support fractional gates with dynamic circuits). So I was really hoping the validation service will be setup in time, although that's not looking likely. Regardless, I'd highly recommend doing the check in a separate PR, if we don't want to wait on the validation service.

PR #1714, opened 2024-05-29
Prepare release 0.24

kt474:
### Details and comments
Fixes #

PR #1720, opened 2024-05-31
Add private_endpoint to constructor

kt474:
### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -73,6 +73,7 @@ class QiskitRuntimeService:
         proxies: Optional[dict] = None,
         verify: Optional[bool] = None,
         channel_strategy: Optional[str] = None,
+        private_endpoint: Optional[bool] = False,
     ) -> None:
         """QiskitRuntimeService constructor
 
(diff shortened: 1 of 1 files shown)


raulotaolea (approved):
LGTM 👍🏼

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -73,6 +73,7 @@ def __init__(
         proxies: Optional[dict] = None,
         verify: Optional[bool] = None,
         channel_strategy: Optional[str] = None,
+        private_endpoint: Optional[bool] = False,
```
jyu00, line 76: Shouldn't the default be `None`? So I can choose _not_ to connect to private endpoint even if my saved account says otherwise? 

```suggestion
        private_endpoint: Optional[bool] = None,
```

jyu00 (commented):
Can you also add or amend to #1705's release note to say it's an parameter to not just `save_account()` but also initializing the service?

kt474:
> Can you also add or amend to #1705's release note to say it's an parameter to not just `save_account()` but also initializing the service?

I'll update the release note in https://github.com/Qiskit/qiskit-ibm-runtime/pull/1714

PR #1730, opened 2024-06-06
Add client-side validations

gadial:
### Summary
This PR adds a new file, `utils/validations.py` and data validations which can be applied client-side to prevent the user from sending a job that will fail at a later stage.

### Details and comments
The added validations:
1. In `Sampler`, no classical register of size 0 is passed.
2. In `Sampler`, classical register names are valid, non-keyword Python identifiers.
3. In `Estimator`, no empty observable array is passed.


qiskit_ibm_runtime/utils/validations.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Args:
+        pubs: The list of pubs to validate
+
+    Raises:
+        ValueError: If any observable array is of size 0
+    """
+    for pub in pubs:
+        if pub.observables.shape == (0,):
+            raise ValueError("Empty observables array is not allowed")
```
kt474, line 72: We already have some validation methods in `utils/utils.py` do you think they should be moved into this file as well? 

Rest of the PR LGTM

gadial: It seems reasonable. I didn't do it since I didn't want to disturb too much code on my own. I'll do it now.

kt474 (approved):

Issue #1602, opened 2024-04-11
Nested experimental suboptions override non-experimental suboptions


chriseclectic:
**Describe the bug**

If a sub-options field is contained in `experimental` options, it will override any values specified in the main options.

**Steps to reproduce**

This is a minimal example to reproduce, using a non-experimental option initialized as if it were an experimental options:

```python
options = {
    "resilience": {
        "zne_mitigation": True, 
        "zne": {"noise_factors": [1, 1.5, 2]}
    },
}
estimator = EstimatorV2(backend=backend, options=options)
result = estimator.run([pub]).result()
```

We can confirm the noise factors in result metadata
`print(result.metadata["resilience"]["zne"]["noise_factors"])` -> `[1.0, 1.5, 2.0]`

Now if we ad an experimental ZNE option

```python
options = {
    "resilience": {
        "zne_mitigation": True, 
        "zne": {"noise_factors": [1, 1.5, 2]}
    },
    "experimental": {
        "resilience": {"zne": {"extrapolator": ["linear"]}},
    }
}
estimator = EstimatorV2(backend=backend, options=options)
result = estimator.run([pub]).result()
```
The specified noise factors are discarded and the default ones are used:
`print(result.metadata["resilience"]["zne"]["noise_factors"])` -> `[1.0, 3.0, 5.0]`

**Expected behavior**

The second case should merge the experimental ZNE options into the regular ones (and similarly for any other suboptions) without discarding the fields in the original one.

**Suggested solutions**

There is likely a bug in the way experimental options are merged with regular options

**Additional Information**

- **qiskit-ibm-runtime version**: 0.22
- **Python version**: 3.11
- **Operating system**: MacOS

kt474:
fixed in #1731

PR #1731, opened 2024-06-06
Disallow setting flat options

### LINKED ISSUE #1485: Stop allow setting flat option

The primitives today allow setting a nested option by itself. For example, 

```
estimator.options.update(extrapolator="linear")
```

is allowed and would update the "correct" option in `ZneOptions`. However, this only works if all options have unique names, which is no longer true in V2 primitives. So setting something like `num_randomizations` could yield unexpected results. 

**Acceptance criteria**

Only accept nested options in `options.update()`.

jyu00:
### Summary

Deprecate setting options using a flat dictionary. For example, `options={"log_level": "DEBUG"}` is deprecated, and one needs to specify `options={"environment": {"log_level": "DEBUG"}}`.

### Details and comments
Fixes #1485 and #1602


kt474 (approved):
LGTM

PR #1735, opened 2024-06-10
Fix type annotation

EarlMilktea:
### Summary

Fix the annotated type of `filters`.

### Details and comments

In `qiskit.providers.providerutils`, `filters` is used as the first argument of `filter` function against `backends: list[Backends]`.

```python
# 3. Apply acceptor filter.
backends = list(filter(filters, backends))
```

So the annotated type of `filters` must be `Callable[[IBMBackend], bool]` rather than `Callable[[list[IBMBackend]], bool]`.

Note that there are exactly the same lines in `qiskit_ibm_provider`.


kt474 (approved):
Thank you @EarlMilktea!

PR #1740, opened 2024-06-10
Add refresh method to fake backends

### LINKED ISSUE #1707: Allow users to update fake backends with latest data

When we introduced local testing mode, we encouraged people to use fake backends corresponding to the real target backend they will ultimately use. Fake backends were originally designed to be used completely offline, and their data files (such as backend properties) are never updated once created. This means the noise model of a fake backend can be very different from its real counterpart. 

Some users requested that we provide a way to pull the latest backend files, so the noise model would better represent the real systems. 

**Acceptance criteria**

- Add a `refresh()` method to fake backend that'd pull the latest backend data files from its real counterpart, if one still exists. It'd need to take a `service` parameter since we need an account to access the real backend.

ptristan3:
### Details and comments
Fixes #1707


qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -458,6 +462,48 @@ def _get_noise_model_from_backend_v2(  # type: ignore
 
         return noise_model
 
+    def refresh(self, service: QiskitRuntimeService) -> None:
+        """Updated the data files from its real counterpart"""
```
kt474, line 466: It'd be good to explain in the docstring that this method will overwrite the local config/property files

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -458,6 +462,48 @@ def _get_noise_model_from_backend_v2(  # type: ignore
...
+            real_config = real_backend.configuration()
+            real_defs = real_backend.defaults()
+
+            if real_props:
+                new_version = real_props.backend_version
+
+                if new_version > version:
+                    props_path = os.path.join(self.dirname, self.props_filename)
+                    with open(props_path, "w", encoding="utf-8") as fd:
+                        fd.write(json.dumps(real_props.to_dict(), cls=BackendEncoder))
```
kt474, line 483: This overwrites the local files which I think is fine, but some things to keep in mind - 

- The updated config/properties will be maintained through sessions, so once a user updates these files, they will always be updated unless they go manually revert the changes in their local code
- Updating qiskit-ibm-runtime will revert the changes  
- 
If we choose to go down this path, we should make it clear in the docstrings and logs that this is how it works  
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

kt474 (approved):

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -458,6 +462,65 @@ def _get_noise_model_from_backend_v2(  # type: ignore
...
+            real_backend = backends[0]
+
+            real_props = real_backend.properties()
+            real_config = real_backend.configuration()
+            real_defs = real_backend.defaults()
+
+            if real_props:
+                new_version = real_props.backend_version
+
+                if new_version > version:
```
jyu00, line 497: Can we really rely on the `version`? I don't think it gets updated every time new calibration is pushed. And why even check the version if the server side data is already the freshest.

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -458,6 +462,65 @@ def _get_noise_model_from_backend_v2(  # type: ignore
...
+                    logger.info(
+                        "The backend %s has been updated from {version} to %s version.",
+                        self.backend_name,
+                        real_props.backend_version,
+                    )
+                else:
+                    logger.info("There are no available new updates for %s.", self.backend_name)
+
+        except Exception as ex:
+            logger.info("The refreshing of %s has failed: %s", self.backend_name, str(ex))
```
jyu00, line 522: This should raise an exception or at least a warning. Info level log doesn't get surfaced by default.

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -458,6 +462,65 @@ def _get_noise_model_from_backend_v2(  # type: ignore
...
+                    if real_defs:
+                        defs_path = os.path.join(self.dirname, self.defs_filename)
+                        with open(defs_path, "w", encoding="utf-8") as fd:
+                            fd.write(json.dumps(real_defs.to_dict(), cls=BackendEncoder))
+
+                    logger.info(
+                        "The backend %s has been updated from {version} to %s version.",
+                        self.backend_name,
+                        real_props.backend_version,
+                    )
```
jyu00, line 517: This just updates the files but doesn't actually refresh the `target` if it's already built since it's cached.

test/integration/test_fake_backends.py
```diff
@@ -171,3 +171,20 @@ def test_non_cx_tests(self):
...
+        # pylint: disable=no-value-for-parameter
+        super().setUpClass()
+
+    def test_refresh_method(self):
+        """Test refresh method"""
+        old_backend = FakeSherbrooke()
+        with self.assertLogs("qiskit_ibm_runtime", level="INFO"):
+            old_backend.refresh(self.service)
+        new_backend = FakeSherbrooke()
+        self.assertGreaterEqual(old_backend.backend_version, new_backend.backend_version)
```
jyu00, line 190: I don't quite understand this test. If FakeSherbrooke currently has version 1. The server data has version 2. If the `refresh()` code is not working correctly, both `old_backend` and `new_backend` would remain at version 1, and the test would still pass.

jyu00 (commented):
A little bit late, but I think this needs some changes.

PR #1750, opened 2024-06-14
[WIP] support for fixed angle gates in target conversion

nkanazawa1989:
### Summary

Qiskit [Target](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.Target#target) supports gate definition for multiple discrete angles. However our `convert_to_target` converter doesn't nicely support this mechanism. This PR overhaul the converter to support this feature.

### Details and comments

Transpiler (translation) requires gate definition (e.g. matrix), which is directly attached to the Qiskit `Gate` object in the standard library. On the other hand, `GateConfig` in the V1 configuration model doesn't have this definition. The definition is obtained by the name matching, and thus `GateConfig.name` must be defined in the dictionary returned from the `get_standard_gate_name_mapping` function, e.g. sx, x, rx, ...

However, when we define new operation in the Qiskit `Target` via the `.add_instruction` method, one must specify the unique `name` argument when the entry is defined for a specific gate angle. This indicates `GateConfig` must also define unique label for fixed gate angle entries. Since original Qiskit model is not flexible, this PR redefines the similar model with `.label` field in the qiskit-ibm-runtime (to avoid version mismatch issue).

Alternatively, we can still use the Qiskit model with custom name with regex, e.g. `rx_30`, but we need to mange this regex pattern both on client and server side. Also this pattern becomes complicated when the gate has more than one parameters; gate with two parameters and only one parameter is fixed, `u2_1.23_p1` -> `U2(1.23, Parameter("P1"))`. In my opinion this approach is easy to implement but hard to maintain. 

I prefer using the custom model with new `label` field, because this approach allows the client software (qiskit-ibm-runtime) to be agnostic to the gate naming convention in IBM Quantum systems.

I also introduced new backend configuration implementation with pydantic since this is already required package and much more performant and robust compared with the Qiskit implementation (I needed to update configuration to use new `GateConfig` model).

### DIFF
--- a/qiskit_ibm_runtime/models/__init__.py
+++ b/qiskit_ibm_runtime/models/__init__.py
@@ -1,0 +1,15 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2024.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""IBM custom backend model."""
+
+from .backend_configuration import IBMBackendConfiguration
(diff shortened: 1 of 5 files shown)


kt474:
Closing this for now, feel free to reopen

Issue #1577, opened 2024-04-02
cannot import name ValidationInfo from pydantic


ArslanH22:
**Describe the bug**
An error occurs when you try to import QuantumRuntimeServices from qiskit_ibm_runtim. It originates from qiskit_ibm_runtime\options\utlis.py at line 23. It says it cannot import the name 'ValidationInfo' from 'pydantic'.

**Steps to reproduce**
from qiskit_ibm_runtime import QuantumRuntimeServices

**Expected behavior**

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.22.0
- **qiskit version**: 1.0.2
- **Python version**: 3.11.7
- **Operating system**: 64 bit (AMD64)

kt474:
Hi @ArslanH22, this was fixed in #1500 - you can fix the error by installing pydantic `pip install pydantic`

soly33tworks:
> Hi @ArslanH22, this was fixed in #1500 - you can fix the error by installing pydantic `pip install pydantic`

Tried. Doesn't work, getting the exact same error.

Quartermain91:
@ArslanH22  Bump, same issue

Quartermain91:
@kt474  bump still facing this issue

kaisarmasum:
Doesn't work.

kt474:
@Quartermain91 @kaisarmasum can you post your error log?

kaisarmasum:
> @Quartermain91 @kaisarmasum can you post your error log?

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[7], line 1
----> 1 from qiskit_ibm_runtime import QiskitRuntimeService, Sampler, Batch

File ~\anaconda3\Lib\site-packages\qiskit_ibm_runtime\__init__.py:220
    217 from .utils.utils import setup_logger
    218 from .version import __version__
--> 220 from .estimator import (  # pylint: disable=reimported
    221     EstimatorV2,
    222     EstimatorV1,
    223     EstimatorV1 as Estimator,
    224 )
    225 from .sampler import (  # pylint: disable=reimported
    226     SamplerV2,
    227     SamplerV1,
    228     SamplerV1 as Sampler,
    229 )
    230 from .options import Options, EstimatorOptions, SamplerOptions, OptionsV2

File ~\anaconda3\Lib\site-packages\qiskit_ibm_runtime\estimator.py:30
[...]

PR #1759, opened 2024-06-19
Add required pydantic version in setup

kt474:
### Summary

Related to #1577 - users that already have `pydantic` installed in the their environments may run into incompatibility issues because the version of `pydantic` required is not specified in `setup.py`.

### Details and comments
Fixes #


SamFerracin (approved):
👍

PR #1760, opened 2024-06-19
Use error code 1305 instead of hard coding error msg 

### LINKED ISSUE #989: Use status code instead of `Ran too long`

kt474:
### Summary

Instead of hard coding the error reason "RAN TOO LONG", use the error code returned from the API instead.

Example:

```python
~/Documents/qiskit-ibm-runtime/qiskit_ibm_runtime/runtime_job_v2.py in result(self, timeout, decoder)
    137             error_message = self._reason if self._reason else self._error_message
    138             if self._reason_code == 1305:
--> 139                 raise RuntimeJobMaxTimeoutError(error_message)
    140             raise RuntimeJobFailureError(f"Unable to retrieve job result. {error_message}")
    141         if self._status == "CANCELLED":

RuntimeJobMaxTimeoutError: 'Error code 1305; Ran too long'
```

### Details and comments
Fixes #989


qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -239,15 +240,12 @@ def _set_status(self, job_response: Dict) -> None:
         """
         try:
             reason = job_response["state"].get("reason")
-            reason_code = job_response["state"].get("reason_code")
+            reason_code = job_response["state"].get("reasonCode")
```
SamFerracin, line 243: Was it getting anything before?

SamFerracin (approved):

qiskit_ibm_runtime/base_runtime_job.py
kt474: I'm not sure why/how this was changed on the server side, checking with the runtime team

kt474: Turns out the IQP channel returns `reasonCode` while then cloud/paygo channel returns `reason_code`. Going to hold off on merging this for now because handling these inconsistencies on the client side can get messy - following up with IQP.

kt474: Going to just handle both cases on the client side

Issue #1608, opened 2024-04-11
Add local channel


jyu00:
**What is the expected feature or enhancement?**

`QiskitRuntimeLocalService` was created to support local testing mode and is meant to be similar to `QiskitRuntimeService`. Today a user is not expected to instantiate this class directly, since it only has a `run()` method that is called under the cover by the primitives. As we add more methods to it (see #1606 and #1607), users _will_ want to instantiate it. Instead of having to import / create another class, we should just add a "local" channel. 

**Acceptance criteria**

User can do 

```python
from qiskit_ibm_runtime import QiskitRuntimeService, Sampler

service = QiskitRuntimeService(channel="local")  # <--- this will return an instance of `QiskitRuntimeLocalService`
```

jyu00:
I was having second thoughts about whether to use `QiskitRuntimeService(channel="local")`, since we want to allow one to refresh the fake backend's data (see #1707), which requires a real service. So people may think they can do 

```python
service = QiskitRuntimeService(channel="local")
fake_backend = FakeSherbrooke()
fake_backend.refresh(service)
```

But that won't work since `service` here is local service, not a real QRT service. On the other hand, even if we make people to initialize `service = QiskitRuntimeLocalService()`, they may still think it's a "service" they can pass to `refresh()`. 

We should just check if the service passed to `refresh` is a non-local `QiskitRuntimeService`

PR #1764, opened 2024-06-20
Adding backend related methods to `QiskitRuntimeLocalService`

### LINKED ISSUE #1606: Add backend related methods to QiskitRuntimeLocalService

`QiskitRuntimeLocalService` was created to support local testing mode and is meant to be similar to `QiskitRuntimeService`. The initial release of `QiskitRuntimeLocalService` only has the `run()` method, and we immediately got feedback asking for the other methods 🙂. 

This issue is about adding backend related methods to `QiskitRuntimeLocalService`. 

**Acceptance criteria**

Add the following methods to `QiskitRuntimeLocalService`
- backends
- backend
- least_busy

In local testing mode, the only backend accessible are the fake backends, hence these methods should only return fake backends (e.g. `backend("fake_sherbrooke")` would work but not `backend("ibm_sherbrooke")`)

SamFerracin:
### Summary
Add the following methods to QiskitRuntimeLocalService

backends
backend
least_busy

### Details and comments
Fixes #1606

### DIFF
--- a/qiskit_ibm_runtime/fake_provider/local_service.py
+++ b/qiskit_ibm_runtime/fake_provider/local_service.py
@@ -103,8 +103,10 @@ class QiskitRuntimeLocalService:
         if name:
             for b in backends:
                 if b.name == name:
-                    return [b]
-            raise err
+                    backends = [b]
+                    break
+            else:
+                raise err
             
         if min_num_qubits:
             backends = [b for b in backends if b.num_qubits >= min_num_qubits]
--- a/qiskit_ibm_runtime/fake_provider/local_service.py
+++ b/qiskit_ibm_runtime/fake_provider/local_service.py
@@ -19,7 +19,7 @@ import logging
 import warnings
 from dataclasses import asdict
 from random import choice
-from typing import Any, Callable, Dict, List, Literal, Optional, Union
+from typing import Callable, Dict, List, Literal, Optional, Union
 
 from qiskit.primitives import (
     BackendEstimator,
--- a/qiskit_ibm_runtime/fake_provider/local_service.py
+++ b/qiskit_ibm_runtime/fake_provider/local_service.py
@@ -131,8 +131,6 @@ class QiskitRuntimeLocalService:
 
         Args:
             min_num_qubits: Minimum number of qubits the fake backend has to have.
-            instance: This is only supported for ``ibm_quantum`` runtime and is in the
-                hub/group/project format.
             filters: More complex filters, such as lambda functions, that can be defined as for the
                 :meth:`backends` method.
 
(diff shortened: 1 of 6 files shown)


qiskit_ibm_runtime/fake_provider/local_service.py
```diff
@@ -51,6 +54,91 @@ def __init__(
...
+
+        Args:
+            min_num_qubits: Minimum number of qubits the fake backend has to have.
+            filters: More complex filters, such as lambda functions, that can be defined as for the
+                :meth:`backends` method.
+
+        Returns:
+            A fake backend.
+        """
+        return choice(self.backends(min_num_qubits=min_num_qubits, filters=filters))
```
SamFerracin, line 140: @jyu00 this is my best guess of what this function should do, but let me know if you have any thoughts. For example, should it pick a random backend the first time, and then just keep returning that same backend?

kt474 (commented):
Can you also add a release note - instructions are here https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/CONTRIBUTING.md#release-notes

You can see examples [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/release-notes/0.24.0.rst), these later get rendered in the qiskit docs [here](https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/release-notes)

release-notes/0.24.1.rst
```diff
@@ -1,5 +1,9 @@
 0.24.1 (2024-06-18)
 ===================
+New Features
+------------
+
+- Added ``backend``, ``backends``, and ``least_busy`` methods to ``QiskitRuntimeLocalService``. (`1764 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/1764>`__)
```
kt474, line 6: Sorry I gave a bad example here - this `0.24.1.rst` file is what the release notes look like after they are generated during a release. For PRs, the release notes should be in the format `<PR number>.<type>.rst`. So for example, for this PR, it'd be: `1764.feat.rst`

Better examples: https://github.com/jyu00/qiskit-ibm-runtime/tree/a2f6df81c74f9ed9cbaeba7ae75fbd18668d6da7/release-notes/unreleased

SamFerracin: Nice automation flow!

kt474 (approved):

jyu00:
It's too late now since this was already released, but just for future reference, `QiskitRuntimeLocalService` is actually a undocumented class (i.e. it doesn't show up in API reference). It's hidden because it was not meant to be instantiated directly. So one wouldn't know about its existence and thus its new backend methods unless they look at the code, at least until #1608 is released that is.

PR #1765, opened 2024-06-20
Add more fake backend methods

### LINKED ISSUE #1605: Make fake backends more like IBMBackend

The fake backends in this repo are meant to represent IBM backends and be similar to `IBMBackend`. In addition to inheriting `BackendV2`, `IBMBackend` also has a number of additional methods that should be added to fake backends. Note that this update only needs to apply to V2 fake backends (since we are [deprecating V1](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1604)).

**Acceptance criteria**

- Add additional `IBMBackend` methods to V2 fake backends, including
    - service (this should return `QiskitRuntimeLocalService`
    - properties, defaults, configuration
    - status
    - check_faulty

kt474:
### Summary

Added the following methods to `FakeBackendV2`:

- ~~service~~ unable to resolve circular dependencies with adding a service property 
- properties, defaults, configuration
- status
- check_faulty

### Details and comments
Fixes #1605


qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -165,6 +174,101 @@ def _load_json(self, filename: str) -> dict:
...
+            from the local file.
+
+        Returns:
+            The backend properties.
+        """
+        if refresh or (self._props_dict is None):
+            self._set_props_dict_from_json()
+        return BackendProperties.from_dict(self._props_dict)
+
+    def defaults(self, refresh: bool = False) -> PulseDefaults:
```
ptristan3, line 218: The parameter name could be confusing because we have a `refresh` method to update the data files from their real. Maybe `reload` sounds better?

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -165,6 +174,101 @@ def _load_json(self, filename: str) -> dict:
...
+            refresh: If ``True``, re-retrieve the backend defaults from the
+            local file.
+
+        Returns:
+            The backend pulse defaults or ``None`` if the backend does not support pulse.
+        """
+        if refresh or self._defs_dict is None:
+            self._set_defs_dict_from_json()
+        if self._defs_dict:
+            return PulseDefaults.from_dict(self._defs_dict)  # type: ignore[unreachable]
```
ptristan3, line 231: `unreachable` again, I don't understand why.

qiskit_ibm_runtime/fake_provider/fake_backend.py
kt474: I think we should be consistent with the methods in `ibm_backend` - also it's the same refresh because we want to use the files that were just updated by the `refresh()` method

ptristan3 (approved):

PR #1766, opened 2024-06-21
Add Support for Local Websockets and IBM Quantum API

Rcuz8:
### Summary
* Adds support for replacing `http://` with `ws://` (Right now we only replace `https://` with `wss://`) to enable local urls
* Default localhost URLs to the ibm quantum channel


kt474:
We're removing websockets as a part of https://github.com/Qiskit/qiskit-ibm-runtime/issues/2140 so closing this for now

PR #1771, opened 2024-06-25
Add instance property to job

kt474:
### Summary

From a slack conversation - it'd be nice if there was an easy, documented way to get the instance where a job was run. 

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/base_runtime_job.py
+++ b/qiskit_ibm_runtime/base_runtime_job.py
@@ -470,7 +470,9 @@ class BaseRuntimeJob(ABC):
 
     @property
     def instance(self) -> Optional[str]:
-        """Return the instance the job was run on"""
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -468,6 +468,11 @@ def usage_estimation(self) -> Dict[str, Any]:
 
         return self._usage_estimation
 
+    @property
+    def instance(self) -> Optional[str]:
+        """Return the instance the job was run on"""
```
mriedem, line 473: "Instance" is an IQP term, not for the pay-go cloud channel. What happens if this is a cloud channel job? Does this blow up? Does it return null or an empty string?

kt474: Returns `None` for cloud channel jobs - I'll update the docstring to be more clear

PR #1777, opened 2024-06-27
Fake backend refresh method improvements

### LINKED ISSUE #1707: Allow users to update fake backends with latest data

ptristan3:
### Summary

Continuing the [PR](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1740)

### Details and comments
Fixes #1707


test/integration/test_fake_backends.py
```diff
@@ -183,8 +183,25 @@ def setUpClass(cls):
...
+            new_backend.refresh(self.service)
+        self.assertIn("There are no available new updates for fake_sherbrooke", logs.output[0])
+
+        # to verify the refresh can't be done
+        wrong_backend = FakeSherbrooke()
+        # set a non-existent backend name
+        wrong_backend.backend_name = "wrong_fake_sherbrooke"
+        with self.assertLogs("qiskit_ibm_runtime", level="WARNING") as logs:
+            wrong_backend.refresh(self.service)
+        self.assertIn("The refreshing of wrong_fake_sherbrooke has failed", logs.output[0])
```
kt474, line 207: Tested locally and functionally LGTM but is there a way to revert the files that are updated when running this test? Not sure how this works in the github CI but when running the tests locally it'd be nice if they didn't edit local files

kt474 (approved):

PR #1780, opened 2024-06-28
Set _active flag to false if session already closed

### LINKED ISSUE #1597: Set _active flag in Session.from_id()

`Session.from_id()` already queries session details. It can also set the `_active` flag, so if that session is closed, any attempt to submit a new job will immediately be rejected. 

**Acceptance criteria**

Set the session active flag when it's created via `from_session`

kt474:
### Summary

If the session created with `from_id` is already closed, jobs should be rejected immediately. 

### Details and comments
Fixes #1597


SamFerracin (commented):
Is it possible to add tests for this?

SamFerracin (approved):

PR #1793, opened 2024-07-05
Add local channel

### LINKED ISSUE #1608: Add local channel

ptristan3:
### Details and comments
Fixes #1608


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -56,6 +56,27 @@ class QiskitRuntimeService:
...
+        token: Optional[str] = None,
+        url: Optional[str] = None,
+        filename: Optional[str] = None,
+        name: Optional[str] = None,
+        instance: Optional[str] = None,
+        proxies: Optional[dict] = None,
+        verify: Optional[bool] = None,
+        channel_strategy: Optional[str] = None,
+        private_endpoint: Optional[bool] = None,
+    ):
```
kt474, line 71: I think these params could be replaced with *args and **kwargs

release-notes/unreleased/1793.feat.rst
```diff
@@ -0,0 +1,8 @@
+The :class:`QiskitRuntimeLocalService` was created to support a local
+testing mode. To avoid the user creating separately the class the "local"
+has been added to the valid :class:`QiskitRuntimeService` channels.
```
kt474, line 3:
```suggestion
The :class:`QiskitRuntimeLocalService` was created to support a local
testing mode. To avoid having to initialize a separate class, "local"
has been added to the valid :class:`QiskitRuntimeService` channel.
```

kt474 (approved):
LGTM just one comment from the original issue - https://github.com/Qiskit/qiskit-ibm-runtime/issues/1608#issuecomment-2181608504

In the refresh method we can add a better error message if the local service is passed in

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -85,7 +95,7 @@ def __init__(
         values in the loaded account.
 
         Args:
-            channel: Channel type. ``ibm_cloud`` or ``ibm_quantum``.
+            channel: Channel type. ``ibm_cloud``, ``ibm_quantum`` or ``local``.
```
jyu00, line 98: Can we add a little more information on `local` - e.g. 

If `local` is specified, the local testing mode is used, and primitive queries will run on a local simulator. See [Qiskit Runtime local testing mode](https://docs.quantum.ibm.com/guides/local-testing-mode) for more details.

jyu00:
I also just noticed that since this would expose the `QiskitRuntimeLocalService` with this change, we should make its `run()` method private, because we don't want people to do `service.run()`. 

`QiskitRuntimeService.run()` issues a [deprecation message](https://github.com/Qiskit/qiskit-ibm-runtime/blob/674017dbb2882145a77391b4c7c5dfa9155c26bf/qiskit_ibm_runtime/qiskit_runtime_service.py#L830-L836) today. But all of our primitives [call that method](https://github.com/Qiskit/qiskit-ibm-runtime/blob/674017dbb2882145a77391b4c7c5dfa9155c26bf/qiskit_ibm_runtime/base_primitive.py#L157-L158), so wouldn't everyone get that deprecation message? I'm very puzzled as I'm running 0.25 but not seeing the deprecation message. @kt474 do you know what's going on?

kt474:
> `QiskitRuntimeService.run()` issues a [deprecation message](https://github.com/Qiskit/qiskit-ibm-runtime/blob/674017dbb2882145a77391b4c7c5dfa9155c26bf/qiskit_ibm_runtime/qiskit_runtime_service.py#L830-L836) today. But all of our primitives [call that method](https://github.com/Qiskit/qiskit-ibm-runtime/blob/674017dbb2882145a77391b4c7c5dfa9155c26bf/qiskit_ibm_runtime/base_primitive.py#L157-L158), so wouldn't everyone get that deprecation message? I'm very puzzled as I'm running 0.25 but not seeing the deprecation message. @kt474 do you know what's going on?

I believe the way the stacklevels work, the deprecation message will only show up if `service.run()` is called directly - so the message will not show up when calling `service._run()` or from within in the primitives. Having said that I should've updated the primitives to also use the private `_run` method.

Issue #1789, opened 2024-07-04
JSON Error when retrieving job results


osamarais:
**Describe the bug**
JSON error when running on real backend:

  File "/lib/python3.12/site-packages/simplejson/__init__.py", line 533, in loads
    return cls(encoding=encoding, **kw).decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/qiskit_ibm_runtime/utils/json.py", line 331, in __init__
    super().__init__(object_hook=self.object_hook, *args, **kwargs)
TypeError: JSONDecoder.__init__() got an unexpected keyword argument 'encoding'

**Steps to reproduce**
Submit a job and get the results from real backend
result = estimator.run([(ansatz, hamiltonian, [params])]).result()

**Expected behavior**
Should return job data without an error.

**Suggested solutions**
Simply replacing
`import json`
with
`import simplejson as json`
in qiskit_ibm_runtime/utils/json.py
fixes this particular issue.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.23.0
- **Python version**: 3.12.3
- **Operating system**: Ubuntu 24.04 LTS  Linux 6.8.0-36-generic x86-64

kt474:
Hi @osamarais, this has been discussed in a previous issue - https://github.com/Qiskit/qiskit-ibm-runtime/issues/1603. I believe the solution here would be to uninstall simplejson from your environment

Issue #1603, opened 2024-04-11
JSONDecoder error when retrieving IQP job


jyu00:
**Describe the bug**

Retrieving job 

<img width="928" alt="image" src="https://github.com/Qiskit/qiskit-ibm-runtime/assets/28284230/f6d4d0ea-3937-4ebf-b6d6-06d24aeb1654">

raised a JSONDecoder error

![image](https://github.com/Qiskit/qiskit-ibm-runtime/assets/28284230/789dc615-9a11-4979-9e70-551e9b12e38c)

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.22
- **Python version**: 3.10
- **Operating system**:

I didn't look too much into this, but a quick Google search says it may be an incompatibility between `request` and Python 3.10.

danipombo:
For context, the above error was generated when running the code on the IBM Watson Studio. However, It ran smoothly when executing in my local environment.

pacomf:
I replicated the issue in Watson Studio and I resolved uninstalling the simplejson dependency:

`!yes | pip uninstall simplejson`

For context: https://github.com/Qiskit/qiskit-ibm-provider/pull/649#issuecomment-1572362758

kt474:
Thank you @pacomf - yes this same issue we discovered a while back in `qiskit-ibm-provider`. 

>It appears that when simplejson is in the environment requests preferentially imports it. However, the arguments accepted by JSONDecoder are inconsistent between the standard library json and simplejson leading to errors when using the standard json.JSONDecoder when simplejson is in the environment.

Closing as resolved

HuangJunye:
This issue was reported during the quantum challenge when people were running the notebook on qBraid. Is there a plan to add some checks of whether simplejson exist in the environment and raise warnings? This bug is rather hard to debug unless user search on the github repo.

rryoung98:
Thanks @HuangJunye it's certainly a non-obvious bug we'll try and fix it asap.

JaganNatarajan:
@rryoung98 is this issue fixed. I am still seeing this happen in qBraid. Could you please update?

kt474:
Hi @JaganNatarajan I'm working on a fix in #1800

PR #1800, opened 2024-07-10
Fix simplejson incompatibility

kt474:
### Summary

Related to #1789 and #1603 - Attempt to fix the issue where having `simplejson` installed in an environment prevents users from retrieving job results 

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_job.py
+++ b/test/integration/test_job.py
@@ -47,6 +47,17 @@ class TestIntegrationJob(IBMIntegrationJobTestCase):
         self.assertTrue(job.result())
 
     @run_integration_test
+    def test_run_with_simplejson(self, service):
+        """Test retrieving job results with simplejson package installed."""
+        try:
+            __import__("simplejson")
+            job = self._run_program(service=service)
+            job.wait_for_final_state()
+            self.assertTrue(job.result())
+        except ImportError:
+            self.assertRaises(ImportError)
+
+    @run_integration_test
     @quantum_only
     def test_run_program_log_level(self, service):
         """Test running with a custom log level."""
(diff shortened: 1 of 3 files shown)


ptristan3 (commented):
Could we add a simple unit test to verify this fix?

test/integration/test_job.py
```diff
@@ -46,6 +46,17 @@ def test_run_program(self, service):
...
+    @run_integration_test
+    def test_run_with_simplejson(self, service):
+        """Test retrieving job results with simplejson package installed."""
+        try:
+            __import__("simplejson")
+            job = self._run_program(service=service)
+            job.wait_for_final_state()
+            self.assertTrue(job.result())
+        except ImportError:
+            self.assertRaises(ImportError)
```
SamFerracin, line 58: Have you considered pip installing it, running the test, and pip uninstalling it? Not sure what the best approach (installing it VS raising the import error) is in this case

FWIW, the installation takes very little, so the overall test time should be impacted minimally?

kt474: I did consider that and yes I think it would be a more complete test that way but I feel that it's generally not good practice to install/uninstall packages inside a test. 

Also not sure what the best approach is here and if there's a better solution

ptristan3: I agree that modifying the env to run a test seems not the best approach. But we are doing some similar in the [test](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/integration/test_fake_backends.py#L188) so...I'm also not sure what the best way to test here.

kt474: Going to merge for now and closely monitor the main branch

PR #1802, opened 2024-07-16
Remove downloading result from external url

kt474:
### Summary

From a slack conversation with @daka1510 regarding when `_download_external_result` was first added [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/709) to support osprey

> We have now cleaned up the corresponding code on the server-side to support this special result flow. Feel free to remove it from the SDK, too, if no one in this group objects.
Allowing clients to fetch data from COS directly via the SDK might become a priority again, but probably not follow this design with pre-signed URLs. Since it's all unclear right now, it's probably not a good idea to keep this code around.

### Details and comments
Fixes #

### DIFF
--- a/test/unit/test_circuit_runner_job.py
+++ b/test/unit/test_circuit_runner_job.py
@@ -62,9 +62,6 @@ class FakeCircuitRunnerJob(RuntimeJob):
         """
         return JOB_FINAL_STATES
 
-    def _download_external_result(self, response: Any) -> Any:
-        return self._data
-
     def submit(self) -> None:
         """Fake submit"""
         pass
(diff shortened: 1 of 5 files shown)


test/unit/test_circuit_runner_job.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, data: Any) -> None:
         """
         self._data = data
         api_client = MagicMock()
+        api_client.job_results = MagicMock(return_value=self._data)
```
ptristan3, line 37: Why is this line added? It seems to have no relation to the PR we want to revert

kt474: In `test_circuit_metadata_ndarray` we were mocking `_download_external_result` to just return the data given so since that method has been removed, i'm mocking `job_results()`

ptristan3 (approved):

PR #1804, opened 2024-07-16
Deprecate passing backend as a string

### LINKED ISSUE #1787: Deprecate passing a string for backend

Related to #1681. 

Today we allow one to pass the backend name (string) when initializing Sampler/Estimator/Session/Batch, for example, you can do

```python
sampler = SamplerV2(backend="ibm_osaka")
```

This would actually call QiskitRuntimeService() (i.e. use your default account setting) under the cover, which can be a surprise to people if the default channel or h/g/p is not what they want to use.

Furthermore, some people would do

```python
service = QiskitRuntimeService(name="foo")
sampler = SamplerV2(backend="ibm_osaka")
```

In the past this would cause confusion because it would actually use the default account instead of the account named `foo`. To address this, we [added code](https://github.com/Qiskit/qiskit-ibm-runtime/pull/955) to use the last-initialize account instead of the default one. But this was also inferring user's intent and contradicts with the [docstring](https://github.com/Qiskit/qiskit-ibm-runtime/blob/fd3f1055f4a51f4c6d7ffaca574a5cb446a7da08/qiskit_ibm_runtime/sampler.py#L87) which says `If a name is specified, the default account is used`. 

To make the account selection more explicit, we can 1) make both `service` and `backend` required or 2) disallow passing a string for `backend`. 

With the ISA circuit requirement, people are more likely to initialize a backend object now to do transpilation, so (2) seems to be the better choice. 

**Acceptance criteria**

- Deprecate passing a string for `backend` when initializing `Session`, `Batch`, `SamplerV2` and `EstimatorV2`.

kt474:
### Details and comments
Fixes #1787

### DIFF
--- a/qiskit_ibm_runtime/batch.py
+++ b/qiskit_ibm_runtime/batch.py
@@ -18,6 +18,7 @@ from qiskit.providers.backend import BackendV1, BackendV2
 
 from qiskit_ibm_runtime import QiskitRuntimeService
 from .session import Session
+from .utils.deprecation import issue_deprecation_msg
 
 
 class Batch(Session):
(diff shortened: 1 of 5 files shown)


ptristan3 (approved):

jyu00:
@kt474 before we release this, can you also deprecate passing `service` to `Session` and `Batch`? Since we can extract `service` from the `backend` instance.

kt474:
> @kt474 before we release this, can you also deprecate passing `service` to `Session` and `Batch`? Since we can extract `service` from the `backend` instance.

Added in https://github.com/Qiskit/qiskit-ibm-runtime/pull/1826

Issue #1762, opened 2024-06-19
Use attribute docstrings for options


jyu00:
**What is the expected feature or enhancement?**

Today most of the options dataclasses use the same Google style argument docstring as the rest of the modules. Unfortunately, the rendered API reference would list each field under both `Parameters` and `Attributes`, [for example](https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.EstimatorOptions):

<img width="787" alt="image" src="https://github.com/Qiskit/qiskit-ibm-runtime/assets/28284230/ba245969-526e-4581-a0c7-71558ec532de">

<img width="323" alt="image" src="https://github.com/Qiskit/qiskit-ibm-runtime/assets/28284230/2e5ec412-5dab-47f2-a391-48a44ac799b5">

<br></br>

The one exception is [SamplerExecutionOptions](https://github.com/Qiskit/qiskit-ibm-runtime/blob/3026f9f83b23993607a5aa66e481bbaafdfee521/qiskit_ibm_runtime/options/sampler_execution_options.py#L20), which only has attribute docstring. The generated doc therefore only shows the fields under `Attributes`:

<img width="638" alt="image" src="https://github.com/Qiskit/qiskit-ibm-runtime/assets/28284230/88493840-10c0-49c5-99a1-7f4021f19695">

<br></br>

I think this is a much nicer format, as you get both the description and type in one place. 

@Eric-Arellano please let us know if you have better ideas! 

**Acceptance criteria**

- Use attribute docstring instead of argument docstring for options classes. 

Also note that when I played with `SamplerExecutionOptions`, the text `Default: True` was not formatted correctly, so I don't know if there is special convention on specifying defaults.

jyu00:
Reopening since I noticed `ZneOptions` was not converted.

PR #1805, opened 2024-07-17
Adding NoiseLearner class and related functionality

SamFerracin:
### Summary
Adding `NoiseLearner` class, alongside the dedicated options and the result containers.

Missing integration tests, but tested locally on staging


docs/conf.py
```diff
@@ -16,39 +16,39 @@
 import re
 import sys
 
-sys.path.insert(0, os.path.abspath('.'))
```
ihincks, line 19: these appear to be only formatting issues. since this is a big feature we're introducing, could you please open a new PR for this file?

qiskit_ibm_runtime/noise_learner.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+# pylint: disable=unused-import,cyclic-import
+from .session import Session
+from .batch import Batch
+
+logger = logging.getLogger(__name__)
+
+
+class NoiseLearner:
```
ihincks, line 47: I have reservations about how generic this name is, but unfortunately have nothing more constructive to say at this time. The concern is that we'll one day want a different thing to be learned, in which case we'll regret making this name not specific to Sparse Pauli learning.

qiskit_ibm_runtime/noise_learner.py
```diff
@@ -0,0 +1,276 @@
...
+
+            session: Session in which to call the primitive.
+
+                If both ``session`` and ``backend`` are specified, ``session`` takes precedence.
+                If neither is specified, and the primitive is created inside a
+                :class:`qiskit_ibm_runtime.Session` context manager, then the session is used.
+                Otherwise if IBM Cloud channel is used, a default backend is selected.
+
+            options: :class:`NoiseLearnerOptions`. Alternatively, :class:`EstimatorOptions` can be
+                provided, in which case the estimator options get reformatted into noise learner options
```
ihincks, line 126:
```suggestion
                provided for convenience, in which case the estimator options get reformatted into noise learner options
```

qiskit_ibm_runtime/noise_learner.py
```diff
@@ -0,0 +1,276 @@
...
+                and all the irrelevant fields are ignored.
+        """
+        self._mode: Optional[Union[Session, Batch]] = None
+        self._service: QiskitRuntimeService | QiskitRuntimeLocalService = None
+        self._backend: Optional[BackendV1 | BackendV2] = None
+
+        if isinstance(mode, (Session, Batch)):
+            self._mode = mode
+            self._service = self._mode.service
+            self._backend = self._mode._backend
```
ihincks, line 136: I suppose there's no way to do this via public getters? Seems like a flaw in the service/session/batch. 

@kt474 do you agree that this might be an oversight, or is there something that I'm missing?

qiskit_ibm_runtime/noise_learner.py
```diff
@@ -0,0 +1,276 @@
...
+            raise ValueError("A backend or session must be specified.")
+
+        self._set_options(options)
+
+    @property
+    def options(self) -> NoiseLearnerOptions:
+        """The options in this noise learner."""
+        return self._options
+
+    def run(self, tasks: Iterable[Union[QuantumCircuit, EstimatorPubLike]]) -> RuntimeJobV2:
```
ihincks, line 166: I think we should consider this

```suggestion
    def run(self, circuits: Iterable[QuantumCircuit]) -> RuntimeJobV2:
```

with the recommendation

```python
learner.run(pub.circuit for pub in pubs)
```

My argument is that this will be a cleaner path to an interface that only accepts twirl blocks, assuming that's the goal.

qiskit_ibm_runtime/noise_learner.py
```diff
@@ -0,0 +1,276 @@
...
+                tuples ``(circuit, observables)`` or ``(circuit, observables, parameter_values)``.
+                In this case, the pub-like objects are converted to a list of circuits, and all
+                the other fields (such as ``observables`` and ``parameter_values``) are ignored.
+
+        Returns:
+            Submitted job.
+
+        """
+        if not all(isinstance(t, QuantumCircuit) for t in tasks):
+            coerced_pubs = [EstimatorPub.coerce(pub) for pub in tasks]
```
ihincks, line 181: Is this just for validation?

docs/conf.py
SamFerracin: These are unwanted changes that I had missed on the diff. I will just revert them. Thanks for spotting them

qiskit_ibm_runtime/noise_learner.py
SamFerracin: I am not 100% against this, because `pub.circuit for pub in pubs` is simple enough that it is not inconvenient.
But then I would also disallow passing estimator options (because it feels unnatural to allow est options but not pubs), and in that case we don't have a convenient way for users to do something like `learner_options.into_estimator_options`.

So if we make this change, IMO we should just change the narrative: The noise learner takes noise learner options and circuits, up to you to convert them

qiskit_ibm_runtime/noise_learner.py
SamFerracin: Yes, because they may pass actual pubs or tuples, and I want to avoid logic to distinguish between all these cases

qiskit_ibm_runtime/noise_learner.py
ihincks: Maybe a goal of the "experimental" stage of this feature can be to determine whether we want to have pub support. We allow pubs for now, and see how we feel about it as we continue to use it.

qiskit_ibm_runtime/noise_learner.py
```diff
@@ -0,0 +1,276 @@
...
+        if isinstance(mode, (Session, Batch)):
+            self._mode = mode
+            self._service = self._mode.service
+            self._backend = self._mode._backend
+        elif isinstance(mode, IBMBackend):  # type: ignore[unreachable]
+            self._service = mode.service
+            self._backend = mode
+        elif isinstance(mode, (BackendV1, BackendV2)):
+            self._service = QiskitRuntimeLocalService()
+            self._backend = mode
```
jyu00, line 142: This block is for local testing mode (i.e. if an Aer simulator or a FakeBackend is passed in). It should raise an exception because NoiseLearner is not supported in local mode (yet).

qiskit_ibm_runtime/noise_learner.py
SamFerracin: Sounds like a plan

qiskit_ibm_runtime/options/noise_learner_options.py
```diff
@@ -0,0 +1,85 @@
...
+
+from .utils import Unset, UnsetType
+from .twirling_options import TwirlingStrategyType
+from .options import OptionsV2
+from .utils import primitive_dataclass, make_constraint_validator, skip_unset_validation
+
+
+@primitive_dataclass
+class NoiseLearnerOptions(OptionsV2):
+    """Options for :class:`.NoiseLearner`.
```
jyu00, line 27: Since this is a new class, we can start using attribute docstring instead of argument docstring (see https://github.com/Qiskit/qiskit-ibm-runtime/issues/1762) to make the API reference look nicer.

qiskit_ibm_runtime/noise_learner.py
```diff
@@ -0,0 +1,276 @@
...
+        filtering out every option that is not part of the NoiseLearningOptions."""
+        ret = {}
+
+        for key in [
+            "max_layers_to_learn",
+            "shots_per_randomization",
+            "num_randomizations",
+            "layer_pair_depths",
+            "twirling_strategy",
+            "simulator",
```
jyu00, line 268: I'd suggest using the `dataclasses.fields()` method to extract all fields from `NoiseLearnerOptions` then remove the ones you don't want. When things change, it's a lot easier to find an extra parameter (since it'd be rejected by the server), then one that's set and not passed. 

Also I assume you meant to include `experimental`?

test/integration/test_noise_learner.py
```diff
@@ -0,0 +1,43 @@
...
+from ..decorators import run_integration_test
+from ..ibm_test_case import IBMIntegrationTestCase
+
+
+class TestIntegrationNoiseLearner(IBMIntegrationTestCase):
+    """Integration tests for NoiseLearner."""
+
+    def setUp(self) -> None:
+        super().setUp()
+        self.backend = "ibmq_qasm_simulator"
```
jyu00, line 27: this should use `test_eagle_us-east` instead since we don't have cloud simulators (on IQP) anymore.

test/integration/test_noise_learner.py
```diff
@@ -0,0 +1,43 @@
...
+        c2.cx(1, 2)
+
+        self.circuits = [c1, c2]
+
+    @run_integration_test
+    def test_noise_learner_session(self):
+        """Verify if estimator primitive returns expected results"""
+
+        learner = NoiseLearner(self.backend)
+        learner.run(self.circuits)
```
jyu00, line 43: I assume you'll update this test with some asserts once it's working?

qiskit_ibm_runtime/noise_learner.py
SamFerracin: I don't think I fully like the idea of inheriting from `BasePrimitiveV2`, because that would make `NoiseLearner` be a primitive both as an object and conceptually (imagine a user that looks at it for the first time and tries to understand what it is). However, I agree that there is a lot of duplicate code between this implementation of noise learner and base primitives V2, which I like even less than inheriting. 
Therefore, I am okay with inheriting, but raising the point that we may want to think whether we have a missing concept in our codebase and are just shoving it into `BasePrimitiveV2`

qiskit_ibm_runtime/__init__.py
```diff
@@ -216,6 +216,7 @@
 from .utils.utils import setup_logger
 from .version import __version__
 
+from .noise_learner import NoiseLearner
```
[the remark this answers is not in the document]
SamFerracin: @jyu00 I am okay with that.
Related to this, since we are releasing this as an experimental feature, do we want to place it in an `experimental` subdirectory, add warning, and/or something else?

test/integration/test_noise_learner.py
SamFerracin: Yes, that's right!

qiskit_ibm_runtime/__init__.py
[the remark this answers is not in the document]
jyu00: Yeah I forgot to mention it - we should have a warning in the docstring saying it's experimental and that the interface may change in the future without notice. We'll do the same in the actual doc + announcement.

qiskit_ibm_runtime/noise_learner.py
```diff
@@ -0,0 +1,276 @@
...
+                instance. If a name is specified, the default account (e.g. ``QiskitRuntimeService()``)
+                is used.
+
+            session: Session in which to call the primitive.
+
+                If both ``session`` and ``backend`` are specified, ``session`` takes precedence.
+                If neither is specified, and the primitive is created inside a
+                :class:`qiskit_ibm_runtime.Session` context manager, then the session is used.
+                Otherwise if IBM Cloud channel is used, a default backend is selected.
+
```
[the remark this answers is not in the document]
SamFerracin: Thanks! Removed

qiskit_ibm_runtime/noise_learner.py
kt474: Should we update `backend()` to return the backend object (after deprecation period)? Looks like it's worked that way since the session class was created.

qiskit_ibm_runtime/noise_learner.py
jyu00: If the goal is to see how useful pub support is, then it'd be nice to track the usage. For example, we can pass a new  `metadata` parameter to `noise-learner`, or lump it under `experimental`.

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -0,0 +1,251 @@
...
+            if name in options_dict and name not in ignored_names:
+                ret[name] = deepcopy(options_dict[name])
+
+        remove_dict_unset_values(ret)
+        remove_empty_dict(ret)
+
+        # Remove image
+        ret.get("experimental", {}).pop("image", None)
+
+        return ret
```
jyu00, line 251: we should add `support_qiskit` now, even if it's not used. So it'll be there when we are ready to support non-Qiskit callers.

qiskit_ibm_runtime/noise_learner.py
jyu00: > we may want to think whether we have a missing concept in our codebase and are just shoving it into BasePrimitiveV2

Yeah we should have a generic `ProgramRunner` now that we can call more than just primitives. But that can be another PR.

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -0,0 +1,109 @@
...
+        Args:
+            data: The data of a noise learner experiment.
+            metadata: Metadata that is common to all pub results; metadata specific to particular
+                pubs should be placed in their metadata fields. Keys are expected to be strings.
+        """
+        self._data = data
+        self._metadata = metadata or {}
+
+    @classmethod
+    def from_estimator_result(cls, result: EstimatorResult) -> NoiseLearnerResult:
```
jyu00, line 69: Is the noise model always returned if ZNE is specified? Or does it require some option?

SamFerracin: I think that the noise is only specified if the estimator did noise learning. So for example, if the estimator did ZNE with gate folding, it won't be there

qiskit_ibm_runtime/noise_learner/noise_learner.py
SamFerracin: Okay, I added it. For now, the server will just `pop` it and forget about it

qiskit_ibm_runtime/utils/noise_learner_result.py
jyu00: I think that's worth pointing out in the docstring - i.e. not all `EstimatorResult` will contain noise model, and that an empty `NoiseLearnerResult` is returned if it doesn't. 

Also, do we want to return an empty `NoiseLearnerResult` or just raise?

SamFerracin: Honestly, when I added this function I thought I'd need it, but I ended up not using it anywhere else. So:
> do we want to return an empty NoiseLearnerResult or just raise?
I don't know right now, I'd like to see a usecase before I decide. If you are okay with it, I'll remove it for now

I think that there are a few more methods that may turn out to be useful here.`from_estimator_result` is one, but also `to_noise_model` to get a `NoiseModel` object that can be used in a simulator, plotting methods..
I will start working on it as soon as I have time :)

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -0,0 +1,257 @@
...
+
+    The noise learner class allows characterizing the noise processes affecting the gates in one or more
+    circuits of interest, based on the Pauli-Lindblad noise model described in [1].
+
+    The :meth:`run` method allows runnig a noise learner job for a list of circuits. After the job is
+    submitted, the gates are collected into independent layers, and subsequently the resulting layers are
+    are characterized individually. The way in which the gates are collected into layers depends on the
+    ``twirling_strategy`` specified in the given ``options`` (see :class:`NoiseLearnerOptions` for more
+    details).
+
```
ihincks, line 59: I think we should spell out a common case that will arise for early adopters, even at the risk of being repetitive. The common case is that they want to force particular full-width (wrt active qubits) layers. (note also that "twirling_strategy" is not an option)

```suggestion
    are characterized individually.
    
    The way in which the gates are collected into layers depends on the twirling
    ``strategy`` specified in the given ``options`` (see :class:`NoiseLearnerOptions` for more
    details).
    Note that all strategies obey barriers.
    For example, if you have three ISA entangling layers of interest, consider putting them into one circuit
    separated by barriers acting on the qubits you wish to twirl, and select ``strategy="active-circuit"``.
```

ihincks: WARNING: the above snippet is actually incorrect because "active-ignore" has a default value ("barrier", "delay"). Can we actually achieve this behaviour without inserting a bunch of dummy gates/measurements? I think "no"

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -0,0 +1,257 @@
...
+            circuits: An iterable of circuits to run the noise learner program for. Alternatively,
+                pub-like (primitive unified bloc) objects can be specified, such as
+                tuples ``(circuit, observables)`` or ``(circuit, observables, parameter_values)``.
+                In this case, the pub-like objects are converted to a list of circuits, and all
+                the other fields (such as ``observables`` and ``parameter_values``) are ignored.
+
+        Returns:
+            The submitted job.
+
+        """
```
ihincks, line 153: Consider mentioning that the union over all layers from all circuits is taken.

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -0,0 +1,257 @@
...
+    def _set_options(
+        self, options: Optional[Union[Dict, NoiseLearnerOptions, EstimatorOptions]] = None
+    ) -> None:
+        """
+        Sets the options, ensuring that they are of type ``NoiseLearnerOptions``.
+        """
+        if not options:
+            self._options = NoiseLearnerOptions()
+        elif isinstance(options, NoiseLearnerOptions):
+            self._options = options
```
ihincks, line 221: I see `self._options = replace(options)` in BasePrimitiveV2, so this non-copy behaviour may be unexpected.

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -0,0 +1,257 @@
...
+        elif isinstance(options, EstimatorOptions):
+            d = asdict(options.resilience.layer_noise_learning)  # type: ignore[union-attr]
+            d.update({"twirling_strategy": options.twirling.strategy})  # type: ignore[union-attr]
+            d.update({"max_execution_time": options.max_execution_time})
+            d.update({"simulator": options.simulator})
+            d.update({"environment": options.environment})
+            d.update({"experimental": options.experimental})
```
ihincks, line 228: why update instead of direct setting?

qiskit_ibm_runtime/options/noise_learner_options.py
```diff
@@ -0,0 +1,93 @@
...
+    r"""The circuit depths (measured in number of pairs) to use in learning experiments.
+    
+    Pairs are used as the unit because we exploit the order-2 nature of our entangling gates in
+    the noise learning implementation. A value of ``3`` would correspond to 6 layers of the layer
+    of interest, for example. Default: (0, 1, 2, 4, 16, 32).
+    """
+
+    twirling_strategy: Union[UnsetType, TwirlingStrategyType] = Unset
+    r"""The twirling strategy in the identified layers of two-qubit twirled gates.
+
```
ihincks, line 65: note that barriers and delay instructions are ignored when determining whether a qubit is active

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def __init__(self, data: Sequence[NoiseLearnerDatum], metadata: dict[str, Any] | None = None):
+        """
+        Args:
+            data: The data of a noise learner experiment.
+            metadata: Metadata that is common to all pub results; metadata specific to particular
+                pubs should be placed in their metadata fields. Keys are expected to be strings.
+        """
+        self._data = data
+        self._metadata = metadata or {}
```
ihincks, line 66: I'd prefer shallow copies to avoid unexpected problems.
```suggestion
        self._data = list(data)
        self._metadata = metadata.copy() or {}
```

ihincks (approved):

qiskit_ibm_runtime/noise_learner/noise_learner.py
SamFerracin: okay, i can replace!

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -0,0 +1,268 @@
...
+# pylint: disable=unused-import,cyclic-import
+from ..session import Session
+from ..batch import Batch
+
+logger = logging.getLogger(__name__)
+
+
+class NoiseLearner:
+    """Class for executing noise learning experiments.
+
```
jyu00, line 50: If we plan to release this as experimental, then we should have add a warning here.

jyu00 (approved):

jyu00:
Can you also add a release note?

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -0,0 +1,268 @@
...
+    @property
+    def options(self) -> NoiseLearnerOptions:
+        """The options in this noise learner."""
+        return self._options
+
+    def run(self, circuits: Iterable[Union[QuantumCircuit, EstimatorPubLike]]) -> RuntimeJobV2:
+        """Submit a request to the noise learner program.
+
+        This function breaks the given list of circuits into a list of unique layers, following
+        the strategy set by the ``twirling-strategy`` field specified in the ``options`` (see
```
ihincks, line 148:
```suggestion
        the strategy set by the ``twirling.strategy`` field specified in the ``options`` (see
```

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -0,0 +1,268 @@
...
+        :class:`NoiseLearnerOptions` for more details) and sorting them based on the number of
+        times they occur in the various circuits. Then, it runs the noise learning experiment
+        for as many layers as specified by the ``max_layers_to_learn`` field in the ``options``,
+        prioritizing layers that occurr more frequently.
+
+        Args:
+            circuits: An iterable of circuits to run the noise learner program for. Alternatively,
+                pub-like (primitive unified bloc) objects can be specified, such as
```
ihincks, line 156: warning: might break black formatting with line limit

```suggestion
                estimator pub-like (primitive unified bloc) objects can be specified, such as
```

qiskit_ibm_runtime/noise_learner/noise_learner.py
SamFerracin: I think it should be ``twirling-strategy`` because the ``options`` are NoiseLearnerOptions. In the description, they explain what will happen for the various strategies

qiskit_ibm_runtime/noise_learner/noise_learner.py
SamFerracin: There is a note:
```
    .. note::
        Currently, the :class:`.NoiseLearner` is released an experimental feature.
        As such, it is subject to change without notification and its stability is not
        guaranteed.
```
Would you feel safer with a warning, in that it is more explicit?

qiskit_ibm_runtime/noise_learner/noise_learner.py
ihincks: I see, I hadn't noticed the twirling prefix in the new options. I amend the suggestion to:

```suggestion
        the strategy set by the ``twirling_strategy`` field specified in the ``options`` (see
```

With the idea that anything in double backticks to always refer to some kind of code.

qiskit_ibm_runtime/noise_learner/noise_learner.py
SamFerracin: Okay, moved it :) 

```
class NoiseLearner:
    """Class for executing noise learning experiments.

    .. note::
        Currently, the :class:`.NoiseLearner` is released an experimental feature.
        As such, it is subject to change without notification and its stability is not
        guaranteed.
```

PR #1806, opened 2024-07-17
Break up options classes in docs

jyu00:
### Summary

This PR breaks up the classes in `options` package into different sections, to make it easier to find:

image

The table for "Base primitive options" is small because the text is short. @Eric-Arellano or @beckykd, if you know to fix that please let me know.  

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/options/__init__.py
+++ b/qiskit_ibm_runtime/options/__init__.py
@@ -75,7 +75,7 @@ You can also pass dictionaries to each sub-category, for example::
 Classes
 =======
 
-Base Primitive Options
+Base primitive options
 ----------------------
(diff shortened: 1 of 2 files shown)


Eric-Arellano (approved):
Great improvement! @beckykd can you please double check this matches our current style for how to say V1 vs V2? I don't remember what the final approach was.

beckykd:
> Great improvement! @beckykd can you please double check this matches our current style for how to say V1 vs V2? I don't remember what the final approach was.

Yep - "V2 Estimator" and "V1 primitives" are correct. Thanks!

beckykd (approved):

PR #1808, opened 2024-07-18
Scrub RuntimeOptions from docs

jyu00:
### Summary

This PR removes `RuntimeOptions` from docstrings. 

### Details and comments

`RuntimeOptions` was the original options class used when Qiskit Runtime supported only custom programs. Now that we don't support custom programs anymore, exposing that class in the API doc would only cause confusion. Ideally we'd also clean it up from the code, but that requires a bit more work.

### DIFF
--- a/qiskit_ibm_runtime/__init__.py
+++ b/qiskit_ibm_runtime/__init__.py
@@ -196,7 +196,6 @@ Classes
    IBMBackend
    RuntimeJob
    RuntimeJobV2
-   RuntimeOptions
    RuntimeEncoder
    RuntimeDecoder
 """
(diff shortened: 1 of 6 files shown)


kt474 (approved):

qiskit_ibm_runtime/__init__.py
```diff
@@ -196,7 +196,6 @@
    IBMBackend
    RuntimeJob
    RuntimeJobV2
-   RuntimeOptions
```
Eric-Arellano, line 199: This will result in a 404. Do you want the page to redirect to something, or let it 404?

Eric-Arellano: Exactly, https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.RuntimeOptions will 404. https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/0.24/qiskit_ibm_runtime.RuntimeOptions will be fine.

> then I'd say let it 404 since it's normal for classes to be removed

Okay, that's fine with me if there isn't a sensible redirect

PR #1823, opened 2024-07-29
Make PEA option public

ihincks:
### Summary

This PR exposes the PEA mitigation method in the ZNE options. It also exposes another helpful option to do with specifying the points at which extrapolation methods are evaluated.

### Details and comments

Test evidence is incoming.

### DIFF
--- a/qiskit_ibm_runtime/options/zne_options.py
+++ b/qiskit_ibm_runtime/options/zne_options.py
@@ -44,7 +44,7 @@ class ZneOptions:
 
         In particular, suppose an input pub has observable array shape ``obs_shape`` and parameter
         values shape ``par_shape``, with corresponding pub shape
-        ``shape=np.broadcast_shapes(obs_shape, par_shap)``. Then the corresponding pub result will
+        ``shape=np.broadcast_shapes(obs_shape, par_shape)``. Then the corresponding pub result will
         additionally contain:
 
         1. `pub_result.data.evs_extrapolated` and `pub_result.data.stds_extrapolated`,
--- a/qiskit_ibm_runtime/options/zne_options.py
+++ b/qiskit_ibm_runtime/options/zne_options.py
@@ -34,7 +34,52 @@ ExtrapolatorType = Literal[
 
 @primitive_dataclass
 class ZneOptions:
-    """Zero noise extrapolation mitigation options. This is only used by V2 Estimator.
+    """Zero noise extrapolation mitigation options. This is only used by the V2 Estimator.
+
+    ..note::
+
+        Any V2 estimator is guaranteed to return data fields called ``evs`` and ``stds`` that
+        report the desired expectation value estimates and errors, respectively.
+        When ZNE options are enabled in the runtime estimator, additional data is returned.
+
+        In particular, suppose an input pub has observable array shape ``obs_shape`` and parameter
+        values shape ``par_shape``, with corresponding pub shape ``shape``. Then the corresponding
+        pub result will additionally contain:
+
+        1. `pub_result.data.evs_extrapolated` and `pub_result.data.stds_extrapolated`,
+            both with shape ``(*shape, num_extrapolators, num_evaluation_points)``, where
+            ``num_extrapolators`` is the length of the list of
+            ``options.resilience.zne.extrapolators``, and ``num_evaluation_points`` is the length of
+            the list ``options.resilience.extrapolated_noise_factors``. These values provide
+            evaluations of every extrapolator at every specified noise extrapolation value.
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/options/zne_options.py
```diff
@@ -34,7 +34,52 @@
...
+    """Zero noise extrapolation mitigation options. This is only used by the V2 Estimator.
+
+    ..note::
+
+        Any V2 estimator is guaranteed to return data fields called ``evs`` and ``stds`` that
+        report the desired expectation value estimates and errors, respectively.
+        When ZNE options are enabled in the runtime estimator, additional data is returned.
+
+        In particular, suppose an input pub has observable array shape ``obs_shape`` and parameter
+        values shape ``par_shape``, with corresponding pub shape ``shape``. Then the corresponding
```
SamFerracin, line 46:
```suggestion
        values shape ``par_shape``, with corresponding pub shape ``shape = obs_shape * par_shape``. Then the corresponding
```

qiskit_ibm_runtime/options/zne_options.py
```diff
@@ -34,7 +34,52 @@
...
+           ``ensemble_stds_noise_factors`` all have shape ``(*shape, num_noise_factors)`` where
+           ``num_noise_factors`` is the length of ``options.resilience.zne.noise_factors``. These
+           values provide evaluations of the best-fit model at each of the noise amplifications.
+           In the case of no twirling, both ``*stds*`` arrays will be equal, otherwise,
+           ``stds_noise_factors`` is derived from the spread over twirling samples, whereas
+           ``ensemble_stds_noise_factors`` assumes only shot noise and no drift.
+
+        Metadata is formatted as follows:
+
+        - ``PrimitiveResult.metadata["resilience"]["zne"]``
```
SamFerracin, line 65: I'm always a bit hesitant to specify this kind of information because if in a few weeks we change the metadata in the server (e.g., by adding a field), there is a chance that we may forget to update this docstring

qiskit_ibm_runtime/options/zne_options.py
```diff
@@ -34,7 +34,52 @@
...
+            - ``["extrapolated_noise_factors"]`` The extrapolation points corresponding to
+              ``num_evaluation_points`` mentioned above in (1).
+
+        - ``PubResult.metadata["resilience"]["zne"]["extrapolator"]`` which extrapolators were
+          used for this pub in particular.
+
+        Technical note: for single observables with multiple basis terms it might turn out that
+        multiple extrapolation methods are used in _the same_ expectation value, for example, ``XX``
+        gets linearly extrapolated but ``XY`` gets exponentially extrapolated in the observable
+        ``{"XX": 0.5, "XY": 0.5}``. Let's call this a *hetergeneous fit*. The data from (3) is
```
SamFerracin, line 77: What is `(3)`? And in a few lines, what is `(2)`?

qiskit_ibm_runtime/options/zne_options.py
ihincks: We could just remove all of the metadata description.

qiskit_ibm_runtime/options/zne_options.py
ihincks: we'd actually want ``shape = numpy.broadcast_shapes(obs_shape, par_shape)``. Would this be okay?

qiskit_ibm_runtime/options/zne_options.py
ihincks: Ah, they're referencing (1) and (2), but there used to be a (3) in the source I copied this from. Will fix.

qiskit_ibm_runtime/options/zne_options.py
```diff
@@ -48,8 +93,25 @@ class ZneOptions:
             * `"gate_folding_back"` uses 2-qubit gate folding to amplify noise. If the noise
               factor requires amplifying only a subset of the gates, then these gates are selected
               from the back of the topologically ordered DAG circuit.
-
-        noise_factors: Noise factors to use for noise amplification. Default: (1, 3, 5).
+            * `"pea"` uses a technique called probabalistic error amplification (`PEA
```
SamFerracin, line 96: PEA only makes sense if gate twirling is on. Yet, `twirling.enable_gates` is `False` by default. I've forgotten about this many times while playing with PEA, and the performance wasn't great. Do we want to add a note about this, and perhaps even a warning?

ihincks: I believe it turns twirling on for you when pea is enabled. If this is not the case, I think it's a bug.

In any case, yes a note would be worthwhile.

PR #1827, opened 2024-07-30
Add job usage method

### LINKED ISSUE #1819: Provide a usage function on RuntimeJob so metrics don't have to be parsed

The [RuntimeJob.metrics()](https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.RuntimeJob#metrics) method does not document what is in that metrics dictionary.

For the quantum channel the API docs are [here](https://api.quantum-computing.ibm.com/runtime/api#/Jobs/get_job_metrics_jid). For the cloud channel the API docs are [here](https://us-east.quantum-computing.cloud.ibm.com/openapi/#/Jobs/get_job_metrics_jid).

There are 3 different fields in there that talk about usage:

image

It'd be nice if the client didn't have to know which of these to use to get the usage of a terminal-state job and instead a `usage()` method or something like that could abstract that information in the SDK.

This is sort of similar to how `Session.details()` has the `usage_time` field, but at least that's [documented](https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.Session#details).

**Acceptance criteria**

Provide a method on `RuntimeJob` to get the usage time of a job.

kt474:
### Details and comments
Fixes #1819


qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -135,11 +135,23 @@ def cancel_result_streaming(self) -> None:
...
     def metrics(self) -> Dict[str, Any]:
         """Return job metrics.
 
         Returns:
-            Job metrics, which includes timestamp information.
+            A dictionary with job metrics including but not limited to the following:
+
+            * ``timestamps``: Timestamps of when the job was created, started running, and finished.
+            * ``usage``: Job usage metrics
+
```
kt474, line 154: Should we keep this intentionally vague so we have more freedom to update the information returned from the server?

qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -135,11 +135,23 @@ def cancel_result_streaming(self) -> None:
             return
         self._ws_client.disconnect(WebsocketClientCloseCode.CANCEL)
 
+    def usage(self) -> Dict[str, Any]:
+        """Return job usage in seconds."""
+        try:
+            metrics = self._api_client.job_metadata(self.job_id())
+            return metrics.get("usage")
```
kt474, line 142: If we add a `usage()` method what exactly should be returned? This would return:

```
{'quantum_seconds': 0, 'seconds': 0}
```

but if quantum_seconds and seconds are the same, we can just returned one of them

qiskit_ibm_runtime/base_runtime_job.py
jyu00: This is the definition of `usage` from our docs:

> Usage is a measurement of the amount of time the QPU is locked for your workload.

qiskit_ibm_runtime/base_runtime_job.py
jyu00: Yes we should return just 1

ptristan3 (approved):

PR #1828, opened 2024-07-30
Migrate integration tests to test_eagle

### LINKED ISSUE #1683: Update integration tests to use `test_eagle`
With `ibmq_qasm_simulator` officially being sunset, we need to use a different backend for our integration tests. The plan is to use `test_eagle` instead but it is not available on production yet and it returns random results.

In the meantime, we can ask IQP to add the simulator back to the h/g/p's we use for integration tests

ptristan3:
### Summary

Fixes #1683 

### Details and comments

Before merging, the `QISKIT_IBM_QPU` var should be added for all the environments we run the integration tests. Also, the `QISKIT_IBM_INSTANCE` value should be verified.


test/integration/test_ibm_job.py
```diff
@@ -117,7 +122,7 @@ def test_retrieve_completed_jobs(self):
             backend_name=self.sim_backend.name, limit=3, pending=False
         )
         for job in completed_job_list:
-            self.assertTrue(job.status() in ["DONE", "CANCELLED", "ERROR"])
+            self.assertTrue(job.status() in [JobStatus.DONE, JobStatus.CANCELLED, JobStatus.ERROR])
```
kt474, line 125: Shouldn't the job statuses be strings? All jobs should be `RuntimeJobV2`

test/integration/test_retrieve_job.py
```diff
@@ -67,7 +67,8 @@ def test_retrieve_all_jobs(self, service):
         for rjob in rjobs:
             if rjob.job_id() == job.job_id():
                 self.assertEqual(job.program_id, rjob.program_id)
-                self.assertEqual(job.status(), rjob.status())
+                self.assertEqual(job.status(), "QUEUED")
+                self.assertEqual(rjob.status(), JobStatus.QUEUED)
```
kt474, line 71: Same here, shouldn't all jobs be `RuntimeJobV2`?

kt474 (commented):
Thanks! Looks good, just some minor comments.

I'd prefer to hold off on merging until we're more comfortable with the reliability of the test eagle devices

kt474 (approved):

PR #1833, opened 2024-07-31
Execution span classes

### LINKED ISSUE #1832: Add interface to execution span

The interface will allow users to retrieve medium-grained timing information about their sampler jobs.

yaelbh:
### Summary

The new classes `ExecutionSpan` and `ExecutionSpanCollection` will be populated with timing information of sampler jobs.

### Details and comments
Closes #1832.

### DIFF
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -73,14 +73,14 @@ class ExecutionSpanCollection:
     @overload
     def __getitem__(self, idxs: Union[slice, List[int]]) -> "ExecutionSpanCollection": ...
 
-    def __getitem__(self, idxs):
+    def __getitem__(self, idxs: Union[int, slice, List[int]]):
         if isinstance(idxs, int):
             return self._spans[idxs]
         if isinstance(idxs, slice):
             return ExecutionSpanCollection(self._spans[idxs])
         return ExecutionSpanCollection([self._spans[idx] for idx in idxs])
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable[ExecutionSpan]:
         return iter(self._spans)
 
     @property
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -12,7 +12,7 @@
 
 """Execution span classes."""
 
-from typing import Iterable, TypeVar, Union, overload, Tuple, Dict, List
+from typing import Iterable, Union, overload, Tuple, Dict, List
 from datetime import datetime
 from dataclasses import dataclass
 
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -17,15 +17,15 @@ from __future__ import annotations
 import abc
 from datetime import datetime
 import math
-from typing import overload, Iterable, Iterator
+from typing import overload, Iterable, Iterator, Tuple
 
 import numpy as np
 import numpy.typing as npt
 
 
-# Looks like a bug in mypy makes it not recognize `tuple` as an alias,
-# even in Python 3.9, so we disable mypy for the line
-ShapeType = tuple[int, ...]  # type: ignore
+# Python 3.8 does not recognize tuple[<something],
+# in spite of `from __future__ import annotations`
+ShapeType = Tuple[int, ...]
 """A shape tuple representing some nd-array shape."""
 
 
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -119,10 +119,7 @@ class ExecutionSpanSet:
         return str(self.to_list_of_tuples())
 
     def __eq__(self, other: object) -> bool:
-        # TODO: consider changing to a dataclass
-        if not isinstance(other, ExecutionSpanSet):
-            return NotImplemented
-        return self._spans == other._spans
+        return isinstance(other, ExecutionSpanSet) and self._spans == other._spans
 
     def to_list_of_tuples(self) -> List:
         """Return span set in the form of a list of tuples"""
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -83,11 +83,11 @@ class ExecutionSpan:
 class ExecutionSpanSet:
     """A collection of timings for the PUB result."""
 
-    def __init__(self, spans: Sequence[ExecutionSpan]):
+    def __init__(self, spans: Iterable[ExecutionSpan]):
         self._spans = spans
 
     def __len__(self) -> int:
-        return len(self._spans)
+        return len(list(self._spans))
 
     @overload
     def __getitem__(self, idxs: int) -> ExecutionSpan: ...
@@ -98,11 +98,12 @@ class ExecutionSpanSet:
     def __getitem__(
         self, idxs: Union[int, slice, List[int]]
     ) -> Union[ExecutionSpan, "ExecutionSpanSet"]:
+        span_list = list(self._spans)
         if isinstance(idxs, int):
-            return self._spans[idxs]
+            return span_list[idxs]
         if isinstance(idxs, slice):
-            return ExecutionSpanSet(self._spans[idxs])
-        return ExecutionSpanSet([self._spans[idx] for idx in idxs])
+            return ExecutionSpanSet(span_list[idxs])
+        return ExecutionSpanSet(span_list[idx] for idx in idxs)
 
     def __iter__(self) -> Iterator[ExecutionSpan]:
         return iter(self._spans)
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -44,7 +44,7 @@ class ExecutionSpan:
     @property
     def duration(self) -> float:
         """Return the duration"""
-        return (self.stop - self.start).seconds
+        return (self.stop - self.start).total_seconds()
 
     def contains_pub(self, pub_idx: Union[int, Iterable[int]]) -> bool:
         """Returns if a pub is contained"""
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -17,9 +17,6 @@ from datetime import datetime
 from dataclasses import dataclass
 
 
-SliceType = slice
-
-
 @dataclass(frozen=True)
 class ExecutionSpan:
     """Stores an execution time span for a subset of job data."""
@@ -30,7 +27,7 @@ class ExecutionSpan:
     stop: datetime
     """The stop time of the span, in UTC."""
 
-    data_slices: Dict[int, SliceType]
+    data_slices: Dict[int, slice]
     r"""Which data have dependence on this execution span."""
 
     @property
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -18,8 +18,6 @@ from dataclasses import dataclass
 
 
 SliceType = Tuple[int, int]
-ExecutionSpanT = TypeVar("ExecutionSpanT", bound="ExecutionSpan")
-ExecutionSpanCollectionT = TypeVar("ExecutionSpanCollectionT", bound="ExecutionSpanCollection")
 
 
 @dataclass(frozen=True)
@@ -53,7 +51,7 @@ class ExecutionSpan:
         pub_idx = {pub_idx} if isinstance(pub_idx, int) else set(pub_idx)
         return not pub_idx.isdisjoint(self.data_slices)
 
-    def filter_by_pub(self, pub_idx: Union[int, Iterable[int]]) -> ExecutionSpanT:
+    def filter_by_pub(self, pub_idx: Union[int, Iterable[int]]) -> "ExecutionSpan":
         """Returns an ExecutionSpan filtered by pub-"""
         pub_idx = {pub_idx} if isinstance(pub_idx, int) else set(pub_idx)
         slices = {idx: sl for idx, sl in self.data_slices.items() if idx in pub_idx}
@@ -74,7 +72,7 @@ class ExecutionSpanCollection:
         ...
 
     @overload
-    def __getitem__(self, idxs: Union[slice | List[int]]) -> ExecutionSpanCollectionT:
+    def __getitem__(self, idxs: Union[slice | List[int]]) -> "ExecutionSpanCollection":
         ...
         
     def __getitem__(self, idxs):
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -73,8 +73,8 @@ class ExecutionSpan:
         """Construct a span from a tuple"""
         return ExecutionSpan(*list(span_tuple))
 
-    def __str__(self) -> str:
-        return str(self.to_tuple())
+    def __repr__(self) -> str:
+        return f"ExecutionSpan({self.to_tuple()})"
 
 
 class ExecutionSpanSet:
@@ -105,8 +105,8 @@ class ExecutionSpanSet:
     def __iter__(self) -> Iterator[ExecutionSpan]:
         return iter(self._spans)
 
-    def __str__(self) -> str:
-        return str(self.to_list_of_tuples())
+    def __repr__(self) -> str:
+        return f"ExecutionSpanSet({repr(self.to_list_of_tuples())})"
 
     def __eq__(self, other: object) -> bool:
         return isinstance(other, ExecutionSpanSet) and self._spans == other._spans
--- a/qiskit_ibm_runtime/execution_span.py
+++ b/qiskit_ibm_runtime/execution_span.py
@@ -73,7 +73,9 @@ class ExecutionSpanCollection:
     @overload
     def __getitem__(self, idxs: Union[slice, List[int]]) -> "ExecutionSpanCollection": ...
 
-    def __getitem__(self, idxs: Union[int, slice, List[int]]):
+    def __getitem__(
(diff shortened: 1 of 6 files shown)


qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,94 @@
...
+        pub_idx = {pub_idx} if isinstance(pub_idx, int) else set(pub_idx)
+        slices = {idx: sl for idx, sl in self.data_slices.items() if idx in pub_idx}
+        return ExecutionSpan(self.start, self.stop, slices)
+
+
+class ExecutionSpanCollection(Sequence[ExecutionSpan]):
+    """A collection of timings for the PUB result."""
+
+    def __init__(self, spans: Iterable[ExecutionSpan]):
+        self._spans = spans
```
yaelbh, line 64: I think we can either inherit from `Sequence[ExecutionSpan]` or store `_spans`. I prefer the second option, @ihincks what do you think? Anyway we don't need both.

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,94 @@
...
+"""Execution span classes."""
+
+from typing import Union, Iterable, Sequence, TypeVar
+from datetime import datetime
+from dataclasses import dataclass
+
+
+"""The format accepted by ``numpy.ndarray.__getitem__()``."""
+SliceType = tuple[Union[slice, int, list[int]], ...]
+ExecutionSpanT = TypeVar("ExecutionSpanT", bound="ExecutionSpan")
```
yaelbh, line 22: What's the purpose of this line?

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @property
+    def duration(self) -> float:
+        return (self.stop - self.start).seconds
+
+    def contains_pub(self, pub_idx: int | Iterable[int]):
+        pub_idx = {pub_idx} if isinstance(pub_idx, int) else set(pub_idx)
+        return not pub_idx.isdisjoint(self.data_slices)
+
+    def filter_by_pub(self, pub_idx: int | Iterable[int]) -> ExecutionSpanT:
```
yaelbh, line 54:
```suggestion
    def filter_by_pub(self, pub_idx: int | Iterable[int]) -> ExecutionSpan:
```

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,94 @@
...
+
+"""Execution span classes."""
+
+from typing import Union, Iterable, Sequence, TypeVar
+from datetime import datetime
+from dataclasses import dataclass
+
+
+"""The format accepted by ``numpy.ndarray.__getitem__()``."""
+SliceType = tuple[Union[slice, int, list[int]], ...]
```
yaelbh, line 21: I don't see a situation where we'll have something other than `slice` in the data slices. So I think we can remove the definition of `SliceType` here and below replace `SliceType` with `slice`. @ihincks Was there a specific reason to define `SliceType`?

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,94 @@
...
+    :class:`~PubResult`\s :class:`~.DataBin` sharing the same base shape.
+    Therefore, the format of this field is a mapping from pub indexes to 
+    the same format accepted by 
+    NumPy slicing, where each value indicates which slice of each field in the
+    data bin depend on raw data collected during this execution span.
+    """
+
+    @property
+    def duration(self) -> float:
+        return (self.stop - self.start).seconds
```
yaelbh, line 48: I think what we want here is `total_seconds()` and not `seconds`:
```
>>> from datetime import datetime
>>> t1 = datetime.now()
>>> t2 = datetime.now()
>>> t2 - t1
datetime.timedelta(seconds=15, microseconds=241326)
>>> (t2 - t1).seconds
15
>>> (t2 - t1).total_seconds
<built-in method total_seconds of datetime.timedelta object at 0x7f81ce98b1b0>
>>> (t2 - t1).total_seconds()
15.241326
```

qiskit_ibm_runtime/execution_span.py
yaelbh: In fact, it is possible for a batch job to have two different slices, each corresponding to the same pub result. Therefore I think we should go for a list of integers.

qiskit_ibm_runtime/execution_span.py
ihincks: Let's not inherit from `Sequence[ExecutionSpan]`. (Depending on whether you get it from abc.collections or typing, it would buy you different things)

Let's just implement `__iter__ -> Iterable[ExecutionSpan]` , `__len__ -> int`, and

```
@typing.overload
def __getitem__(self, idxs: slice | list[int]) -> ExecutionCollection: ...

@typing.overload
def __getitem__(self, idxs: int) -> ExecutionSpan: ...
```

which will make it implicitly obey the sequence protocol.

qiskit_ibm_runtime/execution_span.py
ihincks: Thanks, I wrote that code blind and didn't check the `time` docs.

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,94 @@
...
+        return min(span.start for span in self._spans)
+
+    @property
+    def stop(self) -> datetime:
+        """The stop time of the entire collection, in UTC."""
+        return max(span.stop for span in self._spans)
+
+    def plot(self):
+        """Show a timing diagram"""
+        pass
```
ihincks, line 94: This is a nice-to-have, and not required initially, if that helps.

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,97 @@
...
+"""Execution span classes."""
+
+from typing import Iterable, TypeVar, Union, overload, Tuple, Dict
+from datetime import datetime
+from dataclasses import dataclass
+
+
+SliceType = Tuple[int, int]
+ExecutionSpanT = TypeVar("ExecutionSpanT", bound="ExecutionSpan")
+ExecutionSpanCollectionT = TypeVar("ExecutionSpanCollectionT", bound="ExecutionSpanCollection")
```
ihincks, line 22: We can safely remove both of these `TypeVar`s. For example, we can change ` -> ExecutionSpanCollectionT` to ` -> "ExecutionSpanCollection"`, noting the quotes.

yaelbh:
@ihincks Maybe we want to add `contains_pub` and `filter_by_pub` methods also to `ExecutionSpanCollection`?

yaelbh:
> @ihincks Maybe we want to add contains_pub and filter_by_pub methods also to ExecutionSpanCollection?

`ExecutionSpanSet.contains_pub` doesn't make sense. On the other hand the example [here](https://github.ibm.com/IBM-Q-Software/requirements-and-priorities/issues/431#issuecomment-87462410) assumes the existence of `ExecutionSpanSet.filter_by_pub`, so I'm implementing it now.

yaelbh:
@ihincks This pull request is ready for review.
@kt474 I'll appreciate if you can prepare a release after this PR is merged.

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,141 @@
...
+            return self._spans[idxs]
+        if isinstance(idxs, slice):
+            return ExecutionSpanSet(self._spans[idxs])
+        return ExecutionSpanSet([self._spans[idx] for idx in idxs])
+
+    def __iter__(self) -> Iterator[ExecutionSpan]:
+        return iter(self._spans)
+
+    def __str__(self) -> str:
+        return str(self.to_list_of_tuples())
```
ihincks, line 101: If only implementing one of repr or str, I usually prefer repr because str calls repr when undefined.

```suggestion
    def __repr__(self) -> str:
        return f"ExecutionSpanSet({self._spans})"
```

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,141 @@
...
+    data bin depend on raw data collected during this execution span.
+    """
+
+    @property
+    def duration(self) -> float:
+        """Return the duration"""
+        return (self.stop - self.start).total_seconds()
+
+    def contains_pub(self, pub_idx: Union[int, Iterable[int]]) -> bool:
+        """Returns if a pub is contained"""
```
ihincks, line 50:
```suggestion
        """Returns whether the pub with the given index has data with dependence on this span.
        
        Args:
            pub_idx: One or more pub indices from the original primitive call.
        
        Returns:
            Whether there is dependence on this span.
        """
```

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,141 @@
...
+        pub_idx = {pub_idx} if isinstance(pub_idx, int) else set(pub_idx)
+        return not pub_idx.isdisjoint(self.data_slices)
+
+    def filter_by_pub(self, pub_idx: Union[int, Iterable[int]]) -> "ExecutionSpan":
+        """Returns an ExecutionSpan filtered by pub"""
```
ihincks, line 55:
```suggestion
        """Return a new span whose slices are filtered to the provided indices.
        
        For example, if this span contains slice information for pubs with indices 1,3,4 and `[1,4]` is provided,
        then the span returned by this method will contain slice information for only those two indices, but
        be identical otherwise.
        
        Args:
            pub_idx: One or more pub indices from the original primitive call.
            
        Returns:
            A new filtered span.
        """
```

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,141 @@
...
+
+    data_slices: Dict[int, SliceType]
+    r"""Which data have dependence on this execution span.
+
+    Data from the primitives are array-based, with every field in a 
+    :class:`~PubResult`\s :class:`~.DataBin` sharing the same base shape.
+    Therefore, the format of this field is a mapping from pub indexes to 
+    the same format accepted by 
+    NumPy slicing, where each value indicates which slice of each field in the
+    data bin depend on raw data collected during this execution span.
```
ihincks, line 41: Re-reading this, it's probably a bit too technical.

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,141 @@
...
+        return ExecutionSpan(self.start, self.stop, slices)
+
+    def to_tuple(self) -> Tuple:
+        """Return span in the form of a tuple"""
+        return (self.start, self.stop, self.data_slices)
+
+    @classmethod
+    def from_tuple(cls, span_tuple: Tuple) -> "ExecutionSpan":
+        """Construct a span from a tuple"""
+        return ExecutionSpan(*list(span_tuple))
```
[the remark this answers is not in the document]
yaelbh: In various places I found myself writing the same code again and again, so I decided to write functions that implement it, and it indeed proved itself more convenient.

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,141 @@
...
+    def to_list_of_tuples(self) -> List:
+        """Return span set in the form of a list of tuples"""
+        return [span.to_tuple() for span in self]
+
+    @classmethod
+    def from_list_of_tuples(cls, list_of_tuples: Sequence[Tuple]) -> "ExecutionSpanSet":
+        """ "Construct span set from a list of tuples"""
+        return ExecutionSpanSet(
+            [ExecutionSpan.from_tuple(span_tuple) for span_tuple in list_of_tuples]
+        )
```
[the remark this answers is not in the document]
yaelbh: See answer in a previous comment.

qiskit_ibm_runtime/execution_span.py
yaelbh: This suggestion only shows the addresses of the execution spans, not their content.

qiskit_ibm_runtime/execution_span.py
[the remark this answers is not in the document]
ihincks: Could you show a couple of examples please? I'm concerned that these are too specific to justify putting them in the namespace.

there is also this pattern, more similar to a `namedtuple`
```python
def __iter__(self):
    yield self.start
    yield self.stop
    yield self.data_slices

...

# convert ExecutionSpan to tuple
tuple(span)

# convert iterable to ExecutionSpan
ExecutionSpan(*it)
```

qiskit_ibm_runtime/execution_span.py
ihincks: Since this PR implements `ExecutionSpan` as a dataclass, I think that the dataclass repr (and the list repr) will propagate, and not show the id.

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,146 @@
...
+        """The total duration of this collection, in seconds."""
+        return (self.stop - self.start).total_seconds()
+
+    def filter_by_pub(self, pub_idx: Union[int, Iterable[int]]) -> "ExecutionSpanSet":
+        """Returns an ExecutionSpanSet filtered by pub"""
+        return ExecutionSpanSet([span.filter_by_pub(pub_idx) for span in self])
+
+    def plot(self) -> None:
+        """Show a timing diagram"""
+        raise NotImplementedError
```
ihincks, line 146:
```suggestion
```

qiskit_ibm_runtime/utils/json.py
```diff
@@ -425,6 +439,14 @@ def object_hook(self, obj: Any) -> Any:
                 return PubResult(**obj_val)
             if obj_type == "PrimitiveResult":
                 return PrimitiveResult(**obj_val)
+            if obj_type == "ExecutionSpan":
+                new_slices = {}
+                for task_id, task_slice in obj_val["data_slices"].items():
+                    new_slices[int(task_id)] = slice(task_slice[0], task_slice[1])
```
ihincks, line 445: This suggestion is made in case we ever choose to start using task_slice.step, in which case this decoder will still work.
```suggestion
                    new_slices[int(task_id)] = slice(*task_slice)
```

qiskit_ibm_runtime/utils/validations.py
```diff
@@ -126,3 +130,39 @@ def validate_job_tags(job_tags: Optional[List[str]]) -> None:
...
+        for task_id, task_slice in exspan.data_slices.items():
+            if task_slice.start != slice_ends[task_id]:
+                return False
+            slice_ends[task_id] = task_slice.stop
+
+    for pub_length, res in zip(slice_ends, result):
+        if len(res.data.shape) == 0:
+            expected_length = 1
+        else:
+            expected_length = prod(res.data.shape)
```
ihincks, line 163: math.prod returns 1 for zero-length inputs
```suggestion
        expected_length = prod(res.data.shape)
```

qiskit_ibm_runtime/execution_span.py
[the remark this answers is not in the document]
yaelbh: I don't incline strongly towards the addition of these methods. I thought they could serve as a nice and useful addition, but if there's a problem with them then I'll just drop.

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,123 @@
...
+    """Stores an execution time span for a subset of job data."""
+
+    start: datetime
+    """The start time of the span, in UTC."""
+
+    stop: datetime
+    """The stop time of the span, in UTC."""
+
+    data_slices: Dict[int, slice]
+    r"""Which data have dependence on this execution span."""
```
ihincks, line 31:
```suggestion
    r"""Which data have dependence on this execution span.
    
    This is a mapping from pub indices to data slices, where:
    
    * A pub index is its 0-indexed position inside of the primitive run call. 
    * A data slice references one or more shots from data, indicated that the corresponding 
      pub result data was measured within the time window reported by this execution span.
      
      In the case of a shaped pub, i.e. a pub where the circuit is parametric and multiple 
      parameter sets are specified in some array, a data slice references the *flattened* pub
      result data, where flattening is done in row-major order and the shot axis is last. So for 
      instance, if 32 shots are drawn from a pub with shape ``(3,)``, then there are ``96`` 
      total shots ordered as 32-32-32, and ``slice(20, 40)`` would reference shots 20 through 
      39 inclusive.
    """
```

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,123 @@
...
+        Returns:
+            A new filtered span.
+        """
+        pub_idx = {pub_idx} if isinstance(pub_idx, int) else set(pub_idx)
+        slices = {idx: sl for idx, sl in self.data_slices.items() if idx in pub_idx}
+        return ExecutionSpan(self.start, self.stop, slices)
+
+
+class ExecutionSpanSet:
+    """A collection of timings for the PUB result."""
```
ihincks, line 69:
```suggestion
    """A collection of timings for pub results.
    
    This class is a list-like containing :class:`~.ExecutionSpan`\\s, where each execution span 
    represents a time window of data collection and points to the data that was collected 
    during that window.
    
    Because these windows sometimes contain pieces of concurrent classical processing, It is possible for 
    them to overlap.
    """
```

qiskit_ibm_runtime/execution_span.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def __getitem__(
+        self, idxs: Union[int, slice, List[int]]
+    ) -> Union[ExecutionSpan, "ExecutionSpanSet"]:
+        span_list = list(self._spans)
+        if isinstance(idxs, int):
+            return span_list[idxs]
+        if isinstance(idxs, slice):
+            return ExecutionSpanSet(span_list[idxs])
+        return ExecutionSpanSet([span_list[idx] for idx in idxs])
```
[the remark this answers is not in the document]
yaelbh: Fixed in 997b3d78

ihincks (approved):

PR #1836, opened 2024-08-01
Fix failing integration tests 

kt474:
### Summary

#1805 accidentally overwrote the deprecation warnings in #1816, adding them back here. 

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_noise_learner.py
+++ b/test/integration/test_noise_learner.py
@@ -13,6 +13,7 @@
 """Integration tests for NoiseLearner."""
 
 from copy import deepcopy
+from unittest import SkipTest
 import numpy as np
 
 from qiskit.circuit import QuantumCircuit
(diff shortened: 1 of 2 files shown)


SamFerracin (approved):

test/integration/test_noise_learner.py
```diff
@@ -33,7 +34,10 @@ class TestIntegrationNoiseLearner(IBMIntegrationTestCase):
 
     def setUp(self) -> None:
         super().setUp()
-        self.backend = "test_eagle"
+        try:
+            self.backend = self.service.backend("test_eagle")
+        except:
+            raise SkipTest("test_eagle not available in this environment")
```
kt474, line 40: Going to skip tests if test_eagle is not available, we can update this as a part of https://github.com/Qiskit/qiskit-ibm-runtime/pull/1828

I'll also need to reach out and make sure the test accounts we use all have access to `test_eagle` and `test_eagle_us-east`

PR #1844, opened 2024-08-06
Reformatting NoiseLearner results

SamFerracin:
### Summary
Before exposing an option to pass noise models to the estimator, `NoiseLearnerResult` needs to be reformatted to match the format required on the server side.


chriseclectic (changes requested):
Suggestions to make it more compatible with upcoming Aer PauliLindbladError

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -14,47 +14,117 @@
...
-            qubits.
-        ValueError: If ``generators`` and ``rates`` have different lengths.
+        ValueError: If ``paulis`` and ``rates`` have different lengths.
     """
 
-    circuit: QuantumCircuit
-    qubits: Sequence[int]
-    generators: PauliList
-    rates: Sequence[float]
+    def __init__(self, paulis: PauliList, rates: Sequence[float]) -> None:
```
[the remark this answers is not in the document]
SamFerracin: I actually prefer coercing

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -14,47 +14,117 @@
...
 from qiskit.quantum_info import PauliList
 
 
-@dataclass(frozen=True)
-class NoiseLearnerDatum:
-    """A container for a single noise learner datum.
+class PauliLindbladError:
+    """
+    A container for the generators and rates of an error channel in the sparse Pauli-Lindblad noise
+    model.
```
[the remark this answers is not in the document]
SamFerracin: I like this! A bunch of methods of the AER one (like the `to_pauli_error`) seems like a nice to have, it will be nice to replace the two classes with one. The surviving one should probably live in qiskit

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -14,47 +14,136 @@
...
+class PauliLindbladError:
+    r"""A Pauli error channel generated by a Pauli Lindblad dissipator.
+
+    This operator represents an N-qubit quantum error channel
+    :math:`E(ρ) = e^{\sum_j r_j D_{P_j}}(ρ)` generated by Pauli Lindblad dissipators
+    :math:`D_P(ρ) = P ρ P - ρ`, where :math:`P_j` are N-qubit :class:`~.Pauli` operators.
+
+    The list of Pauli generator terms are stored as a :class:`~.PauliList` and can be
+    accessed via the :attr:`generators` attribute. The array of dissipator rates
+    :math:`r_j` can be accessed via the :attr:`rates` attribute.
```
ihincks, line 34:
```suggestion
    accessed via the :attr:`~generators` attribute. The array of dissipator rates
    :math:`r_j` can be accessed via the :attr:`~rates` attribute.
```

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -14,47 +14,136 @@
...
+
+    The equivalent Pauli error channel can be constructed as a composition
+    of single-Pauli channel terms
+
+    .. math::
+
+        E = e^{\sum_j r_j D_{P_j}} = \prod_j e^{r_j D_{P_j}}
+        = prod_j \left( (1 - p_j) S_I + p_j S_{P_j} \right)
+
+    where :math:`p_j = \frac12 - \frac12 e^{-2 r_j}`.
```
ihincks, line 44: Link to equation in the paper?

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -14,47 +14,136 @@
...
-    def __post_init__(self) -> None:
-        if len({self.circuit.num_qubits, len(self.qubits), self.generators.num_qubits}) != 1:
-            raise ValueError("Mistmatching numbers of qubits.")
-        if len(self.generators) != len(self.rates):
-            msg = f"``generators`` has length {len(self.generators)}, but "
-            msg += f"``rates`` has length {len(self.rates)}."
+        if len(generators) != len(self._rates):
+            msg = f"``generators`` has length {len(generators)} "
+            msg += f"but ``rates`` has length {len(self.rates)}."
             raise ValueError(msg)
```
ihincks, line 61: This might catch more edge cases
```suggestion
        if (len(generators),) != self._rates.shape:
            raise ValueError(
                f"``generators`` has length {len(generators)} "
                f"but ``rates`` has shape {self._rates.shape}."
            )
```

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -14,47 +14,136 @@
...
+        r"""
+        The number of qubits in this :class:`~.PauliLindbladError`.
+        """
+        return self.generators.num_qubits
+
+    def __repr__(self) -> str:
+        return f"PauliLindbladError(generators={self.generators}, rates={self.rates.tolist()})"
+
+
+class LayerError:
```
ihincks, line 88: What's the argument for for having both `circuit` and `qubits`? Why can't the circuit just be ISA, and only refer to physical qubits?

The circuit inputs to the primitives are ISA, which makes me prefer to reference everything without a layout. qubit mappings tend to be confusing.

SamFerracin: I don't have a good answer for that, rather than "this is how things work in pec-runtime."
Elaborating on this, we'd like to have "block" here, but we don't have a well-established concept of block in qiskit-runtime yet. So next in line is "circuit" and "qubits," which is what a block is made of (where I am assuming that the definition of block that we have in pec-runtime is here to stay). Having ISA circuits with physical qubits would not be in line with the definition of block in pec-runtime, and I worry that we'll end up having too many concepts of block here and there

Also interested in hearing @chriseclectic's opinion though

qiskit_ibm_runtime/utils/noise_learner_result.py
SamFerracin: True

qiskit_ibm_runtime/utils/noise_learner_result.py
ihincks: This is non-blocking for this PR, I just wanted to poke at it. I understand a bit better now that qubits+circuit is more or less supposed to mimic a `CircuitInstruction`, if we had available to us proper circuit blocks.

chriseclectic: The qubits referrer to noise / twirled qubits of a block, which could be in parallel with other blocks, so is very far from an ISA circuit block which would always be a full width layer. Eg if you twirled a two qubit gate on qubits (a,b) out of N this would be a 2-qubit circuit with qubits (a, b).If you put in ISA circuits you are back to the issue of not knowing which subset of qubits was twirled and hence which ones the noise model applies to, unless you pad all your generators to be ISA generators too, but then you can'tdo things like noise model stitching with independent parallel blocks etc

chriseclectic (approved):

PR #1853, opened 2024-08-12
Adding logic to serialize and deserialize noise learner results

SamFerracin:
### Summary
This PR adds the necessary logic to serialize and deserialize noise learner results (specifically, the `PauliLindbladError` and `LayerError` classes). Additionally, it ensures that backward compatibility by modifying the `NoiseLearnerDecoder` class


qiskit_ibm_runtime/utils/json.py
```diff
@@ -350,14 +351,16 @@ def object_hook(self, obj: Any) -> Any:
...
-                return _deserialize_from_settings(
-                    mod_name=obj["__module__"],
-                    class_name=obj["__class__"],
-                    settings=_cast_strings_keys_to_int(obj_val),
-                )
+            if obj_type == "settings":
+                if (
+                    obj["__module__"].startswith("qiskit.quantum_info.operators")
+                    or obj["__module__"] == "qiskit_ibm_runtime.utils.noise_learner_result"
+                ):
```
ihincks, line 358: this won't be correctly black'd, but what about

```suggestion
                if obj["__module__"].startswith(("qiskit.quantum_info.operators", "qiskit_ibm_runtime.utils.noise_learner_result")):
```

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -91,6 +91,11 @@ def num_qubits(self) -> int:
         """
         return self.generators.num_qubits
 
+    @property
+    def settings(self) -> dict:
+        """Return settings."""
+        return {"generators": self.generators, "rates": self.rates}
+
```
ihincks, line 98: I realize this may be out of your (our) hands, but this attribute just looks like namespace pollution to me. It's not obvious to me how these things should be considered "settings", that this attribute should be public, etc. 

For example ``__json__()`` or ``__state__()`` would be nicer to me.

I'm more complaining about the state of affairs of our ser/de rather than trying to block this PR.

qiskit_ibm_runtime/utils/json.py
jyu00: Why does `noise_learner_result` live in `utils` and not `noise_learner`?

qiskit_ibm_runtime/utils/noise_learner_result.py
jyu00: +1. It's `settings` in `quantum_info` due to historical reasons, but we are free to use a different method.

qiskit_ibm_runtime/utils/json.py
```diff
@@ -384,6 +387,10 @@ def object_hook(self, obj: Any) -> Any:
                 if shape is not None and isinstance(shape, list):
                     shape = tuple(shape)
                 return DataBin(shape=shape, **obj_val["fields"])
+            if obj_type == "LayerError":
+                return LayerError(**obj_val)
+            if obj_type == "PauliLindbladError":
+                return PauliLindbladError(**obj_val)
```
jyu00, line 395: These won't actually be used, right? Since they are using `settings` instead.

jyu00 (commented):
Maybe some tests?

ihincks (approved):
+1 to Jessie's suggestion to add ser/de tests for the new things. Other than that, LGTM.

SamFerracin:
@jyu00 
> Why does noise_learner_result live in utils and not noise_learner?
I had try putting it there, but I got stuck in some circular dependencies. IIRC, it had to do with the decoder having to import the types used by the results

jyu00:
> @jyu00
> 
> > Why does noise_learner_result live in utils and not noise_learner?
> 
> I had try putting it there, but I got stuck in some circular dependencies. IIRC, it had to do with the decoder having to import the types used by the results

But the decoder doesn't import `PauliLindbladError` and `LayerError` anymore now that it uses `_json`

PR #1876, opened 2024-08-16
Fix 404 link in release note

Eric-Arellano:
This page was removed from current API docs, so we need to use a historical version.


kt474 (approved):

kt474:
Cherry picked into `stable/0.28`

PR #1881, opened 2024-08-20
Adding methods to draw noise learner results

SamFerracin:
This PR adds methods to draw `LayerError` objects on the coupling map of the desired backend.

Example:
```
from qiskit.quantum_info import PauliList
from qiskit_ibm_runtime.fake_provider.local_service import QiskitRuntimeLocalService
from qiskit_ibm_runtime.utils.noise_learner_result import LayerError, PauliLindbladError

service = QiskitRuntimeLocalService()
backend = service.backend("fake_hanoi")

circuit = QuantumCircuit(4)
qubits = [0, 1, 2, 3]

generators = PauliList(["IIIX", "IIXI", "IXII", "YIII", "ZIII", "XXII", "ZZII"])
rates = [0.04, 0.01, 0.02, 0.005, 0.02, 0.01, 0.02]
error = PauliLindbladError(generators, rates)

layer_error = LayerError(circuit, qubits, error)

fig = layer_error.draw_map(backend)
```

It produces an interactive plot that looks like this:
[example.pdf](https://github.com/user-attachments/files/16681982/example.pdf)


qiskit_ibm_runtime/visualization/utils.py
```diff
@@ -0,0 +1,855 @@
...
+"""
+Utility functions for visualizing qiskit-ibm-runtime's objects.
+"""
+
+from typing import List
+
+
+def get_qubits_coordinates(num_qubits: int) -> List[List[int]]:
+    r"""
+    Returns a list of coordinates in the form ``(row, column)`` for drawing a QPU map with
```
ihincks, line 23:
```suggestion
    Return a list of coordinates in the form ``(row, column)`` for drawing a QPU map with
```

qiskit_ibm_runtime/visualization/utils.py
```diff
@@ -0,0 +1,855 @@
...
+# pylint: disable=too-many-return-statements
+
+"""
+Utility functions for visualizing qiskit-ibm-runtime's objects.
+"""
+
+from typing import List
+
+
+def get_qubits_coordinates(num_qubits: int) -> List[List[int]]:
```
ihincks, line 21: Consider a helper with the following signature:

```python

_Row = namedtuple("_Row", "num_qubits,x_offset,down_links")
"""Tuple specifying a dense horizontal row of the standard heavy hex planar embedding.

Args:
    num_qubits: The number of qubits in the dense row.
    x_offset: The integer x-offset of the row.
    down_links: The positions of dense row, wrt the first qubit, at which to place qubits below.
"""

def _heavy_hex_coords(rows: list[_Row], row_major: bool) -> list[tuple[int, int]]:
    """Generate heavy hex coordinates for the given rows.

    Args:
        rows: The dense rows specifications.
        row_major: Whether qubits should be ordered by their ``(x, y)`` locations (as opposed to ``(y, x)``).

    Returns:
        A list of qubit coordinates, where list position corresponds with qubit index.
    """
```

Under this scheme, something like bangkok (27) can be specified as:
```
_heavy_hex_coords(
    [_Row(0, 0, [3, 7]), _Row(10, 0, [1, 5, 9]), _Row(10, 1, [2, 6])],
    False
)
```
Noting that the first _Row with no qubits is to specify qubits 6 and 17 here: 
image


Likewise, our 127s can be specified as
```
_heavy_hex_coords(
    [
        _Row(14, 0, [4, 8, 12]), 
        a:=_Row(15, 0, [2, 6, 10, 14]), 
        b:=_Row(15, 0, [4, 8, 12]), 
        a,
        b,
        a
        _Row(14, 1, []), 
    ],
    True
)
```

If you find a better scheme, go nuts, but this seems easier than keeping track of huge lists.

qiskit_ibm_runtime/visualization/utils.py
```diff
@@ -0,0 +1,855 @@
...
+            [8, 4],
+            [8, 5],
+            [8, 6],
+            [8, 7],
+            [8, 8],
+            [8, 9],
+            [8, 10],
+        ]
+
+    if num_qubits == 127:
```
ihincks, line 287: Looks like this is missing some herons, like 133.

qiskit_ibm_runtime/visualization/draw_layer_error_map.py
```diff
@@ -0,0 +1,281 @@
...
+        return discreet_colorscale[-1]
+    if rate == 0:
+        return default
+    return discreet_colorscale[int(np.round(rate, 3) * 1000)]
+
+
+def draw_layer_error_map(
+    layer_error: LayerError,
+    backend: BackendV2,
+    coordinates: Optional[List[List[int]]] = None,
```
ihincks, line 77: Stylistically, I'd like to move towards `list` in this repo. You'll need annotations from future.

qiskit_ibm_runtime/visualization/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Functions and classes for visualizing qiskit-ibm-runtime's objects.
```
ihincks, line 14: Please add to sphinx too.

qiskit_ibm_runtime/visualization/draw_layer_error_map.py
```diff
@@ -0,0 +1,281 @@
...
+        raise ValueError("Invalid ``discreet_colorscale.``")
+
+    if rate >= 1:
+        return discreet_colorscale[-1]
+    if rate == 0:
+        return default
+    return discreet_colorscale[int(np.round(rate, 3) * 1000)]
+
+
+def draw_layer_error_map(
```
ihincks, line 74: Pros and cons of not adding this to :class:`LayerError` directly? Analagously, `QuantumCircuit.draw`.

ihincks (commented):
Nice work Sam. Would you mind giving it a shot with the 433 qubit topology and noise everywhere to see if it crashes and burns or not?

qiskit_ibm_runtime/visualization/draw_layer_error_map.py
SamFerracin: I had a good reason, but I cannot recall now. I assume something changed compared to the initial implementation that I had in mind.

Happy to move it, but I'd call it something other than "draw" (maybe "draw_map"?), because we may want to add more plotters in the future for noise learner results

qiskit_ibm_runtime/visualization/utils.py
SamFerracin: I thought of creating a nice interface, but I think that should live in qiskit rather than on qiskit-ibm-runtime.

Right now, qiskit uses big lists too, and it defines them inside a function ([here](https://github.com/Qiskit/qiskit/blob/main/qiskit/visualization/gate_map.py#L96)], so I cannot even import them, I am forced to copy paste them.

If we add an interface in this PR, I think we should do it with the idea of porting it to qiskit, and that is a project in and of its own (people may see it and come back with comments, requests to change or add stuff..). 
I think that a more sensible way forward may be the following:
- Use big lists in this PR, same as qiskit
- Open a PR in qiskit to add an interface that allows getting rid of the maps
- Once available, start using the interface in qiskit-ibm-runtime

Any thoughts?

ihincks: >  I think that should live in qiskit rather than on qiskit-ibm-runtime[...]we should do it with the idea of porting it to qiskit[...] Open a PR in qiskit
I'm not so sure about this because qiskit is intended to be hardware agnostic, whereas qiskit-ibm-runtime is specifically focused on IBM hardware.

qiskit_ibm_runtime/visualization/draw_layer_error_map.py
ihincks: I kind of like not having to remember what to import from where to draw stuff, so my preference is often on drawing methods. Yes, I agree that just draw is not sensible, and I might make it even more specific then `draw_map`.

qiskit_ibm_runtime/visualization/utils.py
SamFerracin: Mmmh okay.. This changes things then

SamFerracin: As agreed privately, I'll do this in a subsequent PR

qiskit_ibm_runtime/visualization/utils.py
SamFerracin: There's nowhere I can copy/paste them from. Is it okay if I add support for 133 Heron when I refactor the coordinates?

ihincks: sure

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -185,6 +198,53 @@ def num_qubits(self) -> int:
...
         return len(self.qubits)
 
+    def draw_map(
+        self,
+        backend: BackendV2,
+        coordinates: Optional[list[list[int]]] = None,
+        *,
+        colorscale: str = "Bluered",
+        color_no_data: str = "lightgray",
+        edges_n_segs: int = 16,
```
ihincks, line 208: qiskit tends to use the full `num` for numbers.

```suggestion
        edges_num_segs: int = 16,
```

or how about the more verbose

```suggestion
        num_edge_segments: int = 16
```

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -185,6 +198,53 @@ def num_qubits(self) -> int:
...
+        edges_width: float = 4,
+        height: int = 500,
+        plot_bgcolor: str = "white",
```
ihincks, line 211: `plot` seems redundant, and remove acronyms where reasonable

```suggestion
        background_color: str = "white",
```

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -185,6 +198,53 @@ def num_qubits(self) -> int:
...
 
+    def draw_map(
+        self,
+        backend: BackendV2,
+        coordinates: Optional[list[list[int]]] = None,
+        *,
+        colorscale: str = "Bluered",
+        color_no_data: str = "lightgray",
+        edges_n_segs: int = 16,
+        edges_width: float = 4,
```
ihincks, line 209:
```suggestion
        edge_width: float = 4,
```

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -185,6 +198,53 @@ def num_qubits(self) -> int:
         """
         return len(self.qubits)
 
+    def draw_map(
+        self,
+        backend: BackendV2,
+        coordinates: Optional[list[list[int]]] = None,
```
ihincks, line 204: to emphasize the (x,y) innards
```suggestion
        coordinates: Optional[list[tuple[int, int]]] = None,
```

ihincks: It's not obvious to me why coordinates is the only kw argument thats not kw-only.

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -185,6 +198,53 @@ def num_qubits(self) -> int:
...
+        radius: float = 0.25,
+        width: int = 800,
+    ) -> go.Figure:
+        r"""
+        Draws a map view of a this layer error.
+
+        Args:
+            backend: The backend on top of which the layer error is drawn.
+            coordinates: A list of coordinates in the form ``(row, column)`` that allow drawing each
+                qubit in the given backend on a 2D grid.
```
ihincks, line 221: (These comments also apply to ` draw_layer_error_map`)

We could be clearer here about what is needed, why, and the interplay. Also, it seems like the only property of the backend that we actually need is the coupling map, so can we just pass that instead? Argue back if you want, but in any case, I think the docs need to say something about how the backend is needed to get a coupling map.

```suggestion
            coupling_map: A device coupling map.
            coordinates: A list of coordinates in the form ``(row, column)`` specifying the location to draw
                each qubit. If not provided, this method will attempt to retrieve a pre-defined list based on 
                the qubit count and a heavy-hex embedding.
```

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -101,6 +104,16 @@ def num_qubits(self) -> int:
         """
         return self.generators.num_qubits
 
+    def n_body(self, n: int) -> PauliLindbladError:
```
ihincks, line 107: feels to me like there's a verb missing for something that's not an attribute

```suggestion
    def restrict_num_bodies(self, num_bodies: int) -> PauliLindbladError:
```

qiskit_ibm_runtime/utils/noise_learner_result.py
ihincks: As we add more plotting stuff, it seems like we'll probably want coordinates/coupling map combinations more and more. Should we consider adding this to vis.utils ?

```python
class Embedding:
    coupling_map: CouplingMap
    coordinates: list[tuple[int, int]]
 
    @staticmethod
    def from_backend(backend): pass
```

qiskit_ibm_runtime/utils/noise_learner_result.py
SamFerracin: I think that the original reason (which btw is questionable) was that all that was coming after `*` was a plotly thing. For example, `plot_bgcolor` is the name of the field that needs to be changed in a plotly figure to change the background color. But this does not matter anymore now that:
- `plot_bgcolor` changed to `background_color`
- a few args sneaked their way in without being plotly standard fields (such as "radius").
I will just remove the `*`

qiskit_ibm_runtime/utils/noise_learner_result.py
SamFerracin: No, I don't think it's a bad idea. At the end of the day, coupling map and coordinates go hand in hand. It certainly would signal intention much more

ihincks: For the plotter, the type could be `Embedding | BackendV2`, with the idea that if a backend is given, Embedding.from_backend is called, and if they want custom embedding, they supply it?

SamFerracin: I agree with this.
My only concern (and feel free to disagree) is that this info should not be in the visualization utils, but somewhere else. Where? Maybe in utils, or even better inside `Backend`. Therefore, I am a little bit hesitant to add an `Embedding` class here

> or even better inside `Backend`.

Backends already hold the coupling map, which is a "fundamental" property. Coordinates are not fundamental from the point of view of visualization, in the sense that qubits can be placed in different coordinates and that won't change nay important property of the backend itself. Yet, they are needed in many different places (including qiskit), always together with the coupling map, which makes me believe that it would be very convenient to have a `backend.standard_coordinates` type of argument.

So my preference would either be to add them to `Backend` (which may take a bit though, and will add overhead for adding a new backend) or to have an `Embedding` class in `utils` and not in `visualization` (but in this case, I'd like to know if we'll need to add more info to this class in preparation for other visualization tools)

ihincks: If you want embedding in a different module, that's fine with me. I agree that adding it to the backend would be overhead.

qiskit_ibm_runtime/utils/embeddings.py
```diff
@@ -0,0 +1,484 @@
...
+        coupling_map: A coupling map specifying how the qubits in the embedding are connected.
+    """
+
+    def __init__(
+        self,
+        coordinates: List[Tuple[int, int]],
+        coupling_map: Union[List[Tuple[int, int]], CouplingMap],
+    ) -> None:
+        num_qubits = len(coordinates)
+        if any(q0 > num_qubits or q1 > num_qubits for (q0, q1) in coupling_map):
```
ihincks, line 40: I assume `for (q0, q1) in coupling_map)` works if coupling map is a list of tuple or a CouplingMap?

SamFerracin: Yes, that right, CouplingMap is an iterable. This is also tested where I do, e.g., `e = Embedding.from_backend(self.vigo)`

ihincks (approved):

garrison:
This change appears to be causing a CI failure in the Qiskit addon for circuit cutting: https://github.com/Qiskit/qiskit-addon-cutting/actions/runs/10968529221/job/30459985317

The problem is that plotly is being imported by qiskit-ibm-runtime, but it is not listed as a dependency.

PR #1888, opened 2024-08-23
Add "private" environment option

### LINKED ISSUE #1811: Add private job attribute

Add a `private` attribute to runtime jobs to differentiate them from public jobs. 

**Acceptance criteria**

kt474:
### Summary

wip

### Details and comments
Fixes #1811

### DIFF
--- a/qiskit_ibm_runtime/api/clients/runtime.py
+++ b/qiskit_ibm_runtime/api/clients/runtime.py
@@ -60,6 +60,7 @@ class RuntimeClient(BaseBackendClient):
         max_execution_time: Optional[int] = None,
         start_session: Optional[bool] = False,
         session_time: Optional[int] = None,
+        private: Optional[bool] = False,
         channel_strategy: Optional[str] = None,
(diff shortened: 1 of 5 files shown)


ptristan3 (approved):

mriedem:
Does this work for the cloud channel? I don't see a `private` field in the POST /jobs endpoint for the qiskit runtime cloud API. What happens if specified in that case? Does it blow up?

mriedem:
> Does this work for the cloud channel? I don't see a `private` field in the POST /jobs endpoint for the qiskit runtime cloud API. What happens if specified in that case? Does it blow up?

Looks like it doesn't blow up, but probably doesn't do anything either. If this isn't supported for the cloud channel, shouldn't the docs mention that?

kt474:
> Looks like it doesn't blow up, but probably doesn't do anything either. If this isn't supported for the cloud channel, shouldn't the docs mention that?

Yeah it's not supported for the cloud channel - I can update the docs

PR #1890, opened 2024-08-26
Port `Noise_model.from_dict()`

kt474:
### Details and comments
Fixes #


qiskit_ibm_runtime/utils/noise_model.py
```diff
@@ -0,0 +1,187 @@
...
+from qiskit.circuit import Reset
+from qiskit.circuit.library.generalized_gates import PauliGate, UnitaryGate
+
+from qiskit_aer.noise.errors.quantum_error import QuantumError
+from qiskit_aer.noise.errors.readout_error import ReadoutError
+from qiskit_aer.noise.noiseerror import NoiseError
+from qiskit_aer.noise.noise_model import NoiseModel
+
+
+def from_dict(noise_dict: Dict[str, Any]) -> NoiseModel:
```
SamFerracin, line 30: Should a deprecation warning be added as discussed on Slack?
cc'ing @jyu00

jyu00: We don't need to deprecate the function itself, since utility functions are not public and can be removed at any time.

We should, however, issue a deprecation warning if [noise_model](https://github.com/Qiskit/qiskit-ibm-runtime/blob/8c612d1343c2b8fd800fd0562974be783c58bdaa/qiskit_ibm_runtime/options/simulator_options.py#L56) options is used outside of local testing mode, so we can eventually remove this `from_dict` function. But that can also be a separate PR.

SamFerracin (approved):

jyu00:
I'm also a bit concerned that we'd essentially be using an Aer <0.15 `from_dict()` but >0.15 `to_dict()`. So we should have a test case ensuring this combination works, if we don't already have one.

qiskit_ibm_runtime/utils/noise_model.py
kt474: I'll handle that in a separate PR

kt474:
> I'm also a bit concerned that we'd essentially be using an Aer <0.15 `from_dict()` but >0.15 `to_dict()`. So we should have a test case ensuring this combination works, if we don't already have one.

We have a test for that [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/unit/test_data_serialization.py#L133)

PR #1896, opened 2024-08-29
Don't create new session in `from_id()`

### LINKED ISSUE #1895: Getting the details of an existing sessions submits a new session
**Describe the bug**

When trying to get the details of a previously submitted session, it sends a new POST request to the API resulting in the creation of a new empty session (no jobs submitted), which remains pending.

**Steps to reproduce**

```python
from qiskit_ibm_runtime import QiskitRuntimeService, Session
 

past_session = Session.from_id(session_id, service)
past_session_details = past_session.details()

```

This will create a new empty session. You can check the POST request enabling the HTTP debugging logs (before this code):

```python
import logging
import http.client

# enable debugging

logging.basicConfig(level=logging.DEBUG)
http.client.HTTPConnection.debuglevel = 2
```

**Expected behavior**

I have a way to retrieve old session details without side effects.

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.28.0
- **Python version**: 3.12.5
- **Operating system**: Mac OS Sonoma

kt474:
### Summary

`Session.from_id()` calls the POST /Session endpoint which is unnecessary because the session and session id already exist 

### Details and comments
Fixes #1895

### DIFF
--- a/qiskit_ibm_runtime/batch.py
+++ b/qiskit_ibm_runtime/batch.py
@@ -88,6 +88,7 @@ class Batch(Session):
         service: Optional[QiskitRuntimeService] = None,
         backend: Optional[Union[str, BackendV1, BackendV2]] = None,
         max_time: Optional[Union[int, str]] = None,
+        new_session: Optional[bool] = True,
     ):
         """Batch constructor.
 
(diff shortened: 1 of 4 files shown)


ptristan3:
Could you add tests to ensure the `new_session` param works properly?

ptristan3 (approved):

SamFerracin (approved):
Approved, with a comment that we should probably clean `Session` up in future iterations

PR #1907, opened 2024-09-05
Remove 0.24.0 deprecations

### LINKED ISSUE #1911: Documentation has to be updated
Usage examples in our documentation are not always up-to-date. For example: https://github.com/Qiskit/qiskit-ibm-runtime/blob/f12aa74621481c3a8427e5ed361f373b9d97efc7/qiskit_ibm_runtime/estimator.py#L90 issues a deprecation warning about the `backend` argument being replaced by `mode`.

Do we have an tool to automatic tool to check code snippets in documentation?

kt474:
### Summary

Removes/updates everything that was deprecated in the `0.24.0` release. `BackendV1` related things will be taken care of in a separate PR 

Changes:
- The arguments backend and session for Sampler and Estimator have been removed and replaced with "mode"
- Primitive `Session` property has been replaced with `mode`
- Specifying options without the full dictionary structure is no longer supported 
- `Job.program_id()` has been replaced with `Job.primitive_id()` 
- `Service.run()` and `Session.run()` have been replaced with a private method, `_run()` 
- In `Service.backend()`, "name" is now a required parameter 
- `Service.get_backend()` removed, replaced with `backend()`

### Details and comments
Fixes #1911


qiskit_ibm_runtime/base_primitive.py
```diff
@@ -242,17 +230,7 @@ def _set_options(self, options: Optional[Union[Dict, OptionsT]] = None) -> None:
...
-            except ValidationError:
-                self._options = self._options_class(**merge_options(default_options, options))
-                issue_deprecation_msg(
-                    "Specifying options without the full dictionary structure is deprecated",
-                    "0.24.0",
-                    "Instead, pass in a fully structured dictionary. For example, use "
-                    "{'environment': {'log_level': 'INFO'}} instead of {'log_level': 'INFO'}.",
-                    4,
-                )
+            self._options = self._options_class(**merge_options_v2(default_options, options))
```
ptristan3, line 233: We shouldn't keep the try/except?

qiskit_ibm_runtime/options/options.py
```diff
@@ -153,19 +151,8 @@ def _set_attr(_merged: dict) -> None:
...
-            _set_attr(merged)
-            issue_deprecation_msg(
-                "Specifying options without the full dictionary structure is deprecated",
-                "0.24.0",
-                "Instead, pass in a fully structured dictionary. For example, use "
-                "{'environment': {'log_level': 'INFO'}} instead of {'log_level': 'INFO'}.",
-                2,
-            )
+        merged = merge_options_v2(self, kwargs)
+        _set_attr(merged)
```
ptristan3, line 155: same here

qiskit_ibm_runtime/base_primitive.py
kt474: The behavior in the except block is what was deprecated and now being removed so I think we can remove the whole try/except. Only `merge_options` supports merging values from different dict levels.

ptristan3 (approved):

PR #1914, opened 2024-09-09
add custom url resolver to enable using other endpoints

cpkurotori:
### Summary

The service has other environments that need to be accessible via the provider in which the url cannot be properly generated with the current code. To enable using the provider with these environments without hardcoding strategies for these edge cases, I added the option for a custom url resolver.

### Details and comments

Use example:

```
def custom_url(url, instance, private_endpoint = False):
    return "https://my-custom-url"

service=QiskitRuntimeService(channel="ibm_cloud", url="https://test.cloud.ibm.com", instance="crn:v1:my_instance", token="my_token", url_resolver=custom_url)
```

### DIFF
--- a/qiskit_ibm_runtime/api/client_parameters.py
+++ b/qiskit_ibm_runtime/api/client_parameters.py
@@ -12,10 +12,10 @@
 
 """Represent IBM Quantum account client parameters."""
 
-from typing import Dict, Optional, Any, Union
+from typing import Dict, Optional, Any, Union, Callable
 from ..proxies import ProxyConfiguration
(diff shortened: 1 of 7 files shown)


kt474 (approved):
Thanks! LGTM

Can you also add a [release note](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/CONTRIBUTING.md#release-notes)

cpkurotori:
Re-requested review because instead of renaming the util function, I figured it would probably be better to deprecate since it is a breaking change.

PR #1918, opened 2024-09-11
Execution span docs

ihincks:
### Summary

This PR adds the execution span classes to the docs. It also turns the `execution_span` module into a composite module where each class is broken out into its own file. My understanding is that this will not affect downstream unless they are doing wacky stuff.


qiskit_ibm_runtime/execution_span.py
```diff
@@ -1,283 +0,0 @@
-# This code is part of Qiskit.
```
ihincks, line 1: this deleted file contains three classes. they were copy-pasted as-is into the new folder files

Eric-Arellano (approved):
Yeah this should be fine Python-wise and documentation-wise. Previous importers of `qiskit_ibm_runtime.execution_spans` now will import via `execution_spans/__init__.py` rather than `execution_spans.py` - that shouldn't matter to them.

I did not review the actual code itself, so I'm assuming things did not change from before.

PR #1933, opened 2024-09-18
Deprecate returning JobStatus in RuntimeJob

kt474:
### Details and comments
Fixes #


jyu00:
We should make sure internal calls to `status()` don't surface this deprecation warning. For example, `job.logs()` calls `job.status()` under the cover, but that call shouldn't surface this warning. 

Also issuing the warning once is probably good enough. It'd be excessive if someone is using `job.status()` in a loop to poll for status lol.

kt474:
> We should make sure internal calls to `status()` don't surface this deprecation warning. For example, `job.logs()` calls `job.status()` under the cover, but that call shouldn't surface this warning.

With a stack level of 2, other internal calls to `status()` don't show the deprecation warning 

> Also issuing the warning once is probably good enough. It'd be excessive if someone is using `job.status()` in a loop to poll for status lol.

Python deprecation warnings are only shown once per session for a specific call, so looping `job.status()` only shows the warning once

PR #1941, opened 2024-09-23
Update tests with program_id

kt474:
### Summary

Missed these in #1907 

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_job.py
+++ b/test/integration/test_job.py
@@ -144,7 +144,7 @@ class TestIntegrationJob(IBMIntegrationJobTestCase):
     def test_job_program_id(self, service):
         """Test job program ID."""
         job = self._run_program(service)
-        self.assertEqual(self.program_ids[service.channel], job.program_id)
+        self.assertEqual(self.program_ids[service.channel], job.primitive_id)
 
     @run_integration_test
     def test_wait_for_final_state(self, service):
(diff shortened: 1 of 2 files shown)


ptristan3 (commented):
Can we define a way to run the integration test and make it mandatory in the PR evidence section? (Similar we do in primitives with the e2e test running in a custom image). We need to be sure that a PR will not cause some failure with integration tests.

kt474:
> Can we define a way to run the integration test and make it mandatory in the PR evidence section? (Similar we do in primitives with the e2e test running in a custom image). We need to be sure that a PR will not cause some failure with integration tests.

Yeah we used to have the integration tests on every PR but changed it because it was a bit much to have them run on every commit. I think we can do this again or find a way to manually run them from PRs in github actions

ptristan3:
> > Can we define a way to run the integration test and make it mandatory in the PR evidence section? (Similar we do in primitives with the e2e test running in a custom image). We need to be sure that a PR will not cause some failure with integration tests.
> 
> Yeah we used to have the integration tests on every PR but changed it because it was a bit much to have them run on every commit. I think we can do this again or find a way to manually run them from PRs in GitHub actions

I agree we can't include the integration tests with each commit, but we can find a way to trigger (manually) the corresponding GitHub action. For instance, in the primitives repo we use a special commit message to trigger some stages.

PR #1942, opened 2024-09-23
Adding support `LayerError`s with unknown error channels

SamFerracin:
### Summary
This lays the ground for including more information on how the circuit is broken into layers server-side.


qiskit_ibm_runtime/options/resilience_options.py
```diff
@@ -100,8 +100,15 @@ def _validate_options(self) -> "ResilienceOptionsV2":
         # Validate not ZNE+PEC
         if self.pec_mitigation is True and self.zne_mitigation is True:
             raise ValueError(
-                "pec_mitigation and zne_mitigation`options cannot be "
+                "'pec_mitigation' and 'zne_mitigation' options cannot be "
```
SamFerracin, line 103:
```suggestion
                "pec_mitigation and zne_mitigation options cannot be "
```

qiskit_ibm_runtime/options/resilience_options.py
```diff
@@ -100,8 +100,15 @@ def _validate_options(self) -> "ResilienceOptionsV2":
         # Validate not ZNE+PEC
         if self.pec_mitigation is True and self.zne_mitigation is True:
             raise ValueError(
-                "pec_mitigation and zne_mitigation`options cannot be "
+                "pec_mitigation and zne_mitigation options cannot be "
```
SamFerracin, line 103:
```suggestion
                "pec_mitigation and zne_mitigation` options cannot be "
```

qiskit_ibm_runtime/options/resilience_options.py
```diff
@@ -100,8 +100,15 @@ def _validate_options(self) -> "ResilienceOptionsV2":
         # Validate not ZNE+PEC
         if self.pec_mitigation is True and self.zne_mitigation is True:
             raise ValueError(
-                "pec_mitigation and zne_mitigation`options cannot be "
+                "pec_mitigation and zne_mitigation` options cannot be "
```
SamFerracin, line 103:
```suggestion
                "pec_mitigation and zne_mitigation`options cannot be "
```

qiskit_ibm_runtime/options/resilience_options.py
```diff
@@ -100,8 +100,15 @@ def _validate_options(self) -> "ResilienceOptionsV2":
         # Validate not ZNE+PEC
         if self.pec_mitigation is True and self.zne_mitigation is True:
             raise ValueError(
-                "pec_mitigation and zne_mitigation`options cannot be "
+                "'pec_mitigation' and 'zne_mitigation' options cannot be "
                 "simultaneously enabled. Set one of them to False."
             )
 
+        if not isinstance(self.layer_noise_model, UnsetType) and None in [
```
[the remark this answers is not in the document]
SamFerracin: I had that, but I get a mypy error here because "self.layer_noise_model is not iterable." What you suggest here works in practice, but I couldn't find a way of disabling the mypy complaint

chriseclectic: I would change this to allow values of None in the options for two reasons:
1. It lets users pass in the returned option from one estimator/noise_learner run to another Estimator run call without having to manually modify the options and remove any None values 
2. Maybe some uers will want to experiment with manually disabling mitigation for specific layer(s) for debugging/research applications?

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -139,20 +139,28 @@ class LayerError:
...
     Raises:
         ValueError: If ``circuit``, ``qubits``, and ``error`` have mismatching number of qubits.
     """
 
     def __init__(
-        self, circuit: QuantumCircuit, qubits: Sequence[int], error: PauliLindbladError
+        self,
+        circuit: QuantumCircuit,
+        qubits: Sequence[int],
+        error: Optional[PauliLindbladError] = None,
```
chriseclectic, line 153: I cant see from the diff, but is there a doc string that needs updating to say what a value of None correponds to?

qiskit_ibm_runtime/utils/noise_learner_result.py
```diff
@@ -170,9 +178,9 @@ def qubits(self) -> list[int]:
         return self._qubits
 
     @property
-    def error(self) -> PauliLindbladError:
+    def error(self) -> Union[PauliLindbladError, None]:
         r"""
-        The error channel in this :class:`.~LayerError`.
+        The error channel in this :class:`.~LayerError`, or ``None`` if the error channel is unknown.
```
chriseclectic, line 183: If you want to also allow treating None as a user value as disabling maybe something like this?
```suggestion
        The error channel in this :class:`.~LayerError`, or ``None`` if the error channel is either unknown or explicitly disabled.
```

test/unit/test_data_serialization.py
```diff
@@ -340,8 +340,9 @@ def assert_primitive_results_equal(self, primitive_result1, primitive_result2):
 
     def assert_pauli_lindblad_error_equal(self, error1, error2):
         """Tests that two PauliLindbladError objects are equal"""
-        self.assertEqual(error1.generators, error2.generators)
-        self.assertEqual(error1.rates.tolist(), error2.rates.tolist())
+        if error1 or error2:
```
chriseclectic, line 343: This might not be necessary for a test if it isnt one of the cases, but in general this should be
```suggestion
        if error1 and error2:
```
and then adding an `else` clause to handle case where only 1 of the errors is None

chriseclectic (commented):
Mostly LGTM. Just a few nit pics and some suggestions about allowing user None values

qiskit_ibm_runtime/utils/noise_learner_result.py
SamFerracin: Yes:
```
error: The Pauli Lindblad error channel affecting the ``circuit``, or ``None`` if the error
            channel is either unknown or explicitly disabled.
```

qiskit_ibm_runtime/options/resilience_options.py
SamFerracin: I removed this if because it's not needed. Server doesn't crush

joshuasn (approved):

PR #1950, opened 2024-09-25
Debugger API, phase I

SamFerracin:
Phase I Debugger is required to:

- Provide a convenience function to Cliffordize circuits
- Provide a way to compare the results of ideal vs noisy simulations

It is not required to produce noisy results based on the outcomes of noise learner.

## TODO
Consider alternative names for:
- Debugger
- DebuggerResult
- simulate


qiskit_ibm_runtime/debugger/debugger.py
```diff
@@ -0,0 +1,171 @@
...
+        r"""
+        A convenience method that returns the cliffordized version of the given ``pubs``, obtained
+        by run the :class:`.~ConvertISAToClifford` transpiler pass on the PUBs' circuits.
+
+        Args:
+            pubs: The PUBs to turn into Clifford PUBs.
+        Returns:
+            The Clifford PUBs.
+        """
+        coerced_pubs = [EstimatorPub.coerce(pub) for pub in pubs]
```
joshuasn, line 154: why not coerce and convert to clifford in a single loop then validate?

qiskit_ibm_runtime/debugger/debugger.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        return self._noise_model
+
+    def simulate(
+        self,
+        pubs: Sequence[EstimatorPubLike],
+        with_noise: bool = True,
+        seed_simulator: Optional[int] = None,
+        default_precision: float = 0,
+    ) -> list[DebuggerResult]:
```
jyu00, line 94: What do you think about making the return type a class that can be sliced (similar to `PrimitiveResult`), instead of a list? A class would give us a lot more flexibility if we ever want to change the output type, or include additional metadata.

qiskit_ibm_runtime/debugger/debugger.py
```diff
@@ -0,0 +1,170 @@
...
+
+    Args:
+        backend: A backend.
+        noise_model: A noise model for the operations of the given backend. If ``None``, it
+            defaults to the noise model generated by :meth:`NoiseModel.from_backend`.
+    """
+
+    def __init__(self, backend: Backend, noise_model: Optional[NoiseModel] = None) -> None:
+        self._backend = backend
+        self._noise_model = noise_model or NoiseModel.from_backend(
```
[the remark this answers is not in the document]
SamFerracin: I have thought of that, but I'd have to think about it or look for it (presumably in qiskit-aer?). Do you know if any such validation method exists somewhere?

qiskit_ibm_runtime/debugger/debugger.py
```diff
@@ -0,0 +1,170 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""A debugger."""
+
+from __future__ import annotations
+from typing import Optional, Sequence, List
+
+from qiskit_aer.noise import NoiseModel
+from qiskit_aer.primitives.estimator_v2 import EstimatorV2 as AerEstimator
```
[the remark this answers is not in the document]
SamFerracin: Good catch!

qiskit_ibm_runtime/debugger/debugger.py
SamFerracin: I am not opposed to that. But we need to find one more name

qiskit_ibm_runtime/debugger/debugger.py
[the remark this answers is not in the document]
jyu00: I don't know if there is any... but `NoiseModel` itself has attributes like `noise_qubits` and `basis_gates` that can potentially be used to verify against what's in the circuit.

qiskit_ibm_runtime/debugger/debugger.py
jyu00: `DebuggerResult` which has a list of `DebuggerPubResult` 😄

SamFerracin: Okay. I like this, because all we need to change is "Debugger*" (if we want to), otherwise it's all pretty consistent with Estimator

qiskit_ibm_runtime/debug_tools/neat.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def __init__(self, backend: Backend, noise_model: Optional[NoiseModel] = None) -> None:
+        if not HAS_QISKIT_AER:
+            raise ValueError(
+                "Cannot initialize object of type 'Neat' since 'qiskit-aer' is not installed. "
+                "Install 'qiskit-aer' and try again."
+            )
+
+        self._backend = backend
+        self._noise_model = (
```
SamFerracin, line 126:
```suggestion
        self.noise_model = (
```

qiskit_ibm_runtime/debug_tools/neat.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @property
+    def noise_model(self) -> NoiseModel:
+        r"""
+        The noise model used by this analyzer tool for the noisy simulations.
+        """
+        return self._noise_model
+
+    @noise_model.setter
+    def noise_model(self, value: NoiseModel) -> NoiseModel:
```
SamFerracin, line 140: As discussed previously, this could be the place to do some validation checks (which will happen at init time) to make sure that users don't get errors when they hit `simulate`. I'm struggling a bit to understand what exactly we want to validate and what should be left to qiskit-aer. Here are some ideas, feel free to comment and/or add:

- Validate that noise model is "Clifford." This would be useful because we simulate with the "stabilizer" method, and so for example `NoiseModel.from_backend(backend)` -- which has `thermal_relaxation=True` by default and non-Clifford components -- will cause `simulate` to fail. Unfortunately, there's no public check that we can import from qiskit-aer, and the failure seems to come from one of the C++ files. So while this is a useful and nice-to-have check, I don't know how to set it up usefully (I mean.. we could `try .. except ..` with `AerSimulator(noise_model, "stabilizer").run(QuantumCircuit(1))`, but that's not really clean).
- @jyu00 had suggested checking that a noise model "covers" the backend, and issue warnings if, for example, there are some qubits or basis gates in the backend for which the noise model is unspecified (Jessie, please call me out if I'm misrepresenting your comment). So for example, `Neat(backend, NoiseModel())` would warn. I am not sure if this should be done for two reasons.
    1. I this that one of the main use-cases would be users building their own noise models and trying to see how `simulate`'s results compare to QPU's results. If this is the case, then users may just want to add the noise for the qubits/gates that they are interested in using, and so the warning would be issued more often than not.
    2. Qiskit-aer has no problem with noise models not "fully covering" backends. It seems to me that we'd be doing qiskit-are's job, in a way, meaning that this sort of warnings should live in qiskit-aer if they should live at all, and not be migrated to qiskit-ibm-runtime.

ihincks: By the way, now that this class is not called Debugger, I don't have concerns about the 1-1 relationship between an instance and a noise model. Recall that the concern was a user with the comment "that's weird, i'm not used to having two debuggers on the go at once".

qiskit_ibm_runtime/debug_tools/neat.py
```diff
@@ -0,0 +1,228 @@
...
+        This function uses ``qiskit-aer``'s ``Estimator`` class to simulate the estimation task
+        classically.
+
+        .. note::
+            To ensure scalability, every circuit in ``pubs`` is required to be a Clifford circuit,
+            so that it can be simulated efficiently regardless of its size. For estimation tasks
+            that involve non-Clifford circuits, the recommended workflow consists of mapping
+            the non-Clifford circuits to the nearest Clifford circuits using the
+            :class:`.~ConvertISAToClifford` transpiler pass, or equivalently, to use the Neat's
+            :meth:`to_clifford` convenience method.
```
ihincks, line 173: Could we simplify the workflow by adding a kwarg for automatic conversion for the lazy folk? I haven't thought too hard about the consequences.

```python
    def simulate(
        self,
        pubs: Sequence[EstimatorPubLike],
        with_noise: bool = True,
        cliffordize: bool = True,
        seed_simulator: Optional[int] = None,
        default_precision: float = 0,
    ) -> NeatResult:
```

qiskit_ibm_runtime/debug_tools/neat.py
```diff
@@ -0,0 +1,228 @@
...
+
+        Args:
+            pubs: The PUBs specifying the estimation task of interest.
+            with_noise: Whether to perform an ideal, noiseless simulation (``False``) or a noisy
+                simulation (``True``).
+            seed_simulator: A seed for the simulator.
+            default_precision: The default precision used to run the ideal and noisy simulations.
```
ihincks, line 180: Let's add some clarity about what exactly this argument is doing because it's a cause of questions in the Estimator.

Also, if we want it to match the Estimator.run method:
```suggestion
            precision: The default precision used to run the ideal and noisy simulations.
```

qiskit_ibm_runtime/debug_tools/neat.py
```diff
@@ -0,0 +1,228 @@
...
+    def backend(self) -> Backend:
+        r"""
+        The backend used by this analyzer tool.
+        """
+        return self._backend
+
+    def simulate(
+        self,
+        pubs: Sequence[EstimatorPubLike],
+        with_noise: bool = True,
```
ihincks, line 157: This is a matter of subjective preference, and I'm happy enough with the current setup, but I suspect it would feel slightly more ergonomic to me to split this into two methods along this bool:

```python
great_ratio = neat.ideal_sim(puberoos) / neat.noisy_sim(puberoos)
```

ihincks (commented):
Lookin real nice.

qiskit_ibm_runtime/debug_tools/neat.py
SamFerracin: Not the first person to suggest this, I am happy to go for it

qiskit_ibm_runtime/debug_tools/neat.py
SamFerracin: I have thought about this. The consequence I see is that people may forget about it, end end up doing the following thing:
```
# run non-Cliff pubs on CPU
results = estimator.run(pubs).results()

# debugger (which automatically turns PUBs into Clifford PUBs)
results_deb = Neat(backend).simulate(pubs)

# compare
ratio = results / results_deb
```
So at the very least, if we want to do this, we should set `cliffordize` to `False` by default

qiskit_ibm_runtime/debug_tools/neat.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    validate_estimator_pubs(pubs)
+    validate_isa_circuits([pub.circuit for pub in pubs], backend.target)
+
+    if validate_clifford:
+        for pub in pubs:
+            for instr in pub.circuit:
+                op = instr.operation
+                # ISA circuits contain only one type of non-Clifford gate, namely RZ
+                if op.name == "rz" and op.params[0] not in [0, np.pi / 2, np.pi, 3 * np.pi / 2]:
```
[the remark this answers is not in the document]
SamFerracin: all gone

ihincks (approved):

PR #1951, opened 2024-09-25
Add EstimatorPubResult with plotter for ZNE data

### LINKED ISSUE #1820: Add `EstimatorPubResult` with plotting methods

Just like there exists a `SamplerPubResult`, we should introduce an `EstimatorPubResult`. The value of this is that we can add convenience methods to the pub result class. For example, methods to help plot ZNE data would be especially helpful.

** Details

Here is a snippet that makes a plot. Please note that it is just something I found lying around that may help the assignee get started---it's certainly not meant to represent exactly what we need.

```python
result = service.job(<some zne job id>).result()
pub_result = result[0]

nfs = result.metadata["resilience"]["zne"]["noise_factors"]
nfs_x = result.metadata["resilience"]["zne"]["extrapolated_noise_factors"]

models = result.metadata["resilience"]["zne"]["extrapolator"]
plt.figure(figsize=(5 * len(models), 4))

for idx_model, model in enumerate(models):
    plt.subplot(1, len(models), idx_model + 1)
    for idx in range(pub_result.data.evs_noise_factors.shape[2]):
        lines = plt.errorbar(
            nfs,
            pub_result.data.evs_noise_factors[0,0,idx],
            fmt='.',
            yerr=pub_result.data.stds_noise_factors[0,0,idx]
        )

        plt.plot(
            nfs_x,
            means:=pub_result.data.evs_extrapolated[0,0,idx,idx_model,:],
            color=(color:=lines[0].get_color()),
            alpha=0.3
        )

        plt.fill_between(
            nfs_x,
            means-(stds:=pub_result.data.stds_extrapolated[0,0,idx, idx_model, :]),
            means+ stds,
            color=color,
            alpha=0.1
        )

        plt.ylim([-0.05,1.05])
```

**Acceptance criteria**

 - [ ] Estimator jobs return `EstimatorPubResult`s 
 - [ ] The new pub result class has a plotting method for ZNE
 - [ ] The plotting method raises sensible errors when ZNE data is not present
 - [ ] The plot looks good in a variety of regimes, such as when the fit blows up to infinity because of insufficient data

joshuasn:
### Summary

Adds an EstimatorPubResult class with a method for plotting ZNE data.

### Details and comments
Closes #1820.
The plotter supports two separate views of the data with the `subplots` argument–by default, all expectation values are plotted on a single plot per extrapolator:
Screenshot 2024-09-25 at 5 00 16 PM
Setting `subplots=True` gives a grid view of each value:
Screenshot 2024-09-25 at 5 00 28 PM
Further, we can select a subset of the values with the `indices` argument (here, `(4, 1)`):
Screenshot 2024-09-25 at 5 00 39 PM

The estimator does not currently return EstimatorPubResult so the class needs to be manually instantiated with the right metadata before the method is called:
```
pub_result = EstimatorPubResult(result[0].data, result[0].metadata)
pub_result.metadata["resilience"]["zne"]["noise_factors"] = result.metadata["resilience"]["zne"]["noise_factors"]
pub_result.metadata["resilience"]["zne"]["extrapolated_noise_factors"] = result.metadata["resilience"]["zne"]["extrapolated_noise_factors"]
pub_result.metadata["resilience"]["zne"]["extrapolators"] = result.metadata["resilience"]["zne"]["extrapolator"]
```


qiskit_ibm_runtime/utils/estimator_pub_result.py
```diff
@@ -0,0 +1,83 @@
...
+
+from __future__ import annotations
+
+from typing import Sequence
+import plotly.graph_objects as go
+
+from qiskit.primitives.containers import PubResult
+
+
+class EstimatorPubResult(PubResult):
```
SamFerracin, line 30: What is the workflow for a user that gets estimator results from running on a QPU?
Do they need to construct an `EstimatorPubResult` by themselves? Could you write a snippet that starts with `results = job.result()` so that we can look at it? 🙂

qiskit_ibm_runtime/utils/estimator_pub_result.py
```diff
@@ -0,0 +1,83 @@
...
+    ) -> go.Figure:
+        """Plot the zero noise extrapolation data contained in this estimator pub result.
+
+        Args:
+            indices: The indices of the expectation values to include in the plot. If ``None``, includes
+                all values. See :class:`~.ZneOptions` for information on the indexing scheme.
+            names: The names to assign to the expectation values. If ``None``, the names correspond to
+                the indices.
+            n_stds: The number of standard deviations to include around each fit.
+            mag_tol: The tolerance. If ``evs_extrapolated`` has a greater magnitude than this value, the
```
SamFerracin, line 54: Consider using some other name (maybe `max_evals` and `stds_max`?), since the concept of magnitude of an evs doesn't show up anywhere else in qiskit-ibm-runtime (that I know of)

qiskit_ibm_runtime/visualization/plot_zne.py
```diff
@@ -0,0 +1,288 @@
...
+    Returns:
+        A plotly figure.
+
+    Raises:
+        ValueError: If ``result`` does not contain zero noise extrapolation data.
+        ValueError: If the length of ``names`` is not equal to the length of ``indices``.
+    """
+    resilience = result.metadata.get("resilience")
+    if not resilience:
+        raise ValueError("Result does not contain resilience metadata.")
```
SamFerracin, line 69:
```suggestion
    if not resilience := result.metadata.get("resilience"):
        raise ValueError("Result does not contain resilience metadata.")
```

qiskit_ibm_runtime/visualization/plot_zne.py
```diff
@@ -0,0 +1,288 @@
...
+
+    zne_metadata = resilience.get("zne")
+    if not zne_metadata:
+        raise ValueError("Result does not contain ZNE data.")
```
SamFerracin, line 73:
```suggestion
    if not zne_metadata := resilience.get("zne"):
        raise ValueError("Result does not contain ZNE data.")
```

qiskit_ibm_runtime/visualization/plot_zne.py
```diff
@@ -0,0 +1,288 @@
...
+    indices: Sequence[tuple[int, ...]] | None = None,
+    names: Sequence[str] | None = None,
+    n_stds: int = 1,
+    mag_tol: float = 10,
+    std_tol: float = 0.2,
+    height: int = 500,
+    width: int = 1000,
+    n_cols: int = 4,
+    colorscale: str = "Aggrnyl",
+    subplots: bool = False,
```
SamFerracin, line 38: I think we talked about this at some point, writing this his for the sake of open-sourcing the converstation.

Have you considered if `subplots=True` and `subplots=False` should be different functions, e.g. `plot_zne` and `plot_zne_subplots` (names just made up now, they can be improved)?
The main body of the functions are different (wrapped in a big `if .. else ..`), as are the outputs and (I imagine) the default values that you may want to have. And it's true that they rely on similar pre- and post-processing logic, but this can be incorporated into private functions and used in both places.

I imagine that this is something that comes up pretty often in scatter plots in other python libraries.
I'm not sure if there is a standard pattern though

qiskit_ibm_runtime/visualization/plot_zne.py
```diff
@@ -0,0 +1,288 @@
...
+"""Functions to visualize :class:`~.EstimatorPubResult` ZNE data."""
+
+from __future__ import annotations
+
+from itertools import product
+from typing import Sequence
+from plotly.colors import sample_colorscale
+from plotly.subplots import make_subplots
+import numpy as np
+import plotly.graph_objects as go
```
SamFerracin, line 22:
```suggestion
if TYPE_CHECKING:
    import plotly.graph_objs as go
```

qiskit_ibm_runtime/utils/estimator_pub_result.py
joshuasn: Right now, that's what needs to be done. The intent is for an estimator job to return estimator pub results instead so the method can be called directly.

qiskit_ibm_runtime/visualization/plot_zne.py
joshuasn: Yeah, agreed. I have broken it down into two separate methods. The `subplots=True` case is now `plot_zne` and `subplots=False` is `plot_zne_extrapolators`. Seems natural to want to look at each expectation value on its own rather than having them all on the same plot by default.

qiskit_ibm_runtime/utils/estimator_pub_result.py
SamFerracin: 👍 makes sense
Right now, `SamplerPubResult` and other result containers live in `qiskit.primitives.containers`. Do we want to add `EstimatorPubResult` here and port it later, or write it directly in qiskit? The latter seems more natural (no deprecations needed), I'm not sure when it'd become available to us though

joshuasn: Per offline discussion, we're adding it in runtime as opposed to `qiskit.primitives.containers` as the methods rely on the metadata created by runtime types, e.g., `ZneOptions`, to work.

qiskit_ibm_runtime/utils/estimator_pub_result.py
```diff
@@ -0,0 +1,130 @@
...
+from qiskit.primitives.containers import PubResult
+
+if TYPE_CHECKING:
+    from plotly.graph_objects import Figure as PlotlyFigure
+
+
+class EstimatorPubResult(PubResult):
+    """Result of Estimator Pub."""
+
+    def draw_zne(
```
ihincks, line 35: What happens when this is estimator data with ZNE turned off? Some possible behaviour:
1. fails gracefully with warning empty plot
2. fails gracefully with actionable error message
3. fails with cryptic AttributeError or something
4. not relevant, because this type is only returned (future work) by the runtime if there was ZNE done

joshuasn: Right now, it checks that the right metadata fields are present and if not throws an error (in the `draw_*` functions). I prefer 2 (what is implemented now but will improve the error message, we point people how to use ZNE and this method) or 4 (but would probably make sense to rename the class to ZnePubResult or something).

SamFerracin: I am okay with 1, 2 and 4. Personally, thinking of a scenario where we want to include more functionality for different types of results, I find 1 and 2 simpler on the user side, in that:
- with 2 there is one class of EstimatorPubResult, some of its methods require specific metadata to be present and will fail (or return an empty plot with a warning) otherwise
- with 4 there are different classes of estimator results with potentially different methods, so as a user I need to go into the documentation to check what methods are supported by the result class that I'm looking at

But 4 may be easier to handle on the dev side, so i can be okay with 4 too

qiskit_ibm_runtime/visualization/draw_zne.py
```diff
@@ -0,0 +1,395 @@
...
+
+from .utils import plotly_module
+from ..utils.estimator_pub_result import EstimatorPubResult
+
+if TYPE_CHECKING:
+    from plotly.graph_objects import Figure as PlotlyFigure
+    from plotly.graph_objects import Scatter as PlotlyScatter
+
+
+def draw_zne(
```
SamFerracin, line 29: Since the other one is called `draw_zne_extrapolators`, I would think of calling this one `draw_zne_evs` or something like that. By reading `draw_zne` I get the impression that this one is "better" than the other one and thus it gets to be "the" draw_zne, but I think they're on equal grounds

qiskit_ibm_runtime/utils/estimator_pub_result.py
ihincks: Good points @SamFerracin , you have pushed me into camps 1,2.

joshuasn: I went with 2 and changed the error message to be more descriptive.

qiskit_ibm_runtime/visualization/draw_zne.py
```diff
@@ -0,0 +1,400 @@
...
+            customdata=stds,
+            name=name,
+            mode="lines",
+            line={"color": color},
+            hovertemplate=hovertemplate,
+            legendgroup=legend_group,
+            showlegend=show_legend,
+        ),
+        *(
+            go.Scatter(
```
SamFerracin, line 315: If possible, replace this with a single scatter trace using `None`s as discussed privately

joshuasn: Attempted the above, did not affect the opacity as expected. Marking as resolved.

qiskit_ibm_runtime/utils/estimator_pub_result.py
```diff
@@ -0,0 +1,130 @@
...
+        indices: Sequence[tuple[int, ...]] | None = None,
+        names: Sequence[str] | None = None,
+        n_stds: int = 1,
+        max_mag: float = 10,
+        max_std: float = 0.2,
+        height: int = 500,
+        width: int = 1000,
+        n_cols: int = 4,
+        colorscale: str = "Aggrnyl",
+    ) -> PlotlyFigure:
```
ihincks, line 46: <this comment applies to all functions/methods>

Nit: the prevalent style in `qiskit` is the full `num_` prefix

```suggestion
    def draw_zne_evs(
        self,
        indices: Sequence[tuple[int, ...]] | None = None,
        names: Sequence[str] | None = None,
        num_stds: float = 1.0,
        max_mag: float = 10,
        max_std: float = 0.2,
        height: int = 500,
        width: int = 1000,
        num_cols: int = 4,
        colorscale: str = "Aggrnyl",
    ) -> PlotlyFigure:
```

ihincks (approved):

PR #1952, opened 2024-09-25
Adding a more developer-friendly interface to specify qubit coordinates

SamFerracin:
Addresses part of the technical debt introduced with the visualization PR.

Question:
Could/should `_get_qubit_coordinates` be made public, or perhaps be part of the `Backend` class?

### DIFF
--- a/qiskit_ibm_runtime/utils/embeddings.py
+++ b/qiskit_ibm_runtime/utils/embeddings.py
@@ -102,7 +102,7 @@ class _Row:
     down_links: Sequence[int] = ()
 
 
-def _heavy_hex_coords(rows: Sequence[_Row], row_major: bool = True) -> list[tuple[int, int]]:
+def _heavy_hex_coords(rows: Sequence[_Row], row_major: bool = True) -> List[Tuple[int, int]]:
     """Generate heavy hex coordinates for the given rows.
 
     Args:
--- a/qiskit_ibm_runtime/utils/embeddings.py
+++ b/qiskit_ibm_runtime/utils/embeddings.py
@@ -15,7 +15,8 @@
 Utility class to represent an embedding of a set of qubits in a two-dimensional plane.
 """
 
-from typing import List, Tuple, Union
+from dataclasses import dataclass
+from typing import List, Tuple, Union, Sequence
 
 from qiskit.providers.backend import BackendV2
 from qiskit.transpiler import CouplingMap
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/utils/embeddings.py
```diff
@@ -61,8 +62,12 @@ def from_backend(cls, backend: BackendV2) -> "Embedding":
...
         if not (coupling_map := backend.coupling_map):
             raise ValueError(f"Coupling map for backend '{backend.name}' is unknown.")
-        if not (coordinates := _get_qubits_coordinates(backend.num_qubits)):
-            raise ValueError(f"Coordinates for backend '{backend.name}' are unknown.")
+        try:
+            coordinates = _get_qubits_coordinates(backend.num_qubits)
+        except ValueError as err:
+            raise ValueError(
+                f"Failed to fetch coordinates for backend '{backend.name}' with error: {err}"
+            )
```
ihincks, line 70:
```suggestion
        except ValueError as err:
            raise ValueError(
                f"Failed to fetch coordinates for backend '{backend.name}'."
            ) from err
```

qiskit_ibm_runtime/utils/embeddings.py
```diff
@@ -81,6 +86,54 @@ def coupling_map(self) -> CouplingMap:
         return self._coupling_map
 
 
+@dataclass(frozen=True)
+class _Row:
+    r"""
+    A class representing a dense row of the standard heavy hex planar embedding.
```
ihincks, line 92:
```suggestion
    A class representing a dense row of the standard heavy hex planar embedding.
    
    Here, a "dense row" means a horizontal row of adjacent qubits.
```

qiskit_ibm_runtime/utils/embeddings.py
```diff
@@ -81,6 +86,54 @@ def coupling_map(self) -> CouplingMap:
...
+
+    Args:
+        num_qubits: The number of qubits in the dense row.
+        x_offset: The integer x-offset of the row.
+        down_links: The positions of dense row, wrt the first qubit, at which to place qubits below.
```
ihincks, line 97: I know I suggested this format, so I recognize this is an annoying comment, but what about using `slice` instead of `_Row`?

Eg for 27 qubit devices:
```python
slice(3, 10, 4)
slice(0, 10)
slice(1, 10, 4)
slice(1, 11)
slice(3, 10, 4)
```

Feel free to respond with "quit wasting my time"

SamFerracin: No that's fine, it's just a CTRL+F, no big deal :P 
As long as you suggest a different name, because `slice` is something else already

ihincks: Actually, `range` instead of `slice`, and maybe just `Iterable[int]` instead of `range` while at it. Essentially, this becomes a sparse format for anything that can be specified row-wise.

SamFerracin: Oh, I thought you meant that you wanted a different name for the dataclass

ihincks (approved):

PR #1953, opened 2024-09-26
Enhance the ISA check with validation of rzz angle

### LINKED ISSUE #1949: Validate that rzz angle is between 0 and pi/2
`rzz` in IBM devices was not calibrated to angle values outside the range of `[0, pi/2]`, such values should raise an exception.

yaelbh:
### Summary

During the ISA check, we check the angles of rzz gates, and verify that they are in the permitted range of `[0, pi/2]`.

### Details and comments
Fixes #1949.


yaelbh:
Note: the diff in the conf file, it is difficult to see it in the GitHub platform, it is the addition of the `rzz` gate to the set of basis gates (other fractional gates, like `rzx`, are already there).

PR #1955, opened 2024-09-30
Add backend refresh() method

kt474:
### Details and comments
Fixes #


qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -357,6 +357,12 @@ def target_history(self, datetime: Optional[python_datetime] = None) -> Target:
         self._convert_to_target(refresh=True)
         return self._target
 
+    def refresh(self) -> None:
+        """Refresh the current backend target"""
+        self._get_properties(datetime=python_datetime.now())
+        self._get_defaults()
+        self._convert_to_target(refresh=True)
```
jyu00, line 364: I think this doesn't refresh the backend configuration?

jyu00 (commented):
can you also add a test?

test/integration/test_backend.py
```diff
@@ -119,6 +119,16 @@ def test_backend_target_history(self):
...
 
+    @production_only
+    def test_backend_target_refresh(self):
+        """Test refreshing the backend target."""
+        backend = self.backend
+        with self.subTest(backend=backend.name):
+            old_target = backend.target_history(datetime=datetime.now() - timedelta(60))
+            self.assertIsNotNone(old_target)
+            backend.refresh()
+            self.assertNotEqual(old_target, backend.target)
```
jyu00, line 130: haha I just opened [an issue](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1956) that, when fixed, would make this test less useful. 

I think you can just do 
```
old_target = backend.target
backend.refresh()
new_target = backend.target
```
and assert `old_target` and `new_target` are different objects (to validate target is not cached)

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -357,6 +358,17 @@ def target_history(self, datetime: Optional[python_datetime] = None) -> Target:
...
         return self._target
 
+    def refresh(self) -> None:
+        """Retrieve the newest backend configuration and refresh the current backend target."""
+        if config := configuration_from_server_data(
+            raw_config=self._service._api_client.backend_configuration(self.name, refresh=True),
+            instance=self._instance,
+        ):
+            self._configuration = config
+        self._get_properties(datetime=python_datetime.now())
```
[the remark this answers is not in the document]
kt474: Yeah I think we can consolidate the two methods - i'll do that when fixing https://github.com/Qiskit/qiskit-ibm-runtime/issues/1956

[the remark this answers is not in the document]
jyu00: ok is the plan to use `_get_properties(datetime)` in this PR and then switch to using `properties()` after #1956 is done?

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -357,6 +358,17 @@ def target_history(self, datetime: Optional[python_datetime] = None) -> Target:
...
+        self._get_defaults()
```
jyu00, line 369: Looks like `_get_defaults` also uses cached data if it already exists.

https://github.com/Qiskit/qiskit-ibm-runtime/blob/f8323dc45d1affc189b0b9e5b0684b92c1bb18e5/qiskit_ibm_runtime/ibm_backend.py#L251-L253

test/integration/test_backend.py
```diff
@@ -119,6 +119,16 @@ def test_backend_target_history(self):
...
 
+    @production_only
+    def test_backend_target_refresh(self):
+        """Test refreshing the backend target."""
+        backend = self.backend
+        with self.subTest(backend=backend.name):
+            old_target = backend.target
+            backend.refresh()
+            new_target = backend.target
+            self.assertNotEqual(old_target, new_target)
```
jyu00, line 130: I think this should also check that `configuration()`, `properties()`, and `defaults()` are all different objects after refresh, to ensure the new target is not constructed from cached objects

jyu00 (approved):
The test can be improved a bit, but otherwise LGTM

PR #1973, opened 2024-10-15
Bug fixes and improvements for `LayerError` visualization tool

SamFerracin:
### Improvements
This PR adds new keywords that I found useful while using these tools, such as:
- `highest_rate`: The highest rate used to ocmpute the colorscale.
- `color_out_of_scale`: The color of any rate that is larger than `highest_rate`
- Added example on docs that shows how to draw with custom embeddings (useful when the backend is massive and data is only available for few qubits)

### Fixes
- An indexing issue that gives `KeyError` when the first qubit is not `0`.

### DIFF
--- a/qiskit_ibm_runtime/visualization/draw_layer_error_map.py
+++ b/qiskit_ibm_runtime/visualization/draw_layer_error_map.py
@@ -35,7 +35,7 @@ def draw_layer_error_map(
     num_edge_segments: int = 16,
     edge_width: float = 4,
     height: int = 500,
-    high_scale: Optional[float] = None,
+    highest_rate: Optional[float] = None,
     background_color: str = "white",
     radius: float = 0.25,
     width: int = 800,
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/visualization/draw_layer_error_map.py
```diff
@@ -47,9 +49,12 @@ def draw_layer_error_map(
             to draw the layer error on, or a backend to generate an :class:`~.Embedding` for.
         colorscale: The colorscale used to show the rates of ``layer_error``.
         color_no_data: The color used for qubits and edges for which no data is available.
+        color_out_of_scale: The color used for rates whose value is above ``highest_rate``.
```
joshuasn, line 52:
```suggestion
        color_out_of_scale: The color used for rates with value greater than ``highest_rate``.
```

joshuasn (approved):

PR #1982, opened 2024-10-18
Add DoubleSliceSpan

joshuasn:
### Summary
Adding a new class, DoubeSliceSpan, that allows for handling data that was batched as `(argument_idx, shot_idx)`, or equivalently, flattened according to observables + parameters and shots separately.


qiskit_ibm_runtime/execution_span/double_slice_span.py
```diff
@@ -0,0 +1,76 @@
...
+from .execution_span import ExecutionSpan, ShapeType
+
+
+class DoubleSliceSpan(ExecutionSpan):
+    """An :class:`~.ExecutionSpan` for data stored in a sliceable format.
+
+    This type of execution span references pub result data by assuming that it is a sliceable
+    portion of the data where the shots are the outermost slice and the rest of the data is flattened.
+    Therefore, for each pub dependent on this span, the constructor accepts two :class:`slice` objects,
+    along with the corresponding shape of the data to be sliced.
```
ihincks, line 32:
```suggestion
    along with the corresponding shape of the data to be sliced; in contrast to
    :class:`~.SliceSpan`, this class does not assume that _all_ shots for a particular set of parameter
    values are contiguous in the array of data.
```

qiskit_ibm_runtime/execution_span/double_slice_span.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @property
+    def pub_idxs(self) -> list[int]:
+        return sorted(self._data_slices)
+
+    @property
+    def size(self) -> int:
+        size = 0
+        for shape, args_sl, shots_sl in self._data_slices.values():
+            size += np.zeros((np.prod(shape[:-1]), shape[-1]))[(args_sl, shots_sl)].size
```
ihincks, line 64: Ideally we could do this calculation without allocating these big matrices. Does the following look correct?

This suggestion may exceed line limit.

```suggestion
            size += len(range(math.prod(shape[:-1]))[args_sl]) * len(range(shape[-1])[shots_sl])
```

joshuasn: Good catch, that's right–thanks!

ihincks (approved):

PR #1986, opened 2024-10-21
Fix rzz parameter validation on Qiskit 1.1

wshanks:
The rzz validation check was using `CircuitInstruction.params`, an alias for `CircuitInstruction.operation.params` added in Qiskit 1.2. Here the usage is replaced `CircuitInstruction.operation.params` to keep compatibility with Qiskit 1.1.


wshanks:
I did not add a release note because rzz gates are not available so far any way.

qiskit_ibm_runtime/utils/utils.py
```diff
@@ -73,10 +73,11 @@ def _is_isa_circuit_helper(circuit: QuantumCircuit, target: Target, qubit_map: D
         # errors (beyond pi/2 does not trigger an error down the stack, only may become less
         # accurate).
         if name == "rzz" and (
-            instruction.params[0] < 0.0 or instruction.params[0] > 1.001 * np.pi / 2
```
ihincks, line 76: Out of scope, but a question for @yaelbh when she returns: what if `params[0]` is not numeric, but a `Parameter`?

ihincks (approved):

qiskit_ibm_runtime/utils/utils.py
yaelbh: I'll check. Marking this conversation as resolved, since it won't let merge the branch otherwise.

yaelbh:
@ihincks @wshanks Is there a reason not to merge this PR ?

wshanks:
I think it is fine to merge (I don't have permission to merge myself).

PR #1990, opened 2024-10-22
Add support for noise model and level 1 data to local sampler

wshanks:
This change passes through the `simulator.noise_model` option to the `BackendSamplerV2` as a `noise_model` option of `run_options` if the primitive supports the `run_options` option (support was added in Qiskit 1.3).

Additionally, this change translates the `execution.meas_type` option into `meas_level` and `meas_return` options of `run_options` if the `BackendSamplerV2` supports `run_options`. This change allows support for level 1 data in local testing mode, where otherwise the default is only to return classified data.

### DIFF
--- a/qiskit_ibm_runtime/fake_provider/local_service.py
+++ b/qiskit_ibm_runtime/fake_provider/local_service.py
@@ -216,11 +216,43 @@ class QiskitRuntimeLocalService:
         options_copy = copy.deepcopy(options)
 
         prim_options = {}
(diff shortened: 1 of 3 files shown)


wshanks:
It doesn't make sense to merge this unless https://github.com/Qiskit/qiskit/pull/13357 is also on a path to being merged. If that is the case though, it should be safe to merge this because it should not change behavior if the added options are not used and also it checks that the options are defined before using them, so it doesn't need to wait for https://github.com/Qiskit/qiskit/pull/13357 to be released first.

kt474 (approved):

PR #2011, opened 2024-11-01
Add execution_span.TwirledSliceSpan

ihincks:
### Summary

This new `ExecutionSpan` subclass is added explicitly for the case where a sampler is called with twirling enabled. 

### Details and comments

`SliceSpan` and `DoubleSliceSpan` are insufficient because data from twirled experiments is not contiguous in the right way to make them work, especially when batching constants don't magically align with num_randomizations.

### DIFF
--- a/qiskit_ibm_runtime/execution_span/twirled_slice_span.py
+++ b/qiskit_ibm_runtime/execution_span/twirled_slice_span.py
@@ -29,16 +29,19 @@ class TwirledSliceSpan(ExecutionSpan):
 
     This type of execution span references pub result data that came from a twirled sampler
     experiment which was executed by either prepending or appending an axis to paramater values
-    to account for twirling. The value format of ``data_slices`` begins with the shape of the
-    parameter values accounting for twirling, and the second element indicates whether the
-    twirling axis is the first or the last. ``shape_slice`` and ``shots_slice`` are the same as
-    for :class:`~.DoubleSliceSpan`.
+    to account for twirling. See ``data_slices`` below for details.
 
     Args:
         start: The start time of the span, in UTC.
         stop: The stop time of the span, in UTC.
-        data_slices: A map from pub indices to
-            ``(twirled_shape, at_front, shape_slice, shots_slice)``.
+        data_slices: A map from pub indices to tuples of the form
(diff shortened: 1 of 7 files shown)


qiskit_ibm_runtime/execution_span/twirled_slice_span.py
```diff
@@ -0,0 +1,88 @@
...
+        if at_front:
+            # if the first axis is over twirling samples, push them right before shots
+            ndim = len(twirled_shape)
+            mask = mask.transpose(*range(1, ndim - 1), 0, ndim - 1)
+            twirled_shape = twirled_shape[1:-1] + twirled_shape[:1] + twirled_shape[-1:]
+
+        # merge twirling axis and shots axis before returning
+        return mask.reshape(*twirled_shape[:-2], math.prod(twirled_shape[-2:]))
+
+    def filter_by_pub(self, pub_idx: int | Iterable[int]) -> "TwirledSliceSpan":
```
SamFerracin, line 85: I think that with future annotations this should work:
```suggestion
    def filter_by_pub(self, pub_idx: int | Iterable[int]) -> TwirledSliceSpan:
```

ihincks: I'd like to call this out-of-scope because I have trouble believing this is true for Python 3.8, which we claim to support, but which I'm skeptical that we actually support.

qiskit_ibm_runtime/execution_span/twirled_slice_span.py
```diff
@@ -0,0 +1,88 @@
...
+
+from .execution_span import ExecutionSpan, ShapeType
+
+
+class TwirledSliceSpan(ExecutionSpan):
+    """An :class:`~.ExecutionSpan` for data stored in a sliceable format when twirling.
+
+    This type of execution span references pub result data that came from a twirled sampler
+    experiment which was executed by either prepending or appending an axis to paramater values
+    to account for twirling. The value format of ``data_slices`` begins with the shape of the
```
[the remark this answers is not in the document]
ihincks: just see the latest changes, i had trouble getting sphinx not to error about bullets in an args string.

PR #2012, opened 2024-11-01
Replace fake backend utility methods

### LINKED ISSUE #2010: Don't use fake backend utility functions from Qiskit
User reported that importing a fake backend would trigger this deprecation warning:

```
venv/lib/python3.11/site-packages/qiskit_ibm_runtime/fake_provider/backends/algiers/__init__.py:15: in <module>
    from .fake_algiers import FakeAlgiers
venv/lib/python3.11/site-packages/qiskit_ibm_runtime/fake_provider/backends/algiers/fake_algiers.py:18: in <module>
    from qiskit_ibm_runtime.fake_provider import fake_backend
venv/lib/python3.11/site-packages/qiskit_ibm_runtime/fake_provider/fake_backend.py:34: in <module>
    from qiskit.providers.fake_provider.utils.json_decoder import (
venv/lib/python3.11/site-packages/qiskit/providers/fake_provider/__init__.py:94: in <module>
    from .fake_backend import FakeBackend
venv/lib/python3.11/site-packages/qiskit/providers/fake_provider/fake_backend.py:22: in <module>
    from qiskit.providers.models import BackendProperties
venv/lib/python3.11/site-packages/qiskit/providers/models/__init__.py:80: in __getattr__
    warnings.warn(
E   DeprecationWarning: qiskit.providers.models is deprecated since Qiskit 1.2 and will be removed in Qiskit 2.0. With the removal of Qobj, there is no need for these schema-conformant objects. If you still need to use them, it could be because you are using a BackendV1, which is also deprecated in favor of BackendV2.
```

Looks like when we ported the backend model classes we didn't also port the decoders:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/3a698111ea8ea9f9c7b341d0e8c6da0ff402dd46/qiskit_ibm_runtime/fake_provider/fake_backend.py#L34-L38

We should already have these decoders in qiskit-ibm-runtime, since we need to decode backend data from the server.

kt474:
### Summary

The deprecation warning from the original issue is not from the fake backend utility methods. A couple V1 fake backends were missed in #1946 and those were causing this deprecation warning. The remaining V1 fake backends have also been removed in this PR. 

### Details and comments
Fixes #2010


qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -277,13 +277,13 @@ def _get_value(prop_dict: Dict, prop_name: str) -> Any:
...
-                        name,
-                        qubits,
-                    )
+                    # logger.debug(
+                    #     "Gate calibration for instruction %s on qubits %s is found "
+                    #     "in the PulseDefaults payload. However, this entry is not defined in "
+                    #     "the gate mapping of Target. This calibration is ignored.",
+                    #     name,
+                    #     qubits,
+                    # )
```
jyu00, line 286: Did you mean to comment this out? I don't actually oppose that since we get a lot of this log, but we should open an issue to investigate why it's happening.

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -31,15 +31,15 @@
...
-    decode_backend_configuration,
-    decode_backend_properties,
-    decode_pulse_defaults,
-)
 
 from qiskit.providers.basic_provider import BasicSimulator
 
 from qiskit_ibm_runtime.utils.backend_converter import convert_to_target
+from qiskit_ibm_runtime.utils.backend_decoder import (
+    _decode_backend_configuration,
```
jyu00, line 39: Can we rename the function to not start with `_` if it's to be imported and used elsewhere?

jyu00 (commented):
Can you also add a release note for the deleted backends? We previously got complaints about that 😅.

qiskit_ibm_runtime/utils/backend_converter.py
kt474: Oops nope, I did this to make it easier to debug. I'll change it back and create a separate issue to look into this.

ptristan3 (approved):

PR #2014, opened 2024-11-04
Fix `draw_execution_spans()` hover text placement

ihincks:
### Summary

Prior to this PR, the location of hover text was off by a factor of three because they were not accounting for the fact that every plotted span essentially corresponds to three marker positions: the time start, the time stop, the None value indicating "move to the next one".


SamFerracin (approved):

SamFerracin:
Checked it locally, looks good

PR #2021, opened 2024-11-07
Verify values of parametrized rzz gates

### LINKED ISSUE #2000: Consider handling rzz validation for parametrized angles

yaelbh:
### Summary

Fixes #2000.

This is only a draft. I forgot to write the tests before the code. So the current push contains the logic, but the tests will be added shortly.

### Details and comments

@SamFerracin I have a question about the noise learner. I see that `NoiseLearner._run` calls `validate_isa_circuits`, like `BasePrimitiveV2`. I added in `BasePrimitivesV2` a call to a new function, `validate_rzz_pubs`. But I'm not familiar enough with the noise learner to figure if a similar call should be added there too. I tried to add, but `NoiseLearner._run` receives circuits, unlike `BasePrimitiveV2`, which receives pubs. Should I make `_run` receive pubs and add the call to the new function?


SamFerracin:
@yaelbh it's not entirely clear to me what should be done, I'd have to think about it a little bit. But let me tell you what noise learner does:

NoiseLearner receives one or more circuits. Instead of running noise learning on them directly, it invokes the noise learning steps of the estimator. In more detail:
- It packages the circuits into PUBs by adding dummy observables that will never actually be estimated
- It sets up an estimator with `pec` turned on and `default_shots=0`. In that way, the noise learning part of PEC will be run, but not the estimation part
- It runs the estimator job and extracts the noise data from the returned metadata.

So every validation check performed by pec-runtime's estimator is going to be performed on the noise learner's PUBs

yaelbh:
Following https://github.com/Qiskit/qiskit-ibm-runtime/pull/2035, we should also take into account parameter expressions.

yaelbh:
A couple of comments:
1. The check for fixed angles will take place in both `is_isa_circuit` and `is_rzz_pub`. In `is_isa_circuit` -  because not everyone has a full pub and can call `is_rzz_pub` - see the noise learner. In `is_rzz_pub` - because this way the function makes more sense and is more complete, this will be less confusing for developers, who may naturally expect fixed angles to be covered.
2. Parameter expressions are much more complicated, and will be computationally more expensive. An expression can consist of multiple parameters. We will have to evaluate each expression with all the parameter values for its parameters. At least for now this PR continues to skip parameter expressions.

qiskit_ibm_runtime/utils/utils.py
```diff
@@ -116,6 +120,111 @@ def is_isa_circuit(circuit: QuantumCircuit, target: Target) -> str:
...
+        # accurate).
+        if operation.name == "rzz":
+            angle = instruction.operation.params[0]
+            if isinstance(angle, Parameter):
+                angle_params.add(angle.name)
+            elif not isinstance(angle, ParameterExpression) and (
+                angle < 0.0 or angle > 1.001 * np.pi / 2
+            ):
+                return f"The rzz instruction is supported only for angles in the \
+                range [0, pi/2], but an angle of {angle} has been provided."
```
t-imamichi, line 154: This should be fixed by copying #2035 to remove the redundant white spaces.
The current code include redundant white spaces before `range [0, pi/2]` due to `\`.

qiskit_ibm_runtime/utils/utils.py
```diff
@@ -116,6 +120,113 @@ def is_isa_circuit(circuit: QuantumCircuit, target: Target) -> str:
...
+        # accurate).
+        if operation.name == "rzz":
+            angle = instruction.operation.params[0]
+            if isinstance(angle, Parameter):
+                angle_params.add(angle.name)
+            elif not isinstance(angle, ParameterExpression) and (
+                angle < 0.0 or angle > 1.001 * np.pi / 2
+            ):
+                return (
+                    f"The instruction rzz is supported only for angles in the "
```
t-imamichi, line 154:
```suggestion
                    "The instruction rzz is supported only for angles in the "
```

t-imamichi (approved):
LGTM

qiskit_ibm_runtime/utils/utils.py
```diff
@@ -116,6 +120,113 @@ def is_isa_circuit(circuit: QuantumCircuit, target: Target) -> str:
...
+    rzz_params = list(helper_result)
+
+    param_values = pub.parameter_values
+    # param_values is of the form:
+    # BindingsArray(<shape=(2, 2, 3), num_parameters=4, parameters=['a', 'b', 'c', 'd']>)
+    # param_values.data is a dictionary, whose keys are tuples of parameter names.
+    # For examples, the keys can be: dict_keys([('a', 'b'), ('c',), ('d',)])
+
+    pub_params = list(chain(*[list(param_names) for param_names in param_values.data.keys()]))
+    # pub_params is the list of parameter names in the pub, for example: ['a', 'b', 'c', 'd']
```
ihincks, line 195: these lines are instantiating quite a few more intermediate lists than necessary. also remember that `.keys()` is just about never needed.

```suggestion
    # gather all parameter names, in order
    pub_params = list(chain.from_iterable(pub.parameter_values.data))
```

qiskit_ibm_runtime/utils/utils.py
```diff
@@ -116,6 +120,113 @@ def is_isa_circuit(circuit: QuantumCircuit, target: Target) -> str:
...
+        # rzz gate is calibrated only for the range [0, pi/2].
+        # We allow an angle value of a bit more than pi/2, to compensate floating point rounding
+        # errors (beyond pi/2 does not trigger an error down the stack, only may become less
+        # accurate).
+        if operation.name == "rzz":
+            angle = instruction.operation.params[0]
+            if isinstance(angle, Parameter):
+                angle_params.add(angle.name)
+            elif not isinstance(angle, ParameterExpression) and (
+                angle < 0.0 or angle > 1.001 * np.pi / 2
```
ihincks, line 151: There is an asymmetry here in that rounding errors are compensated on the upper end, but not the lower end. for example, `-1e-12` is invalid whereas `np.pi/2 + 1e-12` is valid.

I'm also  a bit surprised that the level is at 1e-3, very far away from machine epsilon. For example, is there a problem moving this to 1e-10 which is still 100x machine epsilon?

```suggestion
                not -EPSILON < angle < np.pi / 2 + EPSILON
```

wshanks: This mirrors line 82 above. I think I contributed to the original choice of 1.001. The hardware is optimized for pi/2 because the main focus is the cz gate. It doesn't blow up if you go just past pi/2 but you don't want to go much past it (not calibrated and will eventually hit some hardware/physical limit). So it is not critical that the limit be tight on pi/2 but whatever is likely not to run into rounding issues should be okay.

yaelbh: Then, what's the decision, for both ends?

yaelbh: Once there is a decision, I'll of course fix (if needed) not only here, but also in several other places where this check takes place (I've spotted that this check appears in three different places, maybe I'll write a tiny function for it).

wshanks: I am okay with 1e-10. I think we only document pi/2 as the limit and here we are just trying to avoid being too strict where in practice the user won't notice the difference.

There was also the point about the asymmetry with negative angles. I am not sure what to do about that right now. The backend will error right now if it gets a negative number. We could change the backend to round small negative numbers to 0 if we wanted symmetry of behavior for the bounds.

yaelbh: Based on this conversation, I'm keeping the current behavior for the negative angles. Let me know if it's a problem.

yaelbh:
@kt474 @wshanks It's ready for another review. Note that I added release notes.

release-notes/unreleased/2021.feat.rst
```diff
@@ -0,0 +1 @@
+A new function `validate_rzz_pubs` was added. The function verifies that `rzz` parameters are in the range between `0` and `pi/2`, for numeric parameters (e.g. `rzz(np.pi/4, 0)`), and for unbounded parameters (`rzz(theta, 0)`) with values to substitute provided in the pub. Parameter expressions (e.g. `rzz(theta + phi, 0)`) are still not validated.
```
wshanks, line 1: Since this is rst, I think you meant to use double back ticks.
```suggestion
A new function ``validate_rzz_pubs`` was added. The function verifies that ``rzz`` parameters are in the range between ``0`` and ``pi/2``, for numeric parameters (e.g. ``rzz(np.pi/4, 0)``), and for unbounded parameters (``rzz(theta, 0)``) with values to substitute provided in the pub. Parameter expressions (e.g. ``rzz(theta + phi, 0)``) are still not validated.
```

qiskit_ibm_runtime/utils/validations.py
```diff
@@ -102,6 +103,18 @@ def validate_isa_circuits(circuits: Sequence[QuantumCircuit], target: Target) ->
...
 
 
+def validate_rzz_pubs(pubs: Union[List[EstimatorPub], List[SamplerPub]]) -> None:
+    """Validate that rzz angles are always in the range [0, pi/2]
+
+    Args:
+        pubs: A list of pubs.
+    """
+    for pub in pubs:
+        message = is_rzz_pub(pub)
```
wshanks, line 113: Minor, but should it be called `is_not_rzz_pub`? It returns something True-like when the pub does not have rzz.

test/unit/test_sampler.py
```diff
@@ -311,11 +312,85 @@ def test_rzz_validates_only_for_fixed_angles(self):
...
+    def test_rzz_validation_skips_param_exp(self):
+        """Verify that the rzz validation occurs only when the angle is a number or a parameter,
+        but not a parameter expression"""
+        backend = FakeFractionalBackend()
+        param = Parameter("p")
+
+        circ = QuantumCircuit(2)
+        circ.rzz(2 * param, 0, 1)
+
+        # Should run without an error
```
wshanks, line 392: I think this test is just testing that using parameter expressions does not produce an unexpected error? I think ideally this case of using a parameter expression that evaluates to an invalid angle value would be an error. Maybe that should be noted in the comment.

wshanks (commented):
This looks good to me except for a minor formatting change on the release note. I put in two other minor suggestions.

qiskit_ibm_runtime/utils/validations.py
yaelbh: I'll rename to `is_valid_rzz_pub`, it's not about whether there's `rzz` inside but about whether all the `rzz` are valid, I hope that the new name will be clear...

yaelbh: Done in e98bcc13

test/unit/test_sampler.py
yaelbh: Done in bed83268.
I'll open a separate issue for parameter expressions. I'm not sure about the performance implications.

wshanks (approved):
Looks good to me!

qiskit_ibm_runtime/utils/utils.py
wshanks: All conversations must be resolved before merging. Can we mark this as resolved @ihincks? I think the remaining options would be to round up small negative angles/parameters to 0 (a change of scope since the validation code only checks without modifying right now) or to change the backend code to accept small negative angles (out of scope here for now).

PR #2022, opened 2024-11-07
Prepare release 0.33.0

kt474:
### Details and comments
Fixes #

### DIFF
--- a/release-notes/0.33.0.rst
+++ b/release-notes/0.33.0.rst
@@ -1,0 +1,27 @@
+0.33.0 (2024-11-07)
+===================
+
+New Features
+------------
+
+- Added :func:`~.draw_layer_errors_swarm` which draws a swarm plot of one or more
+  :class:`~.LayerError` objects. Also added the convenience method
+  :meth:`~.LayerError.draw_swarm` to invoke the drawing function on a particular instance. (`1988 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/1988>`__)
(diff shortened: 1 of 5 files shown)


release-notes/0.33.0.rst
```diff
@@ -0,0 +1,27 @@
...
+===================
+
+New Features
+------------
+
+- Added :func:`~.draw_layer_errors_swarm` which draws a swarm plot of one or more
+  :class:`~.LayerError` objects. Also added the convenience method
+  :meth:`~.LayerError.draw_swarm` to invoke the drawing function on a particular instance. (`1988 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/1988>`__)
+- Added :class:`.TwirledSliceSpan`, an :class:`ExecutionSpan` to be used when 
+  twirling is enabled in the sampler. In particular, it keeps track of an extra shape
```
abbycross, line 11:
```suggestion
  twirling is enabled in the Sampler. In particular, it keeps track of an extra shape
```

release-notes/0.33.0.rst
```diff
@@ -0,0 +1,27 @@
+0.33.0 (2024-11-07)
+===================
+
+New Features
```
abbycross, line 4: We use sentence case for headers (so the words after the first word wouldn't be upper-case), but I hesitate to suggest fixing this if the release notes have been done this way in many past iterations, meaning a break in consistency ... I leave it to the runtime team to make that call.

abbycross (approved):
Made a couple comments - looks good otherwise!

PR #2027, opened 2024-11-08
Remove 0.26.0 deprecations

### LINKED ISSUE #1681: Inputs to Session and Batch should order of Sampler and Estimator

The inputs to `Session` and `Batch` list the `session` as the first arg followed by `backend`:

`class Session(service=None, backend=None, max_time=None)`

`class Batch(service=None, backend=None, max_time=None)`

which is the opposite of `Sampler` and `Estimator`:

`class SamplerV2(backend=None, session=None, options=None)`

`class EstimatorV2(backend=None, session=None, options=None)`

It would be nice to have a unified calling order.  In particular, the sampler and estimator were written to have backend first because it makes the code cleaner because users are not required to pass `backend=backend` into the init.  The `Session` and `Batch` should follow suit. 

**Acceptance criteria**
The kwargs orders are unified to match the `sampler` and `Estimator` ordering.

### LINKED ISSUE #1727: Give session/batch cm higher precedence than backend

Today if someone does 

```python

sampler1 = SamplerV2(backend=backend)

with Session(backend=backend):
    sampler2 = SamplerV2(backend=backend)
    sampler1.run(...)
    sampler2.run(...)
```

Both `sampler1.run()` and `sampler2.run()` will actually run in job mode, not session, even though session mode is more likely what they want,  given `.run()` is invoked under the `Session` context manager. While this is technically a user error, it's not intuitive that specifying the `backend` keyword would negate the Session context (although it'd help when the `backend` parameter is dropped in favor of `mode`).  

We should make it so that if `run()` is called inside a `Session` or `Batch` context, session/batch is used. Although this is a breaking change so we'd need to do some warning first. 

**Acceptance criteria**

- [x] Issue a deprecation warning if a job inside a session/batch context will actually run in job mode. The warning message should say the job will run in job mode, and the behavior will change to session/batch mode in the near future. 
- [ ] After the deprecation period, make the job to actually run in session/batch mode if it's inside a session/batch context
- [ ] If a backend is specified for both the primitive and session/batch, raises an error, e.g. 
    ```python
    with Session(backend=bar):
        sampler = SamplerV2(backend=foo)
        sampler.run()  # <-- this should raise an error since we don't know if `foo` or `bar` should be used as the backend
    ```

### LINKED ISSUE #1886: `Session.from_id` unexpectedly suppresses deprecation warnings.
**Desribe the bug**

When `Session.from_id` is called, internally [it calls](https://github.com/Qiskit/qiskit-ibm-runtime/blob/8c612d1343c2b8fd800fd0562974be783c58bdaa/qiskit_ibm_runtime/session.py#L379) `cls(service, backend)` to create a `Session` instance, with `backend` [set to a string](https://github.com/Qiskit/qiskit-ibm-runtime/blob/8c612d1343c2b8fd800fd0562974be783c58bdaa/qiskit_ibm_runtime/session.py#L370). Unless it's intentional to suppress warnings from `Session.__init__` when called from `Session.from_id`, this should raise two deprecation warnings (one for [passing a service](https://github.com/Qiskit/qiskit-ibm-runtime/blob/8c612d1343c2b8fd800fd0562974be783c58bdaa/qiskit_ibm_runtime/session.py#L118) and one for [backend as a string](https://github.com/Qiskit/qiskit-ibm-runtime/blob/8c612d1343c2b8fd800fd0562974be783c58bdaa/qiskit_ibm_runtime/session.py#L143)), but raises none. 

**Steps to reproduce**

First, create a session instance and grab the ID:
```
from qiskit_ibm_runtime import Session, QiskitRuntimeService

service = QiskitRuntimeService() # account info previously saved
backend = service.least_busy(operational=True, simulator=False)
session = Session(backend=backend)

session_id = session.session_id
```

Next, regenerate the session from its ID:
```
regenerated_session = Session.from_id(session_id, service) # produces no warnings
```

**Expected behavior**

Deprecation warnings raised for service and backend as a string, as for example when I use `Session.__init__` directly:

```
backend = service.least_busy(operational=True, simulator=False)
session = Session(service=service, backend=backend.name)

/var/folders/7_/rsgnsxr55d32q72f3m991h5r0000gn/T/ipykernel_33497/3846134487.py:6: DeprecationWarning: The service parameter is deprecated as of qiskit-ibm-runtime 0.26.0 and will be removed no sooner than 3 months after the release date. The service can be extracted from the backend object so it is no longer necessary.
  session = Session(service=service, backend=backend.name)
/var/folders/7_/rsgnsxr55d32q72f3m991h5r0000gn/T/ipykernel_33497/3846134487.py:6: DeprecationWarning: Passing a backend as a string is deprecated as of qiskit-ibm-runtime 0.26.0 and will be removed no sooner than 3 months after the release date. Use the actual backend object instead.
  session = Session(service=service, backend=backend.name)
```

**Suggested solutions**

Not sure how to solve the problem in its entirety, but part of the issue seems to be how `issue_deprecation_msg` is used in `Session.__init__`. As used there, the `stacklevel` passed to `warnings.warn` is 3 (default value of 2, [plus one](https://github.com/Qiskit/qiskit-ibm-runtime/blob/8c612d1343c2b8fd800fd0562974be783c58bdaa/qiskit_ibm_runtime/utils/deprecation.py#L81)), which works when a session is created via `__init__` directly. In that case, the stack is: `issue_deprecation_msg` -> `Session.__init__` -> caller. But when `Session.from_id` is used, the stack is too deep: `issue_deprecation_msg` -> `Session.__init__` -> `Session.from_id` -> caller.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.28.0
- **Python version**: 3.11
- **Operating system**: MacOS Sonoma 14.6.1

kt474:
### Summary

Changes in this PR: 
  - Passing in a backend as a string is no longer valid 
  - If a backend is passed to sampler/estimator from within an open session context manager, the job will now run from within that session. If the backend passed in is different from the session backend, an error will be raised. 
  - `Service` is no longer a valid parameter in Session/Batch
  - Refactored `Session.from_id()` because it used to use ``service``. 
  - Removed `QiskitRuntimeService.global_service` because this was only necessary when backends could be given as strings. 
  - Update tests to support all the above changes

I can also split this PR into multiple if that makes it easier to review

### Details and comments
Fixes #1681 
Fixes #1727 
Fixes #1886


qiskit_ibm_runtime/batch.py
```diff
@@ -87,19 +86,13 @@ class Batch(Session):
 
     def __init__(
         self,
-        service: Optional[QiskitRuntimeService] = None,
         backend: Optional[Union[str, BackendV1, BackendV2]] = None,
```
ptristan3, line 89: the `backend` field should support only `BackendV1` or `BackendV2`

qiskit_ibm_runtime/session.py
```diff
@@ -115,43 +108,14 @@ def __init__(
...
-                remedy=(
-                    "The service can be extracted from the backend object so "
-                    "it is no longer necessary."
-                ),
-                period="3 months",
-            )
-
-        self._service = service
         if isinstance(backend, IBMBackend):
             self._service = self._service or backend.service
```
ptristan3, line 112: at this point `self._service` is None so it should be just`backend.service`

ptristan3 (approved):

PR #2036, opened 2024-11-11
Update job status deprecation msg

kt474:
### Details and comments
Fixes #


qiskit_ibm_runtime/runtime_job.py
```diff
@@ -196,6 +196,8 @@ def status(self) -> JobStatus:
                 "In a future release of qiskit-ibm-runtime no sooner than 3 months "
                 "after the release date of 0.30.0, RuntimeJob.status() will be returned as a string "
                 "instead of an instance of `JobStatus`."
```
ihincks, line 198:
```suggestion
                "instead of an instance of `JobStatus`. "
```

ihincks (approved):

Issue #2032, opened 2024-11-10
Transpilation can result in rzz gates with invalid angles


yaelbh:
`rzz` angles are restricted to the range `[0, pi/2]`, but the transpiler doesn't know it. A possible resolution is to flip the sign by adding a transpiler pass that surrounds rzz gates with negative angles by XI on both sides. This can be applied as a post translation plugin in qiskit-ibm-runtime (because the angle limitation is IBM specific).

yaelbh:
We should consider the case where the circuit is parametrized, and the rzz gate in the transpiled circuit is parametrized. Do you have an idea how to handle it?

yaelbh:
Below is an example of a `surprise-rzz` gate, which shows up after transpilation. The user assigns say a negative value to the parameter of the unrestricted gate `crz`, but the transpiled circuit, containing the `surprise-rzz` gate won't be able to run. The pass from #2043 won't help because it doesn't address unbounded parameters.

```python
from qiskit.circuit import QuantumCircuit, Parameter
from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager

from qiskit_ibm_runtime import QiskitRuntimeService


service = QiskitRuntimeService()
backend = service.backend("ibm_fez", use_fractional_gates=True)

p = Parameter("p")

circ = QuantumCircuit(2)
circ.crz(p, 0, 1)


pm = generate_preset_pass_manager(optimization_level=0, target=backend.target)
isa_circ = pm.run(circ)

print(isa_circ)
```

Output:
```

          q_0 -> 0 ────────────■─────────
                   ┌─────────┐ │ZZ(-p/2)
          q_1 -> 1 ┤ Rz(p/2) ├─■─────────
                   └─────────┘
```

yaelbh:
A couple of ideas:
- (credit to Eli) Scan the circuit, if it contains rzz gates with unbounded parameters then drop rzz from the target gates, temporarily, during transpilation.
- (credit to Liran) Transform the pass of #2043 into a pass that can work with unbounded parameters: the pass will generate a circuit starting with two rz gates (for each qubit), followed by an rx gates, an rzz gate, and an additional rx gate. The angles in all these gates will be expressions over the same parameter, for example for the rx gates the angle will be `pi * (-pi/2 < theta < 0 or pi/2 < theta < pi)`.    (we will also have some modulo 2pi here).

[...]

PR #2043, opened 2024-11-13
Add plugin pass to support non-calibrated rzz angles

nkanazawa1989:
### Summary

IBM Quantum recently released [new basis set including rzz gate](https://www.ibm.com/quantum/blog/fractional-gates). The Rzz gate supports continuous angle but it is calibrated only within [0, pi/2]. We have implemented the validation mechanism to detect non-calibrated angles, but user always need to manually tweak circuits to avoid angles outside this range, which is a bit cumbersome.

This PR adds a plugin pass to automatically modifies the input circuit so that end users don't need to consider our hardware constraints. However, a user still needs to consider the constraints when they send parameterized PUB with Rzz gates. The latter case is outside the scope of this PR.

### Details and comments
This PR adds new translation plugin. Alternatively we can add new pass to existing `ibm_dynamic_circuits` entry point, but name is very specific to the dynamic circuits feature and adding fractional feature under this plugin name seems confusing. For now I decided to create new entry point `ibm_fractional` but eventually these must be merged when our hardware supports both features with a single execution path.


yaelbh:
Thanks @nkanazawa1989.

As much as users will certainly be happy to have all angle values covered, it will also be confusing for them to see that values that are supposed to be invalid pass all the checks and run successfully.

Therefore:
1. We should document that fixed angles can have any value, whereas parametrized angles are restricted.
2. We should similarly rephrase the ISA error message.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -0,0 +1,185 @@
...
+            elif pi / 2 < wrap_angle <= pi:
+                replace = _quad2(wrap_angle, node.qargs)
+            else:
+                if angle != wrap_angle:
+                    dag.substitute_node(
+                        node,
+                        RZZGate(wrap_angle),
+                        inplace=True,
+                    )
+                continue
```
yaelbh, line 59: Please explain, thanks

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
```diff
@@ -0,0 +1,63 @@
...
+        ("large_positive_number", 12345),
+        ("large_negative_number", -12345),
+        ("pi/2_pos", pi / 2),
+        ("pi/2_neg", -pi / 2),
+        ("pi_pos", pi),
+        ("pi_neg", -pi),
+        ("quad1", 0.1),
+        ("quad2", pi / 2 + 0.1),
+        ("quad3", -pi + 0.1),
+        ("quad4", -0.1),
```
yaelbh, line 41: The rzz matrix seems to have a period of 4pi. Is it correct? If yes, then I believe the cases of -3pi/2 and 3pi should also be included.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
yaelbh: I understand what confuses me. On the one hand, from what I  see the range to consider should be [-2pi, 2pi] (see https://github.com/Qiskit/qiskit-ibm-runtime/pull/2043/files#r1841666155). On the other hand, I must get something wrong, because according to the code here the period is 2pi.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -0,0 +1,185 @@
...
+
+    This pass preserves the number of Rzz gates, but it may add
+    extra single qubit gate operations.
+    These gates might be reduced by the following
+    single qubit optimization passes.
+    """
+
+    def run(self, dag: DAGCircuit) -> DAGCircuit:
+        # Currently control flow ops and Rzz cannot live in the same circuit.
+        # Once it's supported, we need to recursively check subroutines.
```
yaelbh, line 39: I prefer to already support subroutines, to avoid pitfalls when the time comes and dynamic circuits are allowed.

yaelbh:
What I wrote [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2043#issuecomment-2475554691) is not accurate. An ISA circuit is still a circuit where all rzz angles are in the range [0, pi/2].

How can we let the users know that, if they transpile with a backend that enables fractional gates, then they enjoy the service of the transpiler taking care of the angle values for them?

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
nkanazawa1989: After phase wrapping you always get [-pi, pi]

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -28,35 +28,59 @@ class FoldRzzAngle(TransformationPass):
...
+    Added local gates might be efficiently merged into
+    neighboring single qubit gates by the following single qubit
+    optimization passes.
+
+    This pass allows the Qiskit users to naively use the Rzz gates
+    with angle of arbitrary real numbers.
+
+    .. note::
+        This pass doesn't transform the circuit when the
+        Rzz gate angle is unbound parameter.
```
yaelbh, line 48:
```suggestion
        Rzz gate angle is an unbound parameter.
```

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -28,35 +28,59 @@ class FoldRzzAngle(TransformationPass):
...
+        In this case, the user must assign gate angle before
```
yaelbh, line 49:
```suggestion
        In this case, the user must assign a gate angle before
```

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -28,35 +28,59 @@ class FoldRzzAngle(TransformationPass):
...
+        transpile, or be responsible for choosing parameters
```
yaelbh, line 50:
```suggestion
        transpilation, or be responsible for choosing parameters
```

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
nkanazawa1989: I usually don't like writing unreachable logic because of the maintenance overhead. But the code is in [013a6b4](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2043/commits/013a6b45f54138587ee95ec7ead0e31250d58570)

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
```diff
@@ -61,3 +62,14 @@ def test_folding_rzz_angle_unbound(self):
         pm = PassManager([FoldRzzAngle()])
         isa = pm.run(qc)
         self.assertEqual(qc, isa)
+    
+    def test_identity(self):
+        """Create identity circuit with angles with [-4pi, 4pi] and transpile."""
+        # Angle sum is zero
+        angles = pi * np.linspace(-4, 4, 17)
```
yaelbh, line 69: Did you mean 4pi?

nkanazawa1989:
Thanks @yaelbh for review. I also feel we have vague definition of "ISA". In case of Rzz, the instruction becomes ISA for "IBM Quantum" only when angle in [0, pi/2] (this is why the pass is plugin). I added more document in [a45251a](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2043/commits/a45251afcdd2d24ca44ae5913e54a5e7398014e0). Hope it improves understanding.

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
nkanazawa1989: Yes, it's 4 pi. `pi * np.linspace(-4, 4, 17)` = `np.linspace(-4*pi, 4*pi, 17)`

yaelbh: Oops, I didn't notice the `pi*`.

nkanazawa1989:
> does not mention that you don't have to worry about it if you only transpile your circuit.

+1. @kaelynj can you update the doc?

yaelbh:
> +1. @kaelynj can you update the doc?

Just not yet... Thanks :smile:.

nkanazawa1989:
Thanks for linking the issue. Probably it's safer to merge this first as you mention. Qiskit core doesn't need to consider angle restriction because this is very IBM Quantum specific.

yaelbh:
@kt474 I'm reviewing the transpiler pass, but can you please review the design of the entry point `ibm_fractional`? Maybe we should consult Matthew about it?

yaelbh (changes requested):
I'm not sure exactly how to proceed with this PR, now that Naoki is on leave. I guess that I'll take ownership on it. @wshanks and @kt474, can you please read my comments, see if they make sense, and answer questions therein?

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def test_folding_rzz_angle_unbound(self):
+        """Test skip folding unbound gate angle."""
+        qc = QuantumCircuit(2)
+        qc.rzz(Parameter("θ"), 0, 1)
+        pm = PassManager([FoldRzzAngle()])
+        isa = pm.run(qc)
+        self.assertEqual(qc, isa)
+
+    def test_identity(self):
```
yaelbh, line 66: The test is designed in a way that doesn't catch the bug that I detected [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2043/files#r1855446379).

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -0,0 +1,239 @@
...
+                        inplace=True,
+                    )
+                    modified = True
+                continue
+
+            if not isinstance(node.op, RZZGate):
+                continue
+
+            angle = node.op.params[0]
+            if isinstance(angle, ParameterExpression) or 0 <= angle <= pi / 2:
```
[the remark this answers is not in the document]
wshanks: That sounds reasonable to me.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -0,0 +1,239 @@
...
+
+            # Modify circuit around Rzz gate to address non-ISA angles.
+            modified = True
+            wrap_angle = np.angle(np.exp(1j * angle))
+            if 0 <= wrap_angle <= pi / 2:
+                # In the first quadrant after phase wrapping.
+                # We just need to remove 2pi offset.
+                dag.substitute_node(
+                    node,
+                    RZZGate(wrap_angle),
```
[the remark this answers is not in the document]
wshanks: In the code the angle is wrapped with `np.angle(np.exp(1j * angle))`. In your example, the angle is divided by 2 before or after the wrapping, but I don't understand what the connection is between this division by 2 and the transpiler plugin.

qiskit_ibm_runtime/transpiler/plugin.py
```diff
@@ -96,3 +97,41 @@ def pass_manager(
             plugin_passes += [ConvertConditionsToIfOps()]
 
         return PassManager(plugin_passes) + translator_pm
+
+
+class IBMFractionalTranslationPlugin(PassManagerStagePlugin):
+    """A translation stage plugin for targeting Qiskit circuits
+    to IBM Quantum system with fractional gate support.
+
+    Currently coexistense of fractional gate operations and
```
wshanks, line 106:
```suggestion
    Currently coexistence of fractional gate operations and
```

qiskit_ibm_runtime/transpiler/plugin.py
```diff
@@ -96,3 +97,41 @@ def pass_manager(
...
+    dynamic circuit is not assumed.
+    """
+
+    def pass_manager(
```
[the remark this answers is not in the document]
wshanks: Would @kt474 be the one to decide how to handle the divergence? The behavior here seems reasonable to me.

wshanks (commented):
@yaelbh I think we can take over this PR. I didn't understand the bug you mentioned and I would rather defer to someone else on the consistency of the transpiler plugins. I think otherwise this looks good. I will apply the wording suggestions you made.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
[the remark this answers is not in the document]
yaelbh: I think transpiler passes are expected not to tolerate global phase. This is in order to support controlled gates.

```python
from qiskit.quantum_info import Operator
from qiskit.transpiler.passmanager import PassManager

from qiskit_ibm_runtime.transpiler.passes.basis import FoldRzzAngle


non_isa = QuantumCircuit(2)
non_isa.rzz(7, 0, 1)

pm = PassManager([FoldRzzAngle()])
isa = pm.run(non_isa)

# expected to be True - so far so good - they are equivalent up to global phase
print(Operator.from_circuit(non_isa).equiv(Operator.from_circuit(isa)))


controlled_non_isa = non_isa.to_gate().control()
controlled_isa = isa.to_gate().control()

# False!
print(Operator.from_circuit(controlled_non_isa).equiv(Operator.from_circuit(controlled_isa)))
```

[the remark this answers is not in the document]
wshanks: Thanks, @yaelbh. I see. So this relates to your other question about the 4 pi periodicity. The rzz gate has periodicity of 4 pi but the behavior for (2 pi, 4 pi) is the same as (0, 2 pi) other than an overall -1 factor that is often irrelevant. So would the fix be to keep the current behavior but add an extra `GlobalPhaseGate(pi)` to the circuit if the angle wraps to (pi, 3 pi) when taking the modulus with 4 pi?

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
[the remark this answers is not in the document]
yaelbh: I actually prefer to leave it like this, instead of carrying that `1e-10` all over.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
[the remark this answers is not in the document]
yaelbh: Yes, I'll do it.

yaelbh:
I'll take care of this PR today. Main work items:
1. Apply global phase. 
2. Add an integration test, which will verify that `transpile` invokes the new pass. 
3. Probably improve the tests a bit.
4. Ask Kevin and Shelly to review the plugin. 
5. Go over comments along the PR and make sure that they are all addressed.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -0,0 +1,239 @@
...
+            elif -pi / 2 < wrap_angle < 0:
+                # In the forth quadrant.
+                replace = _quad4(wrap_angle, node.qargs)
+            else:
+                raise RuntimeError("Unreacheable.")
+            dag.substitute_node_with_dag(node, replace)
+        return modified
+
+
+def _quad2(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
```
yaelbh, line 118: I think the `_quad` functions should be static functions inside the `FoldRzzAngle` class, @wshanks do you agree?

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
yaelbh: The new test `test_identity` can miss bugs due to symmetry. Namely, a bug may show up twice and cancel itself. I don't see an issue with adding four more cases to ddt.

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
yaelbh: I think we can remove this test.

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
```diff
@@ -0,0 +1,102 @@
...
+    def test_folding_rzz_angles(self, angle):
+        """Test folding gate angle into calibrated range."""
+        qc = QuantumCircuit(2)
+        qc.rzz(angle, 0, 1)
+        pm = PassManager([FoldRzzAngle()])
+        isa = pm.run(qc)
+
+        self.assertTrue(Operator.from_circuit(qc).equiv(Operator.from_circuit(isa)))
+        for inst_data in isa.data:
+            if inst_data.operation.name == "rzz":
```
yaelbh, line 53: I don't like this `if`. We should be able to predict `isa.data` exactly. In particular, if `isa.data` for some (erroneous) reason does not contain `rzz`, the test will still pass.

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
```diff
@@ -0,0 +1,102 @@
...
+                fold_angle = inst_data.operation.params[0]
+                self.assertGreaterEqual(fold_angle, 0.0)
+                self.assertLessEqual(fold_angle, pi / 2)
```
yaelbh, line 56: We should verify correctness, and not only validity of angle values.

yaelbh:
@kt474 Can you please explain how the transpiler plugins mechanism works?

I tried:
```python
    def test_fractional_pluging(self):
        """ Verify that a pass manager created for a fractional backend applies the rzz folding
        pass"""

        circ = QuantumCircuit(2)
        circ.rzz(7, 0, 1)

        # this commented section is not working either
        #pm = generate_preset_pass_manager(optimization_level=0, target=FakeFractionalBackend().target)
        #isa_circ = pm.run(circ)

        from qiskit import transpile
        isa_circ = transpile(circ, FakeFractionalBackend())

        print(isa_circ)
```
But the pass manager is not applied, what am I doing wrong?

yaelbh:
Note a comment in another issue, that's related to this PR: https://github.com/Qiskit/qiskit-ibm-runtime/issues/2032#issuecomment-2504328147.

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
```diff
@@ -0,0 +1,106 @@
...
+from qiskit.transpiler.passmanager import PassManager
+from qiskit.circuit.parameter import Parameter
+
+from qiskit_ibm_runtime.transpiler.passes.basis import FoldRzzAngle
+from .....ibm_test_case import IBMTestCase
+
+
+@ddt
+class TestFoldRzzAngle(IBMTestCase):
+    """Test FoldRzzAngle pass"""
```
ShellyGarion, line 30: Perhaps it's worth to add tests that also gates that can be decomposed into `RZZGate` (using Qiskit equivalence library) are also being decomposed correctly. 
These gates include:
`CPhaseGate, RXXGate, RYYGate, RZXGate, CRX, CRY, CRZ`

wshanks:
@yaelbh Regarding the transpiler plugin, I think the problem is that `FakeFractionalBackend` inherits from `FakeBackendV2` which inherits from `BackendV2`. I think the extra transpiler passes get added into the preset pass managers through hook methods on the backend like this one on `IBMBackend` (and we probably need to add a hook to `IBMBackend` as well, and perhaps also to `FakeFractionalBackend`):

https://github.com/Qiskit/qiskit-ibm-runtime/blob/639f3071833cef46443adf83eba7ea925b00257c/qiskit_ibm_runtime/ibm_backend.py#L912-L914

qiskit_ibm_runtime/transpiler/plugin.py
```diff
@@ -96,3 +97,41 @@ def pass_manager(
...
+        translator_pm = common.generate_translation_passmanager(
+            target=pass_manager_config.target,
+            basis_gates=pass_manager_config.basis_gates,
+            approximation_degree=pass_manager_config.approximation_degree,
+            coupling_map=pass_manager_config.coupling_map,
+            backend_props=pass_manager_config.backend_properties,
+            unitary_synthesis_method=pass_manager_config.unitary_synthesis_method,
+            unitary_synthesis_plugin_config=pass_manager_config.unitary_synthesis_plugin_config,
+            hls_config=pass_manager_config.hls_config,
+        )
```
alexanderivrii, line 126: We have recently added a new argument ``qubits_initially_zero`` to ``generate_pass_manager``, indicating whether circuit starts in the ``|0>``-state. You may want to propagate its value along with the other arguments. (BTW, I am not sure that I have indented the additional line correctly).

```suggestion
            qubits_initially_zero=pass_manager_config.qubits_initially_zero,
        )
```

qiskit_ibm_runtime/transpiler/plugin.py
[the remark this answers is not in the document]
kt474: This code for the `pass_manager` was copied over from qiskit-ibm-provider over a year ago as a part of the migration and looks like it hasn't been touched. The current behavior in this PR makes sense to me, I can work on possibly consolidating/handling the differences in a separate PR.

yaelbh:
The hook is not working for me also with an `IBMBackend`:
```python

backend = service.backend("ibm_fez", use_fractional_gates=True)

p = Parameter("p")

circ = QuantumCircuit(2)
circ.rzz(7, 0, 1)


pm = generate_preset_pass_manager(optimization_level=0, target=backend.target)
isa_circ = pm.run(circ)

print(isa_circ)
```
This results with `rzz(7)`, in spite of using the code of this PR.

mtreinish:
If you're using the hook points for a backend alternate default stage those only work if you specify the backend to the pass manager constructor. They don't exist on the target.

yaelbh:
Thanks to an advice from @alexanderivrii, I was able to write a test that I'm OK with, in bae92098.

qiskit_ibm_runtime/transpiler/plugin.py
yaelbh: Thanks. I prefer not in this PR (out of scope), so I've opened #2068.

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
yaelbh: Some of the gates listed above don't translate to circuits containing `rzz`. Although they are items in the equivalence library translating them to circuits containing `rzz`, there are also items translating them to other things, and the latter items are sometimes those that apply. As long as we cannot guarantee that the transpiled circuit contains `rzz`, I prefer not to add this kind of test.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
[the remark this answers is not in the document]
yaelbh: @wshanks Does it make sense?

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
yaelbh: Done in 4b08e7e0

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
yaelbh: @wshanks Do you agree to remove the test?

yaelbh:
This would work too: https://github.com/Qiskit/qiskit-ibm-runtime/blob/681aeb990f80f24c13e80bc0948dcd9b9e673959/qiskit_ibm_runtime/transpiler/passes/scheduling/__init__.py#L156-L157

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
wshanks: The parameterized test checks that individual rzz gates are handled correctly. This test checks that many together are handled correctly and produce the right result, but I agree that that doesn't test anything new for the transpiler plugin. It just tests the Qiskit transpiler.

yaelbh: I think it does check the new pass, for the values in `angles`, because this is the only pass in the pass manager. I just think this is already tested in the previous test.

release-notes/unreleased/2043.feat.rst
```diff
@@ -0,0 +1,9 @@
+Added a new transpiler translation plugin :class:`IBMFractionalTranslationPlugin` 
+and a pass :class:`FoldRzzAngle`.
+This plugin is automatically applied for backends
+retrieved with the `use_fractional_gates` optin,
+and the folding pass is added when the backend target includes the `RZZ` gate.
+
+The new pass modifies the input quantum circuit, so that all `RZZ` gates in the
+circuit have an angle parameter within [0, pi/2] which is supported 
+by IBM Quantum processors.
```
wshanks, line 9:
```suggestion
Added a new transpiler translation plugin :class:`~.IBMFractionalTranslationPlugin` 
and a pass :class:`~.FoldRzzAngle`.
This plugin is automatically applied for backends
retrieved with the ``use_fractional_gates`` opt-in,
and the folding pass is added when the backend target includes the ``RZZ`` gate.

The new pass modifies the input quantum circuit, so that all ``RZZ`` gates in the
circuit have an angle parameter within [0, pi/2] which is supported 
by IBM Quantum processors.
```

wshanks (approved):
Here are some minor formatting suggestions for the release note. Otherwise, I think this is ready to merge. If merging the release note suggestions dismisses my approval, I think you can still approve and merge @yaelbh since you did not open the PR.

yaelbh (approved):

yaelbh:
Following https://github.com/Qiskit/qiskit-ibm-runtime/pull/2043#issuecomment-2476345987:
@kaelynj Since the PR is merged, it's time to think of updating the doc. I'm not sure exactly how to do it.
The story is that rzz gates in ISA circuits must satisfy 0 <= theta <= 2pi, as written in the doc.
Transpiling a circuit with an IBM backend will automatically generate a circuit that satisfies this request.

yaelbh:
Also I guess that the doc update should take place at the time of the next release of qiskit-ibm-runtime, and it should be stated that this feature is available only starting from that release.

PR #2044, opened 2024-11-15
Perform fractional gate filtering on raw API dicts

### LINKED ISSUE #2031: Update basis gates according to the fractional gates flag
Currently enabling fractional gates filters adds them to the target (and filters dynamic instructions), but `backend.properties()` and `backend.configuration()` remain unchanged.

Once fixed, we can update the `convert_to_target` accordingly.

wshanks:
This change moves the filtering of fractional gates and dynamic circuit instructions up into the processing of the raw configuration and properties dicts received through the API. Previously, the filtering was performed only during the creation of the `Target` for the backend. So previously an `IBMBackend` with fractional gates would have both fractional gates and dynamic circuit instructions in its configuration and properties data regardless of what value was used for `use_fractional_gates`. This situation led to inconsistent behavior when for example users tried to use `backend.basis_gates` as the reference for which gates were available instead of
`backend.target.operation_names`.

The previous changes to filter instructions during target creation are removed here, but the `use_fractional_gates` backend option is retained because `IBMBackend` loads the properties data lazily and can reload the configuration data after creation with the `refresh()` method.

Closes #2031


qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -349,11 +320,3 @@ def qubit_props_list_from_props(
...
         )
     return qubit_props
-
-
-def is_fractional_gate(gate: Gate) -> bool:
-    """Test if gate is fractional gate familiy."""
-    # In IBM architecture these gates are virtual-Z and delay,
-    # which don't change control parameter with its gate parameter.
-    exclude_list = (RZGate, PhaseGate, U1Gate, Delay)
-    return len(gate.params) > 0 and not isinstance(gate, exclude_list)
```
wshanks, line 359: Since the code here is IBM-specific, I went the simpler route of just hardcoding which gates count as fractional. We could discuss that if necessary. At the point where the filtering is done now, the gates are just strings, though they could be mapped to Qiskit gates.

qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -48,9 +42,6 @@ def convert_to_target(
     configuration: BackendConfiguration,
     properties: BackendProperties = None,
     defaults: PulseDefaults = None,
-    *,
-    include_control_flow: bool = True,
```
nkanazawa1989, line 52: Please make sure this function is not used in the internal software stack.

nkanazawa1989 (dismissed):
Overall the change looks good. I have a couple of concerns as written in inline comments.

qiskit_ibm_runtime/utils/backend_decoder.py
```diff
@@ -66,6 +74,43 @@ def configuration_from_server_data(
     return None
 
 
+def filter_raw_configuration(
+    raw_config: dict, use_fractional_gates: Optional[bool] = False
```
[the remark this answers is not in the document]
wshanks: I am not sure which interface you are saying you prefer to keep. `QiskitRuntimeService.backend` (and `QiskitRuntimeService.backends`) only accepted `use_fractional_gates` before and they still have three behaviors depending on whether `use_fractional_gates` is set to `True`, `False`, or `None`. I had been thinking that was the only public interface related `use_fractional_gates` (besides the `use_fractional_gates` backend option, which I think can still be modified as long as you use `IBMBackend.refresh()` afterward).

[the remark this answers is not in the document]
wshanks: I think you are referring to the two flags of `convert_to_target`. I don't think we care about removing both fractional gates and dynamic circuit instructions from the backend, so that leaves only the three options covered by the `True`, `False`, and `None` values of `use_fractional_gates`. Those options are a little weird versus explicit `include_fractional_gates` and `include_control_flow` options but that is the way it already works for `QiskitRuntimeService.backend()`.

If your concern here is just related to the change to `convert_to_target()`, perhaps we should move the discussion to your other comment there.

qiskit_ibm_runtime/utils/backend_converter.py
wshanks: Hmm, unfortunately, I found that, although this is an undocumented utility function, it is used a few times in the internal software stack. Perhaps it should be documented as a public function.

Luckily, no internal usage of this function references `include_control_flow` or `include_fractional_gates` and the removal of the options here preserves the default behavior so there should be not disruption to the internal usage.

My reasoning was that the filtering of the raw API data was sufficient. There was no need to filter a second time during target creation. Since `convert_to_target` was not documented, I thought I was free to remove the extra options and filtering. I don't feel too strongly though. I can put them back if that seems best.

nkanazawa1989: I think you can at least add warning for usage of these options. But I'm also fine removing them as long as these options are not found by search (since key is required).

qiskit_ibm_runtime/utils/backend_decoder.py
```diff
@@ -51,6 +58,7 @@ def configuration_from_server_data(
         return None
     try:
         decode_backend_configuration(raw_config)
+        filter_raw_configuration(raw_config, use_fractional_gates=use_fractional_gates)
```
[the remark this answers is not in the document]
nkanazawa1989: If it's loaded lazily another option would be adding lazy filtering by consuming the `backend.options.use_fractional_gates` in [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/da0a2d763d5592e8e3e472ecfcc17d7915b065b7/qiskit_ibm_runtime/ibm_backend.py#L271).

tox.ini
```diff
@@ -18,7 +18,7 @@ deps =
     -r{toxinidir}/requirements-dev.txt
     -r{toxinidir}/requirements-visual.txt
 commands =
-  python -m unittest -v
+  python -m unittest discover --verbose --top-level-directory . --start-directory test/unit -p 'test_backend_retrieval.py'
```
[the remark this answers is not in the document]
wshanks: Fixed in 6d46fb82e742dec1105c4c6beb60c806e5053817

qiskit_ibm_runtime/utils/backend_decoder.py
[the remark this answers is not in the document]
nkanazawa1989: Yes, the two flags are what I'm referring to. So you suggest an implicit mapping of

- `include_control_flow=True, include_fractional_gates =True` := `use_fractional_gates =None`
- `include_control_flow=True, include_fractional_gates =False` := `use_fractional_gates =True`
- `include_control_flow=False, include_fractional_gates =True` := `use_fractional_gates =False`

Maybe this is okey but the behavior with `use_fractional_gates =None` is not straightforward, not documented and not tested. So there is no guarantee this behavior is preserved for future release. At least test and document should be added.

Another concern is potential future upgrade of our instruction set. If we need to deal with third kind of instruction group, single flag is not enough and we would need breaking API change. Maybe this is over anxious though.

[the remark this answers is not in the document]
nkanazawa1989: Maybe new implementation is okey with update of documentation.

qiskit_ibm_runtime/utils/backend_decoder.py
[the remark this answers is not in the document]
wshanks: Right, I had meant to do that but I left out the check of `backend.options.use_fractional_gates` and also got the logic of the filtering backwards. I fixed it in 598249e41679c15029e0aa19071b94f04a96e61f. Also, I added some testing in 7cdad8a96141f898ec804b7774159d4afbc745cf that catches the error I made -- it will fail the tests if the properties do not get loaded into the target.

qiskit_ibm_runtime/utils/backend_decoder.py
```diff
@@ -28,16 +28,23 @@
 
 logger = logging.getLogger(__name__)
 
+FRACTIONAL_GATES = ("rzz", "rx")
+DYNAMIC_INSTRUCTIONS = ("if_else", "while_loop", "for_loop", "switch_case")
```
yaelbh, line 32: Is it possible to check whether a gate is an instance of `ControlFlowOp`? I find it more robust than hard-coding instruction names.

qiskit_ibm_runtime/utils/backend_decoder.py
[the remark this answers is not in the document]
wshanks: Yes, the `use_fractional_gates=None` behavior was already present before this PR. It was added in https://github.com/Qiskit/qiskit-ibm-runtime/pull/1938. It is documented in the `IBMBackend.backend()` documentation and it is tested, so I think we are okay here. In any case, I restored the options to `convert_to_target`.

qiskit_ibm_runtime/utils/backend_converter.py
wshanks: I decided just to restore the options to `convert_to_target`. Since `IBMBackend` filters the API data, I did not restore passing the options to `converet_to_target` from it. Passing the options would only matter if `backend.options.use_fractional_gates` were modified but with the caching of the configuration/properties/target the results would be inconsistent unless `refresh()` were used but then the target would be consistent again without passing the options.

qiskit_ibm_runtime/utils/backend_decoder.py
wshanks: No, the mapping of gates to names is always done through hard-coding. For example [here](https://github.com/Qiskit/qiskit/blob/b9d5c9c6aeb4568b6e9fba8943517ebed300886d/qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py#L459-L464) in Qiskit SDK or in this section related to the filtering in `convert_to_target` in this repo:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/da0a2d763d5592e8e3e472ecfcc17d7915b065b7/qiskit_ibm_runtime/utils/backend_converter.py#L79-L84

In Qiskit, there is a public function called `get_standard_gate_name_mapping` that gives a map from string names to quantum gates. It is still effectively [hard-coded](https://github.com/Qiskit/qiskit/blob/b9d5c9c6aeb4568b6e9fba8943517ebed300886d/qiskit/circuit/library/standard_gates/__init__.py#L85) but  at least you can reference it and count on it to be updated independently in the future. Perhaps something like that could be added for these other instructions as well.

We could stop hard-coding `FRACTIONAL_GATES` since I restored `is_fractional_gate`. I didn't mind the hard-coding of rx and rzz here though because `is_fractional_gate` also somewhat arbitrarily hard-codes exceptions to fractional gates and I don't know what the chances are that another gate will need to be supported but it only requires a small change:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/da0a2d763d5592e8e3e472ecfcc17d7915b065b7/qiskit_ibm_runtime/utils/backend_converter.py#L354-L359

wshanks:
I think this is ready for another look.

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -402,6 +392,14 @@ def properties(
...
             if not api_properties:
                 return None
+            if "gates" in api_properties and isinstance(api_properties["gates"], list):
+                if (
+                    self.options.use_fractional_gates is not None
+                    and not self.options.use_fractional_gates
+                ):
+                    api_properties["gates"] = [
+                        g for g in api_properties["gates"] if g.get("gate") not in FRACTIONAL_GATES
+                    ]
```
yaelbh, line 402: You address properties here, and not inside `properties_from_server_data`. But, for the configuration, you did it inside `configuration_from_server_data`. Is there a reason? The code will be more readable if the structure is uniform.

qiskit_ibm_runtime/fake_provider/backends/fractional/conf_fractional.json
```diff
@@ -1 +1 @@
...
...ive", "type": "drive"}, "d4": {"operates": {"qubits": [4]}, "purpose": "drive", "type": "drive"}, "m0": {"operates": {"qubits": [0]}, "purpose": "measure", "type": "measure"}, "m1": {"operates": {"qubits": [1]}, "purpose": "measure", "type": "measure"}, "m2": {"operates": {"qubits": [2]}, "purpose": "measure", "type": "measure"}, "m3": {"operates": {"qubits": [3]}, "purpose": "measure", "type": "measure"}, "m4": {"operates": {"qubits": [4]}, "purpose": "measure", "type": "measure"}, "u0": {"operates": {"qubits": [0, 1]}, "purpose": "cross-resonance", "type": "control"}, "u1": {"operates": {"qubits": [1, 0]}, "purpose": "cross-resonance", "type": "control"}, "u2": {"operates": {"qubits": [1, 2]}, "purpose": "cross-resonance", "type": "control"}, "u3": {"operates": {"qubits": [1, 3]}, "purpose": "cross-resonance", "type": "control"}, "u4": {"operates": {"qubits": [2, 1]}, "purpose": "cross-resonance", "type": "control"}, "u5": {"operates": {"qubits": [3, 1]}, "purpose": "cross-resonance", "type": "control"}, "u6": {"operates": {"qubits": [3, 4]}, "purpose": "cross-resonance", "type": "control"}, "u7": {"operates": {"qubits": [4, 3]}, "purpose": "cross-resonance", "type": "control"}}}
```
yaelbh, line 1: Replace `rzx` with `rzz` in the properties file.

yaelbh (changes requested):
Edit the comment in https://github.com/Qiskit/qiskit-ibm-runtime/blob/ccd4e9137b7779832dd6826dca00323cc5e990f2/qiskit_ibm_runtime/qiskit_runtime_service.py#L548

Another comment to be edited, although it's not from this PR, but related: https://github.com/Qiskit/qiskit-ibm-runtime/blob/ccd4e9137b7779832dd6826dca00323cc5e990f2/qiskit_ibm_runtime/utils/backend_decoder.py#L40

And, of course, thanks a lot for doing this work :smile:

qiskit_ibm_runtime/utils/backend_decoder.py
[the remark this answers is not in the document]
wshanks: Outdated -- the properties filtering should work now.

qiskit_ibm_runtime/utils/backend_decoder.py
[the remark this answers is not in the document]
wshanks: Yes, I think we are good -- I put back the two options to `convert_to_target` and at least didn't make the situation with the three modes of `use_fractional_gates` worse (though I agree it is not the most clear way of working).

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -550,12 +554,16 @@ def _create_backend_obj(
         self,
         backend_name: str,
         instance: Optional[str] = None,
+        use_fractional_gates: Optional[bool] = False,
```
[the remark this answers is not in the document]
wshanks: I removed the default values (I had to remove the default for `instance` which should be okay since this is a private function).

qiskit_ibm_runtime/utils/backend_decoder.py
```diff
@@ -28,16 +28,23 @@
...
+    use_fractional_gates: Optional[bool] = False,
 ) -> Optional[Union[QasmBackendConfiguration, PulseBackendConfiguration]]:
     """Create an IBMBackend instance from raw server data.
 
     Args:
         raw_config: Raw configuration.
         instance: Service instance.
+        use_fractional_gates: Set True to allow for the backends to include
+            fractional gates. See :meth:`~.QiskitRuntimeService.backends`
+                for further details.
```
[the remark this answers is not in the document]
wshanks: I updated the wording.

qiskit_ibm_runtime/utils/backend_decoder.py
```diff
@@ -66,6 +74,43 @@ def configuration_from_server_data(
...
+            ]
+        if "gates" in raw_config:
+            raw_config["gates"] = [
+                g for g in raw_config["gates"] if g.get("name") not in FRACTIONAL_GATES
+            ]
+        if "supported_instructions" in raw_config:
+            raw_config["supported_instructions"] = [
+                i for i in raw_config["supported_instructions"] if i not in FRACTIONAL_GATES
+            ]
+
```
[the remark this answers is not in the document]
wshanks: I made this change, but for a couple backends I checked it seems like `conditional` is always `False` any way. I am confused how this flag is used currently.

qiskit_ibm_runtime/ibm_backend.py
wshanks: Good point. I shifted the filtering into `properties_from_server_data` for consistency.

qiskit_ibm_runtime/fake_provider/backends/fractional/conf_fractional.json
wshanks: Fixed!

qiskit_ibm_runtime/utils/backend_decoder.py
wshanks: I made all the code use the `is_fractional_gate` function. I didn't unify the cases of dynamic circuit instructions but I did discover that Qiskit 1.3 at least provides a variable with all the string names of control flow instructions so I used that for the filter (when it could be imported).

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -458,16 +458,16 @@ def backends(
                         filters=lambda x: ("rz" in x.basis_gates )
                     )
             use_fractional_gates: Set True to allow for the backends to include
-                fractional gates in target. Currently this feature cannot be used
-                simulataneously with dynamic circuits, PEC, PEA, or gate twirling.
-                When this flag is set, control flow instructions are automatically
-                removed from the backend target.
-                When you use the dynamic circuits feature (e.g. if_else) in your
+                fractional gates. Currently this feature cannot be used
+                simulataneously with dynamic circuits, PEC, PEA, or gate
```
yaelbh, line 462:
```suggestion
                simultaneously with dynamic circuits, PEC, PEA, or gate
```

qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -26,19 +26,16 @@
...
 )
 from qiskit.circuit.gate import Gate
-from qiskit.circuit.library.standard_gates import (
-    get_standard_gate_name_mapping,
-    RZGate,
-    U1Gate,
-    PhaseGate,
-)
-from qiskit.circuit.delay import Delay
+from qiskit.circuit.library.standard_gates import get_standard_gate_name_mapping
```
yaelbh, line 29: Where is it used in this file? I expect the linter will ask too.

qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -26,19 +26,16 @@
...
 from qiskit.circuit.parameter import Parameter
 from qiskit.providers.backend import QubitProperties
 from qiskit.providers.exceptions import BackendPropertyError
 from qiskit.transpiler.target import InstructionProperties, Target
 
 from ..models import BackendConfiguration, BackendProperties, PulseDefaults
+# is_fractional_gate used to be defined in this module and might be referenced
+# from here externally
+from .utils import is_fractional_gate  # See comment above before removing
```
yaelbh, line 38: I guess that you'll have to disable the linter's unused import error.

yaelbh (commented):
Added a few small comments, although you didn't say that it's ready for another review

qiskit_ibm_runtime/utils/backend_converter.py
wshanks: This one is used in `convert_to_target` below. This line changed because I removed the use of the gates (in `is_fractional_gate`) but not this function so I wrapped it back to a single line.

qiskit_ibm_runtime/utils/backend_converter.py
wshanks: The function is still used in `convert_to_target` below. It was defined at the bottom of this file, but I wanted to use it in backend_decoder.py, so I moved it to a more standard location in utils.py alongside similar functions like `are_circuits_dynamic` and `is_isa_circuit`. There had been mention about trying to centralize on `is_fractional_gate` as the central function to use for testing if a gate is fractional so I was worried other code could already be using this location to find the function.

wshanks (commented):
Thanks, @yaelbh. I did mean to ask for another review after my last changes. I think it is ready again.

I think I have addressed all the comments, but I didn't mark anything as resolved. Feel free to mark any conversations as resolved if you want to clear up what is still outstanding.

yaelbh (approved):

PR #2046, opened 2024-11-19
Add checks for parameter expressions with gen3-turbo runtime.

blakejohnson:
Adds client-side checks to the Sampler and Estimator to ensure that circuits with `ParameterExpression`s are not used in combination with `"execution_path": "gen3-turbo"`.


blakejohnson:
The linting failures appear spurious to me. I will need to look into how to get pylint to accept these lines.

ihincks (approved):
These changes and tests look correct to me, thanks @blakejohnson .

At some point we'll need to profile all of these isinstance checks that are piling up through validation...out of scope.

Issue #1877, opened 2024-08-19
Weird NoiseLearner.run parameter type in docs


jyu00:
**Describe the bug**

The `NoiseLearner.run` [API page](https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.noise_learner.NoiseLearner) describes the `circuits` input type as 

<img width="811" alt="image" src="https://github.com/user-attachments/assets/f3cda75a-1dee-4999-9538-59c6ee91f9c8">

That's not exactly digestible.

**Steps to reproduce**

**Expected behavior**

**Suggested solutions**

**Additional Information**

@Eric-Arellano Maybe this is trying to reference `EstimatorPubLike` in a weird way? Estimator's API doc doesn't have this issue though. 

cc @SamFerracin 

- **qiskit-ibm-runtime version**:
- **Python version**:
- **Operating system**:

Eric-Arellano:
> Maybe this is trying to reference EstimatorPubLike in a weird way? Estimator's API doc doesn't have this issue though.

Yes, that is what's going on. I'm not sure why Sphinx is expanding inlining `EstimatorPubLike` in https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.noise_learner.NoiseLearner#run but not https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.EstimatorV2#run 🤔

Eric-Arellano:
I have a fix for this in the works 🚀 Working on this discovered another issue that I'm working on and blocks this. I hope to have it up tomorrow.

PR #2052, opened 2024-11-21
Fix type hints in docs for PUBs and improve Sphinx config

Eric-Arellano:
Closes https://github.com/Qiskit/qiskit-ibm-runtime/issues/1877. We now use `EstimatorPubLike` and `SamplerPubLike` for the type hints, rather than Sphinx fully expanding the types into extremely long values shown in #1877.

To fix the type hints, we have to stop using `sphinx-autodoc-typehints` and instead use autodoc's builtin support for type hints. I didn't realize we were still using `sphinx-autodoc-typehints` - the other Qiskit projects migrated off a while ago.

This PR also makes some other improvements to align with changes we made in Qiskit Addons & qiskit-ibm-transpiler:

* always show inheritance in class pages for the base class
* index page should show the PyPI package name in code syntax
* don't show signatures in `autosummary` tables, since they're noisy
* show type hints for parameters that don't have docstring

## Introduces a regression: type hints for properties

Unfortunately, switching off of `sphinx-autodoc-typehints` results in https://github.com/Qiskit/documentation/issues/2346. This is a quirk of autosummary that we cannot easily fix. We have a plan to fix it, although realistically won't get to it until Q1 and possibly Q2.

In the meantime, consider repeating some of the type information in the docstring. For example, in

```
<Attribute id="qiskit_ibm_runtime.EstimatorV2.mode">
    Return the execution mode used by this primitive.
```

You could rewrite to say

```
<Attribute id="qiskit_ibm_runtime.EstimatorV2.mode">
    Return the execution mode used by this primitive, either :class:~`.Session` or ~:class:~`.Batch`.
```


Eric-Arellano:
You can see a preview of the diff caused by this PR at https://github.com/Qiskit/documentation/pull/2343/files

kt474 (approved):

PR #2057, opened 2024-11-25
Local mode job class

### LINKED ISSUE #1963: Make local jobs more like remote jobs

The local testing mode is design to be as closely as possible to remote execution. #1605 made local (fake) backends more similar to `IBMBackend`. We should also make jobs returned in local mode more similar to `RuntimeJobV2`. One example of such discrepancy is #1939. 

**Acceptance criteria**

Job returned in local testing mode should have the same methods as `RuntimeJobV2`, including (but not limited to) `backend()`, `metrics()`, `usage()`, etc

kt474:
### Details and comments
Fixes #1963


ptristan3 (approved):
LGTM

PR #2059, opened 2024-11-27
Replace `jupyter-execute` with `plot` from docs

arnaucasau:
This PR replaces all the `.. jupyter-execute::` calls with `.. plot::`. The change will improve the docs in different ways. For example, it will allow us to leverage the `:alt:` option of the `plot` directive to set an alt text to all the images, which wasn't possible using the `jupyter-execute` directive, and, in addition, it will prevent us from different issues when building the documentation caused by the unreliable nature of `juptyer-execute` as explained in https://github.com/Qiskit/qiskit/pull/9346

In this PR in `qiskit/documentation` we can see the git diff caused by this replacement. The change is mainly the renaming of the images because some of them now have a dedicated code block: https://github.com/Qiskit/documentation/pull/2403/files

### DIFF
--- a/qiskit_ibm_runtime/fake_provider/__init__.py
+++ b/qiskit_ibm_runtime/fake_provider/__init__.py
@@ -33,6 +33,7 @@ Here is an example of using a fake backend for transpilation and simulation.
 
 .. plot::
    :include-source:
+   :context: close-figs
 
    from qiskit import QuantumCircuit
    from qiskit import transpile
@@ -51,10 +52,18 @@ Here is an example of using a fake backend for transpilation and simulation.
    circuit.measure_all()
    circuit.draw('mpl', style="iqp")
 
+.. plot::
+   :include-source:
+   :context: close-figs
+
    # Transpile the ideal circuit to a circuit that can be directly executed by the backend
    transpiled_circuit = transpile(circuit, backend)
    transpiled_circuit.draw('mpl', style="iqp")
 
+.. plot::
+   :include-source:
+   :context: close-figs
+   
    # Run the transpiled circuit using the simulated fake backend
    sampler = SamplerV2(backend)
    job = sampler.run([transpiled_circuit])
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/fake_provider/__init__.py
```diff
@@ -51,10 +52,18 @@
    circuit.measure_all()
    circuit.draw('mpl', style="iqp")
 
+.. plot::
+   :include-source:
+   :context: close-figs
```
arnaucasau, line 57: This is necessary to allow the code to run in the same context as all previous plot directives. Otherwise, we would need to add all the imports and the relevant code again.

By adding `close-figs` to the directive we keep the context but close previous figures before running the code. Without this part, we would see repeated images all over the page.

For more information, see https://matplotlib.org/stable/api/sphinxext_plot_directive_api.html#options

qiskit_ibm_runtime/transpiler/passes/scheduling/__init__.py
```diff
@@ -332,16 +360,9 @@
...
-.. jupyter-execute::
-
-    dd_sequence = [XGate(), XGate()]
-
-    pm = PassManager(
-        [
-            ALAPScheduleAnalysis(durations),
-            PadDynamicalDecoupling(durations, dd_sequence),
-        ]
-    )
```
arnaucasau, line 344: I divided the code into two blocks to display both circuits. Before, even though we ran `.draw()` twice, we were only showing the last circuit as we can see at the end of this page: https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.transpiler.passes.scheduling

qiskit_ibm_runtime/fake_provider/__init__.py
```diff
@@ -33,6 +33,7 @@
 
 .. plot::
```
arnaucasau, line 34: This plot showed three images and I splitting it into three `.. plot::` to be able to set an alt text to each image individually.

Eric-Arellano (approved):
Nice work! Please wait for a Runtime team member to approve before merging

PR #2060, opened 2024-11-27
Add a linter to check the images' alt text

arnaucasau:
In order to meet the [IBM Accessibility Requirements](https://www.ibm.com/able/requirements/requirements) for the API docs, we need to ensure that all images have appropriate alt text so that users can understand any meaningful visuals. The alt text should serve the same purpose as the images on the page, conveying the same meaning rather than describing visual characteristics. In addition, if an image contains important words for understanding the content, the alt text should include those words as well.

This PR adds a linter to check that all images have alt text. The current images have been included in an allowlist that will help us to not break CI before the text is added.

If the script detects an image without alt text, it will show an error indicating the file where the image is located and the line where we are using a sphinx directive like `.. image::` or `.. plot::` that generates an image. Example:

```
💔 Some images are missing the alt text

Errors found in qiskit_ibm_runtime/fake_provider/__init__.py:
- Error in line 34: .. plot::

Alt text is crucial for making documentation accessible to all users. It should serve the same purpose as the images on the page, conveying the same meaning rather than describing visual characteristics. When an image contains words that are important to understanding the content, the alt text should include those words as well.
```

To fix the error, the `:alt: [Your text here]` option must be added under the lines indicated in the error messages. However, If the `.. plot::` directive is used, but no image is generated, you can use the option `:nofigs:`.


tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+# that they have been altered from the originals.
+
+"""Utility script to verify that all images have alt text"""
+
+import multiprocessing
+import sys
+import glob
+
+# Dictionary to allowlist lines of code that the checker will not error
+# Format: {"file_path": [list_of_line_numbers]}
```
arnaucasau, line 21: I wasn't sure about what was the best granularity for the allowlist, but I ended up choosing this format to allow the content team to work one image at a time, if needed, inside a single file. There a files with a lot of images that maybe can't be fixed at once. However, I think making the allowlist to work with files directly is a good approach too.

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2020
```
Eric-Arellano, line 4:
```suggestion
# (C) Copyright IBM 2024
```

tools/verify_images.py
Eric-Arellano: I think this makes a lot of sense to do line numbers because each individual image may take some time. I like it.

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+        328,
+        342,
+        363,
+        382,
+        406,
+    ],
+}
+
+
+def is_image(line) -> bool:
```
Eric-Arellano, line 47: Can you please add type hints for every parameter in this file? Even though we don't run MyPy, it's useful documentation

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+    return line.strip().startswith(".. image:") or line.strip().startswith(".. plot:")
```
Eric-Arellano, line 48: [Python](https://docs.python.org/3/library/stdtypes.html#str.startswith) lets you pass a tuple of prefixes

```suggestion
    return line.strip().startswith((".. image:", ".. plot:"))
```

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def allowlist(filename, line_num) -> bool:
+    return (
+        filename in ALLOWLIST_MISSING_ALT_TEXT and line_num in ALLOWLIST_MISSING_ALT_TEXT[filename]
+    )
```
Eric-Arellano, line 54:
```suggestion
    return line_num in ALLOWLIST_MISSING_ALT_TEXT.get(filename, [])
```

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+        406,
+    ],
+}
+
+
+def is_image(line) -> bool:
+    return line.strip().startswith(".. image:") or line.strip().startswith(".. plot:")
+
+
+def allowlist(filename, line_num) -> bool:
```
Eric-Arellano, line 51: Maybe `in_allowlist`?

```python
if in_allowlist(my_file, 24)
```

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+        filename in ALLOWLIST_MISSING_ALT_TEXT and line_num in ALLOWLIST_MISSING_ALT_TEXT[filename]
+    )
+
+
+def validate_image(file_path) -> tuple[str, list[str]]:
+    """Validate all the images of a single file"""
+    invalid_images: list[str] = []
+
+    with open(file_path, encoding="utf8") as fd:
+        lines = fd.readlines()
```
Eric-Arellano, line 62: More direct is `Path(file_path).read_text().splitlines()`. You avoid the whole context manager and close out the file sooner.

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for index, line in enumerate(lines):
+            if is_image(line) and not allowlist(file_path, index + 1):
```
Eric-Arellano, line 65: You can reduce nesting in this for loop by inverting the conditional and using early returns.

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def validate_image(file_path) -> tuple[str, list[str]]:
+    """Validate all the images of a single file"""
+    invalid_images: list[str] = []
+
+    with open(file_path, encoding="utf8") as fd:
+        lines = fd.readlines()
+
+        for index, line in enumerate(lines):
```
Eric-Arellano, line 64: Maybe call it `line_number` rather than `index`?

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+            if is_image(line) and not allowlist(file_path, index + 1):
+                options: list[str] = []
+                option_idx = index
```
Eric-Arellano, line 67: Maybe `options_line_number`?

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+                    if not option_line.startswith(":"):
+                        break
+
+                    options.append(option_line)
+
+                alt_exists = any(option.startswith(":alt:") for option in options)
+                nofig_exists = any(option.startswith(":nofig:") for option in options)
+
+                # Only `.. plot::`` directives without the `:nofig:` option are required to have alt text.
+                # All `.. image::` directives need alt text and they don't have a `:nofig:` option.
```
Eric-Arellano, line 82: Clarity nit

```suggestion
                # Meanwhile, all `.. image::` directives need alt text and they don't have a `:nofig:` option.
```

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+    return (file_path, invalid_images)
+
+
+def main() -> None:
+    files = glob.glob("qiskit_ibm_runtime/**/*.py", recursive=True)
+
+    with multiprocessing.Pool() as pool:
+        results = pool.map(validate_image, files)
+
+    failed_files = list(filter(lambda x: len(x[1]), results))
```
Eric-Arellano, line 95: This is fine, but more idiomatic in Python is to use a list comprehension rather than `map` and `filter`.

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+
+    if len(failed_files):
```
Eric-Arellano, line 97: Could invert conditional and early return

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+        sys.stderr.write("💔 Some images are missing the alt text\n\n")
```
Eric-Arellano, line 98: You can print to stderr with `print("", file=sys.stderr)`. It will add the last `\n` for you.

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for failed_file in failed_files:
```
Eric-Arellano, line 100: It'll be easier to read if you unpack the tuple directly in the for loop, like `for a, b in my_list`

tools/verify_images.py
```diff
@@ -0,0 +1,117 @@
...
+            sys.stderr.write("Errors found in %s:\n" % failed_file[0])
```
Eric-Arellano, line 101: `%s` is a very old style of interpolation. More modern is an f-string.

Eric-Arellano (commented):
Nice work! None of my concerns are blocking if someone else approves this while I'm OOO

tools/verify_images.py
```diff
@@ -0,0 +1,118 @@
...
+    return line_num in ALLOWLIST_MISSING_ALT_TEXT.get(filename, [])
+
+
+def validate_image(file_path: str) -> tuple[str, list[str]]:
+    """Validate all the images of a single file"""
+    invalid_images: list[str] = []
+
+    lines = Path(file_path).read_text().splitlines()
+
+    for line_number, line in enumerate(lines):
```
Eric-Arellano, line 62: Oh actually calling it `line_number` is inaccurate because of 0-based vs 1-based indexing. So maybe `line_index` and `options_line_index` is more accurate

tools/verify_images.py
```diff
@@ -0,0 +1,118 @@
...
+        if not is_image(line) or in_allowlist(file_path, line_number + 1):
+            continue
+
+        options: list[str] = []
+        options_line_number = line_number
```
Eric-Arellano, line 67: It might be possible to get rid of this and simply use `line_index`? Right now, you end up iterating over the same line twice because `line_index` is no being incremented after identifying that these option lines are not images. You'd replace the `for` loop with a `while` loop sith the same condition as your inner loop on 69.

Imagine this:

```
[line 1] .. image::
[line 2]
[line 3] .. image::
```

The options code would increment `line_index` to 2 and then break out of the inner loop because there were no options.

I think the only edge case to be careful about is when you have a new image as the next line. Imagine this:

```
[line 1] .. image::
[line 2] .. image::
```

Your options code stops at line 2 because there were no options for the first image. You can break out of the inner loop now. Things will work properly as long as the outer loop doesn't increment `line_index` when entering the next iteration: you need to start the iteration at line 2 and not try going to line 3.

arnaucasau: That's an interesting suggestion, thanks! I think we can't remove the `options_line_number` because we need a temporary variable to store the line of the image for the error message. However, we are indeed iterating twice in some lines. I chose that approach because I thought it was more readable and had in mind that the number of options of a directive (:= lines iterated twice) is very small, at most 4/5 I would say.

Regardless, I think it's a good idea to try to iterate only once. If I understand correctly what you are suggesting, I think it would be something like this:

```python
    line_index = 0
    while line_index + 1 < len(lines):
        line = lines[line_index].strip()
        if not is_image(line) or in_allowlist(file_path, line_index + 1):
            line_index += 1
            continue

        options: list[str] = []
        line_index_tmp = line_index

        while line_index + 1 < len(lines):
            line_index += 1
            option_line = lines[line_index].strip()

            if not option_line.startswith(":"):
                break

            options.append(option_line)

         `Check if the image is valid or not...`
```

I think we could iterate once and at the same time remove the inner loop by doing something like this, defining some helper functions:

```python
    line_index = 0
    image_found = False
    image_line = -1
    options: list[str] = []

    while line_index < len(lines):
        line = lines[line_index].strip()

        if image_found and not is_option(line) and not is_valid_image(options):
            invalid_images.append(f"- Error in line {image_line}: {lines[image_line-1].strip()}")
            image_found = False
            options = []
            continue

        if image_found and is_option(line):
            options.append(line)
            line_index += 1
            continue

        image_found = is_image(line) and not in_allowlist(file_path, line_index + 1)
        image_line = line_index + 1
        line_index += 1
```

Do you think either of these approaches is preferable over the current one taking into account performance/readability?

arnaucasau: Thank you for the suggestion! I have been looking at the code, and I think we could even remove the inner loop and still iterate only once per line. For each line, we only need two special cases. If we are looking at an image option we just need to save it, and if there are no more options after seeing an image we create an error message.

```python
    image_found = False
    options: list[str] = []

    for line_index, line in enumerate(lines):
        if image_found and is_option(line):
            options.append(line)
            continue

        if image_found and not is_valid_image(options):
            image_line = line_index - len(options)
            invalid_images.append(f"- Error in line {image_line}: {lines[image_line-1].strip()}")

        image_found = is_image(line)
        options = []
```

kt474 (approved):

tools/verify_images.py
```diff
@@ -0,0 +1,103 @@
...
+    return (file_path, invalid_images)
+
+
+def main() -> None:
+    files = glob.glob("qiskit_ibm_runtime/**/*.py", recursive=True)
+
+    with multiprocessing.Pool() as pool:
+        results = pool.map(validate_image, files)
+
+    failed_files = [x for x in results if len(x[1])]
```
Eric-Arellano, line 80: Python treats `[]` as falsey, unlike JavaScript, so no need for `len()`.

But also this code might be more clear as a dictionary

```suggestion
    failed_files = {file: image_errors for file, image_errors in results if image_errors}
```

You'd need to update line 88 to be `failed_files.items()`.

tools/verify_images.py
```diff
@@ -0,0 +1,103 @@
...
+    options: list[str] = []
+
+    for line_index, line in enumerate(lines):
+        if image_found and is_option(line):
+            options.append(line)
+            continue
+
+        if image_found and not is_valid_image(options):
+            image_line = line_index - len(options)
+            invalid_images.append(f"- Error in line {image_line}: {lines[image_line-1].strip()}")
```
Eric-Arellano, line 66: Normally, I'm not a big fan of nesting. Here, I think it might give more clarity by grouping related control flow

```suggestion
        if image_found:
            if is_option(line):
                options.append(line)
                continue

            # Else, the prior image_found has no more options so we should determine if it was valid.
            #
            # Note that, either way, we do not early exit out of the loop iteration because this `line`
            # might be the start of a new image.
            if not is_valid_image(options):
                image_line = line_index - len(options)
                invalid_images.append(f"- Error in line {image_line}: {lines[image_line-1].strip()}")
```

Eric-Arellano (approved):
Nice improvements!

Eric-Arellano (approved):

PR #2066, opened 2024-12-02
Add interactive CLI to save user account

### LINKED ISSUE #2054: Consider CLI for saving account
## Context

We have a problem with users storing their account tokens in their source code or Jupyter notebook. We're currently writing [a guide](https://github.com/Qiskit/documentation/pull/2322) explaining that you shouldn't do this, but we could probably avoid the situation without additional documentation if we alter the user experience.

A pretty common approach to save user accounts is a CLI. Benefits to user are:
* Less documentation reading needed; they just run a command that guides them through setting up their account
* No temptation to leave token in source code or Jupyter notebook
* Token isn't saved to REPL history

## What is the expected feature or enhancement?

A CLI to save your account, made available on install.

Workflow would be along these lines:

1. User runs command such as:
   ```
   $ qiskit-runtime-save-account
   ```
2. CLI asks whether to use platform or cloud
2. CLI prompts for token
4. CLI prompts again for group and hub. If you wanted to get fancy you could work out which ones they have access to and provide them as options, but just entering the string is fine.
5. Finally, we write this to file and explain it to the user.

The process could look something like:

 ```
 $ qiskit-runtime-save-account
 Would you like to use (1) ibm_quantum or (2) ibm_cloud?
 Type the number and press enter (default ibm_quantum) 1
 Using ibm_quantum
 Please enter your IBM Quantum token, you can get this from https://quantum.ibm.com/
 Please enter the instance you'd like to use (default ibm-q/open/main)
 Account saved to ~/.qiskit/qiskit-ibm.json
 ⚠️ Note your token is saved to disk in plain text. If on a shared computer, make sure to regenerate your token when you're finished.
 ```

This CLI can easily be created using built-in Python functionality.

* [Make CLI available on installing `qiskit-ibm-runtime`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#creating-executable-scripts)
* Use [getpass](https://docs.python.org/3/library/getpass.html) to securely enter the token.

The original `.save_account()` method can still be public, we'd just recommend the CLI as a better user experience.

## Acceptance criteria

Super simple CLI that guides a user through getting account information and saves it to disk. CLI should explain where the information is saved and that it's in plain text.

Nice to have:
* Select from available options (e.g. we could show the user their available instances and have them enter the list number of the one they want)
* Choose storing to environment variables rather than disk (if they're using a shared computer and want variables to disappear after their session).

frankharkins:
- **CLI: First pass**
- **Make types compatible with Python3.8**
- **Add tests**
- **Make CLI module private**
- **Add --help**
- **Add `save-account` as subcommand**
- **Neaten, lint, and format**

### Summary

Closes #2054.

This PR adds an interactive command-line interface (CLI) that guides users through saving their account to disk. This should be easier for users as they can just run one command and be guided through the relevant choices, and it should be more secure as they're discouraged from storing their token in code they might share.

This should also be compatible with `pipx run`, so users can run the following command from any environment.

```
pipx run qiskit-ibm-runtime save-account
```

### Examples

Here's a screenshot of me saving my account correctly.

Screenshot 2024-12-02 at 20 35 29

Here's the output of `--help`:

```
(.venv) ~ $ qiskit-ibm-runtime --help      
usage: qiskit-ibm-runtime [-h] {save-account} ...

Scripts for the Qiskit IBM Runtime Python package

options:
  -h, --help      show this help message and exit

Scripts:
  This package supports the following scripts:

  {save-account}
    save-account  Interactive command-line interface to save your account locally.
```


qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def user_input(message: str, is_valid: Callable[[str], bool]) -> str:
+    """
+    Repeatedly ask user for input until they give us something that satisifies
+    `is_valid`.
+    """
+    while True:
+        response = input(message + " ").strip()
+        if response == "quit":
```
Eric-Arellano, line 186: Maybe `q`? You could accept `q` or `quit` if you want both. Developers are lazy

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+        cls.save_to_disk(
+            {
+                "channel": channel,
+                "token": token,
+                "instance": instance,
+            }
+        )
+
+    @classmethod
+    def print_box(cls, lines: List[str]) -> None:
```
Eric-Arellano, line 102: Maybe move to Format?

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+        message=f"Enter a number 1-{len(options)} and press enter:",
+        is_valid=lambda response: response.isdigit()
+        and int(response) in range(1, len(options) + 1),
+    )
+    choice = options[int(response) - 1]
+    print(f"Selected {Format.greenbold(str(choice))}")
+    return choice
+
+
+class Format:
```
Eric-Arellano, line 211: Nice! It's not always safe to use color codes, such as when piping output or in certain terminals. I recommend renaming this to be `Formatter` and setting an attribute `supports_ansi_codes`. Then, change the methods to be instance methods that use `supports_ansi_codes`.

Determining the color code support is tricky. https://stackoverflow.com/questions/7445658/how-to-detect-if-the-console-does-support-ansi-escape-codes-in-python links to Django's approach in the first answer (click the link for the latest). Imo it's not critical that we always correctly detect things; it can be best effort.

It's also common for CLIs to have an option `--colors` and `--no-colors` to force the CLI to behave a certain way.

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+The `save-account` command-line interface. These classes and functions are not public.
```
Eric-Arellano, line 14: Nit that Python docstring is meant to have one sentence as the summary, then a new paragraph, then any supplemental text. https://peps.python.org/pep-0257/

```suggestion
The `save-account` command-line interface.

These classes and functions are not public.
```

This applies to several docstrings in this file.

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+"""
+
+import argparse
```
Eric-Arellano, line 17: Good choice to use argparse rather than third-party libraries like `click`. Keep the dependency list smaller.

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    This is the entry point for the `qiskit-ibm-runtime` command. At the
+    moment, we only support one script (save-account), but we want to have a
+    `qiskit-ibm-runtime` command so users can run `pipx run qiskit-ibm-runtime
+    save-account`.
+    """
+    # Use argparse to create the --help feature
+    parser = argparse.ArgumentParser(
+        prog="qiskit-ibm-runtime",
+        description="Scripts for the Qiskit IBM Runtime Python package",
```
Eric-Arellano, line 44: Maybe call this commands? That is more conventional with a CLI. Ditto the other references to "scripts"

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+        """Print lines in a box using Unicode box-drawing characters"""
+        width = max(len(line) for line in lines)
+        box_lines = [
+            "╭─" + "─" * width + "─╮",
+            *(f"│ {Format.bold(line.ljust(width))} │" for line in lines),
+            "╰─" + "─" * width + "─╯",
+        ]
+        print("\n".join(box_lines))
+
+    @classmethod
```
Eric-Arellano, line 112: These and the following methods should be `@staticmethod` since they don't use `cls`. That makes the interface more clear that you more quickly realize it's only a utility method. https://www.geeksforgeeks.org/class-method-vs-static-method-python/

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+        }[channel]
+        print(
+            Format.bold("\nPaste your API token")
+            + f"\nYou can get this from {Format.cyan(token_url)}."
+            + "\nFor security, you might not see any feedback when typing."
+        )
+        while True:
+            token = getpass("Token: ").strip()
+            if token != "":
+                return token
```
Eric-Arellano, line 133: Maybe worth making a helper function

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,234 @@
...
+        cls.print_box(
+            [
+                "⚠️ Warning: your token is saved to disk in plain text.",
+                "If on a shared computer, make sure to revoke your token",
+                "by regenerating it in your account settings when finished.",
+            ]
+        )
+
+
+def user_input(message: str, is_valid: Callable[[str], bool]) -> str:
```
Eric-Arellano, line 179: Maybe worth grouping these user input helper functions as static methods in a `UserInput` class. I like the grouping you're doing because it makes this single-file program better organized.

test/unit/test_cli.py
```diff
@@ -0,0 +1,191 @@
...
+from qiskit_ibm_runtime._cli import SaveAccountCLI, select_from_list
+
+from qiskit_ibm_runtime.accounts.account import IBM_CLOUD_API_URL, IBM_QUANTUM_API_URL
+from ..ibm_test_case import IBMTestCase
+
+
+class MockIO:
+    """Mock `input` and `getpass`"""
+
+    # pylint: disable=missing-function-docstring
```
Eric-Arellano, line 28: Imo this can be set at the file level since it's a test.

qiskit_ibm_runtime/_cli.py
frankharkins: Good idea!

qiskit_ibm_runtime/_cli.py
Eric-Arellano: Very sensible. Thanks!

test/unit/test_cli.py
frankharkins: I'm gonna leave it as I think docstring reminders are still useful to encourage explanations of the tests. We can always ignore them in the future.

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    # Use argparse to create the --help feature
+    parser = argparse.ArgumentParser(
+        prog="qiskit-ibm-runtime",
+        description="Commands for the Qiskit IBM Runtime Python package",
+    )
+    parser.add_subparsers(
+        title="Commands",
+        description="This package supports the following commands:",
+        dest="script",
```
Eric-Arellano, line 51: Nit that this should probably be "command"

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,264 @@
...
+        required=True,
+    ).add_parser(
+        "save-account",
+        description=(
+            "An interactive command-line interface to save your Qiskit IBM "
+            "Runtime account locally. This script is interactive-only."
+        ),
+        help="Interactive command-line interface to save your account locally.",
```
Eric-Arellano, line 59: Probably fine to remove this since you already set `description`

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,264 @@
...
+
+    # pylint: disable=missing-function-docstring
+    #
+    def __init__(self, color: bool):
+        self.color = color
+
+    @staticmethod
+    def _skip_if_no_color(
+        method: Callable[["Formatter", str], str]
+    ) -> Callable[["Formatter", str], str]:
```
Eric-Arellano, line 83: Rather than a decorator, a much simpler approach is to have a staticmethod called `_color()`. Have it take `s` and `color_code`, which is either 1, 32, 31, or 36. Note that everything else is the same in the ansi codes.

Keep it simple.

--

Feel free to use a different name of course. Maybe rename the instance var `color` to `color_enabled`

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,264 @@
...
+        return self.green(self.bold(s))
+
+
+class SaveAccountCLI:
+    """
+    This class contains the save-account command and helper functions.
+    """
+
+    def __init__(self, color: bool):
+        self.color = color
```
Eric-Arellano, line 130: Not used

```suggestion
```

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,264 @@
...
+    def greenbold(self, s: str) -> str:
+        return self.green(self.bold(s))
+
+
+class SaveAccountCLI:
+    """
+    This class contains the save-account command and helper functions.
+    """
+
+    def __init__(self, color: bool):
```
Eric-Arellano, line 129: Make it kwarg only

```suggestion
    def __init__(self, *, color: bool) -> None:
```

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,264 @@
...
+        except KeyboardInterrupt:
+            sys.exit()
+
+
+class Formatter:
+    """Format using terminal escape codes"""
+
+    # pylint: disable=missing-function-docstring
+    #
+    def __init__(self, color: bool):
```
Eric-Arellano, line 77: Make it kwarg only
```suggestion
    def __init__(self, *, color: bool) -> None:
```

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,264 @@
...
+        saving it to disk.
+        """
+        print(self.fmt.box(["Qiskit IBM Runtime save account"]))
+        channel = self.get_channel()
+        token = self.get_token(channel)
+        print("Verifying, this might take few seconds...")
+        try:
+            service = QiskitRuntimeService(channel=channel, token=token)
+        except (ApiException, IBMNotAuthorizedError, RequestsApiError) as err:
+            print(
```
Eric-Arellano, line 145: Error messages should print to `file=sys.stderr`

qiskit_ibm_runtime/_cli.py
```diff
@@ -0,0 +1,264 @@
...
+                is_valid=lambda response: response.strip().lower() in ["y", "yes", "n", "no", ""],
+            )
+            if response.strip().lower() in ["y", "yes"]:
+                AccountManager.save(**account, overwrite=True)
+            else:
+                print("Account not saved.")
+                return
+
+        print(f"Account saved to {self.fmt.greenbold(_DEFAULT_ACCOUNT_CONFIG_JSON_FILE)}")
+        print(
```
Eric-Arellano, line 210: Should probably print to sys.stderr since it's a warning

Eric-Arellano (approved):
This looks great from my perspective!

qiskit_ibm_runtime/_cli.py
frankharkins: We need both: `help` appears under `qiskit-ibm-runtime --help` and `description` appears under `qiskit-ibm-runtime save-account --help`.

qiskit_ibm_runtime/_cli.py
frankharkins: Ah yes that is much simpler, thanks

kt474:
Thanks @frankharkins, this looks great! I think for the documentation we should update the docstring in `save_account` and also add a release note explaining the purpose of the CLI and how to use it. 

I was testing the tool and it works well with my IQP production accounts but it didn't seem to work for staging accounts - I'm guessing this is intended because external users won't have staging accounts? 

I was also running into issues with trying to save a cloud account: 
Screenshot 2024-12-18 at 5 48 50 PM

There may be an issue with how the cloud instances are returned.

frankharkins:
I would like to continue this, but I don't have the bandwidth right now. I will wait until IQP account logic is removed and pick this back up then to be cloud-only.

PR #2081, opened 2024-12-13
Updating URL subdomain

francabrera:
### Summary

Updating urls from quantum-computing.ibm.com to quantum.ibm.com and from ibm.com/quantum-computing to ibm.com/quantum

### Details and comments
Fixes #


kt474:
> Not part of this PR, but urls are also saved in users's profiles, which is probably what most people use. We should have code to check for old subdomain and overwrite accordingly.

I can do this in a separate PR

PR #2086, opened 2024-12-17
Improved documentation for the private option

francabrera:
### Summary

Improved documentation for private option.

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/runtime_options.py
+++ b/qiskit_ibm_runtime/runtime_options.py
@@ -70,7 +70,12 @@ class RuntimeOptions:
                 this time limit, it is forcibly cancelled. Simulator jobs continue to use wall
                 clock time.
             session_time: Length of session in seconds.
-            private: Boolean of whether or not the job is marked as private.
+            private: Boolean that indicates whether the job is marked as private. When set to true,
+                input parameters are not returned, and the result can only be read once; after it is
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/runtime_options.py
```diff
@@ -70,7 +70,12 @@ def __init__(
                 this time limit, it is forcibly cancelled. Simulator jobs continue to use wall
                 clock time.
             session_time: Length of session in seconds.
-            private: Boolean of whether or not the job is marked as private.
+            private: Boolean that indicates whether the job is marked as private. When set to true,
+                input parameters are not returned, and the result can only be read once; after it is
+                read, both the input parameters and the results are deleted from the service. If the
+                result is not read, the parameters and results are automatically deleted after a
+                specified time. When set to false, the input parameters and results follow the standard
+                retention behavior.
```
kt474, line 78: Should we also mention that it's not supported on cloud ([comment](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1888#issuecomment-2512949360))?

riley2012: Nit: to be exact, which may not matter in this context: input parameters are deleted when the job is terminated. The result is deleted after it is read

kt474 (approved):

PR #2088, opened 2024-12-17
Add content/accept headers to appropriate http requests

### LINKED ISSUE #2063: Use `Content-Type` header in HTTP requests

This header is essential for indicating the media type of the data being sent in the request body, especially when working with APIs that expect specific formats like JSON. 

It seems that we tried to set that header as default here: https://github.com/Qiskit/qiskit-ibm-runtime/blob/639f3071833cef46443adf83eba7ea925b00257c/qiskit_ibm_runtime/api/rest/base.py#L24

But I think that it's not used.

**Acceptance criteria**

- [ ] Send `Content-Type` header as `application/json` if the request body is a dictionary or list, unless the user specifies a different value.

### LINKED ISSUE #2064: Use `Accept` header in HTTP requests

This `Accept` header allows the client to indicate to the server the content types it can handle in the response, which is crucial when interacting with APIs that can return multiple response formats.

It seems that currently we are sending `'Accept': '*/*'` or in some cases like when we create a Job we are not sending a value for this header.

**Acceptance criteria**

- [ ] Send `Accept` header as `application/json` if the response expected is a dictionary or list, unless the user specifies a different value.

kt474:
### Summary

Added `{"Content-Type": "application/json"}` and `{"Accept": "application/json"}` to the headers of the appropriate http requests. Also added `IBM-API-Version` to the header of all API requests. 

### Details and comments
Fixes #2063 
Fixes #2064 
Also fixes `ws-primitives #220`

### DIFF
--- a/qiskit_ibm_runtime/api/rest/base.py
+++ b/qiskit_ibm_runtime/api/rest/base.py
@@ -25,6 +25,8 @@ class RestAdapterBase:
 
     _HEADER_JSON_ACCEPT = {"Accept": "application/json"}
 
+    _HEADER_API_VERSION = {"IBM-API-Version": "2024-01-01"}
+
     def __init__(self, session: RetrySession, prefix_url: str = "") -> None:
         """RestAdapterBase constructor.
 
@@ -33,6 +35,7 @@ class RestAdapterBase:
             prefix_url: String to be prepend to all URLs.
         """
         self.session = session
+        self.session.headers = self._HEADER_API_VERSION
         self.prefix_url = prefix_url
 
     def get_url(self, identifier: str) -> str:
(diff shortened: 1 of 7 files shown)


ptristan3 (approved):
LGTM. I have only a minor concern related to the specific date.  `"IBM-API-Version": "2024-01-01"`

qiskit_ibm_runtime/api/rest/base.py
```diff
@@ -23,6 +23,10 @@ class RestAdapterBase:
 
     _HEADER_JSON_CONTENT = {"Content-Type": "application/json"}
 
+    _HEADER_JSON_ACCEPT = {"Accept": "application/json"}
+
+    _HEADER_API_VERSION = {"IBM-API-Version": "2024-01-01"}
```
kt474, line 28: @jelcaf just confirming if the api version here is correct?

jelcaf: Yes!

PR #2092, opened 2024-12-23
Add alt text to images (backport #2079)

arnaucasau:
We're trying to make sure that every image has alt text for accessibility.


kt474 (approved):

Eric-Arellano:
@kt474 this needs to be admin merged because the Python 3.10 job doesn't run on this branch (I suspect)

PR #2093, opened 2025-01-02
Extended rzz validation to include parameter expressions

### LINKED ISSUE #2067: Validate parameter expressions in rzz
In rzz gates, all parameter expressions, when evaluated with the corresponding parameter values, must be between 0 and pi/2.

Today we validate only parameters. We scan the circuit to collect a list of parameters that appear in rzz gates, then we check their values. The complexity of the value check is `O(number of rzz parameters * number of parameter sets)`. With parameter expressions, this will increase to `O(number of unique rzz parameter expressions * number of parameter sets)`.

yaelbh:
### Details and comments
Fixes #2067


wshanks (approved):
This looks good to me.

I think a release note would be good since it is a user-facing change (in the sense of the user getting an error in the client instead of an error from the service after a bad parameter binding).

I was not familiar with the internals of the Pub classes before so reading this code I was concerned about whether it handled multiple dimensions correctly. I added some dimensions to the tests and confirmed that it did.

The part that made me worried about dimensions (not being that familiar with the pub classes) was the relationship between `list(chain.from_iterable(pub.parameter_values.data))` and `pub.parameter_values.ravel().as_array()`. It might be nice if the pub class had a `parameter_names` property that was a shorthand for the first expression and if `as_array` documented that its columns matched `parameter_names` when its `parameters` argument was not used so that the connection was more clear.

wshanks:
The note looks clear to me. :+1:

kt474 (approved):

PR #2096, opened 2025-01-07
Use qiskit.qpy.QPY_VERSION (up to 13) instead of always QPY 11

### LINKED ISSUE #2095: Use QPY version 13

Currently we hard code the QPY version used to 11: 

https://github.com/Qiskit/qiskit-ibm-runtime/blob/0f345d847c5db22d7e3ffc1836e25a1bdc098f27/qiskit_ibm_runtime/utils/json.py#L260-L263

QPY 13 (from Qiskit 1.3) made some improvement by removing symengine, so we should be using that version. This implies requiring Qiskit 1.3, which should be fine given there shouldn't be any breaking changes within the same major version. This would have to wait for Qiskit Runtime service to support 1.3 first though, and deprecation warning issued by 1.3 should be fixed. 

**Acceptance criteria**

- [ ] Use QPY 13 once the server side supports it.

1ucian0:
### Summary

This PR always use the Qiskit supported QPY version or 13, if Qiskit SDK is ahead of server. Before merging this, server side needs to confirm that QPY 13 is available.

### Details and comments
Fixes #2095

Additionally, this PR increases the minimum version for `qiskit` to 1.3, making unnecessary the symengine and `_TERRA_VERSION` branches.

`on hold` because the following blockers:
 * #2097 Qiskit 1.3 does not have support for 3.8
 * Confirmation than server-side supports QPY 13


releasenotes/notes/dynamic_qpy_with_13_cap-c77411ff7e286c0b.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade:
+  - |
+    The minimal ``qiskit`` version is now 1.3.
+  - |
+    The QPY payload now is generated based on ``qiskit.qpy.QPY_VERSION`` and capped until version 13.
```
kt474, line 6: We switched to using towncrier for our release notes - https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/CONTRIBUTING.md#adding-a-new-release-note. You can add a release note to the `/unreleased` folder with a file name of `2096.upgrade.rst`.

1ucian0: oops, fixed in fbdc01efcc6e0006749d29d539c0d7f97cc55dbd

1ucian0:
Removing `on hold` because: #2097 got merged and server side has support for QPY 13

release-notes/unreleased/2096.upgrade.rst
```diff
@@ -0,0 +1,2 @@
+The minimal ``qiskit`` required version is now 1.3.  
+The QPY payload is now generated based on ``qiskit.qpy.QPY_VERSION`` and capped at QPY 13, which is supported by Qiskit 1.3.
```
mtreinish, line 2: I'm not sure if this warrants a user note. Mostly the QPY format version used on the wire should be transparent to users and I'm not sure it's worth documenting

qiskit_ibm_runtime/utils/json.py
```diff
@@ -258,9 +252,7 @@ def default(self, obj: Any) -> Any:  # pylint: disable=arguments-differ
         if hasattr(obj, "to_json"):
             return {"__type__": "to_json", "__value__": obj.to_json()}
         if isinstance(obj, QuantumCircuit):
-            kwargs: Dict[str, object] = {"use_symengine": bool(optionals.HAS_SYMENGINE)}
-            if _TERRA_VERSION[0] >= 1:
-                kwargs["version"] = 11
+            kwargs: Dict[str, object] = {"version": min(13, QISKIT_QPY_VERSION)}
```
mtreinish, line 255: I'd really like to see an to query the service endpoint added to query the max QPY version supported for the `13` here so we can do this automatically. But to avoid confusion on where the 13 comes from how about adding a constant variable to the top:

```python
SERVICE_MAX_SUPPORTED_QPY_VERSION = 13
```

to make it clear that the minimum here is between the max version the service understands, and the most recent QPY version that the installed version of qiskit knows how to generate.

release-notes/unreleased/2096.upgrade.rst
1ucian0: my point was to alert about the bump in the dependency requirement. Maybe I'm explaining too much why. Shall I just:
 ```suggestion
```
?

File: qiskit_ibm_runtime/utils/json.py (hunk 3 above)
Comment (1ucian0) [reply qiskit_ibm_runtime/utils/json.py:256]: done in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2096/commits/1b81b5e6265947c3e36c7278a587a1ee40b02e2d

Comment (kt474) [reply qiskit_ibm_runtime/utils/json.py:257]: Should this line here also be updated?

```kwargs = {"version": min(SERVICE_MAX_SUPPORTED_QPY_VERSION, QISKIT_QPY_VERSION)}```

Comment (1ucian0) [reply qiskit_ibm_runtime/utils/json.py]: good catch! done in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2096/commits/d4ff1c8cfc0a1bc943fb91110dc8c1b4b43eb2cd

File: release-notes/unreleased/2096.upgrade.rst (hunk 2 above)
Comment (mtreinish) [reply release-notes/unreleased/2096.upgrade.rst:2]: I would link it a bit more in the text than to explain the relationship.

```suggestion
Qiskit 1.3 introduces a new QPY version 13. This new format version offers better features and the
minimum required version of Qiskit was raised to start leveraging it for serializing circuits in job
submissions.
```

kt474 (approved):

PR #2100, opened 2025-01-09
Handle maintenance message

### LINKED ISSUE #2099: Handling maintenance responses

When the IBM Quantum channel platform is on maintenance, request to the APIs will respond with an HTTP code 503 and a similar body to this:

```json
{
  "error": {
    "statusCode": 503,
    "name": "MAINTENANCE",
    "message": "Planned maintenance. Please check https://docs.quantum.ibm.com/announcements/service-alerts",
    "code": 1999
  }
}
```

The provider should be able to handle this response and display the message in the body.

This was attempted to be fixed previously in https://github.com/Qiskit/qiskit-ibm-runtime/issues/1647, however the solution would have not worked, because:
- the fix was under the assumption that the `auth.quantum-computing.ibm.com/api` would still be responding, which was a bug on the maintenance process of the service not putting this also in maintenance (it was for `auth.quantum.ibm.com/api` but the provider was still using the old subdomain, it has been updated in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2081).
- the provider was not correctly specifying the `Accept` header to receive a JSON response and it was receiving an HTML response instead. This was updated in the provider in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2088.

It's fine to keep the retry logic, as long as at the end of the retries it displays the message in the response.

**Acceptance criteria**

- [ ] A Qiskit user using the `ibm_quantum` channel can see an exception message saying there is a planned maintenance ongoing.

kt474:
### Details and comments
Fixes #2099

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -12,7 +12,6 @@
 
 """Qiskit runtime service."""
 
-import json
 import logging
 import traceback
 import warnings
@@ -161,14 +160,8 @@ class QiskitRuntimeService:
             self._client_params.url = auth_client.current_service_urls()["services"]["runtime"]
             self._client_params.token = auth_client.current_access_token()
             self._api_client = RuntimeClient(self._client_params)
-            try:
-                self._hgps = self._initialize_hgps(auth_client)
-            except json.decoder.JSONDecodeError:
-                raise IBMApiError(
-                    "Unexpected response received from server. "
-                    "Please check if the service is in maintenance mode "
-                    "https://docs.quantum.ibm.com/announcements/service-alerts."
-                )
+            self._hgps = self._initialize_hgps(auth_client)
+
             self._backend_allowed_list = sorted(
                 set(sum([hgp.backends for hgp in self._hgps.values()], []))
             )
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -161,14 +160,8 @@ def __init__(
...
-            try:
-                self._hgps = self._initialize_hgps(auth_client)
-            except json.decoder.JSONDecodeError:
-                raise IBMApiError(
-                    "Unexpected response received from server. "
-                    "Please check if the service is in maintenance mode "
-                    "https://docs.quantum.ibm.com/announcements/service-alerts."
-                )
+            self._hgps = self._initialize_hgps(auth_client)
+
```
kt474, line 164: For maintenance outages the whole service including auth is down so this block would not be reached.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -275,6 +268,10 @@ def _authenticate_ibm_quantum_account(self, client_params: ClientParameters) ->
             Authentication client.
         """
         version_info = self._check_api_version(client_params)
+
+        if "ERROR" in version_info and version_info["ERROR"]["name"] == "MAINTENANCE":  # type: ignore
+            raise IBMApiError(version_info["ERROR"]["message"])  # type: ignore
+
```
kt474, line 274: Would be nice if we had a way to test this too

qiskit_ibm_runtime/api/session.py
```diff
@@ -38,7 +38,6 @@
     503,  # Service Unavailable
     504,  # Gateway Timeout
     520,  # Cloudflare general error
-    521,  # Cloudflare web server is down
     522,  # Cloudflare connection timeout
```
kt474, line 41: Don'y retry on 521, that's the status returned during maintenance

ptristan3 (approved):

PR #2102, opened 2025-01-13
Use IBM Cloud access token to call runtime API 

### LINKED ISSUE #1874: Use bearer token for cloud channel auth

On the Qiskit Runtime server side, when an API Key is provided in the Authorization header, the service goes to IAM to obtain a bearer token before making the authorization request.

This flow fails if the instance's account has Context Based Restrictions (CBR) enabled (example: IP allowlist).

The REST API does support Bearer token authentication (`Authorization` header with `Bearer TOKEN`) which should be able to bypass this.

**Acceptance criteria**
- [ ] an account with CBR enabled should be able to run a job successfully from a device that is allowlisted

**Notes**

This could probably be implemented a number of ways:

1. Provide optional setting to obtain bearer token using api key 
2. Convert auth flow for cloud channel to only use the bearer token flow (bearer token can be cached).

In my opinion option 2 is better because bearer tokens are a better security approach (if a bearer token gets compromised it automatically expires) and we won't need to maintain 2 different code paths (apikey and bearer token)

Please note that staging should be supported as well (endpoint is iam.test.cloud.ibm.com)

Example code block:

```
import requests

apikey = "replace with apikey"
resp = requests.post("https://iam.cloud.ibm.com/identity/token", data={"grant_type": "urn:ibm:params:oauth:grant-type:apikey", "apikey": apikey})
print(resp.status_code)
print(resp.json())
```

### LINKED ISSUE #2070: Call Runtime API using Access Token

The runtime API currently accepts either an apiKey or bearer access token for authorization. 
In the future the API would like to only accept access tokens for authorization. 

To make this transition we should update the client so that we call the IBM cloud API to exchange the provided API key for an access token. This access token can then be passed to the runtime API. 
The access token would be passed in the `Authorization` header with the bearer key word. For example
`Authorization: bearer MY_ACCESS_TOKEN`

The following API from IBM cloud should be used to generate the access token from the API key https://test.cloud.ibm.com/apidocs/iam-identity-token-api#gettoken-apikey

This change would be for the cloud channel. 

**Acceptance criteria**
- [ ] User can continue to pass the QiskitRuntimeService an API key for the cloud channel
- [ ] The requests made to the runtime API use an access token for authorization not apiKey

kt474:
### Summary

Bearer tokens have a `expires_in` value returned from IAM that we need to use to check when/if the token expires 

### Details and comments
Fixes #2070
Fixes #1874


qiskit_ibm_runtime/api/auth.py
```diff
@@ -40,9 +45,24 @@ def __call__(self, r: PreparedRequest) -> PreparedRequest:
...
+                url = STAGING_CLOUD_IAM_URL
+            headers = {"Content-Type": "application/x-www-form-urlencoded"}
+            data = f"grant_type=urn:ibm:params:oauth:grant-type:apikey&apikey={self.api_key}"
+            response = post(url, headers=headers, data=data, timeout=10).json()
+            return response["access_token"]
+        except Exception as ex:  # pylint: disable=broad-except
+            raise IBMNotAuthorizedError(
+                "Unable to retrieve IBM Cloud access token. Please check your credentials"
+            ) from ex
+
```
kt474, line 62: We only need this one endpoint - should this logic be in it's own Cloud auth client?

release-notes/unreleased/2102.other.rst
```diff
@@ -0,0 +1,2 @@
+IBM Cloud accounts will now use an access token to call the Qiskit Runtime API instead of the 
+token provided by the user.
```
ElePT, line 2: From the user perspective, what is the difference between the previous state and the new state? Asking for my own understanding. It feels like users don't need to change the provided token, right?

ElePT (commented):
I went through the PR but I think I'm missing the necessary context to give a proper review. Do you have an idea of who we should ping to review? I see there's an assigned reviewer but there hasn't been any action in a few months, things might have changed since then.

release-notes/unreleased/2102.other.rst
kt474: From the user perspective there should not be any difference. They will still use their cloud api key for initializing their account. They don't need to change their provided key. 

Even though this change shouldn't effect the user, I think it'd still be good to add a release note.

cpkurotori (commented):
lgtm, you might want to consider using the IBM core sdk though for this. Could consider using either [token manager](https://github.com/IBM/python-sdk-core/blob/main/ibm_cloud_sdk_core/token_managers/iam_token_manager.py) for minimal changes to this file. Otherwise could use the [iam authenticator](https://github.com/IBM/python-sdk-core/blob/main/ibm_cloud_sdk_core/authenticators/iam_authenticator.py#L25) but would likely require a little refactor.

kt474:
Thanks @cpkurotori! Addressed your comments - using `IAMTokenManager`, getting the `cname` from the `crn`, and using `expires_in`.

cpkurotori (approved):

PR #2116, opened 2025-01-27
Qiskit 2.0 Compatibility 

kt474:
### Summary

Since support for pulse gates is being sunset, the `Channel` related imports and code have been removed:
- `fake_backend.py`: `AcquireChannel`, `DriveChannel`, `MeasureChannel`, `ControlChannel` 
- `backend_configuration.py`: `AcquireChannel`, `Channel`, `ControlChannel`, `DriveChannel`,`MeasureChannel`
- `ibm_backend.py`: `AcquireChannel`, `ControlChannel`, `DriveChannel`,`MeasureChannel`
- `pulse_defaults.py`: `InstructionScheduleMap` and `PulseQobjDef`. 
- `PulseBackendConfiguration` has been removed, all backend configurations will be returned as `QasmBackendConfiguration`. 
- `c_if` and `condition_bit` removed in https://github.com/Qiskit/qiskit/pull/13506
- `has_calibration_for` and `calibrations` removed in https://github.com/Qiskit/qiskit/pull/13861

Non pulse related: 
- `qiskit.circuit.bit` removed in https://github.com/Qiskit/qiskit/pull/13860
- The deprecated `EfficientSU2` has been replaced with `efficient_su2`

Without `Channel`, `PulseDefaults` are no longer needed for the backend target.

I'll handle the DD related pulse deprecations in a separate PR. Also we should clean up a lot of the backend defaults code since it will no longer be necessary. 

### Details and comments
Fixes #


kt474:
Pulse deprecations: 

`fake_backend.py`: `AcquireChannel`, `DriveChannel`, `MeasureChannel`, `ControlChannel` - used to set the channels map of a fake backend configuration,

wshanks:
Perhaps all the `defs_*.json` files for the fake backends should also be deleted in this PR?

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -462,45 +453,6 @@ def configuration(
...
-
-        Args:
-            qubits: Tuple or list of qubits of the form
-                ``(control_qubit, target_qubit)``.
-
-        Returns:
-            List[ControlChannel]: The Qubit measurement acquisition line.
-        """
-        return self._configuration.control(qubits=qubits)
-
```
kt474, line 503: Can these be removed without a deprecation period? Without pulse support, these methods would essentially be useless (quantum-cloud issue 843)

kt474: @jyu00 Can these be removed?  `drive_channel`, `measure_channel`, `acquire_channel`, `control_channel`

wshanks (approved):
The removals all look reasonable to me. I think there is more that could be removed if the goal is to remove dead code related to pulse (like `PulseBackendConfiguration` and fake `defs_*.json` files). I ran the tests with Qiskit 2.0 (and the changes in #2162 merged in) though and they all passed (after fiddling with Aer which still has issues with Qiskit 2.0), so I think this is good for enabling compatibility with Qiskit 2.0.

eliarbel (changes requested):
Starting Qiskit 2.0, `QuantumCircuit` and `DAGCircuit` no longer support calibration handling,  so the following changes are required: 

Remove this: 
https://github.com/Qiskit/qiskit-ibm-runtime/blob/3fbaa7e5e70b07028429749ef29a02dafa14e4ac/qiskit_ibm_runtime/transpiler/passes/scheduling/block_base_padder.py#L207

Remove the if branch:
https://github.com/Qiskit/qiskit-ibm-runtime/blob/3fbaa7e5e70b07028429749ef29a02dafa14e4ac/qiskit_ibm_runtime/transpiler/passes/scheduling/block_base_padder.py#L272

Remove the try block, use the `gate_length` in the except block:
https://github.com/Qiskit/qiskit-ibm-runtime/blob/3fbaa7e5e70b07028429749ef29a02dafa14e4ac/qiskit_ibm_runtime/transpiler/passes/scheduling/dynamical_decoupling.py#L351

Remove the calibration condition: 
https://github.com/Qiskit/qiskit-ibm-runtime/blob/3fbaa7e5e70b07028429749ef29a02dafa14e4ac/qiskit_ibm_runtime/utils/utils.py#L74

Remove the if branch:
https://github.com/Qiskit/qiskit-ibm-runtime/blob/3fbaa7e5e70b07028429749ef29a02dafa14e4ac/qiskit_ibm_runtime/transpiler/passes/scheduling/scheduler.py#L221


Also, minor doc updates, namely removing pulse schedule mentions:
In
https://github.com/Qiskit/qiskit-ibm-runtime/blob/3fbaa7e5e70b07028429749ef29a02dafa14e4ac/qiskit_ibm_runtime/ibm_backend.py#L287 

And in `README.md`

eliarbel:
One more thing to note @kt474: Due to https://github.com/Qiskit/qiskit/pull/13860/ the import path `qiskit.circuit.bit` doesn't exist anymore. This causes a few unit test against Qiskit 2.0 to fail. This is not directly related to pulse removal (which is the topic of this PR) but just mentioning it here in case you want to include it already in this PR instead of opening a new one

wshanks:
I think I tested with a build of Qiskit 2.0 before https://github.com/Qiskit/qiskit/pull/13860 was merged so I missed the `qiskit.circuit.bit` issue. I think the other things you highlighted (besides the docstring) are addressed in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2162 @eliarbel

qiskit_ibm_runtime/models/backend_configuration.py
```diff
@@ -549,7 +540,6 @@ def __init__(
         display_name: str = None,
         description: str = None,
         tags: list = None,
-        channels: Dict[set, Any] = None,
```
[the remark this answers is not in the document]
kt474: Yeah the main goal was to make `qiskit-ibm-runtime` work with Qiskit 2.0. I agree that there could be a lot more pulse related cleanup. I'll make sure to handle that in a separate PR. 

Created issue: https://github.com/Qiskit/qiskit-ibm-runtime/issues/2179

kt474:
Addressed all comments and merged #2161 into this branch. 

Passes all unit tests against Qiskit 2.0rc1 except the qiskit-aer import issues in the Neat tests which will be fixed in https://github.com/Qiskit/qiskit-aer/pull/2327 (so we'll need a release on the aer side too).

eliarbel (changes requested):
Thanks @kt474 for putting this all together in a single PR, this helps!
I've left some more requests for changes and questions in line.

qiskit_ibm_runtime/ibm_backend.py
wshanks: I recommend not tying removing all the `BackendV1` methods to the task of adding compatibility with Qiskit 2.0. The `.configuration()` and `.properties()` methods still return some information that is not available via the `BackendV2` interfaces. Providers are allowed to extend `BackendV2` with additional methods and we can view these methods in that way. It still might be worth reviewing what is in `configuration()` and `properties()` and seeing how what is missing could be moved into the target or somewhere else in the `BackendV2` interface, but I don't think it has to be tied to this PR. I think the cost of keeping them is fairly low since the API endpoints are for configuration and properties directly, but perhaps we would want to change the format of the data that the API provides to something structured more like the target and then it would make more sense to deprecate the older formats.

test/unit/transpiler/passes/scheduling/test_utils.py
```diff
@@ -22,36 +22,6 @@
 class TestDynamicCircuitInstructionDurations(IBMTestCase):
     """Tests the DynamicCircuitInstructionDurations patching"""
 
-    def test_patch_measure(self):
```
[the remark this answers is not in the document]
kt474: added back

test/unit/transpiler/passes/scheduling/test_utils.py
```diff
@@ -69,16 +39,3 @@ def test_durations_from_target(self):
         self.assertEqual(durations.get("x", (0,)), 160)
         self.assertEqual(durations.get("measure", (0,)), 3200)
         self.assertEqual(durations.get("reset", (0,)), 3200)
-
-    def test_patch_disable(self):
```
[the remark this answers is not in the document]
kt474: added back

qiskit_ibm_runtime/ibm_backend.py
jyu00: +1 on what Will said. I'd be very cautious about removing `configuration()` and `properties()`, given many internal and external code still relies on them. Having those methods also don't break the `BackendV2` interface as additional methods are allowed. Although `IBMBackend` does break the interface by not implementing `run()`... but that's a different story. 

As to @kt474's original question, unfortunately I don't see a way to remove `*_channel()` gracefully, since `drive_channel()`, for example, says it'd return a `DriveChannel`. So even returning a `None` would be a breaking change. I guess we can argue the announcement of pulse gate deprecation on IBM Quantum was the "deprecation warning". And if it has to be a breaking change regardless, we might as well remove them.

jyu00: Also these methods have been broken since we introduced Heron, which never supported pulse gates. E.g. it'd raise an `AttributeError` if you try to call `drive_channel()` with ibm_torino.

qiskit_ibm_runtime/models/backend_configuration.py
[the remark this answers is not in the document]
jyu00: Depending on what you meant by "pulse supporting backends". None of the IBM Quantum backends support pulse gates anymore. Pulse information is still being reported by Eagle devices, but there is an open ticket to remove that, given the information is no longer useful without the ability to run pulse jobs.

wshanks (commented):
The removal of `PulseBackendConfiguration` looks good except for one small thing.

eliarbel (approved):
This LGTM now given the clarification from Jessie about `IBMBackend` API and the removal of `PulseBackendConfiguration`. 
There are still a few removed tests in `test_schedulers.py` which could be kept but I don't think worth blocking this PR on at this stage.

ElePT:
> Addressed all comments and merged #2161 into this branch.
> 
> Passes all unit tests against Qiskit 2.0rc1 except the qiskit-aer import issues in the Neat tests which will be fixed in [Qiskit/qiskit-aer#2327](https://github.com/Qiskit/qiskit-aer/pull/2327) (so we'll need a release on the aer side too).

[Qiskit/qiskit-aer#2327](https://github.com/Qiskit/qiskit-aer/pull/2327) is included in the latest release of Aer that just came out: `qiskit-aer 0.17.0`.

kt474:
> https://github.com/Qiskit/qiskit-aer/pull/2327 is included in the latest release of Aer that just came out: qiskit-aer 0.17.0.

All unit tests passing on both Qiskit 1.4.1 and 2.0rc1 with the latest aer release. Going to merge now, other bugs found will be handled in separate PRs.

PR #2120, opened 2025-01-28
Remove parameter expression validation check for gen3-turbo

ihincks:
### Summary

Remove parameter expression validation check for gen3-turbo

### Details and comments

The gen3-turbo mode now supports this.


yaelbh (approved):
LGTM. Perhaps you can remove `has_param_expressions`, it's not used anymore, at least not in this repository.

ihincks:
Thanks @yaelbh . I'd like to consider whether these functions should have deprecation periods before merging.

ihincks:
Verified that these functions are not documented in the API docs, so we are fine to remove them without deprecation.

PR #2121, opened 2025-01-28
Update deprecation policy to include what is public

jyu00:
### Summary

There was a question this morning on what is considered public API. This PR updates the deprecation to include that section (credit goes to the Qiskit team as it's largely copied from there). 

### Details and comments
Fixes #


kt474 (approved):

ihincks (commented):
Thanks @jyu00 !

PR #2124, opened 2025-01-30
Handle old deprecations

kt474:
### Summary

- Removed support for the simulator option `noise_model`. This option will still work in local mode. We no longer need the `from_dict()` method ported over from Qiskit Aer. FWIW we originally ported `from_dict()` after it was removed from Aer in `0.15.0` but it was just added back to Aer [here](https://github.com/Qiskit/qiskit-aer/pull/2222) in `0.16.0` but may be removed again. 

- Removed deprecated `NoiseLearnerResult` properties `generators` and `rates`. 
- Removed deprecated util function `get_runtime_api_base_url()`. 

### Details and comments
Fixes #


qiskit_ibm_runtime/utils/json.py
```diff
@@ -509,8 +513,8 @@ def object_hook(self, obj: Any) -> Any:
             if obj_type == "to_json":
                 return obj_val
             if obj_type == "NoiseModel":
-                if HAS_AER:
-                    return from_dict(obj_val)
+                if HAS_AER and _AER_VERSION[1] >= 16:
+                    return qiskit_aer.noise.NoiseModel.from_dict(obj_val)
```
kt474, line 517: Should we keep support for serializing `NoiseModel`? It will still work with `aer > 16`

Issue #2125, opened 2025-02-04
Convert pubs to equivalent pubs that are rzz-valid


yaelbh:
Opening a fresh blank issue for an approach to solving #2032, see related discussion also in #2072.

User circuits may contain RZZ gates with invalid angles (i.e., negative or greater than `pi/2`). There are two types of usages of RZZ:
1. Numeric - for example an RZZ gate with angle of -1.3 radians on qubits 4 and 2: `rzz(-1.3, 4, 2)`. We take care of these gates when transpiling for IBM backends, using the [`FoldRzzAngle` pass](https://github.com/Qiskit/qiskit-ibm-runtime/blob/00b8829d172ce2dfb4f0298850f6d84ab4b11d24/qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py#L28).
2. Parametrized - the angle is specified by a parameter or a parameter expression, for example `rzz(p, 4, 2), rzz(p1 + p2, 3, 4)`. The parameters are later instantiated with values, which may result in invalid angles. This issue is about this parametrized case.

#2032 focuses on invalid pubs resulting from transpilation. Another important and frequent use case relates to the original pub, as specified by the user, even before transpilation: the pub can contain many parameter sets, making it impossible for the user to detect and fix faulty parameter sets. This issue applies to both scenarios.

We will make pubs valid by following the construction of the `FoldRzzAngle` pass. We will add to the circuit new `rz` and `rx` gates, and extra parameters for the new gates and for a modified RZZ gate. We will append values of the new parameters to the parameter values.

An interesting question is whether to implement this procedure by writing a function that receives valid pubs and outputs invalid pubs; or by writing a transpiler pass, and accompanying it with a function that modifies the parameter values. A function is a cleaner interface and easier to opt-in. It may however mess up with dynamical decoupling. I'm interested to hear opinions about these two options (@nkanazawa1989 , @wshanks, @ihincks, @jyu00).

A bit more clarification about dynamical decoupling: the user can run the function before or after transpilation. If before then we can get the #2032 surprise. If after then the circuit is altered after dynamical decoupling.

Finally, a smaller things to think of: Automatic addition of new parameters may cause name collision with user parameters.

wshanks:
[...]

PR #2126, opened 2025-02-04
Update parameters to accommodate with RZZ constraints

yaelbh:
### Summary

Solves #2125 through a function that translates from pub to pub.
Only stared - work is still in progress.

### Details and comments
- ~In fact, in order not to have to bother with the different structures parameter values can have in pubs, at least for now we translate a pair of circuit and parameter values to another pair, and require the parameter values to be just a list of tuples. The order inside each tuple matches the order of appearance of parameters in the circuit, just like one of the options users have to specify parameter values when defining pubs.~
- Note the discussion in #2125 about an option that's not implemented here, to do it using a transpiler pass.

### DIFF
--- a/qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
+++ b/qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
@@ -281,8 +281,7 @@ def convert_to_rzz_valid_pub(
     for p_name in pub_params:
         if p_name.startswith("rzz_"):
             raise ValueError(
-                "Original pub is not allowed to contain parameters "
-                "whose names start with rzz_"
+                "Original pub is not allowed to contain parameters whose names start with rzz_"
             )
 
     # first axis will be over flattened shape, second axis over circuit parameters


yaelbh:
I left undone for now:
1. Support for dynamic circuits.
2. Preserving the global phase.

Other than that it's ready to play with.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -244,3 +248,163 @@ def _quad4(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
...
+                        RXGate(param_rx),
+                        (qubits[0],),
+                    )
+                )
+                val_data[f"{param_prefix}rx"] = rx_angles
+
+        if is_rz or is_rx:
+            # param_exp * 0 to prevent an error complaining that the original parameters,
+            # still present in the parameter values, are missing from the circuit
+            param_rzz = param_exp * 0 + Parameter(f"{param_prefix}rzz")
```
wshanks, line 398: I am not sure where the error comes in. It might be worth checking with the Qiskit team if this is the best way to do this. I just worry that a future change could make `0 * param` still drop `param` from the circuit.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -244,3 +248,163 @@ def _quad4(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
...
+
+            if angle % pi >= pi / 2:
+                rx_angles[idx] = pi
+            else:
+                rx_angles[idx] = 0
+
+            rzz_angles[idx] = pi / 2 - abs(mod(angle, pi) - pi / 2)
+
+        rzz_count += 1
+        param_prefix = f"rzz_{rzz_count}_"
```
wshanks, line 337: It might be worth checking if the original circuit has `rzz_1_` parameters. I guess that would only happen if function had already been run on the pub and then all the rzz should already be folded and not need new parameters added.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -244,3 +248,163 @@ def _quad4(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
...
+        qubits = instruction.qubits
+
+        is_rz = False
+        if any(not np.isclose(rz_angle, 0) for rz_angle in rz_angles):
+            is_rz = True
+            if all(np.isclose(rz_angle, pi) for rz_angle in rz_angles):
+                new_data.append(
+                    CircuitInstruction(
+                        RZGate(pi),
```
wshanks, line 346: I wonder if this case and the `is_x` case come up enough to be relevant vs just always using a parameter for consistency.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -244,3 +248,163 @@ def _quad4(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
...
+                    CircuitInstruction(
+                        RXGate(param_rx),
+                        (qubits[0],),
+                    )
+                )
+
+    new_circ.data = new_data
+
+    if program_id == "sampler":
+        return SamplerPub.coerce((new_circ, val_data), pub.shots)
```
wshanks, line 423: At some point, you need to ravel the new rzz_ parameters back to match the shape of the data in the original pub to support parameter arrays with more than one dimension. For example, this code fails for me on this line:

```python
import numpy as np
from qiskit.circuit import Parameter
from qiskit_ibm_runtime.transpiler.passes.basis.fold_rzz_angle import convert_to_rzz_valid_pub

p1 = Parameter("p1")
p2 = Parameter("p2")

circ = QuantumCircuit(3)
circ.rzz(p1 + p2, 0, 1)
circ.rzz(0.3, 0, 1)
circ.x(0)
circ.rzz(p1 - p2, 2, 1)
convert_to_rzz_valid_pub("sampler", (circ, {"p1": np.ones((3, 2)), "p2": np.ones((3, 2))}))
```

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -244,3 +248,163 @@ def _quad4(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
             check=False,
         )
         return new_dag
+
+
+def convert_to_rzz_valid_pub(
```
wshanks, line 255: Since this is not a transpiler pass that will be run automatically, do you want to expose it in a more public way by including it in the documentation and possibly providing a higher level import path than `qiskit_ibm_runtime.transpiler.passes.basis.fold_rzz_angle`?

wshanks (commented):
This looks good to me! My main comments are to handle the case of mutlidimensional parameter arrays and to expose the function publicly. My other comments are just suggestions.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
ElePT: @yaelbh do you have an error trace we can look at? I am also not sure what error this is referring to, but I agree that we should take a look.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -244,3 +248,163 @@ def _quad4(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
...
+    program_id: str, pub: Union[SamplerPubLike, EstimatorPubLike]
```
ElePT: If this is intended to be user-facing, I feel like it would be smoother to allow the function to accept a primitive instance instead of a `program_id`. I don't think that the `program_id` is an externally understood attribute (and it probably shouldn't be), and you can always fetch it from the primitive.

```suggestion
    primitive: Union[SamplerV2, EstimatorV2], pub: Union[SamplerPubLike, EstimatorPubLike]
```

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -244,3 +248,163 @@ def _quad4(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
...
+        if operation.name != "rzz" or not isinstance(
+            (param_exp := instruction.operation.params[0]), ParameterExpression
+        ):
+            new_data.append(instruction)
+            continue
+
+        param_names = [param.name for param in param_exp.parameters]
+
+        col_indices = [np.where(pub_params == param_name)[0][0] for param_name in param_names]
+        # col_indices is the indices of columns in the parameter value array that have to be checked
```
ElePT: super minor suggestion:

```suggestion
        # col_indices is the indices of columns in the parameter value array that have to be checked
        col_indices = [np.where(pub_params == param_name)[0][0] for param_name in param_names]
```

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
```diff
@@ -244,3 +248,163 @@ def _quad4(angle: float, qubits: Tuple[Qubit, ...]) -> DAGCircuit:
...
+                rz_angles[idx] = pi
+            else:
+                rz_angles[idx] = 0
+
+            if angle % pi >= pi / 2:
+                rx_angles[idx] = pi
+            else:
+                rx_angles[idx] = 0
+
+            rzz_angles[idx] = pi / 2 - abs(mod(angle, pi) - pi / 2)
```
ElePT, line 334: When the target allows to store angle bound information, this could be extended to read the bounds from the target.

yaelbh: Let me know when it happens.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
yaelbh: @ElePT 
1. Which import level do you think fits?
2. Where does the relevant documentation reside?

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
yaelbh: Done in e0190510. There is a problem though, that defining a primitive instance requires `mode`. It may confuse users, since in practice we're not using this information.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
yaelbh: c0c08897

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
yaelbh: I don't know. It will happen if all the angles belong to the same quad.

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
yaelbh: The error is correct. A pub must not contain values for a parameter that's not in the circuit. So, for
```python
p = Parameter("p")
circ = QuantumCircuit(1, 1)
circ.measure(0, 0)
sampler.run([(circ, {"p": 1})]).result()
```
we rightfully obtain
```
ValueError: The number of values (1) does not match the number of parameters (0) for the circuit.
```

In the context of this PR, we replace `param_exp` in the rzz gate with `param_rzz`. As a result, it is possible that some (or all) of the parameters in `param_exp` are not contained in the circuit anymore. An alternative to `param_exp * 0` is to check for each parameter in `param_exp` if it's still used, and remove its parameter values if not. It's a few more lines to code, on the other hand it will truncate redundant parameter values. Do you have an opinion?

qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py
yaelbh: Thanks for catching it, it's fixed in d3715101

wshanks (commented):
Looks good to me besides getting feedback from @ElePT on where to document the function.

wshanks (approved):
Approving now because I don't expect @ElePT to be available to give feedback soon and the code all seems good. We can have a follow up PR just to add the function to the public API at a later point and that should be easy to review.

PR #2133, opened 2025-02-10
Adding a version to NoiseLearner

joshuasn:
### Summary

Adds versioning information to NoiseLearner to keep it in sync with Estimator and Sampler primitives.


qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -126,6 +126,8 @@ class NoiseLearner:
 
     """
 
+    version = 2
+
```
ptristan3, line 130: Did we ever have version 1? or did we start with version 2 to balance with Sampler/Estimator?

PR #2137, opened 2025-02-12
Remove e2e test related code

### LINKED ISSUE #1768: There are no e2e tests
There is an `e2e` test folder, an e2e step in CI/Makefile, but there aren't actually any tests.

kt474:
### Summary

There have never been `e2e` tests in this repo, I think we should just remove the mentions of `e2e` tests since we already have integration tests that test on `test_eagle`. 

### Details and comments
Fixes #1768


ptristan3 (approved):

mriedem:
Why didn't the integration tests catch the regression in #2142 then?

kt474:
> Why didn't the integration tests catch the regression in #2142 then?

The integration tests install the dev dependencies so `aer` is included - the error only happens if `aer` isn't installed 

```
pip install -e '.[dev]'
```

PR #2139, opened 2025-02-13
Do not install unnecessary files to site-packages

### LINKED ISSUE #2138: Wheel packages contain unnecessary files
**Describe the bug**

[Official wheel packages available in PyPI](https://pypi.org/project/qiskit-ibm-runtime/) contain unnecessary directories that get installed to the site-packages directory. This might not be a problem when using virtual environments but it is a problem for global installations and also when creating packages for GNU/Linux distros.

**Steps to reproduce**

1. Download [the latest available wheel package](https://pypi.org/project/qiskit-ibm-runtime/#files)
2. Open and check directories
3. Alternatively, clone the repo and run, for example, `python -m build --wheel`

**Expected behavior**

Only the relevant directories should be included in the wheel package, i.e., `qiskit_ibm_runtime` and, at the moment, `qiskit_ibm_runtime-0.35.0.dist-info`.

**Suggested solutions**

Modify the relevant files (I guess `pyproject.toml`) to not include extra directories, i.e., `docs`, `release-notes` and `tools` to site-packages. These can be included in the source distribution tarball.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.35.0
- **Python version**: 3.13.2

iyanmv:
### Summary

Because the `pyproject.toml` has an implicit `setuptools.packages.find`, setuptools considers implicit namespaces by default. This causes directories like `docs`, `tools` and `release-notes` to be included in the wheel packages. If the directory is not clean, and wheel is recreated, the `build` directory is also included.

The solution is to add `namespaces = false` to prevent any folder without an `__init__.py` file from being scanned.

### Details and comments

Fixes #2138


kt474 (approved):
Thank you @iyanmv!

PR #2143, opened 2025-02-14
Cleanup websocket streaming code

### LINKED ISSUE #2140: Websocket streaming not cleaned up
Interim result and job status streaming was removed a while ago. However, a bunch of dead code was not cleaned up, for example:

- There is still code to create the websocket client 
- There are several streaming related methods such as `cancel_result_streaming`
- Also code to establish interim result decoder
- And code to allow user callback 

I don't think a ws stream is actually established anywhere, but it's hard to tell because of all the residual code.

kt474:
### Details and comments
Fixes #2140


qiskit_ibm_runtime/base_primitive.py
```diff
@@ -163,6 +164,15 @@ def _run(self, pubs: Union[list[EstimatorPub], list[SamplerPub]]) -> RuntimeJobV
...
         logger.info("Submitting job using options %s", primitive_options)
 
+        if options_dict.get("environment", {}).get("callback", None):
+            issue_deprecation_msg(
+                msg="The 'callback' option is deprecated",
+                version="0.37.0",
+                remedy="This option will have no effect since interim "
+                "results streaming was removed in a previous release.",
+                stacklevel=3,
+            )
```
kt474, line 174: My understanding is that even though this option doesn't do anything, we still need to properly deprecate before removing.

qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -91,49 +88,39 @@ def __init__(
...
+            self._final_result_decoder = decoder
+
+        if user_callback or client_params:
+            warnings.warn(
+                "Interim results streaming was deprecated and removed in previous releases "
+                "so passing in 'user_callback' or 'client_params' will have no effect. "
+                "These parameters will be removed in a future release.",
+                category=FutureWarning,
+                stacklevel=2,
+            )
```
kt474, line 118: Same here - `user_callback` and `client_params` don't do anything in the job classes but following the deprecation policy I don't think we can just remove them?

PR #2173, opened 2025-03-05
Update use of classes removed in Qiskit 2.0

ElePT:
### Summary
https://github.com/Qiskit/qiskit/pull/13793 removes the `qobj` module. A consequence of this removal is the move of `MeasLevel` and `MeasResult` from `qiskit.qobj.utils` to `qiskit.result`. This PR allows to import from both paths for compatibility with 1.x and 2.x.

The same PR removes configuration and property errors. Because the `BackendConfiguration` and `BackendProperties` classes have been vendored in qiskit-ibm-runtime, the errors should be too.

### Details and comments
Fixes #
This PR will allow to fix: https://github.com/Qiskit/qiskit-neko/issues/52

### DIFF
--- a/qiskit_ibm_runtime/fake_provider/fake_backend.py
+++ b/qiskit_ibm_runtime/fake_provider/fake_backend.py
@@ -52,6 +52,9 @@ from ..models import (
     QasmBackendConfiguration,
     PulseBackendConfiguration,
 )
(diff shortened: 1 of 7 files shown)


kt474:
Cancelling the integration tests because they shouldn't run in PRs

PR #2192, opened 2025-03-21
Use new API URL 

### LINKED ISSUE #2147: New IQP API integration
With the launch of the new IQP API, all URLs used in this lib will be redirected to it. We have to check that the new API works as the previous one. 

@SooluThomas You mentioned a problem with `/versions`. Could you report it here?

kt474:
### Summary

Use the new API URLs

staging: https://quantum.test.cloud.ibm.com/api 
production: https://quantum.cloud.ibm.com/api 

### Details and comments
Fixes #2147


ElePT (commented):
Thanks @kt474, it looks like the new URL is already active and working, right? do you see any blocker for this PR or should we go ahead and merge it?

kt474:
> Thanks @kt474, it looks like the new URL is already active and working, right? do you see any blocker for this PR or should we go ahead and merge it?

I'll follow up with the IQP side, I think we'll have to do more testing with the the URLs
> 
> From the release note perspective, I am debating whether we should add an API Changes release note or not. It seems reasonable to document the change, but it's not the definition of "user facing" that I am used to.

We can document any discrepancies between the APIs that effect the user. But if they're exactly the same then there's no need.

PR #2195, opened 2025-03-24
Refactor `create_new_session` logic when calling `Session.from_id()`

### LINKED ISSUE #2194: Thread safety issue in Session and Batch due to shared _create_new_session class variable
**Describe the bug**

There is a thread-safety issue in [`session.py`](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/session.py), which also affects the `Batch` class via inheritance.

The class variable `_create_new_session = True` ([line 84](https://github.com/Qiskit/qiskit-ibm-runtime/blob/990be8098ea0487d2a3a13b9b9b72263100c87f3/qiskit_ibm_runtime/session.py#L84)) is overwritten in `from_id` ([lines 344–346](https://github.com/Qiskit/qiskit-ibm-runtime/blob/990be8098ea0487d2a3a13b9b9b72263100c87f3/qiskit_ibm_runtime/session.py#L344-L346)). This shared mutable state can result in race conditions when multiple threads call `from_id` or `__init__`, leading to unexpected behavior in how sessions are reused or created.

**Steps to reproduce**

Call `Session.from_id(...)` or instantiate multiple `Session` or `Batch` objects concurrently in different threads and observe inconsistent session reuse or creation.

**Expected behavior**

Session creation logic should be thread-safe. Concurrent access should not result in unexpected session reuse or unexpected creation of new sessions.

**Suggested solutions**

A possible fix is to remove the use of the class-level variable `_create_new_session`, and instead pass this state explicitly via a keyword argument to `__init__` and `_create_session`, e.g.:

```python
def __init__(..., *, create_session: bool = False)
def _create_session(..., *, create_session: bool = False)
```
This avoids shared mutable state and improves thread safety.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.37.0 (exists in previous versions too)
- **Python version**: Any version supporting threading (e.g., 3.10+)
- **Operating system**: Operating system: All (OS-agnostic issue)

kt474:
### Summary

This is a better approach for preventing `Session.from_id()` from calling the POST session endpoint. We don't need a global class variable. 

### Details and comments
Fixes #2194


kt474:
> 1. If `_create_new_session` is no longer used, line 85 in batch.py should be removed (couldn't leave an inline comment).
> 2. Is there any way this fix can be unit-tested?
> 3. Batch and Sessions are publicly documented classes, shouldn't a new argument come with a release note?

Addressed in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2195/commits/60f23aa78a171936f8f810e059b77230017a3dcd - updated an integration test to test that `create_session` is not called in `from_id()`

release-notes/unreleased/2195.bug.rst
```diff
@@ -0,0 +1,7 @@
+The Session and Batch global class variable ``_create_new_session`` has been removed. ``_create_new_session`` was 
+used to control whether or not the POST session API endpoint should be called when initializing a session. 
+For example, when ``Session.from_id()`` is called, we do not want to create a new session. 
+This caused issues when there were multiple sessions or batches open. 
+
+A new parameter ``new_session``, which defaults to ``True``, has been added as a replacement. Initializing a session
+and using ``from_id()`` still work in the same way.
```
ElePT, line 7: We typically don't share details of internal variables in Qiskit and rather focus on the user-facing aspect of the change. In this case, that would be the new parameter and the fix in behavior for `Session.from_id()`. Something like:

```suggestion
A new parameter ``new_session`` has been added to the :class:`.Batch` and :class:`.Session` classes. This parameter allows to control whether the POST session API endpoint should be called when initializing the class. 

It defaults to ``True`` as this is the case for most user-facing workflows. However, methods such as :meth:`.Session.from_id` set it to ``False`` to avoid generating a new session when the original session is still active. This fixed issues where multiple sessions and batches were generated simultaneously when accessing them through ``.from_id()``.
```

We also typically link to the API reference in the release notes using Sphinx directives. I believe this should work in the current setup too.

ElePT (commented):
Thanks a lot @kt474 for the quick updates. Working on the release note suggestions I started wondering if `new_session` is the best name for the new variable, because well, the class name is `Session` (or known to be the counterpart of `Session`). Should we use something more concrete that refers to the API  endpoint and tells it apart from the `Session` object?

release-notes/unreleased/2195.bug.rst
kt474: Thanks! I was having trouble with getting the wording right in this release note. Also, yes I agree, `new_session` is not the best name - maybe `create_new` or `open_new` is better.

ElePT (approved):
Naming is always hard, but `create_new` looks reasonable. Thanks for applying my suggestions.

PR #2222, opened 2025-04-15
Fix session/batch type hint

kt474:
### Summary

`Backend` is not optional and there is already an error raised if no backend is passed in so we should be able to just remove the incorrect type hint. 

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/batch.py
+++ b/qiskit_ibm_runtime/batch.py
@@ -86,7 +86,7 @@ class Batch(Session):
 
     def __init__(
         self,
-        backend: Optional[BackendV2] = None,
+        backend: BackendV2,
         max_time: Optional[Union[int, str]] = None,
     ):
         """Batch constructor.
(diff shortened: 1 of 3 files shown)


ElePT (commented):
The change looks good! I added the backport potential label, and it made me realize that we hadn't discussed backports yet. I have a few questions:
1. Is the process automatized or manual?
2. What are curently the criteria to decide whether a bug fix should get backported?
3. When does the package publish patch releases? (is there a regular schedule or is it rather need based)

kt474:
> 1. Is the process automatized or manual?

Currently it's a manual process (maybe worth looking into mergify)

> 2. What are curently the criteria to decide whether a bug fix should get backported?
> 3. When does the package publish patch releases? (is there a regular schedule or is it rather need based)

We don't have strict criteria for patch releases, it's mostly need based. It's worth discussing though, in the future we should have a more standardized process.

PR #2225, opened 2025-04-16
Update iqp warning text and 0.38 release note 

kt474:
### Summary

`local` is also a valid channel 

### Details and comments
Fixes #


ElePT (commented):
LGTM. The release note also mentioned that ibm_cloud was the only valid channel, I think we should update that too.

PR #2226, opened 2025-04-16
Raise error in `from_id` if the session has no backend

kt474:
### Details and comments
Fixes #


kt474:
> Based off the PR removing the `Optional` type hint from `Session`, it seems clear that sessions should always have backends right? However if a user creates a session and doesn't send any jobs the backend doesn't get set. The error seems fine but I don't think this behavior is very intuitive. I wonder if there is a way to set the backend in the session from the moment it's created without requiring jobs?

Yeah I agree it's not intuitive - it's been along ongoing discussion see `ntc 5063` for more context

PR #2235, opened 2025-04-24
Handle 0.37 deprecations

kt474:
### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/fake_provider/fake_backend.py
+++ b/qiskit_ibm_runtime/fake_provider/fake_backend.py
@@ -278,20 +278,13 @@ class FakeBackendV2(BackendV2):
         return self._target
 
     @property
-    def max_circuits(self) -> int:
-        """(DEPRECATED) The maximum number of circuits
-
-        The maximum number of circuits that the backend supports in a single execution.
-        Note that the actual number of circuits the service allows may be different.
+    def max_circuits(self) -> None:
+        """This property used to return the `max_experiments` value from the
+        backend configuration but this value is no longer an accurate representation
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -278,20 +278,13 @@ def target(self) -> Target:
...
         """
 
-        issue_deprecation_msg(
-            "max_circuits is deprecated",
-            "0.37.0",
-            "Please see our documentation on job limits "
-            "https://quantum.cloud.ibm.com/docs/guides/job-limits#job-limits.",
-        )
-        return self.configuration().max_experiments
+        return None
```
ElePT, line 287: What is the motivation to keep the property? Is it to replace this in the future with some other form of calculating max circuits?

ElePT (commented):
Mostly LGTM. I just left one comment, and we should add a removal release note with these kind of PRs to make sure the changes are easy to track.

qiskit_ibm_runtime/fake_provider/fake_backend.py
kt474: `max_circuits` is an abstract method in the [BackendV2](https://github.com/Qiskit/qiskit/blob/main/qiskit/providers/backend.py#L203) base class so I believe it was decided during deprecation that the method would just return `None`.

ElePT: Oh true, I had forgotten about it. It makes sense then, there aren't many other options.

PR #2239, opened 2025-04-25
WIP Add new channel and allow instance selection with new platform 

kt474:
### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/accounts/account.py
+++ b/qiskit_ibm_runtime/accounts/account.py
@@ -26,7 +26,14 @@ from ..api.auth import QuantumAuth, CloudAuth
 from ..utils import resolve_crn
 
 AccountType = Optional[Literal["cloud", "legacy"]]
-ChannelType = Optional[Literal["ibm_cloud", "ibm_quantum", "local"]]
+ChannelType = Optional[
+    Literal[
+        "ibm_quantum_platform",
+        "ibm_cloud",
+        "ibm_quantum",
+        "local",
+    ]
+]
 
 IBM_QUANTUM_API_URL = "https://auth.quantum.ibm.com/api"
 IBM_CLOUD_API_URL = "https://cloud.ibm.com"
@@ -46,7 +53,7 @@ class Account:
         """Account constructor.
 
         Args:
-            channel: Channel type, ``ibm_cloud`` or ``ibm_quantum``.
+            channel: Channel type, ``ibm_cloud``, ``ibm_quantum``, ``ibm_quantum_platform``.
             token: Account token to use.
             url: Authentication URL.
             instance: Service instance to use.
@@ -111,7 +118,7 @@ class Account:
                 proxies=proxies,
                 verify=verify,
             )
-        elif channel == "ibm_cloud":
+        elif channel in ["ibm_cloud", "ibm_quantum_platform"]:
             return CloudAccount(
                 url=url,
                 token=token,
@@ -123,7 +130,7 @@ class Account:
         else:
             raise InvalidAccountError(
                 f"Invalid `channel` value. Expected one of "
-                f"{['ibm_cloud', 'ibm_quantum']}, got '{channel}'."
+                f"{['ibm_cloud', 'ibm_quantum', 'ibm_quantum_platform']}, got '{channel}'."
             )
 
     def resolve_crn(self) -> None:
@@ -166,10 +173,10 @@ class Account:
     @staticmethod
     def _assert_valid_channel(channel: ChannelType) -> None:
         """Assert that the channel parameter is valid."""
-        if not (channel in ["ibm_cloud", "ibm_quantum"]):
+        if not (channel in ["ibm_cloud", "ibm_quantum", "ibm_quantum_platform"]):
             raise InvalidAccountError(
                 f"Invalid `channel` value. Expected one of "
-                f"['ibm_cloud', 'ibm_quantum'], got '{channel}'."
+                f"['ibm_cloud', 'ibm_quantum', 'ibm_quantum_platform], got '{channel}'."
             )
 
     @staticmethod
@@ -266,7 +273,7 @@ class CloudAccount(Account):
         """
         super().__init__(token, instance, proxies, verify)
         resolved_url = url or IBM_CLOUD_API_URL
-        self.channel = "ibm_cloud"
+        self.channel = "ibm_quantum_platform"
         self.url = resolved_url
         self.private_endpoint = private_endpoint
 
@@ -284,7 +291,7 @@ class CloudAccount(Account):
             CloudResourceNameResolutionError: if CRN value cannot be resolved.
         """
         crn = resolve_crn(
-            channel="ibm_cloud",
+            channel="ibm_quantum_platform",
             url=self.url,
             token=self.token,
             instance=self.instance,
--- a/qiskit_ibm_runtime/base_runtime_job.py
+++ b/qiskit_ibm_runtime/base_runtime_job.py
@@ -371,7 +371,7 @@ class BaseRuntimeJob(ABC):
     @property
     def instance(self) -> Optional[str]:
         """For ibm_quantum channel jobs, return the instance where the job was run.
-        For ibm_cloud, `None` is returned.
+        For ibm_quantum_platform, `None` is returned.
         """
         return self._backend._instance
 
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -84,30 +84,30 @@ class QiskitRuntimeService:
             - Account defined by the input `channel` and `token`, if specified.
             - Account defined by the `default_channel` if defined in filename
             - Account defined by the environment variables, if defined.
-            - Default account for the ``ibm_cloud`` account, if one is available.
+            - Default account for the ``ibm_quantum_platform`` account, if one is available.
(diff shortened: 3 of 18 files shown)


qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -371,7 +371,7 @@ def usage_estimation(self) -> Dict[str, Any]:
     @property
     def instance(self) -> Optional[str]:
         """For ibm_quantum channel jobs, return the instance where the job was run.
-        For ibm_cloud, `None` is returned.
+        For ibm_quantum_platform, `None` is returned.
```
ElePT, line 374: The behavior for `ibm_cloud` should still be documented until the option is removed.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -448,8 +504,8 @@ def backends(
         Args:
             name: Backend name to filter by.
             min_num_qubits: Minimum number of qubits the backend has to have.
-            instance: This is only supported for ``ibm_quantum`` runtime and is in the
-                hub/group/project format.
+            instance: In hub/group/project format if on the ``ibm_quantum`` channel.
+                IBM Cloud account crn if on the ``ibm_quantum_platform`` channel.
```
ElePT, line 508: Same comment on still documenting "ibm_cloud" if it is still a valid channel, which it is until deprecation (then during the deprecation PR we can mark it as deprecated in the docstrings, but not remove the info until the feature is removed).

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -521,10 +577,17 @@ def backends(
...
+                # reset api client
+                self._client_params.instance = self._account.instance
+                self._api_client = RuntimeClient(self._client_params)
+            # do we want backends from all instances in service.backends()?
+            if not self._all_instances:
+                # fetch all instances and all backends in each instance. Save in dict
+                self._all_instances = self._account.list_instances()  # type: ignore[attr-defined]
+                for inst in self._all_instances:
+                    self._discover_backends_from_instance(inst)
+
```
ElePT, line 590: I think that it would make sense to only fetch backends from all instances if no `instance` or `default_instance` have been set.  The code using `list_instances()` doesn't seem to be doing this as far as I can see. Iterating over instances will make fetching backends much slower so users should have a way of calling `backends()` without this additional overhead.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -533,6 +596,37 @@ def backends(
...
+                        # if no default instance set, check all instances and
+                        # fetch one that has the backend
+                        instance_with_backend = self._get_instance_from_grouping(name)
+
+                if instance_with_backend:
+                    # TODO how do we handle this logic here?
+                    # fetching a backend not in the current account instance requires a
+                    # different api client - this api client is tied to everything
+                    self._client_params.instance = instance_with_backend
+                    self._api_client = RuntimeClient(self._client_params)
```
ElePT, line 621: This code seems to be overwriting the api client. How would it behave if a user let's say instantiates the service, asks for backend 1 in instance A, asks for a backend 2 in instance B and decides to start a session with backend 1? Would this work?

ElePT (commented):
Thanks @kt474, I took an initial look at the PR and I see that some details need to be ironed out and in general is a tricky feature to develop. I left a few comments here and there (some really minor), but I think that my main points are:

1. The documentation of the "ibm_cloud" channel should stay while the option is supported, even if it's not the recommended one. We should extend the documentation with the new "ibm_quantum_platform" channel, make sure the new name is used in the examples and deprecation message, and mention that it's the new default and follows the same configuration path as "ibm_cloud". After the alternative is in place, we can deprecate "ibm_cloud" explicitly. The new name is too close to "ibm_quantum" and many users will naively assume that it is replacing it making the migration even more complex.

2. Regarding your 2 questions from the PR description:

> - Should service.backends() return all backends available through all instances?
> - Fetching all instances then fetching backends for every instance is very slow (ntc 5779 would help)

My take is that with any change that we make that adds API calls, we should make sure users are not hit with a performance regression if they aren't getting anything out of it (so if you really need this functionality, sure, you might be willing to sacrifice speed, but some users won't need it). So I would only really fetch all instances if no `instance` or `default_instance` is set, and would only iterate over all instances and backends in that case. Also brainstorming here, this might be a good opportunity to try async API calls, which might mitigate the performance hit a bit?

Sorry if this review is superficial, I will try to take a deeper look tomorrow.

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -46,7 +59,7 @@ def __init__(
         """Account constructor.
 
         Args:
-            channel: Channel type, ``ibm_cloud`` or ``ibm_quantum``.
+            channel: Channel type, ``ibm_cloud``, ``ibm_quantum``, ``ibm_quantum_platform``.
```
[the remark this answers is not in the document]
kt474: There is no "account" when the local channel is used so I think it's ok to leave out here.

qiskit_ibm_runtime/qiskit_runtime_service.py
kt474: That's a good point, I'l have to test that edge case - I don't think that would work. I can't think of any other way to handle this since the backend is ultimately always tied to its instance.

ElePT:
Sharing some conclusions from our internal discussion on how to advance with this PR:

1. We will not add `default_instance`, and use `instance` instead. We will document the parameter in `save_account` as: _“This is an optional parameter. If set, it will define a default instance for service instantiation, if not set, the service will fetch all instances related to a specific token”_ (or something along these lines). We will add this change in behavior to the reno, mentioning that it’s now optional and that not setting it gives access to all instances. We will also let the docs team know about the change.
2. We will add the following new optional parameters to both `save_account`  and `__init__` : `account` , `region` , `plans_preference`. Where if the value is set in init it will overwrite the default from `save_account`.
3. If no instance is set during init, to do authentication, we can call IBM Cloud platform APIs to get all the service instances the user has access to

ElePT (commented):
Sharing some conclusions from our internal discussion on how to advance with this PR:

1. We will not add `default_instance`, and use `instance` instead. We will document the parameter in `save_account` as: _“This is an optional parameter. If set, it will define a default instance for service instantiation, if not set, the service will fetch all instances related to a specific token”_ (or something along these lines). We will add this change in behavior to the reno, mentioning that it’s now optional and that not setting it gives access to all instances. We will also let the docs team know about the change.
2. We will add the following new optional parameters to both `save_account`  and `__init__` : `account` , `region` , `plans_preference`. Where if the value is set in init it will overwrite the default from `save_account`.
3. If no instance is set during init, to do authentication, we can call IBM Cloud platform APIs to get all the service instances the user has access to

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1183,7 +1330,7 @@ def instances(self) -> List[str]:
         """
         if self._channel == "ibm_quantum":
             return list(self._hgps.keys())
-        return []
+        return [t["crn"] for t in self._all_instances]
```
ElePT, line 1333: This line returns an empty list for the case where no instance is given, but I would expect it to list all available instances, as when calling `list_instances()`

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -183,6 +206,42 @@ def __init__(
...
+        # refactor this, is there an easier way to fetch all backends
+        # ntc 5779 would make things a lot faster - get list of backends
+        # from global search API call
+
+        # I don't think we should overwrite the api client like this
+        # but we can't access the backend config without the right instance
+        if self._account.instance != instance:
+            self._client_params.instance = instance
+            self._api_client = RuntimeClient(self._client_params)
+        return self._api_client.list_backends()
```
ElePT, line 221: You may have come up with a better solution to avoid overwriting the client, but in case it helps: I think that you could refactor the client_params object and/or the `RuntimeClient` object (I would have to look into the specifics) to be able to store more than 1 instance in the cloud case. This would allow to keep handles to all instances a user might need instead of imposing one instance at a time. They could be stored together with the backend name, so the client knows which instance to use to access a specific backend.

mtreinish:
So I'm playing with this locally and some observations, right now if you do:

```python
QiskitRuntimeService.save_account(channel="ibm_quantum_platform", token="...")
```

it saves an `ibm_cloud` channel account in the json file. Then when you instantiate the service from the saved credentials it loads up as a `ibm_cloud` channel account and not `ibm_quantum_platform`. This causes a warning to be printed when you load the account because there is no instance set `QiskitRuntimeService()`. I kind of expected this to just all work by default and show all the instances I have since I didn't set a specific one.

kt474:
> it saves an `ibm_cloud` channel account in the json file. Then when you instantiate the service from the saved credentials it loads up as a `ibm_cloud` channel account and not `ibm_quantum_platform`. This causes a warning to be printed when you load the account because there is no instance set `QiskitRuntimeService()`. I kind of expected this to just all work by default and show all the instances I have since I didn't set a specific one.

Fixed the issue where `ibm_cloud` was being saved instead of `ibm_quantum_platform` .  Currently we raise a warning at service initialization because one api can token can have access to multiple accounts. The instances aren't needed until the first `backends()` call.

ElePT (approved):

PR #2242, opened 2025-04-29
Update `test_session_from_id`

kt474:
### Summary

Missed this - on the cloud channel, a job has to finish before the backend is returned in `session.details()` 

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_session.py
+++ b/test/integration/test_session.py
@@ -83,7 +83,8 @@ class TestIntegrationSession(IBMIntegrationTestCase):
         isa_circuit = pm.run([bell()])
         with Session(backend=backend) as session:
             sampler = SamplerV2(mode=session)
-            sampler.run(isa_circuit)
+            job = sampler.run(isa_circuit)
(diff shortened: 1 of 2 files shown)


test/integration/test_session.py
```diff
@@ -83,7 +83,8 @@ def test_session_from_id(self, service):
         isa_circuit = pm.run([bell()])
         with Session(backend=backend) as session:
             sampler = SamplerV2(mode=session)
-            sampler.run(isa_circuit)
+            job = sampler.run(isa_circuit)
+            job.result()
```
ElePT, line 87: Might be missing function, but where is `session.details()` called in this test? is it called in `mock_create_session.assert_not_called()`?

kt474: Sorry my comment was lacking context - details() is called in from_id() - https://github.com/kt474/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/session.py#L337

PR #2246, opened 2025-05-02
Retry on 503

### LINKED ISSUE #2244: Add retries back on 503s
Retries on 503s were removed as a part of https://github.com/Qiskit/qiskit-ibm-runtime/issues/2099 because we wanted to make sure a proper error message was shown where there was a maintenance outage. This was not the best approach because we still want to keep these retries for other network outages (Cloudflare). 

When there is a maintenance outage we can go through the retry logic first and then show the proper error message at the end.

kt474:
### Summary

Should add a test 

### Details and comments
Fixes #2244

### DIFF
--- a/qiskit_ibm_runtime/api/session.py
+++ b/qiskit_ibm_runtime/api/session.py
@@ -35,6 +35,7 @@ from ..version import __version__ as ibm_runtime_version
 STATUS_FORCELIST = (
     500,  # General server error
     502,  # Bad Gateway
+    503,  # Service Unavailable
     504,  # Gateway Timeout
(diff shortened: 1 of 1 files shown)


ElePT (commented):
After reviewing the issue, the PR doesn't seem to be addressing:

> When there is a maintenance outage we can go through the retry logic first and then show the proper error message at the end.

How can a maintenance outage be detected in this case?

ElePT (approved):
Alright, in that case we can probably merge this without mentioning it in the changelog. We just need to update the branch.

Issue #2188, opened 2025-03-19
`Session.close` and `Batch.close` do not close the session and batch on the new IQP


t-imamichi:
**Describe the bug**
`Session.close` and `Batch.close` do not close the session and the batch.

**Steps to reproduce**
(update) I made a reproducible script https://github.com/Qiskit/qiskit-ibm-runtime/issues/2188#issuecomment-2742082563

```
In [10]: from qiskit_ibm_runtime import QiskitRuntimeService, Batch

In [11]: service = QiskitRuntimeService(name="...")  # new IQP

In [12]: session = Batch.from_id('...', service)

In [13]: session.status()
Out[13]: 'Pending'

In [14]: session.close()  # There is no running job

In [15]: session.status()
Out[15]: 'Pending'   # expected to be "Closed"
```

**Expected behavior**
The status become `Closed` as same as the conventional IQP.
I tried the same script on the conventional IQP and the status becomes `Closed` after `close()`.

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.37.0
- **qiskit version**: 1.4.1
- **Python version**: 3.12.9
- **Operating system**: macOS 15.3.2

kt474:
My understanding is that `Session.close()` calls this [endpoint](https://us-east.quantum-computing.cloud.ibm.com/openapi/#/Sessions/update_session_state) which just sets the `accepting_jobs` flag to false. So if there are still are pending jobs in the session, the status will be `Pending`. Only once there are no more pending jobs will the status update to `Closed`.

Getting this for both IQP and the new IQP on cloud 

<img width="998" alt="Image" src="https://github.com/user-attachments/assets/14c684ad-30f3-4145-a2dc-4db9492fe87a" />

t-imamichi:
I made a script to reproduce the behavior. I confirmed that the behavior on new IQP is different from that on the conventional IQP.
The last line of the new IQP is expected to be "Closed" as same as the conventional IQP.

```python
from time import sleep

from qiskit_ibm_runtime import Batch, QiskitRuntimeService, __version__

print(__version__)

for name, iqp in [
    ("primitives", "conventional IQP"),
    ("notebooks-internal", "new IQP"),
]:
    print(iqp)
    service = QiskitRuntimeService(name=name)
    backend = service.backend("test_eagle_us-east")
    batch = Batch(backend)
    sleep(1)
    print(batch.status())
    print("call close()")
    batch.close()
[...]

PR #2247, opened 2025-05-02
Integration test updates

kt474:
### Summary

`test_session_from_id`: This test has been updated multiple times recently because session behavior is different on classic IQP and new IQP (https://github.com/Qiskit/qiskit-ibm-runtime/issues/2188). On the new IQP, one job has to complete first before the backend can be returned in `session.details()`. Also the session does not close right away when `close()` is called. 

After https://github.com/Qiskit/qiskit-ibm-runtime/pull/2102, `test_backend_deepcopy` started failing because of the use of `IAMTokenManager` in the Cloud Auth init function. 

I'll test that the integration tests are passing in my fork 

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_session.py
+++ b/test/integration/test_session.py
@@ -91,9 +91,9 @@ class TestIntegrationSession(IBMIntegrationTestCase):
             mock_create_session.assert_not_called()
 
         self.assertEqual(session._session_id, new_session._session_id)
-        self.assertTrue(new_session._active)
         new_session.close()
         self.assertFalse(new_session._active)
+        self.assertFalse(new_session.details()["accepting_jobs"])
 
         with self.assertRaises(IBMInputValueError):
             Batch.from_id(session_id=session._session_id, service=service)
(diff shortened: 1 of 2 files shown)


test/integration/test_session.py
```diff
@@ -91,9 +91,9 @@ def test_session_from_id(self, service):
             mock_create_session.assert_not_called()
 
         self.assertEqual(session._session_id, new_session._session_id)
-        self.assertTrue(new_session._active)
```
ElePT, line 94: This assertion should hold in the new IQP too, right? the session should be active before calling `.close()`? Or has the behavior of `._active` changed?

kt474: Yeah this holds for the new IQP but fails on IQP classic because once the job completes the session closes right away. 

Not sure if it's some sort of timing discrepancy but this is the current behavior ([from_id code](https://github.com/kt474/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/session.py#L353)): 

Old IQP - job in session completes -> `from_id()` new session -> new session has "state" of `closed` so `._active` is false

New IQP - job in session completes -> `from_id()` new session -> new session has "state" of `active` so `._active` is true

ElePT (approved):
LGTM

PR #2254, opened 2025-05-07
Added a check to verify that observables don't contain projections

### LINKED ISSUE #2250: Validate observables to include only the letters I, X, Y, Z
Currently there is no such check, and the failure is internal. For example,
```python
from qiskit_ibm_runtime import QiskitRuntimeService, EstimatorV2


backend = service.backend("your_favorite_backend")
estimator = EstimatorV2(mode=backend)

circ = QuantumCircuit(1)
job = estimator.run([(circ, "0")])
print(job.job_id())
res = job.result()
```

This is what happens when calling `result()`:
```
qiskit_ibm_runtime.exceptions.RuntimeJobFailureError: 'Unable to retrieve job result. QiskitError: \'Pauli string label "0" is not valid.'
```

The error message is clear, but the failure does not occur at the client. The users have to wait in the queue, and I don't know if they are charged.

We have to validate already in qiskit-ibm-runtime, probably by updating the function `validate_estimator_pubs`.

yaelbh:
### Summary

Fixes #2250

### Details and comments

- This PR's code runs successfully with both Qiskit's main branch and stable branch.
- I changed coefficients in the test to real numbers, because complex coefficients are not allowed, and trigger an exception.
- Similarly, I changed the invalid letters in the test from `J` to `0, 1, +, -, l, r`, because `J` is caught already early, by `SparseObservable`. The goal of the test is to verify that qiskit-ibm-runtime imposes further constraints than `SparseObservable`.


release-notes/unreleased/2250.other.rst
```diff
@@ -0,0 +1 @@
+`IBMInputValueError` inherits from `ValueError`, in addition to the existing parent class `IBMError`. Some validation functions, which previously raised `ValueError` exceptions, now raise `IBMInputValueError` exceptions. We added a validation that observables consist only of Pauli operations, residing in the alphabet {I, X, Y, Z}.
```
ElePT, line 1: Here are some minor formatting suggestions. I would add a feature reno for the new validation too!

```suggestion
:class:`.IBMInputValueError` now inherits from ``ValueError``, in addition to the existing parent class :class:`.IBMError`. Some validation functions, which previously raised ``ValueError`` exceptions, now raise :class:`.IBMInputValueError` exceptions. We added a validation that observables consist only of Pauli operations, residing in the alphabet ``{I, X, Y, Z}``.
```

ElePT (commented):
Thanks @yaelbh, the PR looks good to me, I just left some minor release note suggestions and it should be good to go. I will try to include this in the upcoming 0.39 release to get the new validation out ASAP :)

release-notes/unreleased/2250.other.rst
yaelbh: Thanks @ElePT for the review. Do you suggest to split to `2250.other.rst` and `2250.feat.rst`, where the last sentence ("We added a validation...") foes to the `feat` file?

ElePT: I suggest splitting it, yes!

PR #2261, opened 2025-05-17
Remove backend defaults check in target()

### LINKED ISSUE #2260: Multiple calls to the API if Pulse is False
**Describe the bug**

From @wshanks: The code that decides whether to request the defaults from the API or not keys off the local defaults being None, but when the backend has pulse disabled the defaults from the API are None so the code will always reach out to the API again on every reference to the target/defaults. Because of this we saw degradation in performance of qiskit-experiments for certain backends. 

```
File ~/miniforge3/envs/qiskit_2_0/lib/python3.10/site-packages/qiskit_ibm_runtime/ibm_backend.py:442, in IBMBackend.defaults(self, refresh)
    428 """Return the pulse defaults for the backend.
    429 
    430 The schema for default pulse configuration can be found in
   (...)
    439     The backend pulse defaults or ``None`` if the backend does not support pulse.
    440 """
    441 if refresh or self._defaults is None:
--> 442     api_defaults = self._api_client.backend_pulse_defaults(self.name)
```

**Steps to reproduce**

See above. I was running 0.37.0, but Doug also reproduced. 

**Expected behavior**

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.37.0
- **Python version**:
- **Operating system**:

kt474:
### Summary

`backend.target()` doesn't use backend defaults anymore.

### Details and comments
Fixes #2260

### DIFF
--- a/qiskit_ibm_runtime/ibm_backend.py
+++ b/qiskit_ibm_runtime/ibm_backend.py
@@ -317,7 +317,6 @@ class IBMBackend(Backend):
             Target
         """
         self.properties()
-        self.defaults()
         self._convert_to_target()
         return self._target
 
--- a/release-notes/unreleased/2261.bug.rst
+++ b/release-notes/unreleased/2261.bug.rst
@@ -1,0 +1,2 @@
+The call to :meth:`~.IBMBackend.defaults` in :meth:`~.IBMBackend.target` was removed
+because backend defaults are no longer used in the target generation.


dtmcclure:
I confirmed this fixes the problem we were seeing -- thanks!

ElePT (approved):
LGTM

PR #2263, opened 2025-05-20
Support private jobs on new platform and add `job.private` property

### LINKED ISSUE #2252: Support private jobs for new IQP
On Runtime, we are adding support for private jobs, so we should add the functionality in `qiskit-ibm-runtime` as well. Private jobs will work the same way as they did for classic IQP, we will add a `private` param of type boolean to the `POST /jobs` endpoint. The `private` attribute will also be included in the endpoint response.

### LINKED ISSUE #2253: Update docs to reflect support for private jobs
We will be adding support for private jobs in Runtime (see issue: #2252). The documentation should be updated to reflect this support: https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/runtime-job-v2

kt474:
### Details and comments
Fixes #2252
Fixes #2253

### DIFF
--- a/qiskit_ibm_runtime/options/environment_options.py
+++ b/qiskit_ibm_runtime/options/environment_options.py
@@ -55,11 +55,9 @@ class EnvironmentOptions:
     r"""Boolean that indicates whether the job is marked as private. When set to true, 
         input parameters are not returned, and the results can only be read once. 
         After the job is completed, input parameters are deleted from the service. 
-        After the results are read, they are deleted from the service. 
+        After the results are read, these are also deleted from the service. 
         When set to false, the input parameters and results follow the 
         standard retention behavior of the API.
-        Only returned in the job response if the value is true, 
-        otherwise it is omitted.
 
         Default: False.
     """
--- a/qiskit_ibm_runtime/runtime_options.py
+++ b/qiskit_ibm_runtime/runtime_options.py
@@ -73,11 +73,9 @@ class RuntimeOptions:
             private: Boolean that indicates whether the job is marked as private. When set to true,
                 input parameters are not returned, and the results can only be read once.
                 After the job is completed, input parameters are deleted from the service.
-                After the results are read, they are deleted from the service.
+                After the results are read, these are also deleted from the service.
                 When set to false, the input parameters and results follow the
                 standard retention behavior of the API.
-                Only returned in the job response if the value is true,
-                otherwise it is omitted.
         """
         self.backend = backend
         self.image = image
--- a/release-notes/unreleased/2263.feat.rst
+++ b/release-notes/unreleased/2263.feat.rst
@@ -1,5 +1,5 @@
 The ``private`` option under :class:`EnvironmentOptions` is now supported on the 
-``ibm_cloud`` channel (new IBM Quantum Platform). When this option
+``ibm_cloud`` and ``ibm_quantum_platform`` channels (new IBM Quantum Platform). When this option
 is set to ``True``, the job will be returned without params and results can only
 be retrieved once. 
 


qiskit_ibm_runtime/options/environment_options.py
```diff
@@ -52,12 +52,11 @@ class EnvironmentOptions:
         Default: ``None``.
     """
     private: Optional[bool] = False
-    r"""Boolean that indicates whether the job is marked as private. This is only
```
mriedem, line 55: Removing `This is only supported for ``ibm_quantum`` channel.` is premature until the POST /jobs API supports the private field.

The changes on the server side are being worked such that POST /jobs support is the last thing to be supported and that enables the rest of the feature in the other GET APIs.

qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -116,6 +119,11 @@ def __init__(
                 "results streaming was removed in a previous release.",
             )
 
+    @property
+    def private(self) -> str:
```
mriedem, line 123: >-> str

Should that be bool?

qiskit_ibm_runtime/options/environment_options.py
```diff
@@ -52,12 +52,11 @@ class EnvironmentOptions:
...
-        supported for ``ibm_quantum`` channel. When set to true, input parameters are not
-        returned, and the results can only be read once. After the results are read or after
-        a specified time after the job is completed, the results are deleted from the service.
-        When set to false, the input parameters and results follow the standard retention
-        behavior.
+    r"""Boolean that indicates whether the job is marked as private. When set to true, 
+        input parameters are not returned, and the results can only be read once. 
+        After the results are read or after a specified time after the job is completed, the 
+        results are deleted from the service. When set to false, the input parameters and results 
+        follow the standard retention behavior.
```
mriedem, line 59: >or after a specified time after the job is completed

nit, but for the cloud channel I don't think this is true. I'm not sure if it's ever been true for the quantum channel. Might need to ask Ismael.

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -104,6 +106,7 @@ def __init__(
             session_id=session_id,
             tags=tags,
             version=version,
+            private=private,
```
mriedem, line 109: Where are you ever reading this from a response and populating it?

mriedem (commented):
Looks like there are some issues with this....

qiskit_ibm_runtime/runtime_job_v2.py
ElePT: I think only maintainers can. I can resolve it.

qiskit_ibm_runtime/base_runtime_job.py
ElePT:
```suggestion
    def private(self) -> bool:
```

qiskit_ibm_runtime/options/environment_options.py
```diff
@@ -52,12 +52,14 @@ class EnvironmentOptions:
...
-        When set to false, the input parameters and results follow the standard retention
-        behavior.
+    r"""Boolean that indicates whether the job is marked as private. When set to true, 
+        input parameters are not returned, and the results can only be read once. 
+        After the job is completed, input parameters are deleted from the service. 
+        After the results are read, they are deleted from the service. 
+        When set to false, the input parameters and results follow the 
+        standard retention behavior of the API.
+        Only returned in the job response if the value is true, 
+        otherwise it is omitted.
```
ElePT, line 62: Is this what the last sentence meant? I am not sure what is "only returned", the value of the flag?

```suggestion
        When set to false, the input parameters and results follow the 
        standard retention behavior of the API.
        This parameter is only returned in the job response if the value is true, 
        otherwise it is omitted.
```

qiskit_ibm_runtime/runtime_options.py
```diff
@@ -70,12 +70,14 @@ def __init__(
...
-            private: Boolean that indicates whether the job is marked as private. This is only
-                supported for ``ibm_quantum`` channel. When set to true, input parameters are not
-                returned, and the results can only be read once. After the results are read or after
-                a specified time after the job is completed, the results are deleted from the service.
-                When set to false, the input parameters and results follow the standard retention
-                behavior.
+            private: Boolean that indicates whether the job is marked as private. When set to true,
+                input parameters are not returned, and the results can only be read once.
+                After the job is completed, input parameters are deleted from the service.
+                After the results are read, they are deleted from the service.
```
ElePT, line 76:
```suggestion
                After the results are read, these are also deleted from the service.
```

qiskit_ibm_runtime/options/environment_options.py
ElePT: I actually tested the PR and I don't think it's the parameter being returned:

`PrimitiveResult([SamplerPubResult(data=DataBin(c=BitArray(<shape=(), num_shots=4096, num_bits=4>)), metadata={'circuit_metadata': {}})], metadata={'execution': {'execution_spans': ExecutionSpans([DoubleSliceSpan(<start='2025-05-27 15:55:46', stop='2025-05-27 15:55:46', size=4096>)])}, 'version': 2})`

release-notes/unreleased/2263.feat.rst
```diff
@@ -0,0 +1,7 @@
+The ``private`` option under :class:`EnvironmentOptions` is now supported on the 
+``ibm_cloud`` channel (new IBM Quantum Platform). When this option
```
ElePT, line 2: Once the new channel is introduced, this reno should also talk about "ibm_quantum_platform"

ElePT (commented):
The change looks good to me, I tested it locally and was able to use the `private` flag on production too. My only question is how do we want to time this PR with #2239, to make sure these changes also work with the new "ibm_quantum_platform" channel.

qiskit_ibm_runtime/options/environment_options.py
kt474: Clarified the docs and removed this last sentence in the latest commit since the wording is confusing/misleading. The `private` attribute is not returned in the api job response when it is not set but from the provider, `job.private` will be false.

ElePT (approved):
LGTM

PR #2264, opened 2025-05-21
Update links pointing to IQP Classic

arnaucasau:
This PR updates the links to IQP Classic in the whole repo because the documentation source of truth is moving to https://quantum.cloud.ibm.com/.

This is the only links we can't change yet because the new platform doesn't have announcements.

https://github.com/Qiskit/qiskit-ibm-runtime/blob/ca1b7c283ea5bc58402d3da0213692dd49de9685/qiskit_ibm_runtime/api/session.py#L354

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ pip install -e ".[dev]"
 
 ### Open an issue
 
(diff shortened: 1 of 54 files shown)


ElePT (approved):
LGTM. Thanks a lot for the update.

PR #2269, opened 2025-05-28
IQP migration followup 

### LINKED ISSUE #2271: Allow getting a Job by ID without specifying an instance

Getting a Job by specifying only it's Job ID.

Within the new IBM Quantum Platform ([quantum.cloud.ibm.com](https://quantum.cloud.ibm.com)), it's possible to search for a Job by ID specifying multiple Instances CRNs calling the [/v1/workloads](https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/workloads) method and passing a `search` parameter with the ID of the Job..

> [!NOTE]  
> To specify multiple CRNs for this method, use the `Service-CRN` header separating each CRN with a comma `,`

If the user did not specify an instance when initializing the QiskitRuntimeService, the package should send all the CRNs that the user has access to in the default account of the API Key used to connect to the service. That way the Job can be found within the account instances the user has access to.

The result would include the instance CRN, in case that's needed to query other API methods to retrieve additional data of the Job which may not support specifying multiple CRNs.

**Acceptance criteria**

- [ ] Can retrieve a Job by ID without specifying the CRN or name of the instance the job is part of.

### LINKED ISSUE #2272: Fail to get a backend with fractional gate with qiskit-ibm-runtime 0.40.0
**Describe the bug**

qiskit-ibm-runtime 0.40.0 fails to get a backend with fractional gates in the following example.

**Steps to reproduce**
```python
from qiskit_ibm_runtime import __version__ as runtime_version

print(f"{runtime_version=}")

service = QiskitRuntimeService(name="demo-premium")
backend1 = service.backend("ibm_fez", use_fractional_gates=False)
backend2 = service.backend("ibm_fez", use_fractional_gates=True)
print(backend1.basis_gates)
print(backend2.basis_gates)
```

```
runtime_version='0.40.0'
['cz', 'id', 'rz', 'sx', 'x']
['cz', 'id', 'rz', 'sx', 'x']
```

**Expected behavior**

```
runtime_version='0.39.0'
['cz', 'id', 'rz', 'sx', 'x']
['cz', 'id', 'rx', 'rz', 'rzz', 'sx', 'x']
```

**Suggested solutions**

**Additional Information**

- **Python version**:
- **Operating system**:

kt474:
### Summary

- If there is no instance passed in at initialization, the service cannot make any api calls unless `backend()` or `backends()` is called first. In this case we need to load all instances during initialization. 
- Fixed a bug where if `use_fractional_gates` is set but the backend configuration was already cached, the incorrect configuration could be returned. 

### Details and comments
Fixes #2272 
Fixes #2271

### DIFF
--- a/test/integration/test_backend.py
+++ b/test/integration/test_backend.py
@@ -26,7 +26,7 @@ from qiskit_ibm_runtime import QiskitRuntimeService, SamplerV2 as Sampler
 
 from ..ibm_test_case import IBMIntegrationTestCase
 from ..decorators import run_integration_test, production_only, quantum_only
-from ..utils import bell, get_real_device
+from ..utils import bell
 
 
 class TestIntegrationBackend(IBMIntegrationTestCase):
@@ -324,8 +324,11 @@ class TestIBMBackend(IBMIntegrationTestCase):
 
     def test_use_fractional_gates_flag(self):
         """Test use_fractional_gates returns correct backend config."""
-        real_device_name = get_real_device(self.service)
-        real_device_no_fg = self.service.backend(real_device_name, use_fractional_gates=False)
-        real_device_fg = self.service.backend(real_device_name, use_fractional_gates=True)
+        try:
+            real_device_name = "alt_fez"
+            real_device_no_fg = self.service.backend(real_device_name, use_fractional_gates=False)
+            real_device_fg = self.service.backend(real_device_name, use_fractional_gates=True)
+        except QiskitBackendNotFoundError:
+            self.skipTest("Real backend not available.")
         self.assertIn("rzz", real_device_fg.basis_gates)
-        self.assertNotIn("rzz", real_device_no_fg)
+        self.assertNotIn("rzz", real_device_no_fg.basis_gates)


qiskit_ibm_runtime/utils/utils.py
```diff
@@ -342,12 +342,10 @@ def default_runtime_url_resolver(
...
-            if region == "eu-de":
-                api_host = (
-                    f"{parsed_url.scheme}://{region}" f".quantum.{parsed_url.hostname}/api/v1"
-                )
-            else:
-                api_host = f"{parsed_url.scheme}://" f"quantum.{parsed_url.hostname}/api/v1"
+            region_prefix = "" if region == "us-east" else f"{region}."
+            api_host = (
+                f"{parsed_url.scheme}://{region_prefix}" f"quantum.{parsed_url.hostname}/api/v1"
+            )
```
kt474, line 348: Special url handling should be done for `us-east` as opposed to for `eu-de` because there can be other non default urls that need to be appended.

PR #2274, opened 2025-06-01
Fix failing new IQP tests

kt474:
### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_account.py
+++ b/test/integration/test_account.py
@@ -93,6 +93,7 @@ class TestQuantumPlatform(IBMIntegrationTestCase):
                 token=self.dependencies.token,
                 instance="test_name",
                 channel="ibm_quantum_platform",
+                url=self.dependencies.url
             )
 
         service_instance_name = _get_service_instance_name_for_crn(self.dependencies)
@@ -100,11 +101,13 @@ class TestQuantumPlatform(IBMIntegrationTestCase):
             token=self.dependencies.token,
             instance=service_instance_name,
             channel="ibm_quantum_platform",
+            url=self.dependencies.url
         )
         self.assertEqual(service._account.instance, self.dependencies.instance)
         service_no_instance = QiskitRuntimeService(
             token=self.dependencies.token,
             channel="ibm_quantum_platform",
+            url=self.dependencies.url
         )
 
         backends = service.backends()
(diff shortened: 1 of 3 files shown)


test/integration/test_job.py
```diff
@@ -102,6 +102,7 @@ def test_cancel_job_done(self, service):
     @run_integration_test
     def test_delete_job(self, service):
         """Test deleting a job."""
+        self.skipTest("Deleting jobs not supported on IBM Quantum Platform.")
         sub_tests = ["DONE"]
```
kt474, line 106: Only instance admins can delete jobs so we'll also have to add a warning/proper error in `service.delete_job()`

test/integration/test_retrieve_job.py
```diff
@@ -80,9 +80,6 @@ def test_retrieve_jobs_limit(self, service):
 
         rjobs = service.jobs(limit=2, program_id=self.program_ids[service.channel])
         self.assertEqual(len(rjobs), 2, f"Retrieved jobs: {[j.job_id() for j in rjobs]}")
-        job_ids = {job.job_id() for job in jobs}
-        rjob_ids = {rjob.job_id() for rjob in rjobs}
-        self.assertTrue(rjob_ids.issubset(job_ids), f"Submitted: {job_ids}, Retrieved: {rjob_ids}")
```
kt474, line 83: This sometimes fails because of a race condition where `rjobs` retrieves other jobs running from other tests. I think we can just remove this assert because it's not actually testing the jobs limit.

test/integration/test_account.py
```diff
@@ -56,7 +56,7 @@ def test_initializing_service_no_instance(self):
         """Test initializing without an instance."""
         self._skip_on_ibm_quantum()
         service = QiskitRuntimeService(
-            token=self.dependencies.token, channel="ibm_quantum_platform"
+            token=self.dependencies.token, channel="ibm_quantum_platform", url=self.dependencies.url
```
kt474, line 59: We need the url for staging

ElePT (commented):
The changes LGTM. Does this fix the integration tests fully or just the new failures?

ElePT (approved):

kt474:
> The changes LGTM. Does this fix the integration tests fully or just the new failures?

This fixes the cloud/new IQP tests fully. I didn't bother with the old IQP because internal backends were supposed to be removed yesterday (delayed to the end of this week)

PR #2276, opened 2025-06-02
Clarify rzz out of range error message

wshanks:
Some users reported that the "has been provided" wording sounded like the backend was providing that angle rather than it having been included in the circuit submitted by the user so here "provided" is changed to "requested" to make it more clear that it was the user's request that led to the error.


kt474 (approved):

ElePT (approved):
LGTM

PR #2277, opened 2025-06-03
Add tags and resource group instance preferences

### LINKED ISSUE #2259: Selection of instances to filter instances by resource group or tags

In addition to the options introduced in:

- https://github.com/Qiskit/qiskit-ibm-runtime/issues/2236

It would be great to also restrict which instances can be selected based on:

- Resource group
- Tags

**Acceptance criteria**

- [ ] I can specify a resource group by name or id and instances would only be selected from it when picking one from a backend's name.
- [ ] I can specify a tag or tags that instances need to have when picking one from a backend's name.

kt474:
### Details and comments
Fixes #2259

### DIFF
--- a/qiskit_ibm_runtime/accounts/account.py
+++ b/qiskit_ibm_runtime/accounts/account.py
@@ -81,6 +81,8 @@ class Account:
         self.private_endpoint: bool = False
         self.region: str = None
         self.plans_preference: List[str] = None
+        self.resource_group: str = None
+        self.tags: List[str] = None
 
     def to_saved_format(self) -> dict:
         """Returns a dictionary that represents how the account is saved on disk."""
@@ -104,6 +106,8 @@ class Account:
         private_endpoint = data.get("private_endpoint", False)
         region = data.get("region")
         plans_preference = data.get("plans_preference")
+        resource_group = data.get("resource_group")
+        tags = data.get("tags")
         return cls.create_account(
             channel=channel,
             url=url,
@@ -114,6 +118,8 @@ class Account:
             private_endpoint=private_endpoint,
             region=region,
             plans_preference=plans_preference,
+            resource_group=resource_group,
+            tags=tags,
         )
 
     @classmethod
@@ -128,6 +134,8 @@ class Account:
         private_endpoint: Optional[bool] = False,
         region: Optional[str] = None,
         plans_preference: Optional[List[str]] = None,
+        resource_group: Optional[str] = None,
+        tags: Optional[List[str]] = None,
     ) -> "Account":
         """Creates an account for a specific channel."""
         if channel == "ibm_quantum":
@@ -149,6 +157,8 @@ class Account:
                 region=region,
                 plans_preference=plans_preference,
                 channel=channel,
+                resource_group=resource_group,
+                tags=tags,
             )
         else:
             raise InvalidAccountError(
@@ -290,6 +300,8 @@ class CloudAccount(Account):
         region: Optional[str] = None,
         plans_preference: Optional[List[str]] = None,
         channel: Optional[str] = "ibm_quantum_platform",
+        resource_group: Optional[str] = None,
+        tags: Optional[str] = None,
     ):
         """Account constructor.
 
@@ -304,6 +316,8 @@ class CloudAccount(Account):
             plans_preference: A list of account types, ordered by preference.
             channel: Channel identifier. Accepted values are ``ibm_cloud`` or ``ibm_quantum_platform``.
                 Defaults to ``ibm_quantum_platform``.
+            resource_group: Resource group preference.
+            tags: List of instance tags.
         """
         super().__init__(token, instance, proxies, verify)
         resolved_url = url or (
@@ -314,6 +328,8 @@ class CloudAccount(Account):
         self.private_endpoint = private_endpoint
         self.region = region
         self.plans_preference = plans_preference
+        self.resource_group = resource_group
+        self.tags = tags
 
     def get_auth_handler(self) -> AuthBase:
         """Returns the Cloud authentication handler."""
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/accounts/account.py
```diff
@@ -304,6 +316,8 @@ def __init__(
             plans_preference: A list of account types, ordered by preference.
             channel: Channel identifier. Accepted values are ``ibm_cloud`` or ``ibm_quantum_platform``.
                 Defaults to ``ibm_quantum_platform``.
+            resource_group: Resource group preference.
```
ElePT: This is a single resource group id, right? not a list

```suggestion
            resource_group: Resource group id in string format.
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -101,7 +103,7 @@ def __init__(
         For the ``"ibm_cloud"`` and ``"ibm_quantum_platform"`` channels it is recommended
         to provide the relevant ``instance`` to minimize API calls. If an ``instance`` is not defined,
         the service will fetch all instances accessible within the account, filtered by
-        ``region`` and ``plans_preference``.
+        ``region``, ``plans_preference``, ``resource_groups``, and ``tags``.
```
ElePT: Again, single group, right?

```suggestion
        ``region``, ``plans_preference``, ``resource_group``, and ``tags``.
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -144,6 +146,10 @@ def __init__(
                 for the ``ibm_cloud`` or ``ibm_quantum_platform`` channel.
                 An instance with the first value in the list will be prioritized if an instance
                 is not passed in.
+            Optional[str] resource_group: Set a Resource Group to filter available instances
```
ElePT:
```suggestion
            Optional[str] resource_group: Set a Resource Group ID to filter available instances
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -936,6 +961,11 @@ def save_account(
                 will be prioritized if an instance is not passed in.
             plans_preference: A list of account types, ordered by preference. An instance with the first
                 value in the list will be prioritized if an instance is not passed in.
+            resource_group: Set a Resource Group to filter available instances.
```
ElePT:
```suggestion
            resource_group: Set a Resource Group ID to filter available instances.
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -936,6 +961,11 @@ def save_account(
...
+                An instance with this resource group will be prioritized if an instance is not passed in.
+            tags: Set a list of tags to filter available instances. An instance with this resource group
+                will be prioritized if an instance is not passed in.
```
ElePT:
```suggestion
            tags: Set a list of tags to filter available instances. Instances with these tags
                will be prioritized if an instance is not passed in.
```

ElePT (commented):
Thanks @kt474! I have a couple of questions after reviewing the PR:

- the "resource_group" preference is set with the resource group ID, right? (I actually don't even know where to look for that)
- what will happen if someone sets a tag or resource group that doesn't exist? Do we know what is the intended behavior?
- the resource group unit test seems to be missing

test/integration/test_account.py
```diff
@@ -133,6 +133,21 @@ def test_account_preferences(self):
...
+        """Test tags account preference."""
+        tags = ["services"]
+        service = QiskitRuntimeService(
+            token=self.dependencies.token,
+            url=self.dependencies.url,
+            channel="ibm_quantum_platform",
+            tags=tags,
+        )
+
+        instances = service.instances()
```
ElePT: The docstring of `service.instances()` doesn't specify that the returned instances are filtered by `tags/region/plans_preference`, but they are, right? In this case the docstring should be updated to reflect the behavior.

test/integration/test_account.py
```diff
@@ -133,6 +133,21 @@ def test_account_preferences(self):
...
+        if instances:
+            for instance in instances:
+                self.assertEqual(instance["tags"], tags)
+
```
ElePT, line 217: Could you add a unit test for the error when no matching instances are found?

ElePT (commented):
The changes look good but I left a couple of questions regarding tests. Ideally, we should be able to mock the tag, plan_preference and region to test the backend retrieval directly, right?

test/integration/test_account.py
kt474: Right now `service.instances()` returns all instances, independent of the filters set. This does seem a bit unintuitive though so we can consider changing the behavior.

kt474:
> The changes look good but I left a couple of questions regarding tests. Ideally, we should be able to mock the tag, plan_preference and region to test the backend retrieval directly, right?

Yeah we should be able to mock everything

ElePT (approved):
LGTM! thanks a lot. Just to quickly confirm, in the case where a list of tags contains a "correct" and an "incorrect" tag, as long as a backend is found through the correct tag, there isn't any error being raised, right? Would it make sense to raise a warning mentioning that no backends were found for a specific tag?

ElePT:
I think that the filter should fail if there is any incorrect tag in the list, just as you described. I was asking because I think that my early tests of `plans_preference` would not fail when there was an incorrect plan together with a correct plan. I'm just thinking that we should do a quick check of all filters to make sure that the behavior is more or less consistent.

kt474:
Yeah `plans_preference` works a little differently because it orders the instances based on the plans in the given list. Having an incorrect plan with a correct plan would not raise any errors but I think we could add a warning. Something simple like `No instances with the plan '...' were found`.

PR #2282, opened 2025-06-05
Don't re-retrieve backend in session cm

kt474:
### Summary

Reported in slack - in the case where nothing is passed into the primitive inside a session/batch context manager, the backend is retrieved based on the name. This could cause issues if things in the backend target were modified. 

```python
backend = service.backend("ibm_torino")
backend.target.add_instruction(IfElseOp, name="if_else")

with Batch(backend):
    sampler = SamplerV2() # this would re-init the backend, losing the IfElseOp instruction
    sampler.options.experimental = {"execution_path" : "gen3-experimental"}
    job = sampler.run([isa_circuit])
```

### Details and comments
Fixes #


qiskit_ibm_runtime/base_primitive.py
```diff
@@ -86,19 +86,10 @@ def _get_mode_service_backend(mode: Optional[Union[BackendV2, Session, Batch]] =
...
-            backend = service.backend(
-                name=mode.backend(),  # type: ignore
-                instance=mode._instance,  # type: ignore
-            )
-        return mode, service, backend  # type: ignore
+        service = mode.service
+        backend = mode._backend
+
+        return mode, service, backend
     else:
```
kt474, line 93: I'm not 100% sure why it was done this way in the first place because the backend object is available in the session. I think this was done to cover for the case back when backend was not required or when passing a backend name string was an option.

qiskit_ibm_runtime/base_primitive.py
```diff
@@ -86,19 +86,10 @@ def _get_mode_service_backend(mode: Optional[Union[BackendV2, Session, Batch]] =
...
-                use_fractional_gates=mode._backend.options.use_fractional_gates,  # type: ignore
-            )
-        except (AttributeError, TypeError):
-            backend = service.backend(
-                name=mode.backend(),  # type: ignore
-                instance=mode._instance,  # type: ignore
-            )
-        return mode, service, backend  # type: ignore
+        service = mode.service
+        backend = mode._backend
```
ElePT, line 90: Why not do `mode.backend()`?

kt474: unfortunately `Session.backend()` returns the backend name instead of the object

ElePT: heh, interesting

release-notes/unreleased/2282.bug.rst
```diff
@@ -0,0 +1,4 @@
+Running jobs in a :class:`~.Session` or :class:`~.Batch` with
+a context manager and initializing a primitive without any parameters will reinitialize the
+Session/Batch backend. This could cause issues if the backend target has been modified so instead,
+the actual backend object is now used.
```
ElePT:
```suggestion
Fixed a bug in :class:`.BasePrimitive` where primitives instantiated inside a :class:`~.Session` or :class:`~.Batch` context manager would fetch the backend
from the service (by name) instead of using the backend passed to the :class:`~.Session` or :class:`~.Batch`. This could cause issues when the :class:`~.Session`/:class:`~.Batch`
backend was modified by users (for example, by removing a gate), because the primitives 
would instead fetch the unmodified backend object from the service. After the fix, the
:class:`~.Session`/:class:`~.Batch` backend object is used directly. 
```

ElePT (approved):
LGTM, thanks!

PR #2283, opened 2025-06-06
Create internal copies of backends to avoid modifying their original properties

### LINKED ISSUE #2280: `backend.basis_gates` is not renewed even if the backend is renewed
**Describe the bug**

If I update `backend.basis_gates` by removing a gate, the basis gates persists even if I renewed the backend.

I confirmed that qiskit-ibm-runtime 0.39.0 with `use_fractional_gates=False` renewed the basis gates, but `use_fractional_gates=True` or newer versions  cause the issue.

**Steps to reproduce**
I expect the basis gates of `backend renewed` is same as that of `original`. But it happens only for 0.39.0 and `use_fractional_gates=False`.
 
```python
from qiskit_ibm_runtime import QiskitRuntimeService, __version__

print(f"runtime_version={__version__}")


def test_remove(gate: str, fractional: bool):
    print(f"\nRemove {gate} gate with use_fractional_gates={fractional}")
    service = QiskitRuntimeService(name="demo-premium")
    backend = service.backend("ibm_fez", use_fractional_gates=fractional)
    print("original\t", backend.basis_gates)
    backend.basis_gates.remove(gate)
    print(f"{gate} removed\t", backend.basis_gates)
    backend = service.backend("ibm_fez", use_fractional_gates=fractional)
    print("backend renewed\t", backend.basis_gates)


test_remove("cz", False)
test_remove("cz", True)
```

```
runtime_version=0.39.0

Remove cz gate with use_fractional_gates=False
original	 ['cz', 'id', 'rz', 'sx', 'x']
cz removed	 ['id', 'rz', 'sx', 'x']
backend renewed	 ['cz', 'id', 'rz', 'sx', 'x']

Remove cz gate with use_fractional_gates=True
original	 ['cz', 'id', 'rx', 'rz', 'rzz', 'sx', 'x']
cz removed	 ['id', 'rx', 'rz', 'rzz', 'sx', 'x']
backend renewed	 ['id', 'rx', 'rz', 'rzz', 'sx', 'x']
```

```
runtime_version=0.40.0

Remove cz gate with use_fractional_gates=False
original	 ['cz', 'id', 'rz', 'sx', 'x']
cz removed	 ['id', 'rz', 'sx', 'x']
backend renewed	 ['id', 'rz', 'sx', 'x']

Remove cz gate with use_fractional_gates=True
original	 ['cz', 'id', 'rx', 'rz', 'rzz', 'sx', 'x']
cz removed	 ['id', 'rx', 'rz', 'rzz', 'sx', 'x']
backend renewed	 ['id', 'rx', 'rz', 'rzz', 'sx', 'x']
```

```
runtime_version=0.41.0.dev4+g3a8d49cb

Remove cz gate with use_fractional_gates=False
original	 ['cz', 'id', 'rz', 'sx', 'x']
cz removed	 ['id', 'rz', 'sx', 'x']
backend renewed	 ['id', 'rz', 'sx', 'x']

Remove cz gate with use_fractional_gates=True
original	 ['cz', 'id', 'rx', 'rz', 'rzz', 'sx', 'x']
cz removed	 ['id', 'rx', 'rz', 'rzz', 'sx', 'x']
backend renewed	 ['id', 'rx', 'rz', 'rzz', 'sx', 'x']
```

**Expected behavior**

The basis gates should be renewed when the backend is renewed as same as 0.39.0 and `use_fractional_gates=False`.

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.39.0, 0.40.0, main g3a8d49cb
- **Python version**: 3.12.10
- **Operating system**: macOS 15.5

t-imamichi:
### Summary

This is a PoC to fix #2280

### Details and comments

Test the same script as #2280 and confirmed that `basis_gates` are renewed as expected though there is an overhead of deep copy.

```python
from qiskit_ibm_runtime import QiskitRuntimeService, __version__

print(f"runtime_version={__version__}")


def test_remove(gate: str, fractional: bool):
    print(f"\nRemove {gate} gate with use_fractional_gates={fractional}")
    service = QiskitRuntimeService(name="demo-premium")
    backend = service.backend("ibm_fez", use_fractional_gates=fractional)
    print("original\t", backend.basis_gates)
    backend.basis_gates.remove(gate)
    print(f"{gate} removed\t", backend.basis_gates)
    backend = service.backend("ibm_fez", use_fractional_gates=fractional)
    print("backend renewed\t", backend.basis_gates)


test_remove("cz", False)
test_remove("cz", True)
```

```
runtime_version=0.41.0.dev5+g27fc3158

Remove cz gate with use_fractional_gates=False
original	 ['cz', 'id', 'rz', 'sx', 'x']
cz removed	 ['id', 'rz', 'sx', 'x']
backend renewed	 ['cz', 'id', 'rz', 'sx', 'x']

Remove cz gate with use_fractional_gates=True
original	 ['cz', 'id', 'rx', 'rz', 'rzz', 'sx', 'x']
cz removed	 ['id', 'rx', 'rz', 'rzz', 'sx', 'x']
backend renewed	 ['cz', 'id', 'rx', 'rz', 'rzz', 'sx', 'x']
```

### DIFF
--- a/test/integration/test_backend.py
+++ b/test/integration/test_backend.py
@@ -292,7 +292,7 @@ class TestIBMBackend(IBMIntegrationTestCase):
         """Test renewed backend property"""
         name = self.backend.name
         backend = self.service.backend(name)
-        basis_gates = backend.basis_gates
+        basis_gates = copy.copy(backend.basis_gates)
         # modify a property
         backend.basis_gates.remove(basis_gates[0])
         # renew backend


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -712,6 +713,12 @@ def backends(
         # Set fractional gate flag for use when loading properties or refreshing backend.
         for backend in backends:
             backend.options.use_fractional_gates = use_fractional_gates
+        try:
+            backends = deepcopy(backends)
+        except TypeError:
+            # local simulator cannot be deepcopied due to
+            # `TypeError: cannot pickle '_thread.RLock' object`
+            pass
         return filter_backends(backends, filters=filters, **kwargs)
```
ElePT, line 722: I think that instead of deep-copying the list, we should probably iterate over it and do individual copies of the backends. Else if a simulator is found, none of the backends get copied. I was thinking something like:

```suggestion
        return_backends = []
        for backend in backends:
          try:
             return_backends.append(deepcopy(backend))
          except TypeError:
              # local simulator cannot be deepcopied due to
              # `TypeError: cannot pickle '_thread.RLock' object`
              return_backends.append(backend)
        return filter_backends(return_backends, filters=filters, **kwargs)
```

There's probably a more efficient implementation for it, but this would be the gist. What do you think?

ElePT: I think that instead of deepcopying the full list at once, we should try copying the individual backends. Else if a simulator is found in the list, the real backends don't get copied with it:

```suggestion
        return_backends = []
        for backend in backends:
            backend.options.use_fractional_gates = use_fractional_gates
            try:
                return_backends.append(deepcopy(backend))
            except TypeError:
                # local simulator cannot be deepcopied due to
                # `TypeError: cannot pickle '_thread.RLock' object`
                return_backends.append(backend)
        return filter_backends(return_backends, filters=filters, **kwargs)
```

t-imamichi: Thanks. It makes more sense.

kt474:
PR LGTM but will this still be an issue once the `use_fractional_gates` flag goes away? This change makes fetching backends noticeably slower.

ElePT (commented):
Unfortunately, this issue isn't limited to `use_fractional_gates`. Any modification to the basis set on the user side affects not only the cached backends, but the cached `BackendConfiguration`, which for sure should not happen. I see your concern with the runtime, I ran some profiling with the reproducing code and `cProfile` and `deepcopy` took over a large proportion of the runtime (third item in the list, essentially the slowest operation):

Screenshot 2025-06-19 at 15 48 06

An alternative we could follow is deepcopying the `config` object in the backend construction step, instead of the full backend. I tried this out and the profile looks much better, it's about 6.86x faster:

Screenshot 2025-06-19 at 15 48 24

This would be adding the deepcopy in lines 855 and 862. What do you think @kt474?

kt474:
> This would be adding the deepcopy in lines 855 and 862. What do you think @kt474?

Yeah I think we should do that, similar to how it works with the current IQP classic flow where there's a configuration registry and copy at the api level. 

https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/api/clients/runtime.py#L285

t-imamichi:
Thank you for the discussion. I tried to deep copy configuration instead of backend itself.
It looks good as follows.
```
runtime_version=0.41.0.dev15+g4beaea34

Remove cz gate with use_fractional_gates=False
original	 ['cz', 'id', 'rz', 'sx', 'x']
cz removed	 ['id', 'rz', 'sx', 'x']
backend renewed	 ['cz', 'id', 'rz', 'sx', 'x']

Remove cz gate with use_fractional_gates=True
original	 ['cz', 'id', 'rx', 'rz', 'rzz', 'sx', 'x']
cz removed	 ['id', 'rx', 'rz', 'rzz', 'sx', 'x']
backend renewed	 ['cz', 'id', 'rx', 'rz', 'rzz', 'sx', 'x']
```

I also compared the runtime as follows. The overhead of the latest commit is smaller than the previous approach.
```python
from copy import deepcopy
from timeit import timeit

from qiskit_ibm_runtime import QiskitRuntimeService, __version__

print(f"runtime_version={__version__}")

service = QiskitRuntimeService(name="demo-premium")
_ = service.backends()
print(f"{timeit(lambda: service.backends(), number=10)} sec")
```

```
runtime_version=0.40.1
2.5802090410143137 sec
```

deep copy configuration
```
runtime_version=0.41.0.dev15+g4beaea34
2.631578707951121 sec
```

deep copy backend
```
runtime_version=0.41.0.dev14+g89cddef2
3.1682619999628514 sec
```

kt474 (dismissed):
Thanks @t-imamichi! Just one last thing - can you add a short release note too

test/integration/test_backend.py
```diff
@@ -287,3 +287,14 @@ def test_use_fractional_gates_flag(self):
             self.skipTest("Real backend not available.")
         self.assertIn("rzz", real_device_fg.basis_gates)
         self.assertNotIn("rzz", real_device_no_fg.basis_gates)
+
+    def test_renew_backend_properties(self):
+        """Test renewed backend property"""
+        name = self.backend.name
+        backend = self.service.backend(name)
+        basis_gates = backend.basis_gates
```
kt474:
```suggestion
        basis_gates = copy.copy(backend.basis_gates)
```
The test currently fails because the `basis_gates` reference is also updated when the gate is removed.

t-imamichi:
Thank you for testing it. I applied your suggestion.

ElePT:
Given that the next planned release is 0.41, I do think that we should backport this fix to the 0.40 branch too.

PR #2290, opened 2025-06-16
Integration test cleanup

kt474:
### Summary

- Removes the use of real devices in integration tests. Two integration tests have been using `alt_fez` which has been having issues for a while causing the tests to time out. This way we also don't take up real device time. 
- Also updated `RealAmplitudes` to `real_amplitudes` since it's been deprecated since Qiskit 1.3.

### Details and comments
Fixes #


ElePT (commented):
Would it make sense to use a test backend (`test_eagle`)? We would still be affected if there's an outage but I feel like that's anyways good information to have from an integration test. And test backends are not real devices, so this option would also save device time. I am not too convinced on not having any integration tests using the real service.

ElePT (approved):
Since my previous review I realized that the comment didn't make a lot of sense, because there are obviously more integration tests using `test_eagle`. So overall LGTM.

PR #2289, opened 2025-06-14
Remove IQP Classic

kt474:
### Summary

Things I noticed that are handled in a separate PR (#2295).

TODO before merging: 
- Make sure dates in docs are correct 
- Test account logic and make sure having `ibm_quantum` accounts saved doesn't break anything 
- Double check things that may need proper deprecation 

### Details and comments
Fixes #2285

PR #2296, opened 2025-06-26
Handle new IQP inconsistencies/breaking changes 

kt474:
### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/base_runtime_job.py
+++ b/qiskit_ibm_runtime/base_runtime_job.py
@@ -379,7 +379,7 @@ class BaseRuntimeJob(ABC):
     @property
     def instance(self) -> Optional[str]:
         """For ibm_quantum channel jobs, return the instance where the job was run.
-        For ibm_cloud and ibm_quantum_platform, `None` is returned.
+        For ibm_cloud and ibm_quantum_platform, the instance crn is returned.
         """
         return self._backend._instance
 
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -851,6 +851,7 @@ class QiskitRuntimeService:
             else:
                 # cloud backend doesn't set hgp instance
                 return ibm_backend.IBMBackend(
+                    instance=instance,
                     configuration=config,
                     service=self,
                     api_client=self._active_api_client,
@@ -1235,8 +1236,9 @@ class QiskitRuntimeService:
                 jobs are included. If ``False``, 'DONE', 'CANCELLED' and 'ERROR' jobs
                 are included.
             program_id: Filter by Program ID.
-            instance: This is only supported for ``ibm_quantum`` runtime and is in the
-                hub/group/project format.
+            instance: (DEPRECATED) This is only supported for ``ibm_quantum`` runtime and is in the
+                hub/group/project format. Since this parameter is not supported on ``ibm_cloud``
+                and ``ibm_quantum`` channels, it will be removed in a future release.
             job_tags: Filter by tags assigned to jobs. Matched jobs are associated with all tags.
             session_id: Filter by session id. All jobs in the session will be
                 returned in desceding order of the job creation date.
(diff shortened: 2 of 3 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1316,29 +1319,19 @@ def delete_job(self, job_id: str) -> None:
...
-        try:
-            self._active_api_client.job_delete(job_id)
-        except RequestsApiError as ex:
-            if ex.status_code == 404:
-                raise RuntimeJobNotFound(f"Job not found: {ex.message}") from None
-            raise IBMRuntimeError(f"Failed to delete job: {ex}") from None
+        raise IBMInputValueError(
+            "This method is not supported on the new IBM Quantum Platform. "
+            "It will be deprecated and removed in a future release."
+        )
```
kt474, line 1354: I'm not sure how to deal with the `delete_job()` method. It's not supported on the server side so should we deprecate it first? Or just raise an error for now.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1259,7 +1261,8 @@ def jobs(
         if instance:
             if self._channel in ["ibm_cloud", "ibm_quantum_platform"]:
                 raise IBMInputValueError(
-                    "The 'instance' keyword is only supported for ``ibm_quantum`` runtime."
+                    "The 'instance' keyword is only supported for ``ibm_quantum`` runtime. "
+                    "This parameter will be removed in a future release."
```
kt474: Same with this case - passing in `instance` already raises an error for the new channels. Do we still have to formally deprecate and then remove? Or can we just add to the error message and tell users it's going to be removed.

qiskit_ibm_runtime/qiskit_runtime_service.py
ElePT: I think that the deprecation warning + error combination works for now.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1332,37 +1334,45 @@ def jobs(
...
         try:
             self._active_api_client.job_delete(job_id)
         except RequestsApiError as ex:
             if ex.status_code == 404:
                 raise RuntimeJobNotFound(f"Job not found: {ex.message}") from None
-            raise IBMRuntimeError(f"Failed to delete job: {ex}") from None
+            raise IBMRuntimeError(
+                "Failed to delete job. deleting jobs is not supported on the "
+                f"new IBM Quantum Platform: {ex}"
+            ) from None
```
ElePT: I would keep the generic "failed to delete job" message until `ibm_quantum` is fully removed. It's a small detail but users might bet confused if the error using the old platform refers to the new platform.

qiskit_ibm_runtime/base_runtime_job.py
```diff
@@ -379,7 +379,7 @@ def usage_estimation(self) -> Dict[str, Any]:
     @property
     def instance(self) -> Optional[str]:
         """For ibm_quantum channel jobs, return the instance where the job was run.
-        For ibm_cloud and ibm_quantum_platform, `None` is returned.
+        For ibm_cloud and ibm_quantum_platform, the instance crn is returned.
```
ElePT, line 382: Do we have a unit tests for this property?

ElePT (commented):
Thanks a lot @kt474, the PR looks good, I just had a couple of minor comments and modified a bit the PR description to include details of what it's changing (for future reference).

qiskit_ibm_runtime/base_runtime_job.py
kt474: Update integration test in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2296/commits/80cdc54f39cb35174f4df25096047b894be9a5a3, currently don't have unit tests for these job properties.

ElePT (commented):
Thanks Kevin! the PR GLTM, I am just missing the deprecation warning mentioned in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2289#discussion_r2179587386. Maybe I can quickly push it myself.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1332,17 +1334,25 @@ def jobs(
         return [self._decode_job(job) for job in job_responses]
 
     def delete_job(self, job_id: str) -> None:
-        """Delete a runtime job.
+        """(DEPRECATED) Delete a runtime job.
```
mriedem, line 1337: Why was this deprecated? Deleting a job works fine in the cloud stack, there is a documented API for it and internal integration tests that run against it daily.

https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/jobs

image

PR #2238, opened 2025-04-25
Completely remove pulse defaults

kt474:
### Summary

Can't merge until 3 months after `0.38.0` release ~ 7/16

### Details and comments
Fixes #2179

PR #2298, opened 2025-06-30
Handle 0.38.0 deprecations

kt474:
### Summary

Pulse deprecations removed in a separate PR. This also shouldn't be merged until after ~ 7/16. We should double check that `RuntimeJob` is no longer used anywhere. 

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -108,8 +108,9 @@ class QiskitRuntimeService:
         ``token``. The ``local`` channel doesn't require authentication.
         For non-local channels, it is recommended to always provide the relevant ``instance``
         to minimize API calls. If an ``instance`` is not defined, the service will fetch all
-        instances accessible within the
-        account, filtered by ``region``, ``plans_preference``, and ``tags``.
+        instances accessible within the account, filtered by ``region``, ``plans_preference``,
+        and ``tags``. If ``plans_preference`` is not set, free and trial instances will be prioritized
+        over paid instances.
 
         The service will attempt to load an account from file if (a) no explicit ``token``
         was provided during instantiation  or (b) a ``name`` is specified, even if an explicit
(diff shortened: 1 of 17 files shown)


kt474:
Let's hold off on this PR for now - I'm still a bit unsure about the `RuntimeJob` removal, some older `noise-learner` jobs may still be RuntimeJob instead of RuntimeJobV2

I think #2238 is ready to be merged though.

ElePT:
whoops, I had left my #2238 review pending.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -843,25 +842,17 @@ def _run(
         program_id: str,
         inputs: Dict,
         options: Optional[Union[RuntimeOptions, Dict]] = None,
-        callback: Optional[Callable] = None,
         result_decoder: Optional[Union[Type[ResultDecoder], Sequence[Type[ResultDecoder]]]] = None,
         session_id: Optional[str] = None,
         start_session: Optional[bool] = False,
-    ) -> Union[RuntimeJob, RuntimeJobV2]:
```
ElePT, line 914: should the return type still include `RuntimeJob`?

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -925,15 +916,14 @@ def _run(
...
             program_id=program_id,
-            user_callback=callback,
             result_decoder=result_decoder,
             image=qrt_options.image,
             service=self,
             version=version,
             private=qrt_options.private,
         )
 
-    def job(self, job_id: str) -> Union[RuntimeJob, RuntimeJobV2]:
```
ElePT, line 1000: should the return type still include `RuntimeJob`? (same for following methods in this class)

qiskit_ibm_runtime/session.py
```diff
@@ -146,17 +145,15 @@ def _run(
         program_id: str,
         inputs: Dict,
         options: Optional[Dict] = None,
-        callback: Optional[Callable] = None,
         result_decoder: Optional[Type[ResultDecoder]] = None,
-    ) -> Union[RuntimeJob, RuntimeJobV2]:
```
ElePT, line 151: should the return type still include `RuntimeJob`?

test/integration/test_ibm_job_attributes.py
```diff
@@ -37,7 +37,7 @@ class TestIBMJobAttributes(IBMTestCase):
 
     sim_backend: IBMBackend
     bell: QuantumCircuit
-    sim_job: RuntimeJob
```
ElePT, line 40: Are both accepted or only V2?

ElePT (commented):
Thanks for continuing this PR, I took another look and I'm unsure of how RuntimeJob vs RuntimeJobV2 are expected to be handled. When is which one used? I also noticed the missing release note.

kt474:
> Thanks for continuing this PR, I took another look and I'm unsure of how RuntimeJob vs RuntimeJobV2 are expected to be handled. When is which one used? I also noticed the missing release note.

All primitives should now returned as `RuntimeJobV2` but I think we should keep the `RuntimeJob` class around for a little longer because it's still imported in a lot of places internally.

release-notes/unreleased/2298.update.rst
```diff
@@ -0,0 +1,2 @@
+Since support for streaming results was removed in ``0.32.0``, the deprecated (``0.38.0``) environment option
+``callback`` as well as the ``BaseRuntimeJob`` parameters ``client_params`` and ``user_callback`` have been removed.
```
ElePT: I think that the reno should also mention the removal of `RuntimeJob` and the updated typehints. Technically there should have been a deprecation warning when things were updated server-side, but given that there wasn't one, I consider this typehint update to be more of a fix to better reflect the current situation and I would not slow it down with deprecations.

PR #2301, opened 2025-07-02
Remove `RuntimeOptions`

### LINKED ISSUE #2300: Cleanup `RuntimeOptions`
`RuntimeOptions` was originally only meant for custom programs. Since we removed support for custom programs over a year ago we removed `RuntimeOptions` from the docs ([PR](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1808)) but they are still used to set certain values before running a job. It's confusing to have this extra options class that is separate from the primitive options so we should consider removing this class and refactoring the job flow code.

kt474:
### Summary

Draft PR for now, still need to test which options actually work. Currently more of a POC to show that we don't need the `RuntimeOptions` class anymore and it's confusing to have this extra "options" class. 

### Details and comments
Fixes #2300


ElePT (commented):
Another deprecation slip-up from the runtime programs era. This should have been deprecated when it stopped being useful server-side. At this point, I again see this cleanup as a bugfix rather than a new deprecation, given that the functionality stopped working a long time ago. I am fine with the removal, I would just point out the release note to document it.

ElePT (commented):
I have found a downstream package that depended on RuntimeOptions. Given that we are in no rush to remove the class, I suggest turning this PR into a deprecation PR instead of a removal. I would keep the typehint changes though, they reflect the current status better.

diego-plan9:
Thanks - while `RuntimeOptions` is deprecated, latest changes (new-style options not using RuntimeOptions) and refactors lowered the urgency of actually removing it. Closing this PR in order to tackle the removal separately (and perhaps by the time the old-style options are removed).

PR #2318, opened 2025-07-09
Remove `test_run_program_missing_backend_ibm_cloud`

kt474:
### Summary

This test should've been removed long ago because a backend is required. It was only passing because `service.backend(None)` will just return the first backend from the list of available backends. 

It's been failing for the past few days because the first backend selected, `ibm_quebec`, has been offline. 

### Details and comments
Fixes #


ElePT (approved):
LGTM. Is it intentional to have `service.backend(None)` return a backend? I feel like we should either document it or stop returning a backend if it's not intentional.

PR #2321, opened 2025-07-10
Update `test_jobs_before_backend`

kt474:
### Summary

This test started failing because calling `jobs()` in the case where an instance is not passed in to the service and there is no default instance will just return the jobs from the "last" instance. In this case the last instance is the premium instance and the most recent (`jobs[0]`) job failed so there is no result. 

What should be the proper behavior for this scenario? If there is no instance passed in or selected at any stage then should `jobs()` return jobs from all instances (and from all users)? Currently, users are expected to select a backend first and then the active instance is "correct" so `jobs()` will return jobs from just that instance. @ElePT what do you think? 

### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_account.py
+++ b/test/integration/test_account.py
@@ -234,7 +234,7 @@ class TestQuantumPlatform(IBMIntegrationTestCase):
         jobs = service.jobs()
         self.assertTrue(jobs)
         job = jobs[0]
-        self.assertTrue(job.result())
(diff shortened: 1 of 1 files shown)


kt474:
I agree, lets add a `active_instance` method and then repurpose the instance parameter in the `jobs()` method. I think this PR (adjusting the test) can still be merged because the failure is not related to the test case.

PR #2323, opened 2025-07-15
Update least_busy

### LINKED ISSUE #2240: Improve performance on `least_busy` operation
Explore using the new `GET /backends` endpoint instead of the current path to fetch the `least_busy` device. In the current approach, the  number of API calls scales with the number of devices.

Using `GET/backends` we can access the following backend information for the full list of available devices (for a given instance, I assume) in a single API call:

```
{
  "devices": [
    {
      "name": "ibmq_qasm_simulator",
      "status": {
        "name": "online"
      },
      "is_simulator": true,
      "qubits": 32,
      "clops": {
        "type": "virtual",
        "value": 150
      },
      "processor_type": {
        "family": "Quantum",
        "revision": "RevA",
        "segment": "Segment1"
      },
      "queue_length": 10
    }
  ]
}
```

This allows to filter backends by name, status, is_simulator, num. qubits, clops, processor type and queue_length, which covers most of the use cases I have seen in the unit tests.

This approach does not allow, however, filtering by:

- configuration properties (beyond num.qubits)
- instance (which is more related to IBM Cloud, #2236 and #2237)

If a user wanted to filter by additional configuration properties, we could combine an initial filtering using the information given in `GET/backends` and a second filtering done calling the configuration endpoint, and warn that this is expected to be slower than other filters.

Original proposal in vt-qiskit-technologies #128

**Acceptance criteria**

kt474:
### Summary

Messed up my other two branches. This method is now significantly faster because we don't need to fetch every single backend configuration and compare the number of pending jobs. 
Screenshot 2025-07-15 at 2 43 03 PM

### Details and comments
Fixes #2240


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -200,7 +201,10 @@ def _discover_backends_from_instance(self, instance: str) -> List[str]:
                     new_client = self._create_new_cloud_api_client(instance)
                     self._api_clients.update({instance: new_client})
                     self._active_api_client = new_client
-            return self._active_api_client.list_backends()
+                self._backends_list = self._active_api_client.list_backends()
+            if not self._backends_list:
+                self._backends_list = self._active_api_client.list_backends()
```
ElePT: if list_backends is going to be called anyways outside of the if-else, why not do:

```suggestion
            if not self._backends_list:
                self._backends_list = self._active_api_client.list_backends()
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1157,18 +1161,35 @@ def least_busy(
         Raises:
             QiskitBackendNotFoundError: If no backend matches the criteria.
         """
-        backends = self.backends(
-            min_num_qubits=min_num_qubits, instance=instance, filters=filters, **kwargs
-        )
+        if not self._backends_list:
```
ElePT, line 1165: can't comment on the actual line but I realized that the docstring is outdated (on line 1156 it uses `n_qubits`)

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1157,18 +1161,35 @@ def least_busy(
...
+                candidates.append(backend)
+        if filters:
+            # filters will still be slow because we need the backend configs
+            backends = self.backends(
+                min_num_qubits=min_num_qubits, filters=filters, instance=instance, **kwargs
+            )
+            filtered_backend_names = [back.name for back in backends]
+            for candidate in candidates.copy():
+                if candidate["name"] not in filtered_backend_names:
+                    candidates.remove(candidate)
```
ElePT: I need to look into the filters, but is there any additional filter that we can apply directly to `self._backends_list` without having to get the configs?

ElePT (commented):
Thanks a lot @kt474. The PR looks good to me, I just added a minor documentation comment and a broad question. Apart from this, I think that the PR is ready to merge.

qiskit_ibm_runtime/qiskit_runtime_service.py
kt474: I think that's a confusing example because there is a `n_qubits` property in a backend configuration so 
`least_busy(n_qubits=5, operational=True)` does work

qiskit_ibm_runtime/qiskit_runtime_service.py
kt474: I did it this way because an instance passed to `backends()` could be a different instance from the current active api instance.

qiskit_ibm_runtime/qiskit_runtime_service.py
kt474: Here's what the new backends response returns - so I think we can also apply `is_simulator` and `processor_type`? Those are both in the backend config. 

```
{
      "name": "ibmq_nazca",
      "status": {
        "name": "paused",
        "reason": "Maintenance"
      },
      "is_simulator": false,
      "qubits": 16,
      "clops": {
        "type": "hardware",
        "value": 80
      },
      "processor_type": {
        "family": "Quantum",
        "revision": "RevB",
        "segment": "Segment2"
      },
      "queue_length": 20
    }
```

kt474: Maybe not `is_simulator` because I don't think we even have any simulators anymore.

qiskit_ibm_runtime/qiskit_runtime_service.py
ElePT: hmmm I might be missing something, but it looks like both lines are calling the current active client?

qiskit_ibm_runtime/qiskit_runtime_service.py
ElePT: heh, wonderful

qiskit_ibm_runtime/qiskit_runtime_service.py
ElePT: I'm fine adding these in a follow-up, but I feel like anything we can check without calling `backends()` should be done without calling `backends()`.

kt474: Let's do it in a follow up

qiskit_ibm_runtime/qiskit_runtime_service.py
kt474: I think we need to handle the case where: 

```python

service = QiskitRuntimeService() # No default instance 

service.backends(instance="instance1") # backends_list will only have backends from instance1

service.backends(instance="instance2") # backends_list needs to be refetched because the instance passed in is not the active instance
 ```
If self._backends_list is not `None` but has the backends from a different instance the backends from the wrong instance would be returned. 

There might be a better way to handle this, as the way I did it here is not the most intuitive.

File: qiskit_ibm_runtime/qiskit_runtime_service.py [hunk 4]
```diff
@@ -200,7 +199,8 @@ def _discover_backends_from_instance(self, instance: str) -> List[str]:
                     new_client = self._create_new_cloud_api_client(instance)
                     self._api_clients.update({instance: new_client})
                     self._active_api_client = new_client
-            return self._active_api_client.list_backends()
+            self._backends_list = self._active_api_client.list_backends()
+            return [backend["name"] for backend in self._backends_list]
         # On staging there some invalid instances returned that 403 when retrieving backends
```
kt474, line 204: We still have to store the `/backends` API response somewhere in the service object. I think it's best to handle that here in `_discover_backends_from_instance()`?

ElePT: You meant better to handle that here than in `_discover_backends_from_instance()`, right? I agree

PR #2324, opened 2025-07-15
Post IQP classic removal integration test fixes

kt474:
### Details and comments
Fixes #

### DIFF
--- a/test/integration/test_fake_backends.py
+++ b/test/integration/test_fake_backends.py
@@ -16,7 +16,6 @@
 import itertools
 import operator
 
-from unittest import SkipTest
 from ddt import ddt, data, idata, unpack
 
 from qiskit.circuit import QuantumCircuit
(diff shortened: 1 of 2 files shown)


test/integration/test_fake_backends.py
```diff
@@ -167,23 +166,25 @@ def setUpClass(cls):
...
         # change some configuration
         old_backend.backend_version = "fake_version"
-        # set instance to none to access real backend
-        self.service._account.instance = None
+
+        service = QiskitRuntimeService(
+            token=self.dependencies.token,
+            channel=self.dependencies.channel,
+            url=self.dependencies.url,
+        )
```
kt474, line 179: Sherbrooke is only available on the premium instance so we have to re-initialize the service here. On classic IQP we would set the account instance to `None` to force the service to check all instances.

test/integration/test_ibm_job.py
```diff
@@ -44,10 +43,7 @@ def setUp(self):
 
     def test_cancel(self):
         """Test job cancellation."""
-        if self.dependencies.channel == "ibm_cloud":
-            raise SkipTest("Cloud account does not have real backend.")
         # Find the most busy backend
-        self.service._account.instance = None  # set instance to none to avoid filtering
```
kt474, line 50: Same with these tests - setting the account instance to `None` was the hacky way to force the service to check all instances in IQP classic.

ElePT (approved):
LGTM, I think it's cleaner to have a new service rather than some hacky re-instantiation

PR #2329, opened 2025-07-17
Update usage method to include remaining usage

### LINKED ISSUE #1829: Remaining time of an instance

Pay-go instances in the Cloud channel and Premium instances in the IBM Quantum channel, have or may have an usage limit. Currently, there is no way to query the remaining available time on my instance through Qiskit so I can decide if submit or not a new workload based on that.

For Cloud instances, there is an API in `/instances/configuration` to retrieve the max limit and `/instances/usage` to retrieve the current usage on an instance.

For premium IQP instances, there is an API in `https://api-qcon.quantum.ibm.com/api/Network` that retrieves `capacity` in seconds with the total capacity of the project and `https://api.quantum.ibm.com/runtime/analytics/usage` that retrieves the usage in seconds.

**Acceptance criteria**

A user can query the remaining time on their instance to decide if run or not the next workload.

kt474:
### Summary

Also cleaned up the old usage() methods from IQP classic

### Details and comments
Fixes #1829

### DIFF
--- a/release-notes/unreleased/2329.feat.rst
+++ b/release-notes/unreleased/2329.feat.rst
@@ -1,1 +1,1 @@
-:meth:`.QiskitRuntimeService.usage` has been updated to include a ``usage_remaining`` field.
+:meth:`.QiskitRuntimeService.usage` has been updated to include a ``usage_remaining`` field. This field includes the difference between the reported ``usage_limit_seconds``/``usage_allocation_seconds`` (whatever is higher) and ``usage_consumed_seconds``.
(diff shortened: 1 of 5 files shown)


ElePT (commented):
The method needs to handle the `None` case independently. I tried the new method with an open plan account and it raised a TypeError:

```
>>> service.usage()
  File "<stdin>", line 1, in <module>
  File "/Users/ept/qiskit_workspace/qiskit-ibm-runtime/qiskit_ibm_runtime/qiskit_runtime_service.py", line 1037, in usage
    usage_dict.get("usage_limit_seconds", usage_dict.get("usage_allocation_seconds"))
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

release-notes/unreleased/2329.feat.rst
```diff
@@ -0,0 +1 @@
+:meth:`.QiskitRuntimeService.usage` has been updated to include a ``usage_remaining`` field.
```
ElePT:
```suggestion
:meth:`.QiskitRuntimeService.usage` has been updated to include a ``usage_remaining`` field. This field includes the difference between the reported ``usage_limit_seconds``/``usage_allocation_seconds`` (whatever is higher) and ``usage_consumed_seconds``.
```

ElePT (commented):
LGTM, I just have a final remark, should we call the entry `usage_remaining_seconds` for consistency with the others? I find it a bit strange to suddenly drop the units.

release-notes/unreleased/2329.feat.rst
kt474: I don't think it's the higher value out of `usage_limit_seconds` and `usage_allocation_seconds` - we use `usage_limit_seconds` is it exists (if the instance manager set limits) and if it doesn't, we use `usage_allocation_seconds` (some allocation calculated with the fair share ratio).

PR #2331, opened 2025-07-17
Update error message if no saved accounts are found

kt474:
### Summary

Reported from slack - on `0.40.1`, if a user does not have any accounts with `ibm_quantum_platform` as the channel name saved and then tries QiskitRuntimeService(channel="ibm_quantum_platform"), the service will look for any default accounts saved instead. If an IQP classic account exists (`default-ibm-quantum`), then it will be selected and there will be a misleading error message. This is fixed on the main branch because of the IQP removal PR. 

Now (on the main branch), if saved accounts with `ibm_quantum_platform` don't exist, the service will check for the default `ibm_cloud` account which would still work. We should still update the error message in the `_discover_account()` method to be more helpful. 

@ElePT Let me know if I captured everything correctly here.

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/accounts/management.py
+++ b/qiskit_ibm_runtime/accounts/management.py
@@ -201,8 +201,10 @@ class AccountManager:
             if account_name in all_config:
                 return Account.from_saved_format(all_config[account_name])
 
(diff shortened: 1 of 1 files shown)


ElePT (approved):
I think that this works. Thanks a lot!

PR #2332, opened 2025-07-21
Allow `saved_accounts()` to return even if legacy accounts are saved

kt474:
### Summary

Reported in slack - if a user has `ibm_quantum` channel accounts saved then calling `saved_accounts()` will result in an error because the legacy account can not be initialized (only a bug on main, not `0.40.1`). 

### Details and comments
Fixes #

### DIFF
--- a/test/unit/test_account.py
+++ b/test/unit/test_account.py
@@ -57,17 +57,6 @@ _TEST_IBM_CLOUD_ACCOUNT = Account.create_account(
     ),
 )
 
-# test having an old account saved does not break anything
-_TEST_IBM_QUANTUM_CLASSIC_ACCOUNT = Account.create_account(
-    channel="ibm_quantum_platform",
-    token="token-y",
-    url="https://quantum.cloud.ibm.com/api/v1",
-    instance="crn:v1:bluemix:public:quantum-computing:us-east:a/...::",
-    proxies=ProxyConfiguration(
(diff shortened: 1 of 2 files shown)


test/unit/test_account.py
```diff
@@ -57,17 +57,6 @@
...
-_TEST_IBM_QUANTUM_CLASSIC_ACCOUNT = Account.create_account(
-    channel="ibm_quantum_platform",
-    token="token-y",
-    url="https://quantum.cloud.ibm.com/api/v1",
-    instance="crn:v1:bluemix:public:quantum-computing:us-east:a/...::",
-    proxies=ProxyConfiguration(
-        username_ntlm="bla", password_ntlm="blub", urls={"https": "127.0.0.1"}
-    ),
-)
-
```
kt474, line 70: I actually tried to add a test for this in the IQP removal PR but I didn't change the channel name here 😅

ElePT: And I missed it....

ElePT (approved):
Thanks a lot, I am really happy that this got reported internally, because I remember running into issues loading accounts from files with legacy channels. At least we know this use case is covered.

test/unit/test_account.py
ElePT: All this insisting for tests with legacy accounts...

PR #2335, opened 2025-07-22
Add Experimental warning and decorator

ElePT:
### Details and comments
Fixes #


ElePT:
It's not clear this is a direction we'd like to follow, closing.

PR #2336, opened 2025-07-22
Added pub shots in twirled slice span

### LINKED ISSUE #2312: `TwirledSliceSpan.mask` returns results with incorrect shapes
The shape of `TwirledSliceSpan.mask(pub_idx)` can be different from the result array of the corresponding PUB: the shot axis (last axis) can be longer. This contradicts the docstring of `mask`: https://github.com/Qiskit/qiskit-ibm-runtime/blob/d4112af73970ad2669d195155ab7a2588e9e663a/qiskit_ibm_runtime/execution_span/execution_span.py#L101-L110

This is however almost only a formality: the array returned by `TwirledSliceSpan.mask` supersedes the PUB result array: if a user wants to know which entries of the result array are covered by the span, this information is provided by the entries set to `True` in the output of `mask`.

Options:
1. Fix documentation of `ExecutionSpan.mask` - is it considered as a breaking change?
2. Modify `TwirledSliceSpan.mask` to return an array with the same shape: this also requires changes that can be considered as breaking. We have to provide to `mask` the information about the correct shape, either by adding a parameter to `mask` or by extending the data slice tuples. The parameter to `mask` can be optional, but then we still have to handle the case where it is not set, which takes us back to Option 1.

Note: if the result shape is available to `mask` then we can solve #2311 by returning an array of the given shape, where all the entries are set to `False`.

yaelbh:
### Summary

Fixes #2312.


ihincks (approved):
Thanks @yaelbh  for figuring this out.

PR #2338, opened 2025-07-23
Fail gracefully when pub is not in span

### LINKED ISSUE #2311: Define behavior of execution span mask method for pubs not in the span
Consider the data slices of following twirled spans (originated from a real execution):
```
{0: ((10, 1), True, slice(0, 1, None), slice(0, 1, None)), 1: ((2, 10, 1), False, slice(0, 1, None), slice(0, 1, None))}
{1: ((2, 10, 1), False, slice(10, 11, None), slice(0, 1, None))}
```
The second span doesn't execute anything for PUB 0. As a result, calling this span with `mask(0)` yields an error
```
    def mask(self, pub_idx: int) -> npt.NDArray[np.bool_]:
>       shape, at_front, shape_sl, shots_sl = self._data_slices[pub_idx]
E       KeyError: 0
```

Our options are:
- (my least preferred option) Do nothing - the user receives a clear error message.
- Fail with a nicer error message, by checking if `pub_idx` is in `self._data_slices`.
- Return something, maybe `None`, without raising an exception.

I believe that the same applies also to `SliceSpan` and `DoubleSpan`.

yaelbh:
### Summary

Fixes #2311.


kt474 (approved):
LGTM

PR #2342, opened 2025-07-25
Allow `use_fractional_gates` in fake backends

### LINKED ISSUE #2224: It seems like there is no possibility to use fractional gates with fake Heron r2 backends

In real Heron r2 backends it is possible to use [fractional gates](https://docs.quantum.ibm.com/guides/fractional-gates#fractional-gates) `RZZ` and `RX` by adding `use_fractional_gates=True` flag. I wasn't able to find such a flag for fake providers, does it exists? If not, is it possible to add such a flag? 

Image

**Acceptance criteria**

* It is possible to use `use_fractional_gates=True`  in fake providers.

kt474:
### Details and comments
Fixes #2224

### DIFF
--- a/qiskit_ibm_runtime/execution_span/twirled_slice_span.py
+++ b/qiskit_ibm_runtime/execution_span/twirled_slice_span.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from datetime import datetime
-from typing import Iterable
+from typing import Iterable, Mapping
 
 import math
 import numpy as np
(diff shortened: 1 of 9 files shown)


qiskit_ibm_runtime/utils/backend_decoder.py
```diff
@@ -19,10 +19,7 @@
 import dateutil.parser
 from qiskit.circuit.library.standard_gates import get_standard_gate_name_mapping
 
-try:
-    from qiskit.circuit import CONTROL_FLOW_OP_NAMES
-except ImportError:  # Remove when dropping support for Qiskit < 1.3
-    CONTROL_FLOW_OP_NAMES = frozenset(("for_loop", "while_loop", "if_else", "switch_case"))
```
ElePT, line 25: heh, good catch

ElePT (commented):
LGTM, I would just add a release note for the change in the `refresh` method.

kt474:
This solution only works if the user already has access to the real backend. So with the open plan, only `torino`, `brisbane`, and `sherbrooke` would be supported.

PR #2345, opened 2025-07-29
Inject proxy configuration in CloudAccount.tm attribute

### LINKED ISSUE #2344: Bug - Proxy configuration not initiated in the IAM handling service
**Describe the bug**
Requests using the librairie behind a proxy with the `ibm_quantum_channel` take 3 minutes.

**Steps to reproduce**
1. Set up an isolated environnement that uses a proxy to access internet.
2. initiate a service `QiskitRuntimeService(..., channel="ibm_quantum_channel", proxies = ...)`
3. Make any request (`.backends()/.jobs()`)

**Expected behavior**
Requests take at most 7s to execute.

**Suggested solutions**

Add proxy configuration to the token manager .

**Additional Information**

- **qiskit-ibm-runtime version**: 0.40.1
- **Python version**: 3.12
- **Operating system**: Win11

alodolo:
Fixes #2344


Siddharthgolecha:
Hi @AlexanderAtIBM , as I mentioned in my email, one of the client also applied your patch and this is what they say

The proxy parameters passed to `QiskitRuntimeService` aren’t making it all the way down to the `requests` layer, and the same happens with `verify=False`.

As a workaround, he:

1. Defined the proxy via `http_proxy` and `https_proxy` environment variables.
2. Manually force `verify=False` inside `requests/api.py` around line 55.
```
kwargs["verify"] = False
with sessions.Session() as session:
    return session.request(method=method, url=url, **kwargs)
```

Example workaround:

```
os.environ["http_proxy"] = f"http://{username_ntlm}:{password_ntlm}@{proxy_address}"
os.environ["https_proxy"] = f"http://{username_ntlm}:{password_ntlm}@{proxy_address}"
service = QiskitRuntimeService(..., verify=False)
```

These settings aren’t being forwarded through the intermediate layers. So, I would request you to include these in your patch.

AlexanderAtIBM:
Hi @Siddharthgolecha, 
I added the ssl_verification argument injection in the CloudAuth class.

kt474 (commented):
Thanks! Would it be possible to add a simple release note - [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/CONTRIBUTING.md#adding-a-new-release-note) are the instructions from our contributing guide. 

Can you also add a simple unit test? 

Also @Siddharthgolecha is this patch now working as expected for your client?

alodolo (commented):
Hi @kt474, I've added a unit-test and a release note.

qiskit_ibm_runtime/api/auth.py
```diff
@@ -28,21 +30,39 @@
...
 
-    def __init__(self, api_key: str, crn: str, private: bool = False):
+    def __init__(
+        self,
+        api_key: str,
+        crn: str,
+        proxies: Optional[ProxyConfiguration] = None,
+        private: bool = False,
+        verify: bool = True,
+    ):
```
kt474:
```suggestion
        private: bool = False,
        proxies: Optional[ProxyConfiguration] = None,
        verify: bool = True,
    ):
```
NIT: just for consistency we should keep the order the same

kt474 (dismissed):
LGTM Just one minor comment

kt474 (approved):

PR #2347, opened 2025-07-29
Fix twirled slice span serialization to allow backward compatibility

yaelbh:
Consider the following scenario: the server starts to emit twirled slice spans with data slice version 2, and a user uses an older version of qiskit-ibm-runtime, which is not aware of the data slice versions. In that case the runtime decoder at the user's will crash.

This PR fixes it. Although the old version decoder will not crash, it will return the execution spans as a list (and not as an `ExecutionSpans` object); in the list, twirled slice spans of version 2 will not be decoded, but remain in the raw, encoded format.


ihincks (commented):
In lieu of a review, I opened a PR into this branch on your fork, because 1) I didn't want to leave a comment that had a flaw in the idea that would waste your time, and 2) I felt bad leaving a comment along the lines of "actually, can we rethink the version in the argument" after having agreed to it.

Anyways, after having seen some of the shenanigans here, I wondered if things would be a bit cleaner if we versioned the class itself rather than have an attribute with the version number. You can let me know what you think.

yaelbh:
> I'll fix the CI and add a test.

Done. mypy didn't like the code, for good a good reason. I changed a bit to make mypy happier, but it's still not optimal that the type of the data slices is ambiguous and depends on the class type.

ihincks (dismissed):
TYVM!

yaelbh:
I'm afraid it was wrong to merge the PR about mypy. Pub shots can be different for different slices. We can maintain a `_pub_shots` data member which is a dictionary like `_data_slices`, from pub indices to pub shots.

yaelbh:
No, sorry, it's already a dictionary. So it's probably ok.

ihincks (approved):

PR #2348, opened 2025-07-29
Make channel optional if token passed in service initialization

### LINKED ISSUE #2346: Allow for providing a `token` without specifying a `channel` after IQP classic removal

Now that #2289 is merged, we no longer have two different authentication paths, and the `channel` parameter can feel redundant to users. While this parameter is in theory already optional, in practice it's still required if one wants to specify a `token` (which is a very common user path).

I propose removing this special case handling ( https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/qiskit_runtime_service.py#L308-L312
) and allowing the runtime service to be initialized as:

```
service = QiskitRuntimeService(token=token)
# or
service = QiskitRuntimeService(token=token, instance=instance)

```

**Acceptance criteria**

kt474:
### Summary

Since the `ibm_quantum` channel was removed, we can make the channel parameter optional if a token is also passed in because then the channel should be `ibm_quantum_platform`. 

### Details and comments
Fixes #2346

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -77,45 +77,86 @@ class QiskitRuntimeService:
     ) -> None:
         """QiskitRuntimeService constructor.
 
-        An account is selected in the following order:
-            - If a ``filename`` is specified, account details will be loaded from ``filename``,
-                else they will be loaded from the default configuration file.
-            - If ``name`` is specified, the corresponding account details will be loaded from
-                the configuration file, including ``channel``, ``token``, ``instance``, ``region``,
-                ``plans_preference`, and the advanced configuration parameters: ``url``,
-                ``url_resolver``, ``private_endpoint``,  ``verify``, and  ``proxies``.
-            - If ``channel`` is specified, the default account details for that channel will be
-                loaded from the configuration file, else, the account details will be loaded
-                from the ``default_channel`` defined in the configuration file.
-            - Any loaded details will be overwritten by the corresponding parameter in the
-                service constructor.
+        Recommended uses:
+
+        * Direct instantiation:
+
+            .. code-block:: python
+
+                from qiskit_ibm_runtime import QiskitRuntimeService
+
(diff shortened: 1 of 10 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -305,11 +305,6 @@ def _discover_account(
                 if url:
                     logger.warning("Loading default %s account. Input 'url' is ignored.", channel)
                 account = AccountManager.get(filename=filename, name=name, channel=channel)
-        elif any([token, url]):
-            # Let's not infer based on these attributes as they may change in the future.
-            raise ValueError(
-                "'channel' is required if 'token', or 'url' is specified but 'name' is not."
-            )
```
kt474, line 360: Removing this also makes it so `QiskitRuntimeService(url = '...')` works - do we want to keep it this way?

ElePT: Hmmm do you foresee any problems with the url resolution?? in principle we don't modify it if it's externally provided, right?

kt474: There wouldn't be a problem with url resolution, it'd maybe just be a bit unclear if the user tried passing in a different URL. In the cases without `name` and/or `channel`, the url is just ignored (same with the token).
```python 
# any url would work if there is a valid account saved
service = QiskitRuntimeService(url="https://test.com") 

# the token would be ignored too, the service will just use a saved account 
service = QiskitRuntimeService(token="invalid_token") 
```

kt474: We could add a warning instead? Saying that a default or saved account will be used

ElePT: I am not sure we want to allow the case with `service = QiskitRuntimeService(token="invalid_token") ` to run with the default account. I would find it safer to error in this case. Let's have an offline discussion about the different cases and expected behavior.

PR #2352, opened 2025-07-30
Correctly fail unit tests on Qiskit deprecations

kt474:
### Summary

Turns out `module=r"^qiskit$"` only matches the top Qiskit module so only `qiskit/__init__.py` was being matched. `module=r"^qiskit"` correctly matches all the submodules. 

### Details and comments
Fixes #


diego-plan9:
Thanks - closing as this should be tackled during  #2764 in a separate effort.

PR #2356, opened 2025-08-04
Cleanup backend config fields

### LINKED ISSUE #2355: Cleanup backend data fields

`IBMBackend` configuration currently requires these fields to be present when retrieving data from the server:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/6f9148afb979967c0f6fcb8314dc5439fc4e9ebd/qiskit_ibm_runtime/models/backend_configuration.py#L210-L220

This data model was originally created for generic QPUs. Now that it's for IBM Quantum backends only, some of the fields no longer make sense and should not be required: 

-  local (should default to false)
- simulator (should default to false)
- conditional
- open_pulse (should default to false)
- memory - Ideally this would be renamed to something more intuitive, but that's probably too much work. At the very least the description should be clearer than "backend supports memory" lol 

In addition, the docstring here says 

https://github.com/Qiskit/qiskit-ibm-runtime/blob/6f9148afb979967c0f6fcb8314dc5439fc4e9ebd/qiskit_ibm_runtime/ibm_backend.py#L80

which is not accurate. `tags`, for example, is not an attribute of `IBMBackend`. Some of the other attributes, such as `parametric_pulses` are no longer relevant and should be removed. 

**Acceptance criteria**

- Remove backend configuration data that is no longer relevant from the docstring 
- Make required fields that are no longer relevant optional, so the server can stop reporting them.

kt474:
### Summary

Made memory, open_pulse, conditional, simulator, and local optional in `QasmBackendConfiguration` and updated docstrings. 

### Details and comments
Fixes #2355

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -906,6 +906,7 @@ class QiskitRuntimeService:
 
         Raises:
             QiskitBackendNotFoundError: if no backend could be found.
+            IBMInputValueError: if fractional gates are requested but not supported by the backend.
         """
         backends = self.backends(
             name,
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/models/backend_configuration.py
```diff
@@ -188,20 +188,7 @@ def __repr__(self) -> str:
...
-        n_qubits: number of qubits.
-        basis_gates: list of basis gates names on the backend.
-        gates: list of basis gates on the backend.
-        local: backend is local or remote.
-        simulator: backend is a simulator.
-        conditional: backend supports conditional operations.
-        open_pulse: backend supports open pulse.
-        memory: backend supports memory.
-    """
+    """Class representing an OpenQASM 2.0 Backend Configuration."""
```
kt474, line 191: `QasmBackendConfiguration` isn't actually documented live anywhere from qiskit-ibm-runtime - it was last live in https://quantum.cloud.ibm.com/docs/api/qiskit/1.4/qiskit.providers.models.QasmBackendConfiguration as a part of the Qiskit docs. 

Removing the attributes here because they're repeated down below.

PR #2360, opened 2025-08-11
Update account instance warning message

kt474:
### Summary

In the case where there is no saved default instance and an instance is not passed in at initialization, the warning message should be more detailed. We can list the names of all of the instances available in the account. 

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -126,9 +126,12 @@ class QiskitRuntimeService:
             Optional[Callable] url_resolver: Function used to resolve the runtime url.
             Optional[str] region: Set a region preference. Accepted values are ``us-east`` or ``eu-de``.
                 An instance with this region will be prioritized if an instance is not passed in.
-            Optional[List[str]] plans_preference: A list of account types, ordered by preference.
-                An instance with the first value in the list will be prioritized if an instance
-                is not passed in.
+            Optional[List[str]] plans_preference: A list of account plan names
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -534,8 +534,10 @@ def _resolve_cloud_instances(self, instance: Optional[str]) -> List[Tuple[str, L
             return [(default_crn, self._discover_backends_from_instance(default_crn))]
         if not self._all_instances:
             self._all_instances = self._account.list_instances()
+            instance_names = [instance.get("name") for instance in self._all_instances]
             logger.warning(
-                "Default instance not set. Searching all available instances.",
+                "Default instance is not set. All account instances are available: %s.",
+                ", ".join(instance_names),
```
ElePT: What about moving away from the use of the word "default". I think we are overloading it too much. In this case what we care about is that the instance has been set either through a saved account or explicitly during initialization. We could say sth like:

```suggestion
                "Instance was not set at service instantiation. A relevant instance from all available account ",
                "instances will be selected based on the desired action.",
                "Available account instances are: %s.",
                ", ".join(instance_names),
                "\nIf you need the instance to be fixed, set it explicitly."
```

kt474: Thanks! Agree that we should use default less

kt474:
I think we can merge this and work and add the warning message for switching instances in a separate PR.

ElePT (approved):
Sounds reasonable.

Issue #2328, opened 2025-07-17
Add new visualization tool to interpret timing information


ElePT:
**What is the expected feature or enhancement?**

This issue is part of the "Exposing new timing information" epic described in #2327. Assuming there is a mechanism in place to request additional metadata from the primitives, we need a tool to visualize this timing information. 

The interface of the tool will be refined as we get more information on the format of the metadata. But as a first approach, I imagine a naive implementation like:

```
# run primitive
sampler = SamplerV2(mode=backend, options=options_for_timing_info)
job = sampler.run((pub0, pub1, pub2)
result = job.result()

# grab timing info (the tool could handle this directly)
result_timing = result.metadata.get("timing_info", None)
pub_timings = [pub_result.get("timing_info", None) for pub_result in result]

# call visualization tool
visualize_timings(result_timing, pub_timings)

# Insert visualization here
```

**Acceptance criteria**

To implement this feature, we will need: 

- [ ]  A tentative result format for development and testing (to be discussed)
- [ ]  A new visualization module in `qiskit-ibm-runtime` (analogous to qiskit's, it can live in a new `qiskit_ibm_runtime/visualization` directory
- [ ] The `visualize_timings` tool that takes a primitive result/ already gathered timing information and outputs a visualization. This can go into a file e.g. `timing_visualization.py`

RoyElkabetz:
Thank you @ElePT for opening this, the plan sounds good, I will work on making that happen.
A quick question, in your snippet above, what are the expected differences (in terms of types and structure) between `result_timing` and `pub_timings`?

ihincks:
This is neat. If you do end up making new classes, like TimingInfo, please consider the pattern used by ExecutionSpans.draw(), which calls out to a method in the visualization module. This avoids the need for a user to figure out which thing from qiskit_ibm_runtime.visualization to import and lowers the barrier to entry.

ElePT:
> Thank you [@ElePT](https://github.com/ElePT) for opening this, the plan sounds good, I will work on making that happen. A quick question, in your snippet above, what are the expected differences (in terms of types and structure) between `result_timing` and `pub_timings`?

My understanding is that each individual `pub_timing` would contain the corresponding circuit's compilation step timing info, while `result_timing` would contain the overall execution/linking step timing info.

PR #2361, opened 2025-08-15
[WIP] New visualization tool for compiled circuits scheduling timing information

RoyElkabetz:
### Summary

Related issue: #2328 
This Draft is adding a new visualization tool for plotting timing information of compiled quantum circuits.

### Details and comments

This is still a WIP, however, I would appreciate any comments / suggestions on how to best integrate that with the repo existing visualization module and structures.

**NOTE**
The tool is currently using `pandas`, but since it is not a dependency of the repo I will change that to Numpy.


qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
```diff
@@ -0,0 +1,468 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2024.
```
ihincks:
```suggestion
# (C) Copyright IBM 2025.
```

qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
```diff
@@ -0,0 +1,468 @@
...
+                text=f"{gate_name}_{pulse}",
+                textangle=0,
+            )
+        return (trace, annotation)
+
+    def populate_figure(self, fig: PlotlyFigure) -> PlotlyFigure:
+        """
+        TODO: Add docs
+        """
+        self.color_map = dict(zip(self.instruction_set, colors))
```
ihincks: Consider `cycle` if you don't want to truncate the instruction set due to lack of colours. Maybe it's fine.

```suggestion
from itertools import cycle
...

        self.color_map = dict(zip(self.instruction_set, cycle(colors)))
```

qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
```diff
@@ -0,0 +1,468 @@
...
+            mode="markers",
+            hoverinfo="x+text",
+            name=gate_name,
+            text="<br>".join(
+                [
+                    "Instruction: " + instruction,
+                    "Pulse: " + pulse,
+                    "Start: " + str(t_i),
+                    "Finish: " + str(t_f),
+                    "Duration: " + str(t_f - t_i),
```
ihincks: nit: f-strings are more pythonic here. They also would let you format the numbers, if you want (eg `{t_i:0.4f}`)

```suggestion
                    f"Instruction: {instruction}" + instruction,
                    f"Pulse: {pulse}" + pulse,
                    f"Start: {t_i}",
                    f"Finish: {t_f}",
                    f"Duration: {t_f - t_i}",
```

qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
```diff
@@ -0,0 +1,468 @@
...
+        # Process instructions
+        shift_phase_instructions = []
+        for row in self.df.iterrows():
+            (_, _, _, _, _, pulse, gate_name) = self.unpack_row(row)
+            if "shift_phase" not in pulse:
+                # Process instructions of finite duration
+                trace, annotation = self.trace_finite_duration_instruction(row)
+                self.legend.add(gate_name)
+                self.annotations.append(annotation)
+                fig.add_trace(trace)
```
ihincks: If you ever run into performance problems, something that _might_ help is refactoring to call Scatter once, instead of in a loop. The `None` values in the coordinate lists will distinguish separate line segments.

ihincks (commented):
I'd add the main visualization function to the parent `__init__.py` to make it easier to import.

Is the data that comes back from the job going to be a string or some other object?

RoyElkabetz:
Here is a file with scheduling data that can be used for testing.
[circuit_timing_info.txt](https://github.com/user-attachments/files/21801830/circuit_timing_info.txt)

qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
RoyElkabetz: Done in [c482ec0](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/c482ec0923defc1a525ebe234c7e1b32ee84de49).

qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
RoyElkabetz: Done in [c482ec0](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/c482ec0923defc1a525ebe234c7e1b32ee84de49).

qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
RoyElkabetz: Done in [c482ec0](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/c482ec0923defc1a525ebe234c7e1b32ee84de49).

qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
RoyElkabetz: Done in [e5b37c1](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/e5b37c102508d7bbde5f7c5c89d6bb3b303d8587).

jyu00:
It can be a separate PR, but we talked about giving users the option of saving the figure into a file, to make it easier to zoom in etc and not blocking their notebook.

RoyElkabetz:
> It can be a separate PR, but we talked about giving users the option of saving the figure into a file, to make it easier to zoom in etc and not blocking their notebook.

The `draw_circuit_schedule_timing` function currently returns the figure object without displaying it, following the same pattern as other visualization tools in the repo. This gives users flexibility, they can call `fig.show()` to display it or`fig.write_image("<path.format>")` to save it as needed. 

Since Plotly’s [write_image](https://plotly.github.io/plotly.py-docs/generated/plotly.io.write_image.html) offers a lot of customization options, I think it might be clearer and more flexible to let users use that directly rather than adding a wrapper that only covers a small subset of features.

But I am happy to revisit if we feel a wrapper would provide significant convenience for common use cases.

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,364 @@
...
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Utilities for working with circuit schedule timing information returned from the compiler."""
+
+from itertools import cycle
+from typing import Tuple, List, Set, Dict, Union
```
ElePT: From the to-do below:
```suggestion
from typing import Tuple, List, Set, Dict, Union, TYPE_CHECKING
```

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,364 @@
...
+import os.path
+import numpy as np
+from plotly.graph_objects import Figure as PlotlyFigure, Scatter  # TODO: replace with TYPE_CHECkING
```
ElePT: Missing to-do:

```suggestion
if TYPE_CHECKING:
    from plotly.graph_objects import Figure as PlotlyFigure, Scatter
```

ElePT: I recommend importing `Scatter` as `PlotlyScatter` as well (see [ZNE drawer example](https://github.com/Qiskit/qiskit-ibm-runtime/blob/b45f00aacb80a9cfedd25ec8e8d9c3f3399b5971/qiskit_ibm_runtime/visualization/draw_zne.py#L26))

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,364 @@
...
+
+from ..visualization.utils import plotly_module
+
+go = plotly_module(".graph_objects")
+colors = plotly_module(".colors").qualitative.Plotly
+
+
+READOUT_CHANNEL_PREFIX = "AWGR"
```
ElePT: This is for my own understanding, what does this prefix mean? I feel like it should be documented, as other external users might not know it either.

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,364 @@
...
+BARRIER = "barrier"
+
+
+class CircuitSchedule:
```
ElePT, line 38: Question, would we like users to also manipulate this class, or is this intended to simply represent the internal data structure of the drawer function? Asking because:

1. if the idea is to also make this class external facing, it must be added to the docs
2. if the idea is to also make this class external facing, I think that `draw_circuit_schedule_timing` should also accept inputs of type `CircuitSchedule`

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,364 @@
...
+            "x": ((t_i + t_f) / 2),
+            "y": y_mid,
+            "showarrow": True,
+            "font": {"color": "black", "size": 10},
+            "text": f"{gate_name}_{pulse}",
+            "textangle": 0,
+        }
+        self.annotations.append(annotation)
+
+    def populate_figure(self, fig: PlotlyFigure) -> PlotlyFigure:
```
ElePT, line 354: I tried running the unit test for populate_figure and I noticed that the figure has no axes:

Screenshot 2025-09-09 at 10 45 28

Noting this because, if the idea is to expose this class to users, I think that the units should be specified in the axes.

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,364 @@
...
+            else:
+                raise FileExistsError(f"{circuit_schedule} is missing.")
+        elif isinstance(circuit_schedule, list):
+            data = circuit_schedule
+        else:
+            raise TypeError("CircuitSchedule expects a str or a list.")
+
+        return data
+
+    def _parse(self, raw_data: List[str]) -> None:
```
ElePT, line 103: even if the class is not intended to be user facing, I think that we would benefit from having a better description of the different fields stored in the class, what they mean, and how they are manipulated. This could be added to the init as well as to the relevant class methods.

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,364 @@
...
+
+        raw_data = self._load(circuit_schedule)
+        self._parse(raw_data)
+
+        self.instruction_set: Set[str] = set()
+        self.max_time: int = None
+        self.color_map: Dict[str, str] = {}
+        self.annotations: List[Dict] = []
+        self.legend: Set[str] = set()
+        self.traces: List[Scatter] = []
```
ElePT, line 83: Naive question: what is a trace?

test/unit/visualization/test_draw_circuit_schedule_timings.py
```diff
@@ -0,0 +1,60 @@
...
+        for _, test_case in self.test_cases.items():
+            ((included_channels, filter_readout_channels, filter_barriers), (n_traces)) = test_case
+            fig = draw_circuit_schedule_timing(
+                circuit_schedule=circuit_schedule,
+                included_channels=included_channels,
+                filter_readout_channels=filter_readout_channels,
+                filter_barriers=filter_barriers,
+            )
+            self.assertEqual(len(fig.data), n_traces)
+        self.save_plotly_artifact(fig)
```
ElePT, line 59: Is this saved artifact used anywhere?

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: The functionality is set in a way that the `draw_circuit_schedule_timing()` function does all the layout updating.
The `CircuitSchedule` class is only adding the traces to the figure, but is not responsible on the layout of the figure. So in order to get the axes labels, title and such you need to call the `draw_circuit_schedule_timing()` on your data.

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: The intent is that any kind of manipulation should happen to the data itself (if needed) and not through the class. 
The purpose of the class is to only represent the internal data for plotting, and to also have some data structure that internal ibm tools can import and inherit from so we could test some new functionality before it is being open-sourced.

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: A trace is a single instruction scatter which is scattered using [scatter Traces](https://plotly.com/python/reference/scatter/).

test/unit/visualization/test_draw_circuit_schedule_timings.py
RoyElkabetz: This is a good question 😄 , I was following the unit-tests `draw` functions general structure.
Should I change that to something else if I want to verify the plot is being generated?

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: Arbitrary Wave Generator Readout (AWGR).
I have added a comment in [8311cda](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/8311cda2d0e716b9f280f948d92abbaa3178d0cc).

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: I have added attributes description in [3a0a6c5](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/3a0a6c5bb2bef7730b6bfc8590b250a1046313a8).

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,389 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Utilities for working with circuit schedule timing information returned from the compiler."""
```
mberna: I wouldn't use the word _compiler_ here, could be confused with `qiskit.compiler`. I would say "Qiskit Runtime service" or something similar.

test/unit/visualization/test_draw_circuit_schedule_timings.py
```diff
@@ -0,0 +1,60 @@
...
+        self.circuit_schedule_data = fake_sampler_pub_result.metadata["scheduler_timing_info"][
+            "timing"
+        ].split("\n")
+
+        # test cases structure:
+        #   --------------------   Input Arguments   ------------------       ---  Expected Results  ---
+        # ((included_channels, filter_readout_channels, filter_barriers)     ,     (n_traces))
+        self.test_cases = {
+            1: ((None, False, False), (104)),
+        }
```
mberna: are you planning on adding more tests? Is a dictionary necessary here? If not, I would keep it simple...

test/unit/visualization/test_draw_circuit_schedule_timings.py
mberna: I would rethink this test. The test description says "it makes sure that it produces the right figure" but the assert is only on the number of traces. I agree with Elena that if the artifact is saved, then it should be used somewhere.

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,389 @@
...
+            circuit_schedule: A schedule data as a list of strings where each string corresponds to
+            a single instruction schedule, or a path to a file containing the schedule data.
+
+        Returns:
+            Circuit schedule data.
+        """
+        if isinstance(circuit_schedule, str):
+            if os.path.exists(circuit_schedule) and os.path.isfile(circuit_schedule):
+                with open(circuit_schedule, encoding="utf-8") as file:
+                    data = file.readlines()
```
mberna: Is there corresponding functionality to save the schedule to a file or the user needs to manually copy and paste the schedule from the metadata to a file?

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -0,0 +1,389 @@
...
+            else:
+                raise FileExistsError(f"{circuit_schedule} is missing.")
```
mberna:
```suggestion
                raise FileNotFoundError(f"{circuit_schedule} is missing.")
```

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: Added in [f9785ec](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/f9785ece432cc6fa118d60df10df873338ed981b).

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: Added in [f9785ec](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/f9785ece432cc6fa118d60df10df873338ed981b).

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: Changed in [ec37c6b](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/ec37c6b2de6f4630b98af9463d5d3dd30242fd49).

test/unit/visualization/test_draw_circuit_schedule_timings.py
RoyElkabetz: Added another test realization and added a `ddt` use in [ec37c6b](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/ec37c6b2de6f4630b98af9463d5d3dd30242fd49).

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: I didn't plan to add a `save` function for the data, but it can certainly be added in a follow up PR if we think it could be useful.
However, the data would be saved after processing, so in its current state the `CircuitSchedule` will not be able to consume it because it expects the data to be in its original format (as returned from the server).

qiskit_ibm_runtime/utils/circuit_schedule.py
RoyElkabetz: Changed in [17441cf](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/17441cf8623e3d644ee3d7de6712c3dad129465c).

test/unit/visualization/test_draw_circuit_schedule_timings.py
RoyElkabetz: I went over all the other unit tests that uses that method and couldn't find any checks regarding the saved figure content (or any loads actually). As far as I understand, the purpose of this method is to check that the figure is generated without errors.

test/unit/test_circuit_schedule.py
```diff
@@ -0,0 +1,199 @@
...
+@ddt.ddt
+class TestCircuitSchedule(IBMTestCase):
+    """Tests for CircuitSchedule class."""
+
+    def setUp(self) -> None:
+        """Set up."""
+        fake_sampler_pub_result = FakeCircuitScheduleInputData.sampler_pub_result
+        self.circuit_schedule_data = fake_sampler_pub_result.metadata["compilation"][
+            "scheduler_timing"
+        ]["timing"].split("\n")
```
jyu00: nit - I should be able to pass the `timing` metadata directly to the drawer, without having to do `.split("\n")`

RoyElkabetz: We can certainly make that change, and I totally agree it make more sense to have the tool supporting the format of the data "as is" without needing extra reformatting. 
However, you should know that the data format and content are still in development, which means it may change in the near future. So making that change would tie the tool to the data format in a potentially breakable way. 

We can still make that change and hope for the best 😄 , thoughts?

ElePT (approved):
On my side, the PR looks in a good state to merge. Any missing points you think should block the merge to main @jyu00 @mberna? I assume there are elements that could be tweaked but it might make more sense to do so in follow ups.

test/unit/test_circuit_schedule.py
jyu00: Even if the format changes in the future, ideally people will still be able to use the data "as is" (with the right version of qiskit-ibm-runtime). So I'd say make the change now if it's not too difficult.

ElePT: I agree with Jessie, I think that the change could be as simple as allowing for the full string in `circuit_schedule`:

```
fig = draw_circuit_schedule_timing(
            circuit_schedule=circuit_schedule,
            included_channels=included_channels,
            filter_readout_channels=filter_readout_channels,
            filter_barriers=filter_barriers,
        )
```

and then splitting it inside `draw_circuit_schedule_timing`. This internal logic can be updated to accept changes in the string without changing much the interface.

RoyElkabetz: So at the moment if a string is passed (instead of a list of strings) then this is inferred as a path to a file which contains the data. So I can make that change, but it is not as simple as a quick "find and replace" since I would have to find another way to pass a file (with the schedule data in it).

**Option 1:**

Although it would be a bit ugly, I can add another input argument for a file path if you are ok with that. Essentially breaking the data coming in into two distinct arguments:
- `circuit_schedule: str` - the string data coming from the compiler
- `schedule_file: str` - a file containing the string data

And removing the option of having `circuit_schedule: List[str]` altogether. 

Thoughts?

**Option 2:**

Removing the ability to have a file at the input and then we can use the `str` type as the data instead of a path.

Any preference? or any other suggestion?

RoyElkabetz: Option 2 is implemented with [54ef365](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/54ef3653cfa674f033f1c7e9d048a00c891e6bda).

release-notes/unreleased/2328.feat.rst
```diff
@@ -0,0 +1,23 @@
+Added the new function :meth:`.visualization.draw_circuit_schedule_timing` to plot circuit schedule 
+timing data retrieved from a job when the ``"scheduler_timing"`` execution feature flag is 
+set to ``True``,
```
jyu00: Let's be more specific about the restrictions, e.g. 


> Added a new function :meth:`.visualization.draw_circuit_schedule_timing` to plot circuit schedule timing data returned in job result metadata. This is currently a beta feature and requires the ``"scheduler_timing"`` experimental option to be set to ``True``, as shown below. This is currently only available to ``Sampler`` jobs.

jyu00 (approved):
A minor common on the release note, but otherwise looks good. Thanks @RoyElkabetz!

release-notes/unreleased/2328.feat.rst
RoyElkabetz: Sure, no problem, updated in [eb23ff1](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2361/commits/eb23ff1538a1b339bed16cfc311eb5ae32476532).

PR #2366, opened 2025-08-18
Remove control flow instruction filter when `use_fractional_gates=true`

### LINKED ISSUE #2363: Stop filtering out control flow instructions when use_fractional_gates=true

IBM QPUs now support running a job that contains both fractional gates and conditionals. Hence qiskit-ibm-runtime should stop filtering out control flow instructions when `use_fractional_gates=True` is specified when choosing a backend. 

**Acceptance criteria**

- Control flow instructions are not filtered out when `use_fractional_gates=True` is specified.

kt474:
### Summary

Now that backends support both fractional gates and dynamic circuits we don't need to filter out the control flow instructions when `use_fractional_gates=true` is used. 

### Details and comments
Fixes #2363


qiskit_ibm_runtime/utils/backend_decoder.py
```diff
@@ -86,11 +84,6 @@ def filter_raw_configuration(
...
     gate_map = get_standard_gate_name_mapping()
     if use_fractional_gates:
-        raw_config["conditional"] = False
-        if "supported_instructions" in raw_config:
-            raw_config["supported_instructions"] = [
-                i for i in raw_config["supported_instructions"] if i not in CONTROL_FLOW_OP_NAMES
-            ]
         if "supported_features" in raw_config:
             raw_config["supported_features"] = [
                 g for g in raw_config["supported_features"] if g != "qasm3"
```
jyu00: You don't need this filter either. For historical reasons, `qasm3` in `supported_features` indicates whether the backend supports DC.

ElePT: I think this is worth a comment in the code for future reference

jyu00: fwiw we changed it to `dynamic_circuits` a while ago. But `qasm3` was kept for backward compatibility, since it's used by qiskit-ibm-runtime:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/3d78e094c7c597f4050b797d5ea8c6152741e6cc/qiskit_ibm_runtime/qiskit_runtime_service.py#L499-L506

Although apparently I never opened the issue for runtime to switch to the new flag 🤦‍♀️

kt474: Updated to `dynamic_circuits` in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2366/commits/df5637947b75410be774f9e67d0644df03d968b6

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -464,12 +464,6 @@ def backends(
             IBMInputValueError: If an input is invalid.
             QiskitBackendNotFoundError: If the backend is not in any instance.
         """
-        if dynamic_circuits is True and use_fractional_gates:
-            raise QiskitBackendNotFoundError(
```
jyu00, line 566: I can't comment on those lines, but the docstring needs to be updated as well.

jyu00:
> For `get_translation_stage_plugin()` I think we'll have to deprecate it first even though it's not used anywhere
> 
> https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/ibm-backend#get_translation_stage_plugin

Isn't it automatically picked up when transpiling using an IBMBackend ? The `IBMFractionalTranslationPlugin` says 

https://github.com/Qiskit/qiskit-ibm-runtime/blob/d44ecbde9845160b4706b8a1a9ce2d40716fa356/qiskit_ibm_runtime/transpiler/plugin.py#L120-L121

which sounds to me trying to transpile with both fractional gates & control flow may blow up.

kt474:
> https://github.com/Qiskit/qiskit-ibm-runtime/blob/3d78e094c7c597f4050b797d5ea8c6152741e6cc/qiskit_ibm_runtime/ibm_backend.py#L496-L500

@nkanazawa1989 How should this be handled now that our backends support jobs with both fractional gates are conditionals?

pyproject.toml
```diff
@@ -98,6 +98,7 @@ dependencies = [
 ibm_backend = "qiskit_ibm_runtime.transpiler.plugin:IBMTranslationPlugin"
 ibm_dynamic_circuits = "qiskit_ibm_runtime.transpiler.plugin:IBMDynamicTranslationPlugin"
 ibm_fractional = "qiskit_ibm_runtime.transpiler.plugin:IBMFractionalTranslationPlugin"
+ibm_conditional_and_fractional = "qiskit_ibm_runtime.transpiler.plugin:IBMDynamicFractionalTranslationPlugin"
```
nkanazawa1989: nitpick; can this be "dynamic"? I'm fine if you prefer conditional, but the mix of conditional and dynamic may confuse the other developers.

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -495,9 +495,7 @@ def run(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
 
     def get_translation_stage_plugin(self) -> str:
         """Return the default translation stage plugin name for IBM backends."""
-        if not self.options.use_fractional_gates:
-            return "ibm_dynamic_circuits"
-        return "ibm_fractional"
+        return "ibm_conditional_and_fractional"
```
nkanazawa1989: I think you handle this correctly. Probably you can also deprecate (or pending deprecation) `IBMDynamicTranslationPlugin` and `IBMFractionalTranslationPlugin` as they no longer used standalone.

release-notes/unreleased/2366.feat.rst
```diff
@@ -0,0 +1,2 @@
+Since backends now support running jobs that contain both fractional gates and conditionals, control flow
```
nkanazawa1989:
```suggestion
Since backends now support running jobs that contain both fractional gates and dynamic circuits, control flow 
```

maybe?

kt474:
> This PR looks good to me. Just minor comments.

Thanks! Addressed comments and deprecated `IBMDynamicTranslationPlugin` and `IBMFractionalTranslationPlugin`

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -495,9 +495,7 @@ def run(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
 
     def get_translation_stage_plugin(self) -> str:
         """Return the default translation stage plugin name for IBM backends."""
-        if not self.options.use_fractional_gates:
-            return "ibm_dynamic_circuits"
-        return "ibm_fractional"
+        return "ibm_dynamic_and_fractional"
```
ElePT, line 500: If fractional gates can still be disabled (correct me if I'm wrong), I think that it still make senses to keep the `ibm_dynamic_circuits` standalone plugin, right? So having something like:

```suggestion
        if not self.options.use_fractional_gates:
                return "ibm_dynamic_circuits"
        return "ibm_dynamic_and_fractional"
```

In this case I would "undeprecate" the ibm_dynamic_circuits plugin.

ElePT (commented):
Thanks Kevin! I have one final comment after taking a look at the whole PR again. I don't know why I assumed that `use_fractional_gates` would no longer be used, but now I realized that we still have 2 cases for the transpiler plugin.

PR #2369, opened 2025-08-19
Use job running time in job.properties()

### LINKED ISSUE #2368: Job properties should use running time

`job.properties()` returns the backend properties used for the job. However, it gets the properties at the job creation time:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/93b08fcc320ba6da1f6e9d79cad2e237d0a41fab/qiskit_ibm_runtime/base_runtime_job.py#L197

People using this method are most likely interested in the properties used during circuit execution. Hence it should use the job running time (which is still not exact but much closer). 

**Acceptance criteria**

- Use the job running time instead of creation time when retrieving properties file for the job

kt474:
### Details and comments
Fixes #2368


ElePT (approved):
LGTM

PR #2375, opened 2025-08-22
Update instance selection behavior 

### LINKED ISSUE #2346: Allow for providing a `token` without specifying a `channel` after IQP classic removal

### LINKED ISSUE #2374: Selection of instance enhancements

This issue continues from:

- https://github.com/Qiskit/qiskit-ibm-runtime/issues/2236

We’ve received feedback that users often cannot tell whether a service instance is automatically selected or explicitly specified. Additionally, we want to ensure that automatic selection excludes `pay-as-you-go` or other paid plans by default, unless the user explicitly opts in (by providing a CRN or setting a different plan preference).

**Acceptance criteria**

- When no instance is specified, the system should display a warning message indicating:
  - An instance has not been specified.
  - One will be automatically selected using either the default filters/constraints or the user’s configured filters/constraints (list the actual filters that are going to be applied).

- When an instance is automatically selected, the system should display a warning message with:
  - The name of the selected instance.
  - The plan type of the selected instance.

- Default plan preferences should exclude any paid plans.
  - Use the Global Catalog API [Get child catalog entries of a specific kind](https://cloud.ibm.com/apidocs/resource-catalog/global-catalog#get-child-objects)
 with:
    - Catalog ID: `b6049020-80f4-11eb-a0f7-e35ec9b4054f`
    - Kind: plan
  - From the response, check `metadata.pricing.type`. Plans can be free, trial, paygo, paid, or subscription.
  - By default, only free and trial plans should be included.

kt474:
### Summary

**When no instance is specified:** 

```python
QiskitRuntimeService()
```

Warning: 
`Instance was not set at service instantiation. Free and trial plan instances will be prioritized. Based on the following filters, (tags: None, region: us-east, eu-de) and available plans: (premium, internal), the available account instances are: services-internal-eu, services-premium-eu, services-internal-us, services-premium-us. If you need a specific instance set it explicitly either by using a saved account with a saved default instance or passing it in directly to QiskitRuntimeService().`

**Filtering by `region` and `plans_preference`:** 

```python
QiskitRuntimeService(region="us-east", plans_preference=["internal"])
```

Warning:
`Instance was not set at service instantiation. Based on the following filters, (tags: None, region: us-east, plans_preference: internal), the available account instances are: services-internal-us. If you need a specific instance set it explicitly either by using a saved account with a saved default instance or passing it in directly to QiskitRuntimeService()`

 if `plans_preference` is given, it will be added to list of filters in the warning message. If `plans_preference` is not given, the available plans will be listed instead. 

**Retrieving backends:**

```python
service.backends()
// or 
service.backend('...')
```

Warning: 
`Loading instance: services-internal-eu, plan: internal`

**Using saved account vs. new account**

If using saved account: 

`Loading default saved account` or `Loading saved account: quantum-internal`

If token is passed in: 

`Loading account with the given token. A saved account will not be used.`

### Details and comments
Fixes #2374 
Fixes #2346

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -244,7 +244,7 @@ class QiskitRuntimeService:
                 plans_preference_str = f", plans_preference: {joined_preferences})"
             else:
                 joined_plan_names = ", ".join(instance_plan_names)
-                plans_preference_str = f") , and available plans: ({joined_plan_names})"
+                plans_preference_str = f"), and available plans: ({joined_plan_names})"
 
             filters = f"(tags: {tags_str}, " f"region: {region_str}" f"{plans_preference_str}"
 


kt474:
@ElePT Now that the warning message wording is more or less finalized, I think this is ready for a review

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -393,10 +393,19 @@ def _discover_account(
...
+                url=url,
+                instance=instance,
+                proxies=proxies,
+                verify=verify_,
+            )
+            logger.warning("Loading new account. A saved account will not be used.")
+        elif url:
             raise ValueError(
-                "'channel' is required if 'token', or 'url' is specified but 'name' is not."
+                "'url' is not valid as a standalone parameter. Try also passing in 'token' or 'name'."
```
ElePT, line 412: I think this is a good solution!

ElePT (commented):
Account intialization by token looks good. Let's have an offline discussion regarding instance selection filters before moving forward with the other changes.

kt474:
Updated the PR description so all warning messages are up to date. The behavior has changed that so when no instance is given and `plans_preference` is not set, the available instances will be prioritized by instance pricing type: `free, trial, paygo, paid, subscription`.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -231,6 +232,37 @@ def __init__(
...
+            else:
+                joined_plan_names = ", ".join(instance_plan_names)
+                plans_preference_str = f") and available plans: ({joined_plan_names})"
+
+            filters = f"(tags: {tags_str}, " f"region: {region_str}" f"{plans_preference_str}"
+
+            logger.warning(
+                "Instance was not set at service instantiation. %s"
+                "Based on the following filters, %s, "
+                "the available account instances are: %s. "
```
ElePT: I think that a colon makes sense here:

```suggestion
                "Instance was not set at service instantiation. %s"
                "Based on the following filters: %s, "
                "the available account instances are: %s. "
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -231,6 +232,37 @@ def __init__(
...
+            }
+
+            tags_str = ", ".join(self._tags) if self._tags else "None"
+            region_str = self._region if self._region else "us-east, eu-de"
+            if self._plans_preference:
+                joined_preferences: str = ", ".join(self._plans_preference)
+                plans_preference_str = f", plans_preference: {joined_preferences})"
+            else:
+                joined_plan_names = ", ".join(instance_plan_names)
+                plans_preference_str = f") and available plans: ({joined_plan_names})"
```
ElePT: oxford comma:

```suggestion
                plans_preference_str = f") , and available plans: ({joined_plan_names})"
```

iyanmv:
Why would you print warnings for completely normal behavior like initializing the service with a saved account? This doesn't make any sense to me... By the way, many Jupyter notebooks will have diffs every time they are executed because of the timestamp in those warnings, just by doing `service = QiskitRuntimeService()`.

Would you consider reverting, at least, this line?
https://github.com/Qiskit/qiskit-ibm-runtime/blob/1706d9400625a800b3d432852e6a5ae40d168b17/qiskit_ibm_runtime/accounts/management.py#L185

ElePT:
> Why would you print warnings for completely normal behavior like initializing the service with a saved account? This doesn't make any sense to me... By the way, many Jupyter notebooks will have diffs every time they are executed because of the timestamp in those warnings, just by doing `service = QiskitRuntimeService()`.
> 
> Would you consider reverting, at least, this line?
> 
> https://github.com/Qiskit/qiskit-ibm-runtime/blob/1706d9400625a800b3d432852e6a5ae40d168b17/qiskit_ibm_runtime/accounts/management.py#L185

Thanks for your feedback. You can refer to the [issue](https://github.com/Qiskit/qiskit-ibm-runtime/issues/2374) for more context on why this feature was requested. Balancing different user proficiencies is tricky, it becomes trickier when a migration takes place under the hood and things look the same even when the inner workings aren't. That being said, we will consider your suggestion.

PR #2389, opened 2025-09-04
Add rep_delay validation

### LINKED ISSUE #2387: Validate rep_delay value during submission of a job
**Describe the bug**

Currently, IBM backends do not support `rep_delay` values less than `20e-6s`. However, the runtime client allows users to submit jobs below the minimum value supported. These jobs would eventually fail during execution, raising `6050` errors.

**Steps to reproduce**
```
service=QiskitRuntimeService(channel='ibm_quantum_platform')
backend = service.backend("ibm_brisbane")

num_circuits = 2

qc = QuantumCircuit(2)
qc.x(0)
qc.x(1)
qc.measure_all()

pm = generate_preset_pass_manager(optimization_level=3,backend=backend)
tqc = pm.run([qc]*2)

sampler = Sampler(mode=backend)
sampler.options.execution.rep_delay = 0
job = sampler.run(pubs=tqc,shots=10)
print(job.result())

```

Returns the following error message while retrieving the job results:

` RuntimeJobFailureError: 'Unable to retrieve job result. Error code 6050; Failed to execute program: Input Error: Rep delay 0 outside of valid range 0.000019999999999999998 - 0.0005 `

**Suggested solutions**
It would be good to validate the `rep_delay` value before sending the job to the backend. Allow jobs where `rep_delay` falls within the `backend.rep_delay_range`.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.41.1
- **Python version**: 3.13.2
- **Operating system**: MacOS Sequia 15.6

kt474:
### Summary

Adding client side validation to make sure `rep_delay` value is within the backend's `rep_delay_range`. 

### Details and comments
Fixes #2387


qiskit_ibm_runtime/sampler.py
```diff
@@ -114,10 +114,14 @@ def _validate_options(self, options: dict) -> None:
...
+            ValueError: if validation fails.
         """
-
-        pass
+        if isinstance(rep_delay := options.get("execution", {}).get("rep_delay"), (int, float)):
+            rep_delay_range = self._backend.configuration().rep_delay_range
+            if rep_delay < rep_delay_range[0] or rep_delay > rep_delay_range[1]:
+                raise ValueError(
+                    f"rep_delay {rep_delay} is not in the backend rep_delay_range {rep_delay_range}"
+                )
```
kt474, line 124: I didn't want to repeat this code but I don't think there's a way to do this directly in the `ExecutionOptions` because we need to check the backend config for the `rep_delay_range`

ElePT (commented):
I am in favor of having checks, but I am wondering which validations should be the client's responsibility and which should be the validation service's responsibility. I would have expected this to be one of the parameters checked in the validation process. I think that the overall pipeline should be synchronized a bit better.

diego-plan9:
Thanks. I agree with the sentiment in the last comment: there is a balance between the number of validations client side / server side / user expectations and in general is a trade-off.

Combined with the upcoming move towards a new style of execution and rewriting of the Primitives, closing this in favor of handling `rep_delay` validation either at user side, or with changes server-side.

PR #2399, opened 2025-09-11
Set `exlcude_params` to True when possible

### LINKED ISSUE #2394: Update `exclude_params` when fetching jobs
Since we removed `RuntimeJob` and all primitives should be returned as `RuntimeJobV2`, I don't think we need to have `exclude_params=False` as the default behavior when retrieving jobs. This should speed up retrieving jobs. 

Related issues: 
https://github.com/Qiskit/qiskit-ibm-runtime/issues/1307
https://github.com/Qiskit/qiskit-ibm-runtime/issues/1471

kt474:
### Summary

For retrieving jobs with `job()` and `jobs()` we don't need to set `exclude_params` to True anymore because we don't need to check for the job version. All jobs should be `RuntimeJobV2`. This should result in a small performance improvement. 

### Details and comments
Fixes #2394


ElePT (approved):
LGTM

PR #2402, opened 2025-09-16
Remove references in release notes to old APIs

Eric-Arellano:
As part of https://github.com/Qiskit/documentation/issues/785, we are removing very old versions of Qiskit Runtime docs like 0.19. (This is to address some performance issues the web app is having from having so many API pages)

So, we need to remove links that will no longer exist because the APIs are not in modern versions.


ElePT (approved):
LGTM

PR #2406, opened 2025-09-17
Job methods should not trigger new instance warnings

kt474:
### Summary

I just noticed that the warnings we added in #2375 are triggered by `jobs()` and `job()` and I don't think we want that. We don't need to show the list of instances used in a warning message when retrieving jobs. 

### Details and comments
Fixes #


ElePT (commented):
We should have unit tests for warnings. It's likely that this is not the only place where we are over-warning when changing instances/

PR #2408, opened 2025-09-18
Raise an error for invalid tokens and mismatched accounts

kt474:
### Summary

- Added an error for when an invalid token is given with an instance. Previously this would not raise an error and the initialization level because no API calls that require a token are made. 

- Added another error for when the given instance is not from the same account associated with the API token. The API does allow this token/instance mismatch but it will just result in further confusion. 

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -112,6 +112,9 @@ class QiskitRuntimeService:
         and ``tags``. If ``plans_preference`` is not set, free and trial instances will be prioritized
         over paid instances.
 
+        Also note that only one account per API token can be used. The API token is linked to the
+        account it was created in. If you want to use multiple accounts, you must create multiple API tokens.
+
         The service will attempt to load an account from file if (a) no explicit ``token``
         was provided during instantiation  or (b) a ``name`` is specified, even if an explicit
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -226,6 +230,13 @@ def __init__(
         self._plans_preference = plans_preference or self._account.plans_preference
         self._tags = tags or self._account.tags
         if self._account.instance:
+            if self._account.instance not in [inst["crn"] for inst in self.instances()]:
+                raise IBMInputValueError(
+                    "The given API token is associated with an account that does not have access to "
+                    f"the instance {self._account.instance}. "
+                    "To use this instance, use an API token generated from the account "
+                    "with this instance available."
+                )
```
kt474, line 239: This handles both cases - it will error if an invalid token is passed in an instance crn because of the `self.instances()` call

release-notes/unreleased/2408.update.rst
```diff
@@ -0,0 +1,4 @@
+There will now be a proper error if an invalid API token is given in :class:`.QiskitRuntimeService`.
```
ElePT: Tiny comment, let's say consistent instead of proper:

```suggestion
A consistent error will now be raised during service initialization if an invalid API token is given to :class:`.QiskitRuntimeService`. 
```

PR #2411, opened 2025-09-19
Remove status reason check in `least_busy`

kt474:
### Summary

The status reason field in `/backends` is a free form optional string so we should not depend on it. This issue was discovered because the `reason` returned as both "Available" with an uppercase first letter and "available". 

https://quantum.cloud.ibm.com/api/docs/

### DIFF
--- a/release-notes/unreleased/2411.fix.rst
+++ b/release-notes/unreleased/2411.fix.rst
@@ -1,4 +1,1 @@
-The :meth:`.QiskitRuntimeService.least_busy` method was returning no backends or the incorrect backend because the 
-API response we were depending on was returning strings with an uppercase first letter while we were expecting 
-a fully lowercase string. This has been fixed, we no longer depend on this field. This field is labled as an optional
-string in the backend schema so we never should have depended on it in the first place.
+The :meth:`.QiskitRuntimeService.least_busy` method's behavior was flaky because it expected a fixed response from the API for the `"reason"` field. This has been fixed by no longer depending on this field, as it is labeled as optional in the API schema.


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1266,7 +1266,7 @@ def least_busy(
 
         candidates = []
         for backend in all_backends:
-            if backend["status"]["name"] == "online" and backend["status"]["reason"] == "available":
+            if backend["status"]["name"] == "online":
```
mriedem, line 1280: Which API is this response coming from? This one? https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/backends

Because I don't see `status` as being an object, but instead it's a string:

image

Or is it `GET /backends` https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/backends ?

image

I guess it must be the latter.

mriedem (commented):
The code change looks OK to me, but is there no test for this code?

qiskit_ibm_runtime/qiskit_runtime_service.py
kt474: It's https://quantum.cloud.ibm.com/api/v1/backends 

Screenshot 2025-09-23 at 12 00 35 PM

test/unit/test_backend_retrieval.py
```diff
@@ -118,23 +118,30 @@ def test_filter_least_busy(self):
...
+                "name": "test_backend4",
+                "status": {"name": "online", "reason": "available"},
+                "queue_length": 15,
+            },
         ]
         fake_backends = [
             self._get_fake_backend_specs(**{**default_stat, "backend_name": "test_backend1"}),
             self._get_fake_backend_specs(**{**default_stat, "backend_name": "test_backend2"}),
+            self._get_fake_backend_specs(**{**default_stat, "backend_name": "test_backend3"}),
+            self._get_fake_backend_specs(**{**default_stat, "backend_name": "test_backend4"}),
```
mriedem, line 144: OK so test_backend3 should be filtered out because it's offline, and of the other 3, test_backend1 has the shortest queue. LGTM.

mriedem: Also, if we reverted the code fix, test_backend1 and test_backend2 would also be filtered out and the test would fail because it'd pick test_backend4, so it's a good regression test.

release-notes/unreleased/2411.fix.rst
```diff
@@ -0,0 +1,4 @@
+The :meth:`.QiskitRuntimeService.least_busy` method was returning no backends or the incorrect backend because the 
+API response we were depending on was returning strings with an uppercase first letter while we were expecting 
+a fully lowercase string. This has been fixed, we no longer depend on this field. This field is labled as an optional
+string in the backend schema so we never should have depended on it in the first place.
```
ElePT: I would keep the release note more high level here:

```suggestion
The :meth:`.QiskitRuntimeService.least_busy` method's behavior was flaky because it expected a fixed response from the API for the `"reason"` field. This has been fixed by no longer depending on this field, as it is labeled as optional in the API schema.
```

PR #2412, opened 2025-09-19
Update `test_rzz_validation_param_exp`

kt474:
### Summary

Seems like only on python 3.12 and Qiskit 2.2, this test fails because the order of values in the error message [here](https://github.com/qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/utils/utils.py#L206) is not guaranteed. 

### Details and comments
Fixes #

### DIFF
--- a/test/unit/test_sampler.py
+++ b/test/unit/test_sampler.py
@@ -328,7 +328,10 @@ class TestSamplerV2(IBMTestCase):
         if val2 == 0:
             SamplerV2(backend).run(pubs=[(circ, [val1, val2])])
         else:
-            with self.assertRaisesRegex(IBMInputValueError, f"p2={val2}, p1={val1}"):
+            # order of the values is not guaranteed
+            with self.assertRaisesRegex(
+                IBMInputValueError, (rf"p2={val2}, p1={val1}" rf"|p1={val1}, p2={val2}")
+            ):
                 SamplerV2(backend).run(pubs=[(circ, [val1, val2])])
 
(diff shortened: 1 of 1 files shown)


kt474:
@yaelbh If you get a change can you take a look? [Here's](https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/17875604872/job/50836553039#step:5:2605) an example of the test failing. We're checking for `p2=2.0, p1=1.0` but `p1=1.0, p2=2.0` is in the error message.

test/unit/test_sampler.py
```diff
@@ -328,7 +328,10 @@ def test_rzz_validation_param_exp(self, val1, val2):
         if val2 == 0:
             SamplerV2(backend).run(pubs=[(circ, [val1, val2])])
         else:
-            with self.assertRaisesRegex(IBMInputValueError, f"p2={val2}, p1={val1}"):
+            # order of the values is not guaranteed
+            with self.assertRaisesRegex(
+                IBMInputValueError, (rf"p2={val2}, p1={val1}" rf"|p1={val1}, p2={val2}")
```
yaelbh, line 333: How do you parse this line? How about
```python
rf"(p2={val2}, p1={val1}|p1={val1}, p2={val2})"
```

yaelbh (commented):
Thanks @kt474   (-:

test/unit/test_sampler.py
ElePT: I think they are equivalent, both strings are just concatenated.

yaelbh (approved):

PR #2432, opened 2025-10-07
[WIP] Add calibration_id support

mtreinish:
### Summary

This commit adds support for setting a custom calibration id on a backend when instantiating it. This custom calibration will be used for constructing the target and also used when executing primitive jobs on the backend.

### DIFF
--- a/test/unit/test_backend_retrieval.py
+++ b/test/unit/test_backend_retrieval.py
@@ -278,3 +278,15 @@ class TestGetBackend(IBMTestCase):
         self.assertIn("rx", backend_with_fg.target)
 
         self.assertIsNot(backend_with_fg, backend_without_fg)
+
+    def test_backend_with_custom_calibration(self):
+        """Test getting a backend with a custom calibration."""
+        service = FakeRuntimeService(
+            channel="ibm_quantum",
+            token="my_token",
+            backend_specs=[FakeApiBackendSpecs(backend_name="FakeFractionalBackend")],
+        )
+
+        backend_with_calibration = service.backend("fake_torino", calibration_id="abc1234")
+        self.assertEqual(backend_with_calibration.calibration_id, "abc1234")
+        # TODO: Assert mock has api client calls with cal id set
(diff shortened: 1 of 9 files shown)


test/integration/test_backend.py
```diff
@@ -271,3 +271,13 @@ def test_renew_backend_properties(self):
...
+
+    def test_backend_calibration_id(self):
+        """Test calibration_id is used when fetching the configuration."""
+        name = self.backend.name
+        calibration_id = "invalid_id"
+        with self.assertLogs("qiskit_ibm_runtime", level="WARNING") as log:
+            with self.assertRaises(QiskitBackendNotFoundError):
+                self.service.backend(name, calibration_id=calibration_id)
+
+        self.assertTrue(any(calibration_id in record for record in log.output))
```
kt474, line 283: This test will only pass on production after the deployment

test/unit/mock/fake_runtime_client.py
```diff
@@ -292,6 +292,7 @@ def program_run(
         start_session: Optional[bool] = None,
         session_time: Optional[int] = None,
         private: Optional[int] = False,  # pylint: disable=unused-argument
+        calibration_id: Optional[str] = None,  # pylint: disable=unused-argument
```
ElePT, line 295: Is this `calibration_id` used in the test below?

test/unit/test_backend_retrieval.py
```diff
@@ -278,3 +278,15 @@ def test_backend_with_and_without_fractional_from_same_service(self):
...
+        """Test getting a backend with a custom calibration."""
+        service = FakeRuntimeService(
+            channel="ibm_quantum_platform",
+            token="my_token",
+            backend_specs=[FakeApiBackendSpecs(backend_name="FakeTorino")],
+        )
+
+        backend_with_calibration = service.backend("fake_torino", calibration_id="abc1234")
+        self.assertEqual(backend_with_calibration.calibration_id, "abc1234")
+        # TODO: Assert mock has api client calls with cal id set
```
ElePT, line 292: The mock doesn't really "make" api calls, right?

mtreinish: The mock object tracks calls to methods with the arguments. What I meant by this comment is that we should assert the mock object for the inner api client is actually being called with the `cal_id` parameter set. So we validate that the calibration id is making it through all the layers of plumbing when requesting a backend.

PR #2439, opened 2025-10-15
Handle 0.41.0 deprecations

kt474:
### Summary

Remove deprecations from the `0.41.0` release 

### Details and comments
Fixes #


setup.cfg
```diff
@@ -2,7 +2,7 @@
 source = qiskit_ibm_runtime
 
 [mypy]
-python_version = 3.8
```
ElePT, line 5: heh

ElePT (dismissed):
LGTM. I will keep an eye for potential uses of the deprecated methods in downstream packages just in case.

PR #2455, opened 2025-10-28
Set circuit schedule drawer annotation box dynamically

RoyElkabetz:
### Summary

This PR sets plotly's dropdown annotation box location in the figure dynamically.

### Details and comments

The plot height is determined from the number of channels being plotted, therefore the annotation box location should take that information into account when being set. This is needed, because we place the annotation box outside of the plot, which has a know flaky behavior in plotly, but can be corrected dynamically.


ElePT (approved):
Sounds reasonable. LGTM

PR #2456, opened 2025-10-28
Add executor / noise learner v3

SamFerracin:
Add executor / noise learner v3 features, with the changes required for running tutorials and challenges for QDC2025.

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/quantum_program.py
+++ b/qiskit_ibm_runtime/quantum_program/quantum_program.py
@@ -18,7 +18,6 @@ import abc
 import math
 from typing import TYPE_CHECKING, Any
 from collections.abc import Iterable
-import warnings
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/options/__init__.py
```diff
@@ -85,18 +88,21 @@
 
 """
 
+from .dynamical_decoupling_options import DynamicalDecouplingOptions
 from .environment_options import EnvironmentOptions
+from .estimator_options import EstimatorOptions
 from .execution_options import ExecutionOptionsV2
+from .executor_options import ExecutorOptions
```
mriedem, line 95: It'd be nice to also do this in the top level `__init__.py`:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/__init__.py#L225-L230

PR #2457, opened 2025-10-29
Add input type support to draw circuit schedule function

RoyElkabetz:
### Summary

This PR adds support for a `CircuitSchedule` at the input of the `draw_circuit_schedule_timing()` function in addition to the current support of a circuit schedule as a string. 

### Details and comments

With this change a user would be able to create a new class that inherits from the package class in order to override one of its methods, for example the `preprocess` one like so,

```python 
class MyCircuitSchedule(CircuitSchedule):
    def __init__(self, ...):
        ....
    
    def preprocess(self)
        "This method override the parent class preprocess..."
```
and still be able to call `draw_circuit_schedule_timing()` with that new class at its input.

### DIFF
--- a/release-notes/unreleased/2457.other.rst
+++ b/release-notes/unreleased/2457.other.rst
@@ -1,0 +1,2 @@
+Add :class:`CircuitSchedule` as an optional input to :meth:`draw_circuit_schedule_timing`.
+Allows for manual preprocessing of the circuit schedule timing data outside of its draw function.


ElePT (commented):
The changes look good, my only comment is that I think you should add a release note documenting the change to the drawer function.

ElePT (approved):
LGTM

PR #2458, opened 2025-10-29
Raise warning if instance limit is reached

### LINKED ISSUE #2453: Inform the user when submitting a job to an instance that has reached its limit

Many times, users submit jobs to instances which have already reached their limit for the current billing cycle without realizing that, and just experiencing delays on the execution of these jobs not knowing what the cause was.

We would like that `qiskit-ibm-runtime` actively notifies the user when a job is submitted to an instance which has reached their limit.

This information can be obtained using [Get instance usage](https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/instances#tags__instances__operations__get_usage)

**Acceptance criteria**

The user is notified when they submit a new job to an instance which has already reached their limit (possibly a warning?).

kt474:
### Summary

Adding warnings before job submission if the instance limit has been reached. 

### Details and comments
Fixes #2453


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1192,6 +1194,30 @@ def usage(self) -> Dict[str, Any]:
...
+    def _check_instance_usage(self) -> None:
+        """Raise warning if instance usage has been reached."""
+        usage_dict = self._active_api_client.cloud_usage()
+        limit_reached = usage_dict.get("usage_limit_reached", False)
+        usage_remaining = usage_dict.get(
+            "usage_limit_seconds", usage_dict.get("usage_allocation_seconds")
+        ) - usage_dict.get("usage_consumed_seconds", 0)
+
+        if limit_reached:
+            if not usage_dict.get("usage_limit_seconds") or usage_remaining > 0:
```
ElePT: One question, what if usage_dict.get("usage_limit_seconds") is 0? In this case I believe that `not usage_dict.get("usage_limit_seconds")` will evaluate to True, but according to the issue we only want it to evaluate to True if the value is None. Did I follow this correctly?

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1192,6 +1194,30 @@ def usage(self) -> Dict[str, Any]:
             usage_dict["usage_remaining_seconds"] = usage_remaining
         return usage_dict
 
+    def _check_instance_usage(self) -> None:
+        """Raise warning if instance usage has been reached."""
+        usage_dict = self._active_api_client.cloud_usage()
+        limit_reached = usage_dict.get("usage_limit_reached", False)
+        usage_remaining = usage_dict.get(
+            "usage_limit_seconds", usage_dict.get("usage_allocation_seconds")
```
francabrera: I don't understand why `usage_allocation_seconds` is used here? If the instance is limited the limit will be `usage_limit_seconds`. 

If it's there is no limit in the instance, but `usage_limit_reached` is `true`, then we always display the message for the instance's plan on the account.

diego-plan9: It seems related to #1829 - the logic for the existing `usage()` method also involves `usage_allocation_seconds()`:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/d03882d19decadd8e6d5b294252b09aefc57da98/qiskit_ibm_runtime/qiskit_runtime_service.py#L1185-L1199

@francabrera , can you verify that taking into account `usage_allocation_seconds` (if `usage_limit_seconds`) for the calculation is correct? If it is, I think we can just take advantage of the existing method and piggyback on it.

qiskit_ibm_runtime/qiskit_runtime_service.py
diego-plan9: I agree with you on how the `0` case will be handled - depending on how the above conversation evolves, this line might not be present anymore.

PR #2468, opened 2025-11-06
Patch for executor jobs run prior to latest schemas changes

SamFerracin:
This is a temporary patch to allow loading `Executor` results that were collected before `chunk_timing` was added to the results metadata


ihincks:
This was fixed by a change to the schema field validator instead, but this would have been a nice fix too.

PR #2473, opened 2025-11-07
Fix guide rendering

SamFerracin:
This PR removes pluralization of inline code

### DIFF
--- a/docs/guides/executor_basic.rst
+++ b/docs/guides/executor_basic.rst
@@ -208,13 +208,13 @@ The outputs of the Executor
 
 :class:`~.qiskit_ibm_runtime.quantum_program.QuantumProgramResult` is an iterable. It contains one
 item per circuit task, and the items are in the same order as the items in the program. Every one of
-these items is a dictionary from strings to a ``np.ndarray`` with boolean ``dtype``. Let us take a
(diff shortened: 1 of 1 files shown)


docs/guides/executor_basic.rst
```diff
@@ -208,13 +208,13 @@ The outputs of the Executor
 
 :class:`~.qiskit_ibm_runtime.quantum_program.QuantumProgramResult` is an iterable. It contains one
 item per circuit task, and the items are in the same order as the items in the program. Every one of
-these items is a dictionary from strings to ``np.ndarray``s of ``bool``s. Let us take a look at the
-three items in ``result`` to understand the meaning of their key-value pairs.
+these items is a dictionary from strings to a ``np.ndarray`` with boolean ``dtype``. Let us take a
```
joshuasn: should it be `an np.ndarray` as it's `EN DEE array`?

SamFerracin: No idea. In my mind, I always read it as "a numpy array." But probably that's just me 😆

joshuasn: sure let's say it's `numpy nd array`

wshanks (approved):

PR #2478, opened 2025-11-16
[executor] A function to remove parameter expressions

yaelbh:
The executor does not accept parameter expressions. This PR provides a function that users can run before they instantiate a quantum program. The function will transform their circuit and parameter values such that parameter expressions are removed.

For the PR to be ready:
- [x] Build a table of parameter expressions, to avoid redundant creation of identical parameters.
- [x] Support multiple parameters in a gate.
- [x] Support dynamic circuits.
- [x] Check if boxes are supported, and fix if not.
- [x] Check if the case of one parameter in the entire circuit is supported, and fix if not.
- [x] Allow the user to control parameter names - I actually think this one is not needed, so I'm marking it as done although not done.
- [x] Save the time of parameter creation and binding when the parameter is of type `Parameter`.
- [x] Use `bind_all` instead of projecting and using `bind`.
- [x] Write tests.

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/utils.py
+++ b/qiskit_ibm_runtime/quantum_program/utils.py
@@ -64,7 +64,11 @@ def _replace_parameter_expressions(
             new_op_params.append(new_param)
 
         new_op = type(instruction.operation)(*new_op_params)
-        new_data.append(CircuitInstruction(operation=new_op, qubits=instruction.qubits, clbits=instruction.clbits))
+        new_data.append(
+            CircuitInstruction(
+                operation=new_op, qubits=instruction.qubits, clbits=instruction.clbits
+            )
+        )
 
     new_circuit.data = new_data
     return new_circuit


qiskit_ibm_runtime/quantum_program/utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+from __future__ import annotations
+
+import numpy as np
+
+from qiskit.circuit import Parameter, QuantumCircuit, ParameterExpression
+
+
+def _remove_parameter_expressions_in_blocks(
+    circ: QuantumCircuit,
```
SamFerracin:
```suggestion
    circuit: QuantumCircuit,
```

qiskit_ibm_runtime/quantum_program/utils.py
```diff
@@ -0,0 +1,96 @@
...
+        new_gate = instruction.operation.copy()
+        new_gate.params = new_op_params
+        new_data.append(instruction.replace(params=new_op_params, operation=new_gate))
+
+    new_circ.data = new_data
+    return new_circ
+
+
+def remove_parameter_expressions(
+    circ: QuantumCircuit, param_values: np.ndarray
```
SamFerracin:
```suggestion
    circuit: QuantumCircuit, param_values: np.ndarray
```

qiskit_ibm_runtime/quantum_program/utils.py
```diff
@@ -0,0 +1,96 @@
...
+) -> tuple[QuantumCircuit, np.ndarray]:
+    """Create an input to the quantum program that's
+    free from parameter expressions."""
```
SamFerracin:
```suggestion
    """A helper to replace a circuit's parameter expressions with parameters."""
```
This is a function on circuits, not on programs or program items. Also, more than removing stuff, we replace stuff, so consider using `replace_*` instead of `remove_*`

Additionally, I would expand a bit the docstring, to add a short explanation of what we do, how we recurse into conditionals, and how we transform the parameter values.

qiskit_ibm_runtime/quantum_program/utils.py
```diff
@@ -0,0 +1,96 @@
...
+    parameter_table: dict[str, Parameter] = {}
+    new_param_value_cols: list[np.ndarray] = []
+
+    new_circ = _remove_parameter_expressions_in_blocks(
```
SamFerracin: The "_in_blocks" part of `_remove_parameter_expressions_in_blocks` here is a bit confusing, because the circuit may not have any blocks. For readability, I would consider removing "_in_blocks" and calling it `_remove_parameter_expressions`

test/unit/quantum_program/test_utils.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        p1 = Parameter("p1")
+        p2 = Parameter("p2")
+        param_values = np.array(
+            [
+                [[1, 2], [3, 4], [5, 6], [7, 8]],
+                [[9, 10], [11, 12], [13, 14], [15, 16]],
+                [[17, 18], [19, 20], [21, 22], [23, 24]],
+            ]
+        )
```
SamFerracin, line 45:
```suggestion
        param_values = np.arange(1, 25).reshape((3, 4, 2))
```

Generally, the params will be floats. Therefore, I would actually favour something like this:
``` suggestion
        param_values = np.random.random((3, 4, 2))
```

test/unit/quantum_program/test_utils.py
```diff
@@ -0,0 +1,189 @@
...
+        self.assertEqual(new_circ.parameters[0], p1)
+
+        self.assertEqual(param_values.shape[:-1], new_values.shape[:-1])
+        param_values_flat = param_values.reshape(-1, param_values.shape[-1])
+        new_values_flat = new_values.reshape(-1, new_values.shape[-1])
+        for param_set_1, param_set_2 in zip(param_values_flat, new_values_flat):
+            self.assertTrue(
+                Operator.from_circuit(circ.assign_parameters(param_set_1)).equiv(
+                    Operator.from_circuit(new_circ.assign_parameters(param_set_2))
+                )
```
SamFerracin, line 66: nice

SamFerracin (commented):
Thanks, Yael!
I left a few comments. Additionally, I +1 Josh's comment that it would be nice to structure parameter handling exactly as in Samplomatic, which would make it easier to maintain

test/unit/quantum_program/test_utils.py
yaelbh: I find it more readable if the array is written explicitly. How about keeping the original array, and only adding `dtype=float`?

qiskit_ibm_runtime/quantum_program/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def replace_parameter_expressions(
+    circuit: QuantumCircuit, parameter_values: np.ndarray
+) -> tuple[QuantumCircuit, np.ndarray]:
+    """
+    A helper to replace a circuit's parameter expressions with parameters.
+
+    The function tranverses the circuit and collects all the parameters and parameter expressions.
+    A new parameter is created for every parameter expression that is not a parameter.
```
SamFerracin, line 84: I think that you shouldn't worry about replacing _only_ the parameter expressions.

```suggestion
    The function tranverses the circuit and replaces all the parameters and parameter expressions with new parameters.
```

This would allow you to simplify massively this function (no need for `parameter_expressions_to_new_parameters_map`) and to make a more standard use of the parameter table where you simply:
1. Append every parameter/parameter expression you find to it
2. Call evaluate at the end

If there are some concerns that are missing, we should improve the parameter expression table directly, rather than writing more code in places where we use it (cc'in @joshuasn here, as he's the one who wrote the table in the first place)

qiskit_ibm_runtime/quantum_program/utils.py
```diff
@@ -0,0 +1,119 @@
...
+                    new_param = param_exp
+                else:
+                    new_param = Parameter(str(param_exp))
+                parameter_table.append(param_exp)
+                parameter_expressions_to_new_parameters_map[param_exp] = new_param
+            new_op_params.append(new_param)
+
+        new_gate = instruction.operation.copy()
+        new_gate.params = new_op_params
+        new_data.append(instruction.replace(params=new_op_params, operation=new_gate))
```
SamFerracin: I'm not sure what's the overhead of all these steps (copying, replacing, ...), so I would replace this block with something more standard like this (might not work, but will give you an idea):

```suggestion
        new_op = type(op := instruction.operation)(new_op_params)
        new_circuit.append(new_op, op.qubits, op.clbits)
```

qiskit_ibm_runtime/quantum_program/utils.py
yaelbh: I still need `parameter_expressions_to_new_parameters_map`, because the same parameter / parameter expression can appear several times in a circuit. So, if `a+b` appears 100 times, I want to create one parameter; later when I see `a+b` in my circuit I want to know with which parameter to replace it, and I do it by querying the map.

qiskit_ibm_runtime/quantum_program/utils.py
yaelbh: I prefer to copy, for this reason: doing something in the spirit of this suggestion requires that I'm aware of all the attributes of `CircuitInstruction` and `Operation` objects, such as `label`, and that I take care of setting them to their values in the original instruction/operation. Setting them is sometimes done by setting a constructor argument and sometimes by using a setter method, depending on the attribute. If, one day in the future, a new attribute is added to these classes, our code here will break.

Having said that, my solution, the one that copies, also feels somewhat fragile, however less fragile than the suggested alternative.

PR #2485, opened 2025-11-24
Delay imports of optional dependencies

wshanks:
`qiskit_aer` and `scipy` were eagerly imported in the `utils.json` module (with fallbacks for when they were not installed) which is imported along with `import qiskit_ibm_runtime`. Here these imports are delayed until JSON serialization or deserialization are attempted. In the serialization case, the packages would have already been imported before objects from them would need to be serialized any way.

Avoiding these imports saves roughly half a second (but the number depends on the computer running the test).

### DIFF
--- a/qiskit_ibm_runtime/utils/json.py
+++ b/qiskit_ibm_runtime/utils/json.py
@@ -393,7 +393,7 @@ class RuntimeEncoder(json.JSONEncoder):
             return {"__type__": "PauliLindbladMap", "__value__": out_val}
         if "qiskit_aer" in sys.modules:
             # Guard import so qiskit_aer is not imported unnecessarily. If a
-            # NoiseModel instance has been created, qiskit_aer must alrady have
+            # NoiseModel instance has been created, qiskit_aer must already have
             # been imported.
             from qiskit_aer.noise import NoiseModel  # pylint: disable=import-outside-toplevel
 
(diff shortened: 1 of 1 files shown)


wshanks:
I don't have any urgency for merging this. It was something I noticed while debugging qiskit-experiments' imports (that scipy and qiskit-aer were hardly used but eagerly imported), so I thought I would see what it looks like to avoid importing them.

By the way, deserialization of `spmatrix` was using scipy without a guard. I am not sure if that even comes up in practice any more but I put in a guard like the qiskit-aer case had.

ihincks (commented):
Nice thinking! I have not done any investigating yet about whether this would fit in qiskit-ibm-runtime specifically, but I'm wondering if any of [qiskit's tooling for optionals ](https://quantum.cloud.ibm.com/docs/en/api/qiskit/utils#optional-dependency-checkers)might help us here, or if you think it would be better to just do it manually as this PR currently implements?

I found it convenient in samplomatic to declare [consts of optional dependencies](https://github.com/Qiskit/samplomatic/blob/main/samplomatic/optionals.py) and [to use them](https://github.com/Qiskit/samplomatic/blob/98608c642f985ff7e01e4c8b74cfca6125cbd80c/samplomatic/visualization/plot_graph.py#L76).

ihincks (approved):
Thanks @wshanks

Issue #2488, opened 2025-11-28
Bump minimal Python version to 3.10


diego-plan9:
**What is the expected feature or enhancement?**

With Python 3.9 reaching EOL, and Qiskit dropping support [recently](https://github.com/Qiskit/qiskit/issues/15139), the package should follow suit:
* bump minimal versions declared in `pyproject.toml`
* update workflows to run on 3.10
* general sanity check (mypy / tox / docs)

**Acceptance criteria**

The package enforces 3.10 as the minimal supported Python version, and the tooling reflects it.

PR #2491, opened 2025-12-01
Bump mypy to 3.10-fully-compatible version, use pyproject.toml

diego-plan9:
### Summary

As part of having Python 3.10 as the default version, bump `mypy` to the latest version, to ensure that it is fully compatible with 3.10 syntax. This paves the way towards making the annotations in the codebase follow 3.10 type hint practices (as having and older `mypy` version prevents doing so):
* bump the required version to `1.19.0`
* fix or silence the new errors (using ample usage of `# type: ignore`)
* revise the configuration (new option added, non-supported option removed)
* move the configuration to `pyproject.toml` to unify (as well as the `coverage` configuration), dropping `setup.cfg`

### Details and comments

Related to #2488

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,7 @@ issues = "https://github.com/Qiskit/qiskit-ibm-runtime/issues"
 [project.optional-dependencies]
 
 common = [
-    "mypy==0.931",
+    "mypy==1.19.0",
     "pylint==3.0.0",
     "pproxy==2.7.8",
     "nbqa==1.5.3",


ElePT (approved):
LGTM. The type-ignores are not ideal but I consider them the lesser evil here.

qiskit_ibm_runtime/transpiler/passes/scheduling/block_base_padder.py
```diff
@@ -72,12 +72,12 @@ def __init__(
         target: Optional[Target] = None,
     ) -> None:
 
-        self._node_start_time = None
-        self._node_block_dags = None
+        self._node_start_time: Optional[Dict[DAGNode, Tuple[int, int]]] = None
+        self._node_block_dags: Optional[Dict[DAGNode, DAGCircuit]] = None
```
yaelbh, line 76: Note that each node is assigned one start time and one block dag, but one variable is called `_node_start_time` in singular (makes sense - a single start time per node) and the other variables is called `_node_block_dags` in plural (makes sense - all the nodes together have multiple block dags).

qiskit_ibm_runtime/utils/circuit_schedule.py
```diff
@@ -70,7 +70,7 @@ def __init__(
         """
         self.channels: List = None
         self.type_to_idx: Dict[str, int] = None
-        self.circuit_scheduling = None
+        self.circuit_scheduling: np.array = None
```
yaelbh, line 73:
```suggestion
        self.circuit_scheduling: np.array | None = None
```

yaelbh: `mypy` didn't complain because `strict_optional` is set to `false`.

pyproject.toml
```diff
@@ -52,6 +52,29 @@ write_to = "qiskit_ibm_runtime/VERSION.txt"
...
+warn_redundant_casts = true
+warn_unreachable = true
+strict_equality = true
+disallow_untyped_calls = true
+disallow_untyped_defs = true
+disallow_incomplete_defs = true
+strict_optional = false
+show_error_codes = true
+no_site_packages = true
+disable_error_code = "import-untyped"
```
yaelbh, line 68: I understand that these flags were copied from the other file, still I suggest to revisit the relaxations of `ignore_missing_imports`, `strict_optional`, and `disable_error_code` (this last one was not present in the original file).

diego-plan9: `disable_error_code` was needed to silence some new `mypy` warnings due to not including [library stubs](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker) - I placed the focus on keeping this PR lean and opting for the simplest approach in general.

I agree with revising the `mypy` approach overall eventually (trimming down the number of ignores, re-evaluating the options, and deciding if the project is to use typing in a more comprehensive way), but I see it as something less critical and for the future.

qiskit_ibm_runtime/transpiler/passes/scheduling/block_base_padder.py
diego-plan9: I'll double check as part of the upcoming PR: I reused the definition from its usages.

Issue #2438, opened 2025-10-15
`ValueError: id must be provided` from `list_instances`


mriedem:
**Describe the bug**

I'm trying to submit a job and getting this error:

```
.venv/lib/python3.11/site-packages/flaky/flaky_pytest_plugin.py:146: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
conftest.py:581: in provider
    return qiskit_provider()
           ^^^^^^^^^^^^^^^^^
conftest.py:525: in iqp_cloud_provider
    return QiskitRuntimeService(
.venv/lib/python3.11/site-packages/qiskit_ibm_runtime/qiskit_runtime_service.py:233: in __init__
    if self._account.instance not in [inst["crn"] for inst in self.instances()]:
                                                              ^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/qiskit_ibm_runtime/qiskit_runtime_service.py:1314: in instances
    self._all_instances = self._account.list_instances()
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/qiskit_ibm_runtime/accounts/account.py:350: in list_instances
    catalog_result = catalog.get_catalog_entry(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ibm_platform_services.global_catalog_v1.GlobalCatalogV1 object at 0x720d75648d10>, id = None, account = None, include = None, languages = None
complete = None, depth = None, kwargs = {}

    def get_catalog_entry(
        self,
        id: str,
        *,
        account: Optional[str] = None,
        include: Optional[str] = None,
        languages: Optional[str] = None,
        complete: Optional[bool] = None,
        depth: Optional[int] = None,
        **kwargs,
[...]

PR #2493, opened 2025-12-03
[WIP] List Instances will skip bad instances with a warning

jan-an:
### Summary
- Fixes issue #2438  by modifying `list_instances` in account.py
- Need to add tests and documentation


diego-plan9:
Thanks Janani! Following our recent discussion, I believe the root issue has been fixed by #2603 - and I did not realize that and your assigned issue were targeting the same root cause at the time.

Issue #2479, opened 2025-11-18
Improve tests for NoiseLearnerV3


SamFerracin:
**Acceptance criteria**

- [ ] Improve coverage, ensuring that the most relevant parts of NLV3 and related features are covered by a unit tests.
- [ ] Write e2e tests.

yaelbh:
Cover:
- [ ] The files under `qiskit_ibm_runtime/noise_learner_v3`.
- [ ] `options/noise_learner_v3_options.py` and `options/post_selection_options.py`.

yaelbh:
Updated list of files to test:
- [x] `qiskit_ibm_runtime/noise_learner_v3/converters`
- [x] `qiskit_ibm_runtime/noise_learner_v3/find_learning_protocol.py`
- [ ] `qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_decoders.py`
- [ ] `qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_result.py`
- [ ] `qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py`
- [x] `qiskit_ibm_runtime/noise_learner_v3/validation.py`
- [ ] `qiskit_ibm_runtime/quantum_program/converters.py`
- [ ] `qiskit_ibm_runtime/quantum_program/quantum_program_result.py`
- [ ] `qiskit_ibm_runtime/quantum_program/quantum_program_decoders.py`
- [ ] `qiskit_ibm_runtime/quantum_program/quantum_program.py`
- [ ] `qiskit_ibm_runtime/executor.py`
- [ ] `qiskit_ibm_runtime/utils/executor_result_decoder.py`
- [ ] `qiskit_ibm_runtime/options/executor_options.py`
- [ ] `qiskit_ibm_runtime/options/noise_learner_v3_options.py`
- [ ] `qiskit_ibm_runtime/options/post_selection_options.py`

PR #2496, opened 2025-12-04
[executor-preview] Tests for validation of NLV3

yaelbh:
### Summary

Part of #2479.
Tests for the functions in the file `qiskit_ibm_runtime/noise_learner_v3/validation`.

### DIFF
--- a/test/unit/noise_learner_v3/test_validation.py
+++ b/test/unit/noise_learner_v3/test_validation.py
@@ -48,6 +48,8 @@ class TestValidation(IBMTestCase):
         circuit = QuantumCircuit(target.num_qubits)
         with circuit.box(annotations=[Twirl()]):
             circuit.cx(0, 1)
+        with circuit.box(annotations=[Twirl()]):
+            circuit.measure_all()
 
         validate_instruction(circuit.data[0], target)
         validate_instruction(circuit.data[1], target)


yaelbh:
Ready for review. Note that the PR contains a bug fix.

test/unit/noise_learner_v3/test_find_learning_protocol.py
```diff
@@ -23,7 +23,7 @@
 
 
 class TestConverters(IBMTestCase):
```
SamFerracin:
```suggestion
class TestFindLearningProtocol(IBMTestCase):
```

test/unit/noise_learner_v3/test_validation.py
```diff
@@ -0,0 +1,109 @@
...
+            n_qubits=1e100,
+            basis_gates=["rx"],
+            gates=[],
+            local=False,
+            simulator=False,
+            conditional=True,
+            open_pulse=False,
+            memory=True,
+            coupling_map=[],
+        )
```
SamFerracin: Why not passing `FakeAlgiers.configuration()`?

test/unit/noise_learner_v3/test_validation.py
```diff
@@ -0,0 +1,109 @@
...
+        validate_options(options=options, configuration=configuration)
+
+        options.post_selection = {"enable": False, "x_pulse_type": "xslow"}
+        validate_options(options=options, configuration=configuration)
+
+        options.post_selection = {"enable": True, "x_pulse_type": "xslow"}
+        with self.assertRaisesRegex(ValueError, "xslow"):
+            validate_options(options=options, configuration=configuration)
+
+    def test_validate_instruction(self):
```
SamFerracin: This is a big test function with many assert. In general, I prefer breaking big test functions into many smaller test functions with asserts targeting a single concern, for example:
- `test_validate_instruction_passes_when_options_valid`
- `test_validate_instruction_without_twirl_raises`
- `test_validate_instruction_without_box_raises`
- ...

The benefit is that if we make a change that invalidate one of these asserts, we immediately get a hint of what caused the break by reading the name(s) of the broken tests, as opposed to having to dig through

SamFerracin (commented):
Thanks! A couple of small comments

test/unit/noise_learner_v3/test_validation.py
yaelbh: 260f9126

test/unit/noise_learner_v3/test_validation.py
yaelbh: Please check the solution of using `subTest` (2340427b).

PR #2498, opened 2025-12-08
[executor-preview]  Test noise_learner_v3.py

yaelbh:
### Summary

Part of #2479.
Tests for the functions in the file `qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py`.

### DIFF
--- a/qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
+++ b/qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
@@ -82,9 +82,10 @@ class NoiseLearnerV3:
         ):
             self._options.experimental = {}
 
-        self._session, self._service, self._backend = get_mode_service_backend(  # type: ignore[assignment]
+        self._session, self._service, self._backend = get_mode_service_backend(
             mode
-        )
+        )  # type: ignore[assignment]
+
         if isinstance(self._service, QiskitRuntimeLocalService):  # type: ignore[unreachable]
             raise ValueError("``NoiseLearnerV3`` is currently not supported in local mode.")
 


test/unit/noise_learner_v3/test_noise_learner_v3.py
```diff
@@ -0,0 +1,43 @@
...
+    """Tests the `NoiseLearnerV3` class."""
+
+    def test_init_with_backend(self):
+        """Test `NoiseLearnerV3.init` when the input mode is an IBMBackend."""
+        backend = get_mocked_backend()
+        service = backend.service
+        service.reset_mock()
+        noise_learner = NoiseLearnerV3(mode=backend)
+        assert noise_learner._backend == backend
+        assert noise_learner._service == service
```
SamFerracin: Consider making these properties public. Do the other primitives do that?

yaelbh: They are private for the other primitives.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -77,48 +76,16 @@ def __init__(
...
-            self._backend = self._session._backend
-            self._service = self._session.service
-        elif isinstance(mode, BackendV2):
-            self._backend = mode
-            self._service = self._backend.service
-        else:
-            raise ValueError(
-                "A backend or session/batch must be specified, or a session/batch must be open."
-            )
-        self._mode, self._service, self._backend = _get_mode_service_backend(mode)
```
SamFerracin: Was the part above redundant, because`_get_mode_service_backend` just undoes what was done above?

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -144,15 +111,14 @@ def run(self, instructions: Iterable[CircuitInstruction]) -> RuntimeJobV2:
...
-            target = getattr(self._backend, "target", None)
-            if target and not is_simulator(self._backend):
-                for instruction in instructions:
-                    validate_instruction(instruction, target)
-
-            configuration = getattr(self._backend, "configuration", None)
-            if configuration and not is_simulator(self._backend):
-                validate_options(self.options, configuration())
+        target = getattr(self._backend, "target", None)
+        if target and not is_simulator(self._backend):
```
SamFerracin: I'm not sure what happens if `self._backend` is a simulator, I would expect it to raise somewhere (init or run? 🤷). The truth is that we never supported simulators for NoiseLearner, and the same applies to NoiseLearnerV3.

If this is indeed the case, I would prefer to add a check **to the init** along the lines of:
```
if is_simulator(self._backend):
    raise ValueError("Simulator backend is not supported.")
```

And then here:
- We would not need to check `is_simulator` anymore
- We would need to raise ValueError if target or configuration is None, because real backends do have a target and a config

test/unit/noise_learner_v3/test_noise_learner_v3.py
SamFerracin: I can't think of a reason for them to be private, so I would expose them as properties (obviously without setters). Accessing private attributes is something that we should avoid in every case

test/unit/noise_learner_v3/test_noise_learner_v3.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertEqual(noise_learner._session, None)
+        self.assertEqual(noise_learner._backend, backend)
+        self.assertEqual(noise_learner._service, service)
+
+    def test_init_with_session(self):
+        """Test ``NoiseLearnerV3.init`` when the input mode is a session."""
+        backend_name = "ibm_hello"
+        session = get_mocked_session(get_mocked_backend(backend_name))
+        session.reset_mock()
+        session.service.reset_mock()
```
SamFerracin: I wonder why do we need to reset (twice) something that we just initialized. We should check (in another issue) if we can improve this mock

test/unit/noise_learner_v3/test_noise_learner_v3.py
```diff
@@ -0,0 +1,92 @@
...
+        backend = get_mocked_backend()
+        service = backend.service
+        service.reset_mock()
+        with Session(backend=backend) as session:
+            noise_learner = NoiseLearnerV3(mode=backend)
+            self.assertEqual(noise_learner._session, session)
+            self.assertEqual(noise_learner._backend, backend)
+            self.assertEqual(noise_learner._service, service)
+
+    def test_run_of_session_is_selected(self):
```
SamFerracin, line 27: In this test and in the next one you are manually monkey-patching by overwriting `service._run`. This is not good practice as it can have undesirable side effects, such as leakage (the manual change persists beyond the single test case that made it). 

Please use `unittest.mock`, which takes care of the cleanup. There are some examples in other files

test/unit/noise_learner_v3/test_noise_learner_v3.py
yaelbh: I'll do it. Anyway in a test, when we want to verify that the internals of a class were properly initialized, I think it does make sense to query private variables. I have the same story also in #2501.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
yaelbh: The part above was some copying of the content of `_get_mode_service_backend`, with some editing, that introduced bugs. Some cases were addressed (correctly or incorrectly) by the part above, without ever reaching the call to `_get_mode_service_backend`. Other cases did reach the call, and went through the same process again inside the function.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
yaelbh: - We already raise, in init, if we fall on the `QiskitRuntimeLocalService`, perhaps it covers the case of a simulator?
- If real backends always have a target and a config then why check it?

test/unit/noise_learner_v3/test_noise_learner_v3.py
yaelbh: I copied it from the primitives v2 tests. We should check what happens in this regard when tests run in parallel.

yaelbh: I'll check

test/unit/noise_learner_v3/test_noise_learner_v3.py
yaelbh: There was already a `backend` function, not a property. It's better as a property but I don't want to cause a breaking change. In 4de1380f I wrote `session` and `service` as properties, maybe change them to functions too, just for uniformity with `backend`?

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
yaelbh: I don't think it can be a simulator at this point, so I removed the check in 4b988268. I also think the check if target and config are None is redundant, but why not. We can also revert and think about all this in a separate issue.

test/unit/noise_learner_v3/test_noise_learner_v3.py
yaelbh: Indeed I don't see a reason for any of the calls to `reset_mock` in `test_noise_learner_v3.py`, and for most of the calls in `test_ibm_primitives_v2.py`. To be on the safe side, as you say, we'll face it in another issue.

test/unit/noise_learner_v3/test_noise_learner_v3.py
yaelbh: I don't see how a leakage can occur. The mocked session is local only to this test. But I agree that the good practice is do something like done in c47364cf. By the way note that we override session data members inside `get_mocked_session`.

PR #2506, opened 2025-12-11
Fix bug with NLV3 and Executor decoders

SamFerracin:
This PR fixes two issues with the decoders.

The first fix ensures that the NLV3 creates results of the correct type. Previous to this change, some dictionaries with keys of type `int` were actually returned as dictionaries with keys of type `str`.

The second fix ensures that the Executor can support more than just "v0_1" models

### DIFF
--- a/qiskit_ibm_runtime/noise_learner_v3/converters/version_0_1.py
+++ b/qiskit_ibm_runtime/noise_learner_v3/converters/version_0_1.py
@@ -89,14 +89,14 @@ def noise_learner_v3_result_from_0_1(
             NoiseLearnerV3Result.from_generators(
                 generators=[
                     QubitSparsePauliList.from_sparse_list(
-                        [tuple(term) for term in sparse_list], datum["num_qubits"]
+                        [tuple(term) for term in sparse_list], datum.num_qubits
                     )
-                    for sparse_list in datum["generators_sparse"]
+                    for sparse_list in datum.generators_sparse
                 ],
-                rates=F64TensorModel(**datum["rates"]).to_numpy(),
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_decoders.py
```diff
@@ -40,8 +44,8 @@ def decode(cls, raw_result: str):  # type: ignore[no-untyped-def]
             raise ValueError("Missing schema version.")
 
         try:
-            decoder = AVAILABLE_DECODERS[schema_version]
+            decoder, model = AVAILABLE_DECODERS[schema_version]
```
SamFerracin, line 47: @yaelbh I considered a few different approaches, but eventually converged on what you suggested from the start

qiskit_ibm_runtime/noise_learner_v3/converters/version_0_1.py
```diff
@@ -89,14 +89,14 @@ def noise_learner_v3_result_from_0_1(
...
                     )
-                    for sparse_list in datum["generators_sparse"]
+                    for sparse_list in datum.generators_sparse
                 ],
-                rates=F64TensorModel(**datum["rates"]).to_numpy(),
-                rates_std=F64TensorModel(**datum["rates_std"]).to_numpy(),
-                metadata=datum["metadata"],
+                rates=datum.rates.to_numpy(),
+                rates_std=datum.rates_std.to_numpy(),
+                metadata=datum.metadata,
```
yaelbh, line 98:
```suggestion
                metadata=datum.metadata.model_dump(),
```

PR #2511, opened 2025-12-17
[executor-preview] Merge 0.44 into executor preview branch

diego-plan9:
### Summary

Do some groundwork for merging `0.44.0` into the `executor_preview` branch:
* resolve most of the conflicts
* apply ruff UP fixes (same approach as #2495)
* fix mypy warnings (same approach as  #2491)

This leaves the branch nearly ready for pulling from `main`, with just one conflict unresolved:
```
git pull upstream stable/0.44.0
git checkout --ours qiskit_ibm_runtime/utils/noise_learner_result_decoder.py
git add qiskit_ibm_runtime/utils/noise_learner_result_decoder.py
git commit
```

The intent is to be able to have a regular merge commit in the branch.

### Details and comments

Alternative to #2508

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/converters.py
+++ b/qiskit_ibm_runtime/quantum_program/converters.py
@@ -44,7 +44,7 @@ def quantum_program_to_0_1(program: QuantumProgram, options: ExecutorOptions) ->
         chunk_size = "auto" if item.chunk_size is None else item.chunk_size
         if isinstance(item, CircuitItem):
             model_item = CircuitItemModel(
-                circuit=QpyModelV13ToV16.from_quantum_circuit(item.circuit),
+                circuit=QpyModelV13ToV16.from_quantum_circuit(item.circuit, qpy_version=16),
                 circuit_arguments=F64TensorModel.from_numpy(item.circuit_arguments),
                 chunk_size=chunk_size,
             )
@@ -60,7 +60,7 @@ def quantum_program_to_0_1(program: QuantumProgram, options: ExecutorOptions) ->
                     else:
                         arguments[name] = value
             model_item = SamplexItemModel(
-                circuit=QpyModelV13ToV16.from_quantum_circuit(item.circuit),
+                circuit=QpyModelV13ToV16.from_quantum_circuit(item.circuit, qpy_version=16),
                 samplex=SamplexModelSSV1.from_samplex(item.samplex, ssv=1),
                 samplex_arguments=arguments,
                 shape=item.shape,
--- a/.github/workflows/integration-tests-qiskit-main.yml
+++ b/.github/workflows/integration-tests-qiskit-main.yml
@@ -25,7 +25,7 @@ jobs:
       # avoid cancellation of in-progress jobs if any matrix job fails
       fail-fast: false
       matrix:
-        python-version: [ 3.9 ]
+        python-version: [ "3.10" ]
         os: [ "ubuntu-latest" ]
         environment: ["ibm-cloud-production", "ibm-cloud-staging" ]
     environment: ${{ matrix.environment }}
--- a/qiskit_ibm_runtime/noise_learner_v3/converters/version_0_1.py
+++ b/qiskit_ibm_runtime/noise_learner_v3/converters/version_0_1.py
@@ -46,7 +46,7 @@ def noise_learner_v3_inputs_to_0_1(
         circuit.append(instr, instr.qubits, instr.clbits)
 
     return ParamsModel(
-        instructions=QpyModelV13ToV16.from_quantum_circuit(circuit),
+        instructions=QpyModelV13ToV16.from_quantum_circuit(circuit, qpy_version=16),
         options=options.to_options_model(),
     )
 
--- a/qiskit_ibm_runtime/quantum_program/quantum_program_result.py
+++ b/qiskit_ibm_runtime/quantum_program/quantum_program_result.py
@@ -75,7 +75,7 @@ class QuantumProgramResult:
         self._data = data
         self.metadata = metadata or Metadata()
 
-    def __iter__(self) -> Iterable[dict[str, np.ndarray]]:
+    def __iter__(self) -> Iterator[dict[str, np.ndarray]]:
         yield from self._data
 
     def __getitem__(self, idx: int) -> dict[str, np.ndarray]:
--- a/test/unit/quantum_program/__init__.py
+++ b/test/unit/quantum_program/__init__.py
@@ -1,0 +1,11 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2026.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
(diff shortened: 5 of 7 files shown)


qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_result.py
```diff
@@ -26,9 +24,9 @@
 from samplomatic import InjectNoise
 from samplomatic.utils import get_annotation
 
-MetadataLeafTypes = Union[int, str, float]
-MetadataValue = Union[MetadataLeafTypes, "Metadata", list["MetadataValue"]]
-Metadata = dict[str, MetadataValue]
+MetadataLeafTypes = int | str | float
+MetadataValue = 'MetadataLeafTypes | "Metadata" | list["MetadataValue"]'  # pylint: disable=invalid-name
+Metadata = dict[str, MetadataValue]  # type: ignore[valid-type]
```
ihincks: the invalid name errors _may_ go away if you specify them as type aliases

```suggestion
MetadataValue: TypeAlias = 'MetadataLeafTypes | "Metadata" | list["MetadataValue"]'
Metadata: TypeAlias = dict[str, MetadataValue]
```

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_result.py
```diff
@@ -10,13 +10,11 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-# type: ignore[arg-type]
-
 """Noise learner V3 results."""
 
 from __future__ import annotations
```
ihincks, line 15: Does this file use anything such that this line is needed in Python 3.10? Same question for all files.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_result.py
```diff
@@ -162,9 +160,10 @@ def to_dict(
             ValueError: If some of the instructions have no inject noise annotation and ``require_refs``
                 if ``True``.
         """
-        if len(instructions) != len(self.data):
+        if len(instructions) != len(self.data):  # type: ignore[arg-type]
```
ihincks, line 164: This type error should either not be ignored, or the type should be changed. That is, the existing code before your PR technically has a bug. I don't know if you'd prefer to fix this in your PR or a different one.

(one can't generally assume things with type `Iterable[T]` can be iterated more than once, or that they have a length. We need to cast it to a list or something along those lines)

qiskit_ibm_runtime/options/noise_learner_v3_options.py
```diff
@@ -76,22 +77,24 @@ class NoiseLearnerV3Options(BaseOptions):
         This field is ignored by TREX experiments.
     """
 
-    post_selection: Union[PostSelectionOptions, Dict] = Field(default_factory=PostSelectionOptions)
+    post_selection: PostSelectionOptions | Dict = Field(default_factory=PostSelectionOptions)
```
ihincks:
```suggestion
    post_selection: PostSelectionOptions | dict = Field(default_factory=PostSelectionOptions)
```

qiskit_ibm_runtime/options/noise_learner_v3_options.py
```diff
@@ -144,9 +147,9 @@ def get_callback(self) -> Callable | None:
     # Reason not to implement OptionsV3: I don't feel like committing to an API for it.
 
     # Options not really related to primitives.
-    max_execution_time: Union[UnsetType, int] = Unset
-    environment: Union[EnvironmentOptions, Dict] = Field(default_factory=EnvironmentOptions)
-    simulator: Union[SimulatorOptions, Dict] = Field(default_factory=SimulatorOptions)
+    max_execution_time: UnsetType | int = Unset
+    environment: EnvironmentOptions | Dict = Field(default_factory=EnvironmentOptions)
+    simulator: SimulatorOptions | Dict = Field(default_factory=SimulatorOptions)
```
ihincks:
```suggestion
    environment: EnvironmentOptions | dict = Field(default_factory=EnvironmentOptions)
    simulator: SimulatorOptions | dict = Field(default_factory=SimulatorOptions)
```

qiskit_ibm_runtime/quantum_program/converters.py
```diff
@@ -72,7 +72,7 @@ def quantum_program_to_0_1(program: QuantumProgram, options: ExecutorOptions) ->
 
     return ParamsModel(
         quantum_program=QuantumProgramModel(shots=program.shots, items=model_items),
-        options=asdict(options.execution),
+        options=asdict(options.execution),  # type: ignore[call-overload]
```
ihincks, line 75: educated guess, but I think this would work?

```suggestion
        options=options.execution.model_dump()
```

qiskit_ibm_runtime/options/noise_learner_v3_options.py
diego-plan9: Oh, I was wondering why this was not picked up by `ruff` - and it seems it is using `.utils.Dict`. Thanks for catching it!

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_result.py
diego-plan9: This was previously hidden by the:
```
# type: ignore[arg-type]
```

at the top of the file, which seems to not be supported in the recent `mypy` version:
```
error: type ignore with error code is not supported for modules; use `# mypy: disable-error-code="arg-type"`  [syntax]
```

Instead of replacing the top line, I went with more granular type ignores. I do prefer to tackle it as a separate PR indeed.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_result.py
diego-plan9: I don't think it does, but will strive for gaining full confidence - and it that is the case, I'll tackle it for all the codebase in a separate PR.

qiskit_ibm_runtime/quantum_program/converters.py
diego-plan9: It did not do the trick, unfortunately:

```
qiskit_ibm_runtime/quantum_program/converters.py:75: error: "ExecutionOptions" has no attribute "model_dump"  [attr-defined]
```

It seems `mypy` is not able to understand `options.execution` is a Pydantic dataclass.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_result.py
diego-plan9: Applied at 8b5ee29 - it did work, thanks!

qiskit_ibm_runtime/options/noise_learner_v3_options.py
diego-plan9: Applied at 8b5ee29

qiskit_ibm_runtime/options/noise_learner_v3_options.py
diego-plan9: Applied at 8b5ee29

PR #2518, opened 2025-12-22
Include ruff as part of the tooling

### LINKED ISSUE #2513: Add ruff as part of the tooling

During #2488, `ruff` was used for helping coming up with bumping the codebase to Python 3.10. It is a popular tool for linting in Python-land, convenient, and is [already used](https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#style-and-lint) in experimental capacity by Qiskit since a couple of years ago.

This issue is about adding `ruff` to the project, alongside the existing tools:
* ideally, using the same version as Qiskit, for easier aligning
* starting with a minimal ruleset (rules that homogenize on Python 3.10, provide convenience)
* evaluate adopting Qiskit ruleset

The goal is to ease having a maintainable codebase, with minimal impact in performance.

**Acceptance criteria**

`ruff` is added to the list of dev dependencies, and can be invoked by developers and CI.

diego-plan9:
### Summary

Introduce `ruff` as an additional linter, in order to cover some more checks (and automated fixes) with minimal performance hit:
* the ruleset is a combination of Qiskit and samplomatic, tuned to this project
* changes are a result of the `--fix` option, except for hand-introduced `noqa` entries
* invocation is part of the `make lint` target

### Details and comments
Fixes #2513

### DIFF
--- a/qiskit_ibm_runtime/accounts/account.py
+++ b/qiskit_ibm_runtime/accounts/account.py
@@ -205,7 +205,7 @@ class Account:
         """Assert that the URL is valid."""
         try:
             urlparse(url)
-        except:  # noqa: E722
+        except:  # noqa: E722 bare-except
             raise InvalidAccountError(f"Invalid `url` value. Failed to parse '{url}' as URL.")
 
     @staticmethod
@@ -333,7 +333,7 @@ class CloudAccount(Account):
                     search_cursor=search_cursor,
                     limit=100,
                 ).get_result()
-            except:  # noqa: E722
+            except:  # noqa: E722 bare-except
                 raise InvalidAccountError(
                     "Unable to retrieve instances. "
                     "Please check that you are using a valid API token."
--- a/qiskit_ibm_runtime/fake_provider/fake_provider.py
+++ b/qiskit_ibm_runtime/fake_provider/fake_provider.py
@@ -11,7 +11,7 @@
 # that they have been altered from the originals.
 
 # pylint: disable=wildcard-import,unused-argument
-# ruff: noqa: F405
+# ruff: noqa: F405 undefined-local-with-import-star-usage
 
 """
 Fake provider class that provides access to fake backends.
@@ -20,7 +20,7 @@ Fake provider class that provides access to fake backends.
 from typing import Any
 from qiskit.providers.exceptions import QiskitBackendNotFoundError
 
-from .backends import *  # noqa: F403
+from .backends import *  # noqa: F403 undefined-local-with-import-star
 from .fake_backend import FakeBackendV2
 
 
--- a/qiskit_ibm_runtime/transpiler/passes/scheduling/dynamical_decoupling.py
+++ b/qiskit_ibm_runtime/transpiler/passes/scheduling/dynamical_decoupling.py
@@ -362,7 +362,7 @@ class PadDynamicalDecoupling(BlockBasePadder):
                     if self._target:
                         try:
                             gate_length = self._target[gate.name].get((physical_index,)).duration
-                        except:  # pylint: disable=bare-except # noqa: E722
+                        except:  # pylint: disable=bare-except # noqa: E722 bare-except
                             gate_length = None
                     else:
                         gate_length = self._durations.get(gate, physical_index)


pyproject.toml
```diff
@@ -75,6 +75,33 @@ disallow_untyped_defs = false
...
+    # Rules in alphabetic order
+    "C4",    # flake8-comprehensions
+    "E",     # pycodestyle
+    "F",     # pyflake
+    "G",     # flake8-logging-format
+    "T10",   # flake8-debugger
+    "UP",    # pyupgrade
+]
+ignore = [
+    "E501",  # line-too-long: black handles line length
```
diego-plan9, line 93: This is as currently there are a number of docstrings, strings and similar above the 100 char limit. As `black` does a good job of handling these, I resorted to ignoring this rule

pyproject.toml
```diff
@@ -75,6 +75,33 @@ disallow_untyped_defs = false
...
+    "UP032", # f-string: requires manual revision of non f-string formatting
```
diego-plan9, line 94: Potentially this could be enabled, but requires a bit more manual work to ensure the changes do not hit log lines or similar cases. So I resorted to ignoring / deferring

pyproject.toml
```diff
@@ -75,6 +75,33 @@ disallow_untyped_defs = false
...
+
+[tool.ruff.lint]
+select = [
+    # Rules in alphabetic order
+    "C4",    # flake8-comprehensions
+    "E",     # pycodestyle
+    "F",     # pyflake
+    "G",     # flake8-logging-format
+    "T10",   # flake8-debugger
+    "UP",    # pyupgrade
```
diego-plan9, line 90: From [Qiskit](https://github.com/Qiskit/qiskit/blob/main/pyproject.toml#L223-L234):
* `EXE` - removed as we don't have scripts (other than the ones in `tools/`)
* `F*`- generalized to `F`

From [samplomatic](https://github.com/Qiskit/samplomatic/blob/main/pyproject.toml#L71-L79):
* `F`, `E`, `UP`- adopted
* `D`, `SLF` - removed as requires too many changes
* `I` - I would love to have it, but requires fine tuning of the imports (cyclic, redefined, etc)

pyproject.toml
```diff
@@ -142,6 +169,7 @@ common = [
     "nbconvert>=5.3.1",
     "qiskit-aer>=0.17.0",
     "black~=24.1",
+    "ruff~=0.14.10",
```
diego-plan9, line 172: The original plan was to stick to the version used in `Qiskit`, but it is quite stale.

pyproject.toml
yaelbh: I don't understand. If `black` handles these cases well, then how comes they currently exist? `black` does not handle strings, as you say.

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -205,7 +205,7 @@ def _assert_valid_url(url: str) -> None:
         """Assert that the URL is valid."""
         try:
             urlparse(url)
-        except:
+        except:  # noqa: E722
```
yaelbh: It can be nice to write a add a few words after the error code, so a developer knows immediately what's to be ignored by ruff

qiskit_ibm_runtime/accounts/management.py
```diff
@@ -122,12 +122,12 @@ def _matching_default(account_name: str) -> bool:
...
-        all_account_objects = map(
-            lambda kv: (
+        all_account_objects = (
+            (
                 kv[0],
                 Account.from_saved_format(kv[1]),
-            ),
-            filter_legacy_accounts_dict.items(),
+            )
+            for kv in filter_legacy_accounts_dict.items()
```
yaelbh, line 130: Are the diffs in this file automatic fixes by ruff?

qiskit_ibm_runtime/accounts/storage.py
```diff
@@ -27,7 +27,7 @@ def save_config(
     logger.debug("Save configuration data for '%s' in '%s'", name, filename)
     _ensure_file_exists(filename)
 
-    with open(filename, mode="r", encoding="utf-8") as json_in:
+    with open(filename, encoding="utf-8") as json_in:
```
yaelbh, line 30: Why does ruff bother to make the effort to make such a change? It was even more readable before

test/unit/test_data_serialization.py
```diff
@@ -241,9 +241,8 @@ def test_decoder_import(self):
                 cmd = ["python", temp_fp.name, encoded]
                 proc = subprocess.run(
                     cmd,
-                    stdout=subprocess.PIPE,
-                    stderr=subprocess.PIPE,
-                    universal_newlines=True,
+                    capture_output=True,
+                    text=True,
```
yaelbh, line 245: What a strange diff, what's going on here?

test/unit/test_job_retrieval.py
```diff
@@ -82,8 +81,6 @@ def test_jobs_skip(self, service):
     @run_cloud_fake
     def test_backend_instance_warnings(self, service):
         """Test backend instance warnings do not appear."""
-        if sys.version_info < (3, 10):
-            self.skipTest("assertNoLogs is not supported")
```
yaelbh, line 86: Is this a case that we know for sure to be unreachable?

qiskit_ibm_runtime/accounts/storage.py
diego-plan9: https://docs.astral.sh/ruff/rules/redundant-open-modes/

qiskit_ibm_runtime/accounts/management.py
diego-plan9: Yes, all the changes in the PR are automatic fixes (the individual commits point to which rules were applied).

This change in particular comes from https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4, which are mostly aimed at performance / simplification.

qiskit_ibm_runtime/accounts/storage.py
yaelbh: Do you see what's the confusion that they're trying to avoid? And why does this confusion occur specifically with open modes, and not every time that some function parameter is set to its default value? Anyway if that's ruff's decision then let it be.

test/unit/test_data_serialization.py
diego-plan9: This was colorful indeed - it is https://docs.astral.sh/ruff/rules/replace-stdout-stderr/

test/unit/test_job_retrieval.py
diego-plan9: Yes, this probably should have been included around https://github.com/Qiskit/qiskit-ibm-runtime/pull/2439. Since that PR, we no longer support Python lower than 3.10 - so we cannot hit those `ifs`.

qiskit_ibm_runtime/accounts/storage.py
diego-plan9: I _guess_ the rule takes an assumption that `open` is such a widespread call, that developers assuming that "read" is the default mode makes sense. In general, I'm aiming for a "fire and forget" approach, where we rely on the ruff developers making sensible decisions most of the time, so we can just run `ruff ... --fix` for as many things as possible.

pyproject.toml
diego-plan9: `black`  enforces the line length to be 100 (as per [this](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/pyproject.toml#L6)), but allows for comments and docstrings to go over the limit. `E501` enforces it also for comments and docstrings, so enabling it yields a relative large of errors on those lines.

yaelbh: I see. I didn't understand the original comment, I understand now.

qiskit_ibm_runtime/accounts/account.py
diego-plan9: Revised at 8d2a811.

I think we can just count on some developer diligence (ie. just checking in ruff rules for the code if they need to revisit an offending line for some reason), but since these are the first introduction of `noqa` in the codebase, let's give it a try and re-evaluate after a while.

diego-plan9:
> We should also use only ruff, there's no reason to spend time with two linters. Not in this PR of course.

I like that thought actually - `pylint` is more strict, but `ruff` supports a subset (plus has more rulesets that can be enabled and might match the desires). Part of a longer conversation indeed!

PR #2521, opened 2025-12-23
Include check for rx gates for fractional gates support

### LINKED ISSUE #2520: QiskitRuntimeService should also check for rx in the backend basis when fractional gates=True
**Describe the bug**

`QiskitRuntimeService` should check for both Rx and Rzz gates for backends supporting fractional gates, but only checks for Rzz. 

**Steps to reproduce**

```
backend_name = "ibm_miami"
backend = service.backend(backend_name, use_fractional_gates=True)

```
raises 

`IBMInputValueError: "Backend 'ibm_miami' does not support fractional gates, but use_fractional_gates was set to True."
`
**Expected behavior**

`ibm_miami` supports Rx gate, so it should not raise this error

**Suggested solutions**

Service should also check for Rx gates in the backend basis.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.44

jan-an:
- Fixes #2520  
- `QiskitRuntimeService.backend` also checks for `Rx` gates in the backend basis when setting `use_fractional_gates=True` for input validation ([commit](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2521/commits/83050380e9366513127f804ac67fd9a12684117f))
- Test updated ([commit](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2521/commits/b217ec5360891fa115f3b87782f2b8f4e589f3a8))
- Release notes added ([commit](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2521/commits/570a09d1cd776dffbd0bab0f78ee2ce4fd5799ff))


yaelbh (approved):
Thanks @jan-an.
Note that another bug will arise if you run `backend` with `use_fractional_gates=True`, for a backend that has only `rx` and no `rzz`, then afterwards call `backend` again with the same backend name but `use_fractional=False` - a cached configuration will be taken instead of a refreshed one ([this](https://github.com/Qiskit/qiskit-ibm-runtime/blob/cf542856f08a0e035da8cae2363db0cc98782ef2/qiskit_ibm_runtime/qiskit_runtime_service.py#L723) is the line). Similarly, the line before will cause the configuration to be refreshed again and again for repeated calls to `backend` with `use_fractional=True`, this is not a functional bug but a performance one. I'll open an issue about these bugs.

PR #2525, opened 2025-12-25
[executor-preview]  Fix the executor and test it

yaelbh:
Part of #2479.

This PR copies code fixes and tests not-yet-approved from the open PR #2498. I suggest to first review #2498, let me fix here in parallel according to the comments of #2498, and only then review this PR.


qiskit_ibm_runtime/executor.py
```diff
@@ -51,43 +52,11 @@ class Executor:
...
-            self._service = self._session.service
-
-        elif isinstance(mode, IBMBackend):
-            self._backend = mode
-            self._service = self._backend.service
-
-        else:
-            raise ValueError(
-                "A backend or session/batch must be specified, or a session/batch must be open."
-            )
```
SamFerracin, line 80: Is this equal to what's inside `get_mode_service_backend`, or are there differences?

test/unit/test_ibm_primitives_v2.py
```diff
@@ -650,3 +651,46 @@ def _assert_dict_partially_equal(self, dict1, dict2):
             dict_paritally_equal(dict1, dict2),
             f"{dict1} and {dict2} not partially equal.",
         )
+
+
+class TestGetModeServiceBackend(IBMTestCase):
```
SamFerracin, line 656: Should this be in `test_base_primitive.py`?

test/unit/executor/test_executor.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Tests the `NoiseLearnerV3` class."""
```
SamFerracin:
```suggestion
"""Tests the `Executor` class."""
```

test/unit/test_ibm_primitives_v2.py
yaelbh: Yes, because the function that it tests resides in `base_primitive.py`. I tried to move around the functions and its test in several ways, by starting new files, moving to existing utils files, etc. Everything that I did ended in circular imports that required substantial modifications to unresolve, resulting in effort and amount of modifications that were not justified.

qiskit_ibm_runtime/executor.py
yaelbh: There is one difference: if `get_cm_session()` is not `None` and `mode` is `None`, previously there was an error, now it will take the backend from the session.

SamFerracin (approved):

PR #2526, opened 2025-12-28
[executor-preview]  Test `CircuitItem`

yaelbh:
### Details and comments
Part of #2479

### DIFF
--- a/test/unit/quantum_program/__init__.py
+++ b/test/unit/quantum_program/__init__.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2025.
+# (C) Copyright IBM 2026.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/quantum_program/quantum_program.py
```diff
@@ -80,15 +80,14 @@ def __init__(
...
-            circuit_arguments = np.empty((circuit.num_parameters,), dtype=float)
-        else:
-            circuit_arguments = np.array(circuit_arguments, dtype=float)
+            if circuit.num_parameters:
+                raise ValueError(
+                    f"{repr(circuit)} is parametric, but no 'circuit_arguments' were supplied."
+                )
+            circuit_arguments = []
+
+        circuit_arguments = np.array(circuit_arguments, dtype=float)
```
yaelbh, line 90: There was nothing wrong with the old code, the new code is equivalent, it's only nicer in my opinion.

PR #2527, opened 2025-12-29
[executor-preview] Test `SamplexItem` (including small changes in the code)

yaelbh:
### Summary

Part of #2479

PR #2528, opened 2025-12-30
[executor-preview]  Test `QuantumProgram`

yaelbh:
### Summary

Part of #2479.

### Details and comments
Note that, in addition to testing, there are also some changes in the code.

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/quantum_program.py
+++ b/qiskit_ibm_runtime/quantum_program/quantum_program.py
@@ -234,7 +234,7 @@ class QuantumProgram:
             samplex_arguments: A map from argument names to argument values for the samplex. If this
                 value is provided, a samplex must be present, and ``circuit_arguments`` must not be
                 supplied.
-            samplex_shape: A shape tuple to extend the implicit shape defined by
+            shape: A shape tuple to extend the implicit shape defined by
                 ``samplex_arguments``. Non-trivial axes introduced by this extension enumerate
                 randomizations. If this value is provided, a samplex must be present, and
                 ``circuit_arguments`` must not be supplied.
@@ -246,7 +246,7 @@ class QuantumProgram:
             if samplex_arguments is not None:
                 raise ValueError("'samplex_arguments' cannot be supplied when no samplex is given.")
             if shape is not None:
-                raise ValueError("'samplex_shape' cannot be supplied when no samplex is given.")
+                raise ValueError("'shape' cannot be supplied when no samplex is given.")
             self.items.append(
                 CircuitItem(
                     circuit,
--- a/test/unit/quantum_program/test_quantum_program.py
+++ b/test/unit/quantum_program/test_quantum_program.py
@@ -131,7 +131,7 @@ class TestQuantumProgram(IBMTestCase):
             )
 
         with self.assertRaisesRegex(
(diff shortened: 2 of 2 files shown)


qiskit_ibm_runtime/quantum_program/quantum_program.py
```diff
@@ -137,7 +137,7 @@ def __init__(
         circuit: QuantumCircuit,
         samplex: Samplex,
         *,
-        samplex_arguments: dict[str, np.ndarray | PauliLindbladMap] | None = None,
+        samplex_arguments: dict[str, Any] | None = None,
```
yaelbh, line 140: `np.ndarray | PauliLindbladMap` is wrong, the correct type for the dictionary values is `np.ndarray | dict[str, PauliLindbladMap]`. I prefer `Any`, which is less cumbersome, and also in order no to restrict types of new arguments that we may add in the future.

qiskit_ibm_runtime/quantum_program/quantum_program.py
```diff
@@ -219,34 +219,34 @@ def append(
         *,
         samplex: Samplex | None = None,
         circuit_arguments: np.ndarray | None = None,
-        samplex_arguments: dict[str, np.ndarray] | None = None,
-        shape: tuple[int, ...] | None = None,
+        samplex_arguments: dict[str, Any] | None = None,
+        samplex_shape: tuple[int, ...] | None = None,
```
yaelbh: I prefer this argument name, hope it's ok. `shape` is too general.

qiskit_ibm_runtime/quantum_program/quantum_program.py
```diff
@@ -219,34 +219,34 @@ def append(
...
         chunk_size: int | None = None,
     ) -> None:
         """Append a new :class:`QuantumProgramItem` to this program.
 
         Args:
             circuit: The circuit of this item.
-            samplex: An (optional) samplex to draw random parameters for the circuit.
+            samplex: A samplex to draw random parameters for the circuit.
```
yaelbh, line 230: We don't specify about optional arguments that they're optional, it's encoded in the `| None` of the type hints.

qiskit_ibm_runtime/quantum_program/quantum_program.py
TsafrirA: I initially agreed with your premise (but found the new name to be inaccurate). After some more thought, I think the name is OK, because it is going to represent the shape of the item. It's just that you can only set it manually when dealing with a `SamplexItem`. So, I think the name is OK, but the doc string should change a bit to be clearer.

In any case, the new name is not accurate as the shape is not the shape of the `Samplex` but rather that of the expected samples.

TsafrirA (commented):
This looks good. Left one comment on the `shape` argument name.

qiskit_ibm_runtime/quantum_program/quantum_program.py
yaelbh: To be on the safe side, I think of reverting back to `shape`, here and in #2527.

yaelbh: I renamed back to `shape`, with the intent, in a separate PR after this one and others, to try to replace `append` with `add_circuit` and `add_samplex`.

TsafrirA (approved):
LGTM

PR #2531, opened 2025-12-31
[executor-preview] Test `QuantumProgramResult`

yaelbh:
Part of #2479

### DIFF
--- a/test/unit/quantum_program/__init__.py
+++ b/test/unit/quantum_program/__init__.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2025.
+# (C) Copyright IBM 2026.
 #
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/quantum_program/quantum_program_result.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, data: list[dict[str, np.ndarray]], metadata: Metadata | None
         self._data = data
         self.metadata = metadata or Metadata()
 
-    def __iter__(self) -> Iterable[dict[str, np.ndarray]]:
+    def __iter__(self) -> Iterator[dict[str, np.ndarray]]:
```
yaelbh, line 78: `Iterable` was also good, but `Iterator` is more accurate

PR #2532, opened 2026-01-05
Support executor + noise_learner_v3 schema v0.2

rainerschoe:
### Summary
`ibm-quantum-schemas` will soon introduce new schema versions `v0.2` for executor and noise-learner.

This PR will use the new schema version in runtime-client to encode the input and decode the output for those programs.

### Details and comments
I replaced v0.1 with v0.2, as I did not see a reason for still supporting v0.1 after v0.2 is released. Not sure if this assumption is correct, as there were some counter-indications in the structure of the code.

Fixes #


pyproject.toml
```diff
@@ -89,7 +89,7 @@ dependencies = [
     "urllib3>=1.21.1",
     "python-dateutil>=2.8.0",
     "ibm-platform-services>=0.22.6",
-    "ibm-quantum-schemas @ git+https://github.com/Qiskit/ibm-quantum-schemas.git@d579621311e62b5c342e804a605e4e4e96d88811#egg=ibm-quantum-schemas",
+    "ibm-quantum-schemas @ git+https://github.com/Qiskit/ibm-quantum-schemas.git@1b6b76605f482ae0e42cdf1d53c93fca36a53c67#egg=ibm-quantum-schemas",
```
rainerschoe, line 92: TODO: this currently points to a feature branch. After this is merged, we should point this to a commit on `main`

qiskit_ibm_runtime/noise_learner_v3/converters/version_0_2.py
```diff
@@ -46,26 +46,26 @@ def noise_learner_v3_inputs_to_0_1(
...
 
     return ParamsModel(
-        instructions=QpyModelV13ToV16.from_quantum_circuit(circuit),
+        instructions=QpyModelV13ToV17.from_quantum_circuit(circuit),
         options=options.to_options_model(),
     )
 
 
-def noise_learner_v3_inputs_from_0_1(
+def noise_learner_v3_inputs_from_0_2(
```
rainerschoe, line 54: This function is not needed except in unit-tests. Does not look like production code. Should we move this into tests directory?

qiskit_ibm_runtime/noise_learner_v3/converters/version_0_2.py
```diff
@@ -46,26 +46,26 @@ def noise_learner_v3_inputs_to_0_1(
...
+    """Convert a V0.2 model to noise learner V3 inputs."""
     instructions = list(model.instructions.to_quantum_circuit())
     options = NoiseLearnerV3Options(
         **{key: val for key, val in model.options.model_dump().items() if val}
     )
     return instructions, options
 
 
-def noise_learner_v3_result_to_0_1(
+def noise_learner_v3_result_to_0_2(
```
rainerschoe, line 65: This function is not needed except in unit-tests. Does not look like production code. Should we move this into tests directory?

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_decoders.py
```diff
@@ -19,11 +19,11 @@
 
 # pylint: disable=unused-import,cyclic-import
 from ..utils.result_decoder import ResultDecoder
-from .converters.version_0_1 import noise_learner_v3_result_from_0_1
+from .converters.version_0_2 import noise_learner_v3_result_from_0_2
 
 logger = logging.getLogger(__name__)
 
-AVAILABLE_DECODERS = {"v0.1": noise_learner_v3_result_from_0_1}
+AVAILABLE_DECODERS = {"v0.2": noise_learner_v3_result_from_0_2}
```
rainerschoe, line 26: What is the idea of this dict? I could not see a reason to support multiple decoders as part of this PR.

ihincks: It is common for users to submit a job in one version of qiskit-ibm-runtime, and then wish to re-download the result by job id later in time (possibly months later) to look at it again. Without this dispatch table, they'd have to look around for a compatible qiskit-ibm-runtime version and forcefully downgrade in order to see their data. If they wanted to do some kind of A/B comparison of old data vs new data, they wouldn't be able to easily do it within a single Python environment. 

In short, the user experience would effectively feel like a breaking change between minor versions in certain data analysis workflows.

This is not symmetric in that we do not need to maintain multiple encoders because this same argument doesn't apply. think @mtreinish may have advocated in the past to do so (I'll let him correct me), but whatever those reasons, they I imagine that they are different than the above.

TsafrirA (commented):
Like yourself, I am also a bit puzzled by the backward compatibility here.
Generally speaking, I believe we wanted `qiskit_ibm_runtime` to be able to emit (at least a sub set of) previous schema versions (in case the server side doesn't support a given new version). Given this is not an officially released version of `qiskit_ibm_runtime` I am not sure if we want to be this strict, but it does create a race condition between this PR and the server side PR and deployment. Note that this branch is the only active interface with `Executor` and `NoiseLearnerV3`.

rainerschoe:
Closing as this is covered by different other PRs in the meantime

PR #2541, opened 2026-01-07
Reintroduce delete_job

### LINKED ISSUE #2523: Restore delete_job method
**Describe the bug**

The `delete_job` method was deprecated in #2296 for unknown reasons, just saying it wasn't supported by IQP on Cloud but it is:

https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/jobs

Image

It was removed in release 0.44.0 and now our internal integration tests are breaking as a result.

**Steps to reproduce**

Upgrade to 0.44.0 and try to delete a job.

**Expected behavior**

Should be able to delete a job.

**Suggested solutions**

Restore the `delete_job` method.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.44.0
- **Python version**: n/a
- **Operating system**: n/a

diego-plan9:
### Summary

Reintroduce the `delete_job` endpoint, reverting part of the changes in #2439.

### Details and comments
Fixes #2523

### DIFF
--- a/release-notes/unreleased/2541.feature.rst
+++ b/release-notes/unreleased/2541.feature.rst
@@ -1,0 +1,1 @@
+The ``delete_job()`` method has been reintroduced.


test/integration/test_job.py
```diff
@@ -84,6 +87,35 @@ def test_cancel_job_done(self, service):
         with self.assertRaises(RuntimeInvalidStateError):
             job.cancel()
 
+    @run_integration_test
+    def test_delete_job(self, service):
+        """Test deleting a job."""
+        self.skipTest("Deleting jobs not supported on IBM Quantum Platform.")
```
mriedem, line 93: This is misleading, and probably why the method was removed in the first place.

The delete job API requires a more elevated role in IAM, one that the CI running this probably doesn't have. I'd re-word this at least (and below).

A better pattern would be trying to run the test and if delete job fails with a 401 or 403 error, then skip the test because whoever is running it doesn't have the correct RBAC privileges.

mriedem: I opened issue #2548 for this.

diego-plan9: Oh, good catch! Will follow-up in the issue

PR #2547, opened 2026-01-08
Correct fractional gates configuration check for cached usage of backend object

### LINKED ISSUE #2524: Configuration caching is wrong if a backend contains `rx` but not `rzz`
The [logic to decide whether to refresh the configuration or fetch from a cache](https://github.com/Qiskit/qiskit-ibm-runtime/blob/cd49da0bfb406f3cfdaa54b9d498c44e458422ec/qiskit_ibm_runtime/qiskit_runtime_service.py#L721-L725) assumes that `use_fractional_gates is True` is equivalent to `"rzz" in config.basis_gates`, ignoring cases where `rx` is the only backend's fractional gate. This leads to bugs in both direction: configuration is not refreshed when it should, and vice versa.

To reproduce, call `QiskitRuntimeService.backend` with `ibm_miami` and `use_fractional_gates=True`. Call it again with `use_fractional_gates=True` and debug - you will see that the cache was not used. Now call it with `use_fractional_gates=False`, and see that the configuration still contains `rx`.

jan-an:
## Summary 

- Fixes #2524 
- Caching mechanism tests for both rx/rzz in basis gates to create or reuse backend object from cache

### DIFF
--- a/release-notes/unreleased/2524.bug.rst
+++ b/release-notes/unreleased/2524.bug.rst
@@ -1,5 +1,5 @@
 While creating backend objects, `.QiskitRuntimeService` used to check for only ``rzz`` 
-gates in the backend's basis to decide whether to use cache or create a new backend 
+gates in the backend's basis gates to decide whether to use cache or create a new backend 
 object. Now `.QiskitRuntimeService` will check for both ``rzz`` and ``rx`` gates in the 
 backend's basis gates to decide whether to refresh the configuration or fetch 
 from cache.


release-notes/unreleased/2524.bug.rst
```diff
@@ -0,0 +1,3 @@
+`.QiskitRuntimeService` will now check for both ``rzz`` and ``rx`` gates in the 
+backend's basis gates to decide whether to refresh the configuration or fetch 
+from cache
```
yaelbh:
```suggestion
from cache.
```

yaelbh: Please add a sentence that explains what's the change here, i.e., describe how it was before.

test/integration/test_backend.py
```diff
@@ -276,6 +276,24 @@ def test_backend_fractional_gates_error(self):
             ):
                 self.service.backend(backend.name, use_fractional_gates=True)
 
+    def test_backend_fractional_gates_cache_behavior(self):
+        """Test backend config cache/refresh logic for use_fractional_gates."""
+        try:
+            real_device_name = "ibm_miami"
+            backend_fg = self.service.backend(real_device_name, use_fractional_gates=True)
+            backend_fg2 = self.service.backend(real_device_name, use_fractional_gates=True)
+            backend_no_fg = self.service.backend(real_device_name, use_fractional_gates=False)
```
yaelbh, line 285: Consider adding:
```python
backend_no_fg2 = self.service.backend(real_device_name, use_fractional_gates=False)
backend_fg_3 = self.service.backend(real_device_name, use_fractional_gates=True)
```
and the corresponding assertions below. This way all four cases of (previous fg, current fg) will be covered.

yaelbh (commented):
Thanks @jan-an, it's very good and my comments are small.

release-notes/unreleased/2524.bug.rst
jan-an: Added sentence ([commit](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2547/commits/27632ab0645f943a39f77e760e2e5dbf8c724f3b)) 👍

test/integration/test_backend.py
jan-an: Added additional backend objects and assert statements as suggested ([commit](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2547/commits/a8a8b1e4d19dcdbfa9976eca4e18ce8c31bf68e9))

release-notes/unreleased/2524.bug.rst
jan-an: Hi @yaelbh, thank you for your comments and suggestions! I've made the changes, can you kindly take a look?

test/integration/test_backend.py
```diff
@@ -276,6 +276,34 @@ def test_backend_fractional_gates_error(self):
...
+
+        self.assertIsNot(
+            backend_fg, backend_no_fg, "Config refresh when use_fractional_gates changed"
+        )
+
+        self.assertIs(backend_no_fg, backend_no_fg2, "Cache used to create backend object")
+
+        self.assertIsNot(
+            backend_no_fg2, backend_fg3, "Configuration refreshes when use_fractional_gates changes"
+        )
```
yaelbh, line 299: This assertion can be deleted, it's a direct consequence of the next assertion.

test/integration/test_backend.py
yaelbh: The error messages were changed in this last commit, and I think the previous version was the correct one, please check.

release-notes/unreleased/2524.bug.rst
```diff
@@ -0,0 +1,5 @@
+While creating backend objects from cache when ``use_fractional_gates`` flag is used,
```
yaelbh: When we start to create we still don't know if it's from cache. Also, the fix applies to both when the flag is used and when it's not.

release-notes/unreleased/2524.bug.rst
```diff
@@ -0,0 +1,5 @@
+While creating backend objects from cache when ``use_fractional_gates`` flag is used, 
+`.QiskitRuntimeService` used to check for only ``rzz`` gates in the backend's basis.
```
yaelbh: Check for what? (check for whether to take from cache).

release-notes/unreleased/2524.bug.rst
```diff
@@ -0,0 +1,5 @@
+While creating backend objects from cache when ``use_fractional_gates`` flag is used, 
+`.QiskitRuntimeService` used to check for only ``rzz`` gates in the backend's basis. 
+Now `.QiskitRuntimeService` will now check for both ``rzz`` and ``rx`` gates in the
```
yaelbh:
```suggestion
Now `.QiskitRuntimeService` will check for both ``rzz`` and ``rx`` gates in the 
```

test/integration/test_backend.py
jan-an: Hi, I was writing some comments to follow what each assert was doing but forgot to negate them when I copied them for raising the error message, that's why this happened. Sorry about that, I've corrected the tests and the error messages again, they should be okay now.

release-notes/unreleased/2524.bug.rst
jan-an: Made changes for to the release note 👍

release-notes/unreleased/2524.bug.rst
```diff
@@ -0,0 +1,5 @@
+While creating backend objects, `.QiskitRuntimeService` used to check for only ``rzz`` 
+gates in the backend's basis to decide whether to use cache or create a new backend
```
yaelbh:
```suggestion
gates in the backend's basis gates to decide whether to use cache or create a new backend 
```

yaelbh (approved):

PR #2550, opened 2026-01-08
Update 0.44 release notes

diego-plan9:
### Summary

While preparing for a new release, it seems a number of changes for `0.44.0` were not included as the fragments did not have the names expected by `towncrier`. This PR updates the `0.44.0` release notes retroactively, by renaming the files and re-running `towncrier` (and copy-pasting the result into the right sections).

### Details and comments
Fixes #

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,7 @@ language = "en"
 # The short X.Y version
 version = ""
 # The full version, including alpha/beta/rc tags
-release = '0.44.0'
+release = '0.45.0'
 
 # -- General configuration ---------------------------------------------------
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ zip-safe = false
 root = "."
 write_to = "qiskit_ibm_runtime/VERSION.txt"
 version_scheme = "release-branch-semver"
-fallback_version = "0.44.0"
+fallback_version = "0.45.0"
 
 [tool.mypy]
 python_version = "3.10"
--- a/release-notes/0.45.0.rst
+++ b/release-notes/0.45.0.rst
@@ -1,0 +1,22 @@
+0.45.0 (2026-01-08)
+===================
+
+New Features
+------------
+
+- ``qiskit-ibm-runtime`` is now compatible with Qiskit v2.3.
+- The ``delete_job()`` method has been reintroduced. (`2541 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2541>`__)
+
+
+Bug Fixes
+---------
+
+- Bugfix for :meth:`.QiskitRuntimeService.backend` when the flag ``use_fractional_gates=True``
+  checks for either or both ``Rx`` and ``Rzz`` gates as opposed to only ``Rzz`` gates in the backend basis. (`2520 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2520>`__)
+
+
+Other Notes
+-----------
+
+- The minimal version of Qiskit required is now ``2.0.0``, as the ``1.x`` Qiskit
+  series is no longer supported. (`2543 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2543>`__)
--- a/release-notes/unreleased/2520.bug.rst
+++ b/release-notes/unreleased/2520.bug.rst
@@ -1,2 +1,0 @@
-Bugfix for :meth:`.QiskitRuntimeService.backend` when the flag ``use_fractional_gates=True``
-checks for either or both ``Rx`` and ``Rzz`` gates as opposed to only ``Rzz`` gates in the backend basis.
--- a/release-notes/unreleased/2541.feature.rst
+++ b/release-notes/unreleased/2541.feature.rst
@@ -1,1 +1,0 @@
-The ``delete_job()`` method has been reintroduced.
--- a/release-notes/unreleased/2543.other.rst
+++ b/release-notes/unreleased/2543.other.rst
@@ -1,2 +1,0 @@
-The minimal version of Qiskit required is now ``2.0.0``, as the ``1.x`` Qiskit
-series are no longer supported.


release-notes/0.44.0.rst
```diff
@@ -1,6 +1,39 @@
 0.44.0 (2025-12-16)
 ===================
 
+Upgrade Notes
+-------------
+
+- If a backend that does not support fractional gates is requested by setting the flag
+  ``use_fractional_gates=True`` in :meth:`.QiskitRuntimeService.backend`, 
+  an ``IBMInputValueError`` will be raised. (`2405 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2405>`__)
```
abbycross:
```suggestion
 ``IBMInputValueError`` will be raised. (`2405 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2405>`__)
```

release-notes/0.44.0.rst
```diff
@@ -1,6 +1,39 @@
...
+- The URL resolver was updated so that both ``ibm_cloud`` and ``ibm_quantum_platform`` channels now generate the 
+  same runtime entrypoint URL (``https://quantum.cloud.ibm.com/api/v[x]``).
+  Previously, the ``ibm_cloud`` channel produced a decommissioned URL and relied on redirects. This change removes 
+  that dependency and aligns URL generation across channels. There are no user-facing changes. (`2486 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2486>`__)
+- Impose a y-axis channel ordering for :meth:`draw_circuit_schedule_timing` according
```
abbycross:
```suggestion
- A y-axis channel ordering for :meth:`draw_circuit_schedule_timing` has now been imposed, according 
```

release-notes/0.44.0.rst
```diff
@@ -1,6 +1,39 @@
...
+- If a backend that does not support fractional gates is requested by setting the flag
+  ``use_fractional_gates=True`` in :meth:`.QiskitRuntimeService.backend`, 
+  an ``IBMInputValueError`` will be raised. (`2405 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2405>`__)
+- The URL resolver was updated so that both ``ibm_cloud`` and ``ibm_quantum_platform`` channels now generate the 
+  same runtime entrypoint URL (``https://quantum.cloud.ibm.com/api/v[x]``).
+  Previously, the ``ibm_cloud`` channel produced a decommissioned URL and relied on redirects. This change removes 
+  that dependency and aligns URL generation across channels. There are no user-facing changes. (`2486 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2486>`__)
+- Impose a y-axis channel ordering for :meth:`draw_circuit_schedule_timing` according 
+  to the order of channels in the ``included_channels`` input argument. (`2492 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2492>`__)
+- Add a new ``merge_common_instructions`` input argument to :meth:`draw_circuit_schedule_timing` that
```
abbycross:
```suggestion
- Added a new ``merge_common_instructions`` input argument to :meth:`draw_circuit_schedule_timing` that, 
```

release-notes/0.44.0.rst
```diff
@@ -1,6 +1,39 @@
...
+  ``use_fractional_gates=True`` in :meth:`.QiskitRuntimeService.backend`, 
+  an ``IBMInputValueError`` will be raised. (`2405 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2405>`__)
+- The URL resolver was updated so that both ``ibm_cloud`` and ``ibm_quantum_platform`` channels now generate the 
+  same runtime entrypoint URL (``https://quantum.cloud.ibm.com/api/v[x]``).
+  Previously, the ``ibm_cloud`` channel produced a decommissioned URL and relied on redirects. This change removes 
+  that dependency and aligns URL generation across channels. There are no user-facing changes. (`2486 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2486>`__)
+- Impose a y-axis channel ordering for :meth:`draw_circuit_schedule_timing` according 
+  to the order of channels in the ``included_channels`` input argument. (`2492 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2492>`__)
+- Add a new ``merge_common_instructions`` input argument to :meth:`draw_circuit_schedule_timing` that 
+  if set to ``True`` will merge instructions of the same type based on temporal continuity.
```
abbycross:
```suggestion
  if set to ``True``, will merge instructions of the same type based on temporal continuity.
```

release-notes/0.44.0.rst
```diff
@@ -1,6 +1,39 @@
...
+
+  Example,
```
abbycross:
```suggestion
  Example:
```

release-notes/0.44.0.rst
```diff
@@ -14,6 +47,67 @@ New Features
...
   600% improvement in circuit depth compared to a Heron backend. (`2476 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2476>`__)
+- The :mod:`.qiskit_ibm_runtime.transpiler.passes` module has been extended with a 
+  new transpiler pass: :class:`.ConvertMidCircMeas`. This pass detects non-terminal 
+  :class:`~qiskit.circuit.Measure` instructions in a circuit and converts them into 
+  mid-circuit measurement instructions supported by the target.
+  The pass is constructed with two arguments:
+
+  - ``target`` (required): The backend target defining supported instructions.
+  - ``mcm_name`` (optional): The name of the mid-circuit measurement instruction to use. 
+      Defaults to ``"measure_2"``. The name must follow the pattern "measure_[X]" (e.g., "measure_3", "measure_reset").
```
abbycross:
```suggestion
      Defaults to ``"measure_2"``. The name must follow the pattern "measure_[X]" (for example, "measure_3", "measure_reset").
```

release-notes/0.44.0.rst
```diff
@@ -14,6 +47,67 @@ New Features
...
+
+  During initialization, the pass validates that the specified target supports the given 
+  mid-circuit measurement instruction on at least one qubit. If not, construction fails.
+  When executed, the pass:
+
+  - Runs on an already routed circuit.
+  - Replaces only non-terminal :class:`.~qiskit.circuit.Measure`` instructions with the specified mid-circuit
```
abbycross:
```suggestion
  When executed, the pass does the following:

  - It runs on an already routed circuit.
  - It replaces only non-terminal :class:`.~qiskit.circuit.Measure`` instructions with the specified mid-circuit 
```

release-notes/0.44.0.rst
```diff
@@ -14,6 +47,67 @@ New Features
...
+
+  - Runs on an already routed circuit.
+  - Replaces only non-terminal :class:`.~qiskit.circuit.Measure`` instructions with the specified mid-circuit 
+  measurement instruction, provided the target supports it on the corresponding physical qubit.
+
+  It does not:
+
+  - Replace terminal measurements.
+  - Replace existing mid-circuit measurement instructions (e.g., "measure_2" -> "measure_3").
+  - Convert mid-circuit measurements back to terminal measurements.
```
abbycross:
```suggestion
  It does not do the following:

  - It does not replace terminal measurements.
  - It does not replace existing mid-circuit measurement instructions (for example, ``measure_2`` to ``measure_3``).
  - It does not convert mid-circuit measurements back to terminal measurements.
```

abbycross: Not sure how the -> will be handled by the auto translation. I suggest replacing it with the unicode (which would be `&rarr;`)- but I don't know if that is ok to insert into this .rst source file? @arnaucasau do you know?

release-notes/0.44.0.rst
```diff
@@ -14,6 +47,67 @@ New Features
...
+
+  Example usage::
```
abbycross: Is the double colon here and on line 99 intentional?

diego-plan9: A couble colon is present in the [original fragment](https://github.com/Qiskit/qiskit-ibm-runtime/blob/ae939a99b4df69d1b877040a1de2f2d37ab201a0/release-notes/unreleased/2317.feature.rst). I assume it is for marking a [literal block](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) in pure ReST when rendering, but I don't have visibility on how it is treated by the full documentation machinery.

release-notes/0.44.0.rst
arnaucasau: Unfortunately, I don't think we can use `&rarr;` in `.rst`. After testing it, It seems like the arrow as text ("->") is not a problem for translations (at least, in this specific case), but maybe we can replace it by "to" to make sure it will work.

Also, something that will be hard to translate is the "measure_2" and "measure_3"  references. Could we write them as code by using backticks, please? I think the syntax should be \`\`measure_2\`\` and \`\`measure_3\`\`

release-notes/0.44.0.rst
abbycross: I checked some older release notes in this repo and confirmed that when it's a double colon in the release notes, it renders as a single colon in the published platform docs - all good! ✅

release-notes/0.44.0.rst
abbycross: Thanks @arnaucasau - I edited the suggestion snippet to reflect those changes.

release-notes/0.44.0.rst
```diff
@@ -14,6 +47,67 @@ New Features
...
   600% improvement in circuit depth compared to a Heron backend. (`2476 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2476>`__)
+- The :mod:`.qiskit_ibm_runtime.transpiler.passes` module has been extended with a 
+  new transpiler pass: :class:`.ConvertMidCircMeas`. This pass detects non-terminal 
+  :class:`~qiskit.circuit.Measure` instructions in a circuit and converts them into 
+  mid-circuit measurement instructions supported by the target.
+  The pass is constructed with two arguments:
+
+  - ``target`` (required): The backend target defining supported instructions.
+  - ``mcm_name`` (optional): The name of the mid-circuit measurement instruction to use. 
+      Defaults to ``"measure_2"``. The name must follow the pattern "measure_[X]" (for example, "measure_3", "measure_reset").
```
arnaucasau: As mentioned in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2550/files#r2673253171, not using inline code might mess the automatic translations. I also removed the double quotes of `measure_2` in the suggestion because I think they could be redundant with the back-ticks:

```suggestion
      Defaults to ``measure_2``. The name must follow the pattern ``measure_[X]`` (for example, ``measure_3``, ``measure_reset``).
```

arnaucasau (commented):
Thanks!

release-notes/0.44.0.rst
arnaucasau: Awesome! I also added this suggestion from another part of the file -> https://github.com/Qiskit/qiskit-ibm-runtime/pull/2550#discussion_r2675589946

abbycross (approved):
Thanks - looks good!!

SamFerracin (approved):

PR #2551, opened 2026-01-09
Fix branch ruff warnings (UP007, C401, E741)

diego-plan9:
### Summary

`0.45.0` includes `ruff` as part of the tooling, and enforces it to pass as part of the `make lint` stage - this PR takes care of the warnings that will be emitted when syncing the branch to `0.45.0`, for having an easier time later on.

### Details and comments
Fixes #N/A

PR #2552, opened 2026-01-09
Fix other branch ruff warnings

diego-plan9:
### Summary

In a similar spirit to #2551, upon closer inspection there were more `ruff` warnings being emitted preparing for syncing with `0.45.0`. This PR fixes them - all of them were autofixed by `--fix`.

### Details and comments
Fixes #

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/quantum_program_decoders.py
+++ b/qiskit_ibm_runtime/quantum_program/quantum_program_decoders.py
@@ -35,7 +35,7 @@ class QuantumProgramResultDecoder(ResultDecoder):
     @classmethod
     def decode(cls, raw_result: str):  # type: ignore[no-untyped-def]
         """Decode raw json to result type."""
-        decoded: dict = super().decode(raw_result)
+        decoded: dict[str, str] = super().decode(raw_result)
 
         try:
             schema_version = decoded["schema_version"]


qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -168,7 +168,7 @@ class IBMBackend(Backend):
     def __init__(
         self,
         configuration: QasmBackendConfiguration,
-        service: "qiskit_runtime_service.QiskitRuntimeService",
+        service: qiskit_runtime_service.QiskitRuntimeService,
```
diego-plan9, line 171: This (and other changes in this file) are being picked up by `ruff` in the branch, but not when running `ruff` in `main`. I have not been able to pinpoint the cause (for what is worth, both in the branch and in `main` other quoted references are _not_ picked up, such as https://github.com/diego-plan9/qiskit-ibm-runtime/blob/f0ed4a674ea06e18ba748d82ec78d198577215d1/qiskit_ibm_runtime/base_runtime_job.py#L59).

My thinking is to go ahead with this changes (as long as tooling passes), and after reconciling the branches, apply the changes in this file to `main` - and investigate what is going on.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_decoders.py
```diff
@@ -32,7 +31,7 @@ class NoiseLearnerV3ResultDecoder(ResultDecoder):
     @classmethod
     def decode(cls, raw_result: str):  # type: ignore[no-untyped-def]
         """Decode raw json to result type."""
-        decoded: Dict = super().decode(raw_result)
+        decoded: dict = super().decode(raw_result)
```
SamFerracin: I'm surprised you can get away with this, but if it ever yells, this should be:

```suggestion
        decoded: dict[str, str] = super().decode(raw_result)
```

qiskit_ibm_runtime/quantum_program/quantum_program_decoders.py
```diff
@@ -36,7 +35,7 @@ class QuantumProgramResultDecoder(ResultDecoder):
     @classmethod
     def decode(cls, raw_result: str):  # type: ignore[no-untyped-def]
         """Decode raw json to result type."""
-        decoded: Dict = super().decode(raw_result)
+        decoded: dict = super().decode(raw_result)
```
SamFerracin: Same here:
```suggestion
        decoded: dict[str, str] = super().decode(raw_result)
```

qiskit_ibm_runtime/ibm_backend.py
SamFerracin: sounds good

PR #2556, opened 2026-01-11
[executor-preview] Arrange NLV3 options

yaelbh:
### Summary

Part of #2479.

### Details and comments

- Deleted all unused methods of `NoiseLearnerV3Options`.
- Added support to multiple schema versions in `to_options_model`. 
- Changed `to_runtime_options` (see more detailed comment below).


qiskit_ibm_runtime/options/noise_learner_v3_options.py
```diff
@@ -99,47 +94,29 @@ def _nonnegative_list(cls, value: list[int], info: ValidationInfo) -> list[int]:
...
-
-        remove_dict_unset_values(filtered_options)
-        return filtered_options
-
-    def get_callback(self) -> Callable | None:
-        """Get the callback."""
-        options_dict = asdict(self)
-        remove_dict_unset_values(options_dict)
-        return options_dict.get("environment", {}).get("callback", None)
+        return self._get_runtime_options(asdict(self))
```
yaelbh, line 119: Note that this is a change of behavior for `to_runtime_options`. Before the change, when `max_execution_time` was `Unset`, the resulting dictionary didn't contain `max_execution_time`. After the change it contains `"max_execution_time": None`.

qiskit_ibm_runtime/options/noise_learner_v3_options.py
```diff
@@ -151,70 +128,11 @@ def get_callback(self) -> Callable | None:
...
-        remove_empty_dict(output_options)
-
-        inputs = {
-            "options": output_options,
-        }
-        if options_copy.get("resilience_level", Unset) != Unset:
-            inputs["resilience_level"] = options_copy["resilience_level"]
-
-        return inputs
+        raise NotImplementedError()
```
SamFerracin: Maybe let's add a message?
```suggestion
        raise NotImplementedError("Not implemented by `NoiseLearnerV3Options`.")
```

test/unit/options/test_noise_learner_v3_options.py
```diff
@@ -0,0 +1,51 @@
...
+
+from qiskit_ibm_runtime.options.noise_learner_v3_options import NoiseLearnerV3Options
+
+from ...ibm_test_case import IBMTestCase
+
+
+class TestNoiseLearnerV3Options(IBMTestCase):
+    """Tests the ``NoiseLearnerV3Options`` class."""
+
+    def test_to_options_model(self):
```
SamFerracin, line 23: So we'll end up parametrizing this when we have more schemas?

qiskit_ibm_runtime/options/noise_learner_v3_options.py
SamFerracin: This shouldn't be an issue, correct?

yaelbh: Yes, to my understanding it's the correct behavior now.

test/unit/options/test_noise_learner_v3_options.py
yaelbh: Yes

PR #2561, opened 2026-01-12
Make tests on delete_job skippable

### LINKED ISSUE #2548: Change how delete_job tests are skipped to avoid confusion
This is misleading, and probably why the method was removed in the first place.

The delete job API requires a more elevated role in IAM, one that the CI running this probably doesn't have. I'd re-word this at least (and below).

A better pattern would be trying to run the test and if delete job fails with a 401 or 403 error, then skip the test because whoever is running it doesn't have the correct RBAC privileges.

_Originally posted by @mriedem in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2541#discussion_r2672479809_

diego-plan9:
### Summary

Tweak the reintroduced `delete_job` tests (from #2541) so they are skipped in case the credentials do not allow for deletion of jobs. 

### Details and comments

Fixes #2548

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -135,7 +135,7 @@ def cancel_job_safe(job: RuntimeJobV2, logger: logging.Logger) -> bool:
         raise
 
 
-def wait_for_status(job, status: str, poll_time=1, time_out=20):
+def wait_for_status(job: RuntimeJobV2, status: str, poll_time: int = 1, time_out: int = 20) -> None:
     """Wait for job to reach a certain status."""
     wait_time = 1 if status == "QUEUED" else poll_time
     while job.status() not in ["DONE", "CANCELLED", "ERROR", status] and time_out > 0:


test/integration/test_job.py
```diff
@@ -90,29 +91,35 @@ def test_cancel_job_done(self, service):
...
         for status in sub_tests:
             with self.subTest(status=status):
                 job = self._run_program(service)
                 wait_for_status(job, status)
-                service.delete_job(job.job_id())
+                try:
+                    service.delete_job(job.job_id())
+                except IBMRuntimeError as ex:
+                    if "403" in ex.message or "401" in ex.message:
+                        self.skipTest("Credentials do not have delete job priviledges")
```
yaelbh:
```suggestion
                        self.skipTest("Credentials do not have delete job privileges")
```

test/integration/test_job.py
```diff
@@ -90,29 +91,35 @@ def test_cancel_job_done(self, service):
...
         isa_circuit = pm.run([bell()])
         _ = self._run_program(service, circuits=isa_circuit, backend=real_device_name)
         job = self._run_program(service, circuits=isa_circuit, backend=real_device_name)
         wait_for_status(job, "QUEUED")
-        service.delete_job(job.job_id())
+        try:
+            service.delete_job(job.job_id())
+        except IBMRuntimeError as ex:
+            if "403" in ex.message or "401" in ex.message:
+                self.skipTest("Credentials do not have delete job priviledges")
```
yaelbh:
```suggestion
                self.skipTest("Credentials do not have delete job privileges")
```

diego-plan9:
Ah, I misread and thought you were asking to avoid skipping if the HTTP error was `401`. I pushed f9e1f2f with a less ambiguous matching: unfortunately, the real `RequestsApiError` exception does not get raised `from` the `IBMRuntimeError` so we are limited to comparing the strings, but hopefully this will indeed prevent it picking up other strings unintendedly.

test/integration/test_retrieve_job.py
```diff
@@ -78,14 +78,13 @@ def test_retrieve_jobs_limit(self, service):
...
         job = self._run_program(service)
-        wait_for_status(job, JobStatus.RUNNING)
+        wait_for_status(job, "RUNNING")
         rjobs = service.jobs(pending=True)
         after_status = job.status()
         found = False
         for rjob in rjobs:
             if rjob.job_id() == job.job_id():
                 self.assertEqual(job.primitive_id, rjob.primitive_id)
-                self.assertEqual(job.inputs, rjob.inputs)
```
diego-plan9, line 88: This relaxes the assertion (as the pubs contain an empty numpy array, which cannot be compared directly). Takes advantage of the "sibling" test `test_retrieve_returned_jobs()` not having the assertion already, so assuming comparing just on the job id provide enough confidence.

yaelbh: Why can't they be compared? And what is there to compare in empty arrays?

diego-plan9: Due to numpy conventions:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()]
```

Since the existing test for retrieving returned jobs only asserts on the job id, I think it makes sense to make the same assumption for this test, rather than investing into a comparison function that can cope with empty arrays.

yaelbh: Numpy arrays are compared with `np.array_equal`, which works for arrays of any length. You can omit the comparison if you think it's redundant, but only because you think it's redundant, and not because of considerations related to inconvenience caused by the comparison, which is supposed to be easy.

diego-plan9: I don't think it is redundant, I think it is not necessary and aligned with the other test - a trade of in the end.

yaelbh: We're running in circle here... Ok to be aligned with the other test, it's a good reason to remove something that's not necessary. No tradeoff though, because running `np.array_equal` doesn't cost anything.

diego-plan9: > as the pubs contain an empty numpy array, which cannot be compared directly

It is a tradeoff - the original assert is over the inputs (ie. the full dict, which has nested items, and end up having an empty array on the pubs list), not just on the arrays, so it requires coming up with a custom crafted comparison for the inputs with arguably no strong gain.

PR #2562, opened 2026-01-12
Handling missing supported_instructions

### LINKED ISSUE #2560: Change the fallback for backends whose configuration lacks `supported_instructions`
Backends are normally supposed to have supported instructions, i.e., an attribute called `supported_instructions` in their configuration. Only a small set of old fake backends don't have it:
```
fake_almaden
fake_burlington
fake_cambridge
fake_essex
fake_johannesburg
fake_london
fake_poughkeepsie
fake_rochester
fake_singapore
```

When they are missing, [we treat them as an empty list](https://github.com/Qiskit/qiskit-ibm-runtime/blob/de4b6354050f9d5e0900e7d79d342928f9663976/qiskit_ibm_runtime/utils/backend_converter.py#L111). This default should be replaced with `["measure", "reset", "delay"]`.

An alternative is to update the old fake backends, and raise an error if the configuration is missing the `supported_instructions` attribute. This alternative provides more robustness, but I prefer the previous approach, being a one-line code change.

yaelbh:
### Summary

Fixes #2560.

### Details and comments

@diego Can we run the integration tests before we merge the PR?

### DIFF
--- a/test/integration/test_fake_backends.py
+++ b/test/integration/test_fake_backends.py
@@ -14,11 +14,8 @@
 # pylint: disable=missing-module-docstring
 
 from qiskit_ibm_runtime import QiskitRuntimeService
+from qiskit_ibm_runtime.fake_provider import FakeSherbrooke
 
-from qiskit_ibm_runtime.fake_provider import (
-    FakeProviderForBackendV2,
-    FakeSherbrooke,
-)
 from ..ibm_test_case import IBMIntegrationTestCase
 from ..decorators import production_only
 


test/integration/test_fake_backends.py
```diff
@@ -13,151 +13,18 @@
...
     FakeProviderForBackendV2,
-    FakeMumbaiV2,
     FakeSherbrooke,
-    FakePrague,
 )
-from ..ibm_test_case import IBMTestCase, IBMIntegrationTestCase
+from ..ibm_test_case import IBMIntegrationTestCase
 from ..decorators import production_only
 
 FAKE_PROVIDER_FOR_BACKEND_V2 = FakeProviderForBackendV2()
```
diego-plan9: Can you double check if this is needed (along with the import)? Seems not referenced with the rest of the lines moved.

yaelbh: It wouldn't pass lint if it wasn't used. It's used in line 28, where we define
```python
class TestRefreshFakeBackends(IBMIntegrationTestCase):
```

diego-plan9: I mean
```
FAKE_PROVIDER_FOR_BACKEND_V2 = FakeProviderForBackendV2()
```

yaelbh: Thanks for catching, I've removed it.

Issue #2522, opened 2025-12-24
Make integration tests not depend on specific backends


yaelbh:
The [noise learner tests](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/integration/test_noise_learner.py) consist of circuits with `ecr` instructions. They assume that the backend has `ecr` as a basis gate, and that the edges that appear in the circuit belong to the backend's coupling map. We would like to make the tests more robust to changes in backends.

My first idea was to check which of `ecr, cx, cz` is present in the backend's set of basis gates, and construct the circuits accordingly, taking into account also the coupling map. Something like this might work, but it will not be elegant, because the checks that we perform in the tests depend on the actual circuits. For example, for noise learning, the partition to layers can be different for different (equivalent) circuits.

One can also propose to transpile the circuits, but then again, the tests behave differently for different (equivalent) transpiled circuits.

To summarize, we can try to implement something similar to the "first idea", but a new totally different idea will be appreciated. I even tend to prefer to just allow the tests to depend on the backend, i.e., to keep assuming that some 2Q gate is a basis gate (preferably `cz`, as it is what we have in our Heron systems), and restrict the tests to run only on those backends. Dealing with the coupling map is much easier, and we can, with small effort, alleviate the current assumption of a specific coupling map.

All of the above may apply also for other current and future tests, not necessarily noise learning tests. However I'm not aware of any such other tests. For many such other tests (future ones or current ones that I'm not aware of) I expect transpilation to form a simple solution (not applicable for noise learning though, as explained above).

PR #2565, opened 2026-01-14
Fix integration tests issues

diego-plan9:
### Summary

A number of integration tests are failing, for different reasons. This PR aims to get the integration tests to pass due to conditionally skipping and/or some fixes. More details in comments inline.

### Details and comments

Related to #2522 (not marking as fixes, as this circumvents around the issue by skipping those tests).


test/utils.py
```diff
@@ -89,7 +89,11 @@ def most_busy_backend(
     """
     backends = service.backends(simulator=False, operational=True, instance=instance)
     return max(
-        (b for b in backends if b.configuration().n_qubits >= 5),
+        (
+            b
+            for b in backends
+            if b.configuration().n_qubits >= 5 and b.status().status_msg == "active"
+        ),
```
diego-plan9, line 96: This prevents picking up a backend that is currently paused and not working correctly.

test/integration/test_noise_learner.py
```diff
@@ -58,7 +59,11 @@ def test_with_default_options(self, service):  # pylint: disable=unused-argument
         options = NoiseLearnerOptions()
         learner = NoiseLearner(mode=self._backend, options=options)
 
-        job = learner.run(self.circuits)
+        try:
+            job = learner.run(self.circuits)
+        except IBMInputValueError as ex:
+            if "The instruction ecr on qubits (0, 1) is not supported" in ex.message:
+                self.skipTest("Backend does not meet requirements")
```
diego-plan9: These are the bits relevant to #2522.

test/integration/test_fake_backends.py
```diff
@@ -180,6 +181,12 @@ def test_refresh_method(self):
             url=self.dependencies.url,
         )
 
+        # This tests needs access to the real device, and it might not be available.
+        try:
+            service.backend("sherbrooke")
+        except QiskitBackendNotFoundError:
+            self.skipTest("Credentials do not have access to sherbrooke")
```
SamFerracin: Two comments:

1. I think this is what you want..?
```suggestion
        try:
            service.backend("ibm_sherbrooke")
        except QiskitBackendNotFoundError:
            self.skipTest("Credentials do not have access to ibm_sherbrooke")
```

2. What happens to this test when Sherbrooke is retired? Will it keep passing because if the try/except?

test/integration/test_noise_learner.py
SamFerracin: Instead of skipping the test, I would simply transpile:


```suggestion
        pm = generate_preset_pass_manager(backend=backend, optimization_level=0)
        circuit = pm.run(self.circuit)
```

I'm a little bit nervous about the `3` in `self._verify(job, self.default_input_options, 3)`, which I expect corresponds to the number of instructions learned by the noise learner. If we transpile, this number *may* become larger depending on the backend. However, I would prefer to address that (even with a check that the # of instructions is >=3) than skipping the test

SamFerracin: The same comment applies to all the tests below

test/integration/test_fake_backends.py
diego-plan9: Yes it is, thanks!

> What happens to this test when Sherbrooke is retired? Will it keep passing because if the try/except?

It will be skipped, so we have to remember to update it when sherbrooke is retired. We might have other hard-coded backend names in the tests, and maybe just have a convention to grep over the backend name when a backend is retired. I see this as part of a bigger test revamp to be tackled outside this issue, and I think we can live with this risk for the time being.

test/integration/test_noise_learner.py
diego-plan9: Thanks - I applied the suggested change (not verbatim, as it needed some imports and tweaks) to all tests but one at 82ff53e41f02b1a3a619065b9e576ec1b25e382c, and it is passing in both integration tests. 

`test_learner_plus_estimator` requires further tweaks: when transpiling the circuits (the ones for the noise learner, and the one for the estimator), the assertions at the end of the test fail. I suggest tackling it as part of #2522, and aiming this PR for getting the integration tests to pass, even if there is some skipping.

Issue #2514, opened 2025-12-19
Revise typing ignores


diego-plan9:
**What is the expected feature or enhancement?**

The codebase has a number of `# type: ignore[...]` that might hide actual bugs or hinder reasoning about code. It would be great to gradually aim to reduce the number of them, and in the process get the typing annotations of the package more useful and expressive.

This is a general issue for improving the situation incrementally.

**Acceptance criteria**

The number of `# type: ignore[...]` statements is reduced.

rajnisht7:
as i checked, appears like around 90 can be removed  by refactoring and remaining are due to third party libraries, for 90 around pushing PR in batches seems to be better options thats those with similar modification

diego-plan9:
Indeed - this is more of a "tackle bit by bit" issue, specially as the changes can be noisy and prone to cause conflicts with other PRs in motion. If interested in a specific subset, please comment before tackling for better coordination.

PR #2571, opened 2026-01-20
Remove some mypy ignores

diego-plan9:
### Summary

Remove some `mypy` ignores:
* unify `tags` types
* explicitly declare some types to allow `mypy` to cope with them

### Details and comments
Related to #2514


TsafrirA (approved):
LGTM

PR #2573, opened 2026-01-26
[executor-preview]  Split `QuantumProgram.append` to `add_circuit_item` and `add_samplex_item`

yaelbh:


yaelbh:
I returned `append` and added a deprecation warning. Do you think we'd better emit a warning that's not a deprecation warning? Deprecation warnings are not emitted by default.

SamFerracin (approved):
Thanks 😄 I think the deprecation warning is good enough, as it's the standard

PR #2574, opened 2026-01-26
Emit warning if using circuit calibrations

### LINKED ISSUE #2570: Add ISA check for circuits with custom calibrations
**Description**
Calibration support was dropped with the release of Qiskit 2.0. However, when users of qiskit-ibm-runtime install a version of Qiskit < 2.0.0, they can still submit jobs that contain custom calibrations.

Steps to reproduce:
1. Install qiskit-ibm-runtime with Qiskit version < 2.0, for example via `pip install qiskit-ibm-runtime qiskit==1.4 numpy`
2. Initialize a Qiskit Runtime session and connect to a physical hardware backend (e.g., ibm_kyoto).
3. Construct a QuantumCircuit and include a standard delay instruction on a target qubit.
4. Define a custom qiskit.pulse schedule (e.g., a GaussianSquare pulse with a frequency shift) that represents the non-standard Hamiltonian to be injected.
5. Call qc.add_calibration("delay", [qubit_index], pulse_schedule, [duration]) to bind the payload to the ISA gate.
6. Submit the circuit to SamplerV2. 

Add a check in the ISA validation to ensure that:
- either users get a warning of the same type as that raised by QPY ("UserWarning: Support for loading pulse gates has been removed in Qiskit 2.0. If `delay` is in the circuit it will be left as an opaque instruction.");
- or they get an error that says that pulses are no longer supported.

**Acceptance criteria**
Step 6 should raise a warning or an error.

diego-plan9:
### Summary

When using Qiskit < 2, it is possible to append calibrations to a circuit. As the service does not act upon those calibrations even if present, this PR makes a warning to be emitted under those circumstances.

### Details and comments

Fixes #2570

### DIFF
--- a/qiskit_ibm_runtime/utils/utils.py
+++ b/qiskit_ibm_runtime/utils/utils.py
@@ -122,7 +122,7 @@ def is_isa_circuit(circuit: QuantumCircuit, target: Target) -> str:
     if getattr(circuit, "calibrations", None):
         warnings.warn(
             "Submitting circuits with non-empty circuit.calibrations is deprecated and will be "
-            "removed in qiskit-ibm-provider 0.45.",
+            "removed in qiskit-ibm-runtime 0.45.",
             DeprecationWarning,
             stacklevel=2,
         )


qiskit_ibm_runtime/utils/utils.py
```diff
@@ -117,6 +118,15 @@ def is_isa_circuit(circuit: QuantumCircuit, target: Target) -> str:
             f"but the target system requires {target.num_qubits} qubits."
         )
 
+    # Check if there are circuit.calibrations (only available in Qiskit < 2).
+    if getattr(circuit, "calibrations", None):
+        warnings.warn(
```
diego-plan9, line 123: A `warning` felt more natural than raising - this way users are informed, and we avoid a (technically) breaking change.

test/unit/test_sampler.py
```diff
@@ -161,6 +161,12 @@ def test_sampler_validations(self):
             ):
                 inst.run([(circ,)])
 
+            circ = QuantumCircuit(QuantumRegister(2), ClassicalRegister(2))
+            # Mimic `circuit.add_calibrations()` effects, only available in Qiskit < 2.
+            circ.calibrations = {"delay": {((0,), ()): None}}
```
diego-plan9: This is brittle, but I consciously avoided using the actual API as we don't have a good way of testing against older Qiskit versions automatically. I opened https://github.com/Qiskit/qiskit-ibm-runtime/issues/2575 ... but even if acted upon, it would be using `2.0`.

release-notes/unreleased/2574.feat.rst
```diff
@@ -0,0 +1,2 @@
+When a circuit contains calibrations is executed, a warning is emitted to
```
yaelbh:
```suggestion
When a circuit that contains calibrations is executed, a warning is emitted to
```

qiskit_ibm_runtime/utils/utils.py
```diff
@@ -117,6 +118,15 @@ def is_isa_circuit(circuit: QuantumCircuit, target: Target) -> str:
             f"but the target system requires {target.num_qubits} qubits."
         )
 
+    # Check if there are circuit.calibrations (only available in Qiskit < 2).
+    if getattr(circuit, "calibrations", None):
```
yaelbh, line 122: Did you verify that the warning is not emitted when there are no calibrations? What does normally happen in the old Qiskit when there are no calibrations - I can imagine three options, the third one may cause false positives:
- `circuit` does not have a `calibrations` attribute
- `circuit` has a `calibrations` attribute, set to `None`
- `circuit` has a `calibrations` attribute, set to an empty container

diego-plan9: This should cover the three cases, right (with the third one not entering the `if` as it is an empty dict)? In Qiskit 1.x, `circuit.calibrations` is empty on regular circuits:

```
In [34]: z = random_circuit(50,2)
In [35]: z.calibrations
Out[35]: {}
```

PR #2569, opened 2026-01-15
Fix broken links

beckykd:
### Details and comments
Fixes #2553

PR #2589, opened 2026-02-04
Beckykd fix links

beckykd:
Evidently, we can't ignore broken links in release notes files, so they need to be updated to something that works.


arnaucasau:
Note for the maintainers:

We are bringing the change into Qiskit/documentation via https://github.com/Qiskit/documentation/pull/4506. In my opinion we can just update the docs on our end without needing a new patch release. Same case as https://github.com/Qiskit/qiskit-ibm-runtime/pull/2569

P.S. This change needs to be backported into `stable/0.45` :)

PR #2590, opened 2026-02-05
[Executor/NLV3] Pass backend's calibration id to job runs

ihincks:
### Summary

Since we don't inherit from the base primitive class, we need to do this ourselves (NLV2 already does).

Without this fix, calibration ids are never actually passed in the POST request.

### Details and comments
Fixes #


TsafrirA (approved):
LGTM

PR #2593, opened 2026-02-06
host verification disable now works

### LINKED ISSUE #2592: going through a proxy and disabling host verification fails
**Bug description**

Disabling host verification on an https connection fails

*Context:*

Proxies doing packet inspection do their own ssl handshake with the target host and with the user machine. As a consequence, the user machine sees the proxy certificate rather than the target host certificate. For this reason it is convenient to disable the host verification on the user machine.

**how to reproduce**

*prerequisite:* use https connection going through a proxy doing packet inspection

        channel='ibm_quantum_platform',
        token=token,
        proxies=proxies,
        verify=False,  # Set to False if you have certificate issues
        instance='Instance_2026'
    )

**Expected behavior**

Connection happens without error

**environment of execution**

 - qiskit-ibm-runtime 0.45.9
 - Python Version 3.11
 - Name of service used: QiskitRuntimeService
 - Windows 11

**related issue**
https://github.com/IBM/platform-services-python-sdk/issues/316

joseluu:
## PR summary
fixes  https://github.com/Qiskit/qiskit-ibm-runtime/issues/2592
This is a 3 lines change:
* GlobalCatalogV1 object now carries a disable_ssl_verification flag
* GlobalSearchV2 now initializes the BaseService parent with the proper value of disable_ssl_verification

## Current vs new behavior  
disabling host verification now works

## Does this PR introduce a breaking change?    
- [ ] Yes
- [x] No

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -378,7 +378,7 @@ class QiskitRuntimeService:
     ) -> Account:
         """Discover account for ibm_cloud and ibm_quantum_platform channels."""
         account = None
-        verify_ = verify or True
+        verify_ = verify
         if name:
             if filename:
                 if any([channel, token, url]):
--- a/qiskit_ibm_runtime/accounts/account.py
+++ b/qiskit_ibm_runtime/accounts/account.py
@@ -314,9 +314,10 @@ class CloudAccount(Account):
     def list_instances(self) -> list[dict[str, Any]]:
         """Retrieve all crns with the IBM Cloud Global Search API."""
         iam_url = get_iam_api_url(self.url)
-        authenticator = IAMAuthenticator(self.token, url=iam_url)
+        authenticator = IAMAuthenticator(self.token, url=iam_url, disable_ssl_verification=(not self.verify))
         client = GlobalSearchV2(authenticator=authenticator)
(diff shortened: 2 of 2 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -378,7 +378,7 @@ def _discover_account(
     ) -> Account:
         """Discover account for ibm_cloud and ibm_quantum_platform channels."""
         account = None
-        verify_ = verify or True
+        verify_ = verify
```
diego-plan9, line 381: Is this change intentional? While the `verify` flag in the signature of this function (and others) could use a revision for ensuring it is always a bool, I believe this line mostly coerces `None` into an actual boolean and should still be kept for preserving previous behavior.

diego-plan9 (commented):
Thanks @joseluu ! This would require a new version of `platform-services-python-sdk` that contains your fix to be released before working, right?

diego-plan9 (commented):
Thanks - unfortunately, a related PR was started in the meantime, #2607, which provides a more comprehensive fix (extending the ssl verification flag to calls to other services, and includes handling of the proxy arguments too). As it is a superset of this PR, I'm inclined to focus on landing #2607 if it gets into mergeable state before the next release.

diego-plan9:
Thanks again - closing as #2607 was merged.

PR #2596, opened 2026-02-10
Update custom documentation warning

diego-plan9:
### Summary

As the branch is soon to be merged, this PR updates the warning in the documentation, so it gets rendered for the last time. This should allow users depending on the temporary documentation site to find their way to the new place of the documentation (ie. the regular documentation page).

### Details and comments
Fixes N/A

PR #2598, opened 2026-02-10
Cleanup executor branch documentation mechanism

diego-plan9:
### Summary

Finalize the `executor_preview` branch:
* remove the mechanism for rendering to github pages
* remove extra artifacts on README and pyproject.toml due to the temporary nature of the branch
* disable rendering of documentation (mostly tweaking indexes and entries), to better re-introduce it once the branch is merged

### Details and comments
Fixes N/A

Please note this should not be merged before #2596

### DIFF
--- a/docs/apidocs/index.rst
+++ b/docs/apidocs/index.rst
@@ -8,10 +8,8 @@
    :maxdepth: 1
 
    runtime_service
-   quantum_program
    noise_learner
    noise_learner_result
-   noise_learner_v3
    options
    transpiler
    qiskit_ibm_runtime.transpiler.passes.scheduling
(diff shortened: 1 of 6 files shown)


docs/apidocs/index.rst
```diff
@@ -8,10 +8,8 @@
    :maxdepth: 1
 
    runtime_service
-   quantum_program
    noise_learner
    noise_learner_result
-   noise_learner_v3
```
yaelbh, line 14: Since you delete this line, I think it's safe to retain the files `docs/apidocs/noise_learner_v3.rst` and `docs/apidocs/quantum_program.rst`, instead of deleting them.

diego-plan9: I gave it a try, but leaving them would cause sphinx to issue a warning `WARNING: document isn't included in any toctree [toc.not_included]`. For the guides I used the `:orphan:` directly to appease it, but in this case as it is an automodule it causes extra-warnings (`WARNING: Field list ends without a blank line; unexpected unindent. [docutils]`) that I was not able to solve. 

So I think we can just live with the files being deleted (as they are 4-liners), and reintroduce them when the time comes along with the changes to the index.

diego-plan9:
Closing to reissue against main

Issue #2382, opened 2025-08-06
Invalid instance


tralala87:
### Environment

Google colab

### What is happening?

When I run:

'''
from qiskit_ibm_runtime import QiskitRuntimeService
service = QiskitRuntimeService.save_account(
  token="my_ApiKey",
  set_as_default=True,
  overwrite=True,
)
service = QiskitRuntimeService(instance="my_cnr")
backend = service.backend("ibm_torino")
'''

I get:
qiskit_runtime_service._discover_backends_from_instance:WARNING:2025-08-06 09:33:18,930: Invalid instance ...

I have tried to delete and create new instances, but every time I get the same error.

### How can we reproduce the issue?

See section 'What is happening?'

### What should happen?

See section 'What is happening?'

### Any suggestions?

No idea. Followed qiskit documentation multiple times, but no answer.

kt474:
Hi @C-NLTX can you show how you're initializing `QiskitRuntimeService()`? Also is the instance you're trying to access listed on https://quantum.cloud.ibm.com/instances?

C-NLTX:
Hi @kt474 , thanks, sure here is my code:

from qiskit_ibm_runtime import QiskitRuntimeService
filename = 'path/file'
QiskitRuntimeService.save_account(name='account', token='token', channel='ibm_quantum_platform', filename=filename, overwrite=True)
service = QiskitRuntimeService(name='account', filename=filename)
backend = service.backend(service.backends()[0].name, instance=service.instances()[0]['crn']

While instantiating a service, I am relying on the saved account credentials uisng an explicit file path. While instantiating a backend, I am trying to avoid as much as possible using concrete text labels that I cannot guarantee whether they exist for the runtime. For the above code, this is the error that I am receiving now:  <class 'qiskit_ibm_runtime.accounts.exceptions.InvalidAccountError'>: Unable to retrieve instances.

C-NLTX:
PS: I modified a bit my code (below), now the error is "list out of range" while instantiating a backend using the first label on the backends list:

from qiskit_ibm_runtime import QiskitRuntimeService
filename = 'file/path'
instance='instance'
token='token'
name='account'
QiskitRuntimeService.save_account(name=name, token=token, instance=instance, filename=filename, overwrite=True)
service = QiskitRuntimeService(name=name, instance=instance, filename=filename)
backend = service.backend(service.backends()[0].name)

Error received: <class 'IndexError'>: list index out of range

[...]

PR #2607, opened 2026-02-12
Propagate proxies and ssl_verification args to cloud services

alodolo:
### Summary

This pull request should fix the following issues: #2382 and #2592.
It injects ssl_verification and proxy arguments to outgoing requests.
~~It also patches the proxy integration test.~~

### Details and comments

Remarques: 

* Proxy and ssl_verification injection
The requests parameter injection for `GlobalSearchV2` and `GlobalCatalogV1` is done during the search method. The PR #2593 is adding ssl_verification at the service instanciation level, however those services don't have (yet) the constructors to disable `ssl_verification`.

* IAM service
It seems that the class that handle the IAM service is instanciated multiple times in different ways (`IAMAuthenticator` in `CloudAccount`, and  `IAMTokenManager` in `CloudAuth`). I would recommend some refactoring to use only 1 configured IAM service in the code.

* ~~Integration testing - proxy~~
~~It seems that integration tests for the proxy were skipped because it was conditioned only for `ibm_cloud` services.
On top of that the `pproxy` lib have bugs with the latest versions of python and is not maintained anymore (https://github.com/qwj/python-proxy/pull/202). I switched the lib for `mitmproxy` (MIT license, python API, 40k stars and up to date - https://github.com/mitmproxy/mitmproxy).~~

Fixes #
- proxy config
- ssl_verification flag
- ~~integration tests for proxy~~


pyproject.toml
```diff
@@ -192,10 +191,16 @@ documentation = [
 
 visualization = ["plotly>=5.23.0"]
 test = ["ddt>=1.2.0,!=1.4.0,!=1.4.3"]
+integration_test_proxy = ["mitmproxy"]
```
diego-plan9: Instead of a new extra section, could you add the `mitmproxy` dependency to the existing `tests` dependency list?

 It fits nicely into the purpose of such extra (as in, the dependencies that are needed for running the tests: thanks for spotting that it should not belong in `common`!) - and it allows for keepin the workflow (both automated and for developers) more homogeneous (I see the point of integration tests being slightly unique and trying to avoid a dependency / be more granular, but we assume that `pip install -e .[dev]` should provide devs with everything they need).

diego-plan9: Also, do you know if `mitmproxy` abides by semantic versioning? It should be safe (and preferable) to pin the `mitmproxy` (either to `12.x` or to the latest `12.2.1`), for a more predictable environment.

test/integration/test_proxies.py
```diff
@@ -33,40 +33,50 @@ class TestProxies(IBMTestCase):
...
 
     def setUp(self):
-        """Initial test setup."""
         super().setUp()
-        # launch a mock server.
-        command = ["pproxy", "-v", "-l", "http://{}:{}".format(ADDRESS, PORT)]
-        self.proxy_process = subprocess.Popen(command, stdout=subprocess.PIPE)
+        # Command to start mitmproxy in non-interactive mode
+        self.proc = subprocess.Popen(
+            ["mitmdump", "--ssl-insecure", "--listen-port", str(PORT), "--listen-host", ADDRESS],
```
diego-plan9: Have you evaluated launching `mitmproxy` as a regular Python library, instead of a separate process? There seems to be a `mitmproxy.master.Master` (and `DumpMaster`) available, as well as addons, that might provide the building blocks - with a big disclaimer that I only scratched the surface of the docs.

If it is possible (and not too complex / costly), we could potentially have more control and easiness of debugging / extending (everything in Python land, access to the port from the tests, likely also the option to do some asserts over the requests that passed through it).

.github/workflows/ci.yml
```diff
@@ -120,7 +120,7 @@ jobs:
       # avoid cancellation of in-progress jobs if any matrix job fails
       fail-fast: false
       matrix:
-        python-version: [ '3.10' ]
+        python-version: [ '3.11' ]
```
diego-plan9: Can you revert this change? We default to the lowest Python version supported for the CI runners (unless we want to test a specific Python version).

diego-plan9 (commented):
Thanks @alodolo ! As #2593 might require some time, it makes sense to move this PR forward with the goal of closing the issue.

And thanks for bringing the proxy testing situation into a better state as part of this PR - actually, this initial review is focused on `mitmproxy` usage and the infra around it, preparing for next steps.

pyproject.toml
alodolo: Fixing `mitmproxy` with a 12.x version would create an incompatibility with lower versions of python.

test/integration/test_proxies.py
alodolo: I will look into it, right now I rolled back to the old integration tests.

.github/workflows/ci.yml
alodolo: Rolled back changes, `mitmproxy. mitmdump` has some issues with python 3.10, I will take a deeper look if there is an another mitmproxy API that would be compatible with mitmproxy.

pyproject.toml
alodolo: Reverted the changes on mitmproxy

.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
```
diego-plan9: Can you revert this change? We indeed should support `3.14`, but requires a few more changes - and we better tackle it as a separate PR.

CONTRIBUTING.md
```diff
@@ -334,4 +334,4 @@ the `main` branch, such as from `stable/0.21` to `main`.
 [black]: https://github.com/psf/black
 [pylint]: https://www.pylint.org/
 [ruff]: https://github.com/astral-sh/ruff
-[mypy]: http://mypy-lang.org/
+[mypy]: http://mypy-lang.org/
```
diego-plan9: Small nit - can you revert? This seems to be removing the endline at the end of file.

pyproject.toml
```diff
@@ -199,4 +199,4 @@ dev = [
     "qiskit-ibm-runtime[common]",
     "qiskit-ibm-runtime[documentation]",
     "qiskit-ibm-runtime[visualization]",
-]
+]
```
diego-plan9: Similar to the change in `CONTRIBUTING.md`- can you revert? This seems to be removing the endline at the end of file.

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -314,14 +331,14 @@ def resolve_crn(self) -> None:
 
     def list_instances(self) -> list[dict[str, Any]]:
```
diego-plan9, line 332: Can you cover this changes with an additional test in `test/integration/test_proxies.py`? Similar to `test_proxies_cloud_runtime_client`, but instantiating a `QiskitRuntimeService` (so ultimately it triggers this changes, and we have confidence that instantiating the main class through a proxy works).

Suggestion (not meant to be working code, and actually inspired on your previous commits):

```python
    @integration_test_setup(supported_channel=["ibm_quantum_platform"], init_service=False)
    def test_proxies_qiskit_runtime_service(self, dependencies: IntegrationTestDependencies) -> None:
        """Should reach the proxy using QiskitRuntimeService."""
            instance=dependencies.instance,
            token=dependencies.token,
            channel=dependencies.channel,
            verify=False,
            proxies={ "urls": VALID_PROXIES }
        )
        service.jobs(limit=1)

        # Get lines and terminate proxy similar to the other test
        # And very similar assert (maybe some extra ones for the catalog URLs or the amount of hits, etc)
```

If you take the chance to reintroduce the `supported_channel=["ibm_quantum_platform"]` to the existing test, would also be much appreciated - this way we can get both the tests running as part of the integration tests, instead of skipped.

In local tests, I had to use an ugly `sleep(2)` as part of the `pproxy` initialization, to give it enough time to be ready before the requests hit.

release-notes/unreleased/2592.bug.rst
```diff
@@ -0,0 +1,3 @@
+The method `.accounts.CloudAccount.resolve_crn ` did not inject proxy and ssl_verification
+arguments to outgoing request. Added the arguments to `GlobalSearchV2`, `GlobalCatalogV1` 
+and `IAMAuthenticator` services.
```
diego-plan9:
```suggestion
The ``proxies`` and ``ssl_verification`` arguments for ``QiskitRuntimeService``
are nowpropagated to the underlying HTTP requests to ```GlobalSearchV2``,
``GlobalCatalogV1`` and ``IAMAuthenticator`` services, allowing to instantiate
``QiskitRuntimeService`` correctly when using a proxy.

```

Mostly for making it a bit more descriptive for regular users.

diego-plan9 (commented):
Thanks @alodolo !

> I reverted all the changes link to the switch to `mitmproxy` and the proxies integration test and I will create an another dedicated issue & PR. It seems that `pproxy` has compatibility issues with the latest version of python and `mitmproxy` has issues with `python 3.10`.

Thanks for looking into it - perhaps a less recent `mitmproxy` version might fit the bill when it comes to supporting the right Python versions, as it would be great to replace ``pproxy`` eventually. Indeed, the goal is getting the neat fix in place: looking forward for the follow-up issue and PR, with lower priority.

Other than some nits, it is just one test away from being ready! As another small nit: can you rename the PR to something a bit more human-readable, so it can be easier to parse when going through history?

test/integration/test_proxies.py
```diff
@@ -38,6 +41,17 @@ def setUp(self):
         # launch a mock server.
         command = ["pproxy", "-v", "-l", "http://{}:{}".format(ADDRESS, PORT)]
         self.proxy_process = subprocess.Popen(command, stdout=subprocess.PIPE)
+        """Time for the proxy to start"""
```
diego-plan9: Can you turn this into a comment?

test/integration/test_proxies.py
```diff
@@ -38,6 +41,17 @@ def setUp(self):
...
+        sleep(2)
+        """Block all traffic not routed to the proxy"""
```
diego-plan9: Nice addition for extra confidence! Can you turn this into a comment?

test/integration/test_proxies.py
```diff
@@ -50,19 +64,39 @@ def tearDown(self):
...
-        params.proxies = ProxyConfiguration(urls=VALID_PROXIES)
-        client = RuntimeClient(params)
-        client.jobs_get(limit=1)
-        api_line = pproxy_desired_access_log_line(params.url)
+    @integration_test_setup(supported_channel=["ibm_quantum_platform"], init_service=False)
+    def test_proxies_qiskit_runtime_service(
+        self, dependencies: IntegrationTestDependencies
+    ) -> None:
+        """Should reach the proxy using QiskitRuntimeService."""
+        # Allow some time for `pproxy` to receive requests.
```
diego-plan9: This one seems that it can be removed now?

pyproject.toml
```diff
@@ -140,7 +140,7 @@ dependencies = [
     "numpy>=1.26",
     "urllib3>=2.4.0",
     "python-dateutil>=2.9.0",
-    "ibm-platform-services>=0.55.3",
+    "ibm-platform-services>=0.61.1",
```
diego-plan9, line 143: Thanks - for future reference, this is due to `GlobalCatalogV1.get_catalog_entry()` and other methods not propagating the kwargs until https://github.com/IBM/platform-services-python-sdk/pull/286/changes#diff-8c42907892d2dd2139f88503f2d0049bcec384c7def602131f1a46ba3e750995L292-R316 landed.

diego-plan9 (approved):
Thanks @alodolo!

PR #2608, opened 2026-02-12
Add post-processing stage for the new SamplerV2

### LINKED ISSUE #2585: Write a result decoder implementation for the new SamplerV2
### Relates Issues
#2582 

### Details
The new `SamplerV2` implementation will run the `Executor`, and so the job results will be of type `QuantumProgramResult` the decoder should convert this output to the expected `SamplerResult`. The decoder needs to be used when `.run()` method is used.

TsafrirA:
### Summary
Closes #2585

### Test evidence
Testing on `ibm_fez` with
```python
circ = QuantumCircuit(2)
circ.h(0)
circ.cx(0,1)
circ.rx(Parameter("p1") + Parameter("p2"), 0)
circ.measure_all()
```
and params:
```python
params = list(product(range(0, 3), repeat=2))
```
as a single pub yields

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,6 +110,7 @@ jobs:
           flag-name: unit-tests_python${{ matrix.python-version }}-${{ matrix.os }}
           parallel: true
           file: coverage.lcov
+          fail-on-error: false
   integration-tests:
     if: github.event_name == 'push' && github.repository_owner == 'Qiskit'
     # only kick-off resource intensive integration tests if unit tests and all basic checks succeeded
@@ -157,3 +158,4 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           parallel-finished: true
+          fail-on-error: false
--- a/qiskit_ibm_runtime/options/executor_options.py
+++ b/qiskit_ibm_runtime/options/executor_options.py
@@ -86,3 +86,6 @@ class ExecutorOptions:
 
     execution: ExecutionOptions = Field(default_factory=ExecutionOptions)
     """Low-level execution options."""
+
+    experimental: dict = Field(default_factory=dict)
+    """Experimental options that are passed to the executor."""
--- a/qiskit_ibm_runtime/quantum_program/converters/converters_0_2.py
+++ b/qiskit_ibm_runtime/quantum_program/converters/converters_0_2.py
@@ -75,6 +75,13 @@ def quantum_program_to_0_2(program: QuantumProgram, options: ExecutorOptions) ->
             raise ValueError(f"Item {item} is not valid.")
         model_items.append(model_item)
 
+    # Build options dict starting with execution options
+    options_dict = asdict(options.execution)  # type: ignore[call-overload]
+
+    # Add experimental options if provided
+    if options.experimental:
+        options_dict["experimental"] = options.experimental
+
     return ParamsModel(
         quantum_program=QuantumProgramModel(
             shots=program.shots,
@@ -82,7 +89,7 @@ def quantum_program_to_0_2(program: QuantumProgram, options: ExecutorOptions) ->
             meas_level=program.meas_level,
             passthrough_data=program.passthrough_data,
         ),
-        options=asdict(options.execution),  # type: ignore[call-overload]
+        options=options_dict,
     )
 
 
--- a/test/unit/executor/test_executor.py
+++ b/test/unit/executor/test_executor.py
@@ -99,6 +99,29 @@ class TestExecutorOptions(IBMTestCase):
         self.assertEqual(executor.options.environment.log_level, "WARNING")
         self.assertFalse(executor.options.execution.init_qubits)
 
+    def test_experimental_options_default_empty(self):
+        """Test that experimental options default to empty dict."""
+        executor = Executor(mode=get_mocked_backend())
+        self.assertEqual(executor.options.experimental, {})
+
+    def test_experimental_options_from_dict(self):
+        """Test constructing with experimental options in dict."""
+        opts_dict = {"experimental": {"foo": "bar", "baz": 123}}
+        executor = Executor(mode=get_mocked_backend(), options=opts_dict)
+        self.assertEqual(executor.options.experimental, {"foo": "bar", "baz": 123})
+
+    def test_experimental_options_from_instance(self):
+        """Test constructing with an ExecutorOptions instance with experimental options."""
+        opts = ExecutorOptions(experimental={"custom_key": "custom_value"})
+        executor = Executor(mode=get_mocked_backend(), options=opts)
+        self.assertEqual(executor.options.experimental, {"custom_key": "custom_value"})
+
+    def test_experimental_options_setter(self):
+        """Test setting experimental options via the setter."""
+        executor = Executor(mode=get_mocked_backend())
+        executor.options = {"experimental": {"test": "value"}}
+        self.assertEqual(executor.options.experimental, {"test": "value"})
+
 
 class TestExecutor(IBMTestCase):
     """Tests the ``Executor`` class."""
--- a/test/unit/quantum_program/converters/test_converters_0_2.py
+++ b/test/unit/quantum_program/converters/test_converters_0_2.py
@@ -88,13 +88,17 @@ class TestQuantumProgramConverters(IBMTestCase):
             chunk_size=7,
         )
 
-        options = ExecutorOptions(execution=ExecutionOptions(init_qubits=False))
+        experimental_opts = {"custom_option": "test_value", "another_key": 123}
+        options = ExecutorOptions(
+            execution=ExecutionOptions(init_qubits=False), experimental=experimental_opts
+        )
 
         params_model = quantum_program_to_0_2(quantum_program, options)
 
         self.assertEqual(params_model.schema_version, "v0.2")
         self.assertEqual(params_model.options.init_qubits, False)
         self.assertEqual(params_model.options.rep_delay, None)
+        self.assertEqual(params_model.options.experimental, experimental_opts)
 
         quantum_program_model = params_model.quantum_program
         self.assertEqual(quantum_program_model.shots, shots)


qiskit_ibm_runtime/executor/routines/sampler_v2/__init__.py
```diff
@@ -13,3 +13,4 @@
 """Executor-based SamplerV2"""
 
 from .sampler import SamplerV2
+from .sampler_post_processors import SAMPLER_POST_PROCESSORS
```
SamFerracin: I think we can keep this in its own file, since it's something that user won't need to interact with
```suggestion
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler_post_processors.py
```diff
@@ -0,0 +1,93 @@
...
+
+from collections.abc import Callable
+
+from qiskit.primitives import PrimitiveResult
+from qiskit.primitives.containers import BitArray, DataBin, SamplerPubResult
+
+from ....quantum_program.quantum_program_result import QuantumProgramResult
+
+# Type alias for sampler post-processor functions
+PostProcessorFunc = Callable[[QuantumProgramResult], PrimitiveResult]
```
SamFerracin, line 25:
```suggestion
PostProcessorFunc = Callable[[QuantumProgramResult], QuantumProgramResult]
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler_post_processors.py
```diff
@@ -0,0 +1,93 @@
...
+    def decorator(func: PostProcessorFunc) -> PostProcessorFunc:
+        SAMPLER_POST_PROCESSORS[name] = func
+        return func
+
+    return decorator
+
+
+@register_post_processor("v1")
+def sampler_v2_post_processor_v1(qp_result: QuantumProgramResult) -> PrimitiveResult:
+    """Convert QuantumProgramResult to SamplerV2 PrimitiveResult.
```
SamFerracin, line 50: Could you apply similar changes to every docstring?

```suggestion
    """Convert :class:`~.QuantumProgramResult` to :class:`~.SamplerV2` PrimitiveResult.
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler_post_processors.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def decorator(func: PostProcessorFunc) -> PostProcessorFunc:
+        SAMPLER_POST_PROCESSORS[name] = func
+        return func
+
+    return decorator
+
+
+@register_post_processor("v1")
+def sampler_v2_post_processor_v1(qp_result: QuantumProgramResult) -> PrimitiveResult:
```
SamFerracin: NIT:
```suggestion
def sampler_v2_post_processor_v1(result: QuantumProgramResult) -> PrimitiveResult:
```
since the "qp" part is implicit in the type-hint

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -137,8 +150,63 @@ def result(  # pylint: disable=arguments-differ
...
+
+        # Priority 2: Post-processor from passthrough_data
+        if not post_processor_fn and result.passthrough_data:
+            post_processor_info = result.passthrough_data.get("post_processor")
+            if post_processor_info:
+                context = post_processor_info.get("context")
+                if context == "sampler_v2" and (version := post_processor_info.get("version")):
+                    from .executor.routines import SAMPLER_POST_PROCESSORS
+
+                    post_processor_fn = SAMPLER_POST_PROCESSORS.get(version)
```
SamFerracin: We should raise more aggressively here. In the way this is currently set up, if for some reason a post-processor is not found, then `job.result()` would return an executor result, which breaks the contract set by the SamplerV2 abstraction.

```suggestion
        if result.passthrough_data.get("context") == "sampler_v2":
            try:
                version = result.passthrough_data["post_processor"]["version"]
            except KeyError:
                raise ValueError("Could not determine a post-processor version.")
                
            from .executor.routines import SAMPLER_POST_PROCESSORS
            try:
                post_processor_fn = SAMPLER_POST_PROCESSORS[version]
            except KeyError:
                raise ValueError(f"No post-processor found for version {version}.")
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler_post_processors.py
TsafrirA: I will fix it collectively after we have the full feature branch.

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -181,3 +192,44 @@ def options(self) -> None:
         # Return a minimal options object for compatibility
         # In future phases, this will return a proper SamplerOptions instance
         return None
+
+    @staticmethod
+    def quantum_program_result_to_primitive_result(result: QuantumProgramResult) -> PrimitiveResult:
```
SamFerracin, line 195: I see this more as a util function than a static method of SamplerV2. What do you think?

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler_post_processors.py
SamFerracin: okay, but let's not forget

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
TsafrirA: I have a slight preference to have it directly tied to the `SamplerV2`, but I don't really mind changing it.

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -179,3 +190,44 @@ def options(self) -> None:
...
+            PrimitiveResult containing SamplerPubResult objects.
+
+        Raises:
+            ValueError: If data is malformed or inconsistent
+        """
+        # Build SamplerPubResult for each pub
+        pub_results = []
+        for idx, item_data in enumerate(result):
+            # Validate that measurement data exists
+            if not item_data:
```
SamFerracin, line 211: Do we need this check or not? I can't tell

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -179,3 +190,44 @@ def options(self) -> None:
...
+            # Infer pub_shape from the first classical register's data
+            # meas_data shape: (...pub_shape..., num_shots, num_bits)
+            first_meas_data = next(iter(item_data.values()))
+            pub_shape = first_meas_data.shape[:-2]
+
+            # Create BitArray for each classical register found in the data
+            bit_arrays = {}
+            for creg_name, meas_data in item_data.items():
+                # Create BitArray from measurement data (bit array format)
+                # meas_data shape: (..., num_shots, num_clbits)
```
SamFerracin: You underestimate your skills as a writer: This code is clean and readable, inline comments are redundant
```suggestion
            bit_arrays = {}
            for creg_name, meas_data in item_data.items():
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -179,3 +190,44 @@ def options(self) -> None:
...
+                bit_array = BitArray.from_bool_array(meas_data)
+                bit_arrays[creg_name] = bit_array
+
+            data_bin = DataBin(**bit_arrays, shape=pub_shape)
+
+            pub_result = SamplerPubResult(data=data_bin, metadata={})
+            pub_results.append(pub_result)
+
+        # Create and return PrimitiveResult with preserved metadata
+        return PrimitiveResult(pub_results, metadata={"quantum_program_metadata": result.metadata})
```
```suggestion
        return PrimitiveResult(pub_results, metadata={"quantum_program_metadata": result.metadata})
```

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -109,15 +112,25 @@ def result(  # pylint: disable=arguments-differ
...
         decoder: type[ResultDecoder] | None = None,
+        post_processor: Callable[[QuantumProgramResult], Any] | None = None,
     ) -> Any:
         """Return the results of the job.
 
+        The result will be post-processed if it is a QuantumProgramResult and
+        a post-processor is configured, according to the following precedence:
+        1. post_processor parameter (if provided)
+        2. Post-processor specified in result's passthrough_data
+        3. No post-processing (return raw decoded result)
```
SamFerracin: Reworded it a bit:
```suggestion
        If the decoded job results are of type QuantumProgramResult, and if a post processor is specified (explicitly by the users, or via the results' passthrough data),
        the results are post-processed before being returned. If
        provided, the user-specified `post_processor` takes precedence over the post processor specified in the 
        passthrough data.
```

SamFerracin: we'll need to describe what it means for a post-processor to be specified in the passthrough data, but let's leave this for a later doc PR

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
TsafrirA: The iterator call that follows would be at risk without this.
Eventually we want to allow empty pubs, so this will have to go. I have it on my todo list, and will take care of it collectively.

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -109,15 +112,24 @@ def result(  # pylint: disable=arguments-differ
...
+        (explicitly by the users, or via the results' passthrough data), the results are post-processed
+        before being returned. If provided, the user-specified `post_processor` takes precedence over the
+        post processor specified in the passthrough data.
+
         Args:
             timeout: Number of seconds to wait for job.
             decoder: A :class:`ResultDecoder` subclass used to decode job results.
+            post_processor: Post-processor callable to apply to results.
+                Only applies to QuantumProgramResult. If provided, takes precedence
+                over post_processor specified in passthrough_data.
```
SamFerracin: i think that this will render as italic otherwise
```suggestion
                over ``post_processor`` specified in ``passthrough_data``.
```

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -137,8 +149,75 @@ def result(  # pylint: disable=arguments-differ
...
+        result: Any,
+        post_processor_override: Callable[[QuantumProgramResult], Any] | None = None,
+    ) -> Any:
+        """Apply post-processing to the decoded result.
+
+        Post-processing is only applied if the result is a QuantumProgramResult
+        (from Executor jobs). Other result types are returned unchanged.
+
+        Post-processing precedence:
+        1. post_processor_override parameter (if provided)
```
SamFerracin:
```suggestion
        1. ``post_processor_override`` parameter (if provided)
```

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -137,8 +149,75 @@ def result(  # pylint: disable=arguments-differ
...
+        2. Post-processor from result's passthrough_data (if available)
```
SamFerracin:
```suggestion
        2. Post-processor from result's ``passthrough_data`` (if available)
```

PR #2614, opened 2026-02-13
Avoid implicit optionals

diego-plan9:
### Summary

Enable `RUF013`(https://docs.astral.sh/ruff/rules/implicit-optional/), to ensure explicit marking of arguments that can be `None` - applying the automated fixes (plus 5 instances manually fixed in the last commit). This provides better insight into typing, and avoids warnings with other tools (such as pylance).

### Details and comments
Related to #2514

Ideally, this should be combined with removing `strict_optional = false` from the `mypy` configuration, but this opens up a large number of items that need to be fixed manually - so this PR is for the automated tweaks and enforcing the rule.

### DIFF
--- a/qiskit_ibm_runtime/executor.py
+++ b/qiskit_ibm_runtime/executor.py
@@ -36,13 +36,55 @@ logger = logging.getLogger()
 
 
 class Executor:
-    """Executor for :class:`~.QuantumProgram`\\s."""
+    """Class for running :class:`~.QuantumProgram`\\s.
+
+    The :meth:`run` method can be used to submit a quantum program to be executed on a backend.
+
+    .. code-block:: python
+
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -83,12 +83,13 @@ target-version = "py310"
...
-    "G",     # flake8-logging-format
-    "T10",   # flake8-debugger
-    "UP",    # pyupgrade
+    "C4",     # flake8-comprehensions
+    "E",      # pycodestyle
+    "F",      # pyflake
+    "G",      # flake8-logging-format
+    "T10",    # flake8-debugger
+    "UP",     # pyupgrade
+    "RUF013", # implicit-optional
```
SamFerracin, line 92: I guess this will catch it now?

diego-plan9: Yes, that's the plan!

Issue #2582, opened 2026-02-04
[Wrapper sampler MVP1] Write an implementation of SamplerV2 around Executor


TsafrirA:
### Background
The new `Executor` primitive provides more powerful and efficient tools to interact with a quantum computer. However, the `Sampler` provides a convenient abstraction for users to use. To have the best of both worlds, we want to create a new `SamplerV2` implementation which will utilize the `Executor` under the hood.

### Workplan
The new implementation will initially live side by side with the existing implementation at `qiskit_ibm_runtime.executor.routines.sampler_v2`.

#### Phase 1: Minimal viable implementation
In the first phase, the new sampler will only be able to handle basic PUBs, with no advanced options (Twirling, etc.). It will run the `Executor` (with `CircuitItem`) and results will be converted from `QuantumProgramResult` to `SamplerV2Result`.

- [x] #2583
- [x] #2585

#### Phase 2: Options Support & Feature Parity
In the second phase, we will gradually match the options provided by the existing `SamplerV2`. Twirling options will trigger a `Samplomatic` boxing and building step, and result in a `SamplexItem` to be run via the `Executor`.

- [x] #2586
- [x] #2587
- [x] #2588

#### Future
While outside the scope of this epic, it should be noted that the new implementation opens up avenues for further improvements like accepting pre-boxed circuits.

We expect to eventually deprecate and remove the old implementation of `SamplerV2`.

jyu00:
Thanks @TsafrirA! A couple of questions

1. #2586 only mentioned execution and experimental. Presumably environment options (like `job_tags`) will go there as well? 
2. Presumably the new SamplerV2 will continue to support [local testing mode](https://quantum.cloud.ibm.com/docs/en/guides/local-testing-mode), given it uses BackendSampler right now which still exists?

diego-plan9:
Marking as completed, with the follow-up work captured by #2681

PR #2617, opened 2026-02-16
Add minimal options for the executor-based `SamplerV2`

### LINKED ISSUE #2586: Support general options in the new SamplerV2 implementation (execution, experimental)
### Related Issues
#2582 

### Details
The new `SamplerV2` should accept an options argument, and properly handle `execution` and `experimental` options by passing them on to the `Executor`.

TsafrirA:
### Related Issues
Closes #2586 
#2582

### Summary
This PR adds minimal options support for the new `SamplerV2`. Basic options are mapped to the `Executor`, while others raise a `NotImplementedError`.

### DIFF
--- a/qiskit_ibm_runtime/executor/routines/options/dynamical_decoupling_options.py
+++ b/qiskit_ibm_runtime/executor/routines/options/dynamical_decoupling_options.py
@@ -14,8 +14,9 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass
 from typing import Literal
+
+from pydantic.dataclasses import dataclass
 
 
 @dataclass
--- a/qiskit_ibm_runtime/executor/routines/options/environment_options.py
+++ b/qiskit_ibm_runtime/executor/routines/options/environment_options.py
@@ -14,8 +14,11 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field, asdict
 from typing import Literal
+
+from pydantic import Field
+from pydantic.dataclasses import dataclass
+
 from ....options.executor_options import EnvironmentOptions as ExecutorEnvironemntOptions
 
 
@@ -37,9 +40,13 @@ class EnvironmentOptions:
     """
 
     log_level: Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] = "WARNING"
-    job_tags: list[str] = field(default_factory=list)
+    job_tags: list[str] = Field(default_factory=list)
     private: bool = False
 
     def to_executor_environment_options(self) -> ExecutorEnvironemntOptions:
         """Converts the environment options to executor environment options."""
-        return ExecutorEnvironemntOptions(**asdict(self))
+        return ExecutorEnvironemntOptions(
+            log_level=self.log_level,
+            job_tags=self.job_tags,
+            private=self.private,
+        )
--- a/qiskit_ibm_runtime/executor/routines/options/sampler_options.py
+++ b/qiskit_ibm_runtime/executor/routines/options/sampler_options.py
@@ -14,8 +14,11 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field, asdict
 from typing import Literal
+
+from dataclasses import asdict
+from pydantic import Field
+from pydantic.dataclasses import dataclass
 
 from .dynamical_decoupling_options import DynamicalDecouplingOptions
 from .twirling_options import TwirlingOptions
@@ -75,15 +78,15 @@ class SamplerOptions:
     """
 
     default_shots: int | None = 4096
-    dynamical_decoupling: DynamicalDecouplingOptions = field(
+    dynamical_decoupling: DynamicalDecouplingOptions = Field(
         default_factory=DynamicalDecouplingOptions
     )
-    execution: SamplerExecutionOptions = field(default_factory=SamplerExecutionOptions)
-    twirling: TwirlingOptions = field(default_factory=TwirlingOptions)
+    execution: SamplerExecutionOptions = Field(default_factory=SamplerExecutionOptions)
+    twirling: TwirlingOptions = Field(default_factory=TwirlingOptions)
     experimental: dict | None = None
 
     max_execution_time: int | None = None
-    environment: EnvironmentOptions = field(default_factory=EnvironmentOptions)
+    environment: EnvironmentOptions = Field(default_factory=EnvironmentOptions)
 
     def to_executor_options(self) -> ExecutorOptions:
         """Map SamplerOptions to ExecutorOptions.
--- a/qiskit_ibm_runtime/executor/routines/options/twirling_options.py
+++ b/qiskit_ibm_runtime/executor/routines/options/twirling_options.py
@@ -14,8 +14,9 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass
 from typing import Literal
+
+from pydantic.dataclasses import dataclass
 
 
 @dataclass


qiskit_ibm_runtime/executor/routines/options/sampler_options.py
```diff
@@ -0,0 +1,63 @@
...
+
+from dataclasses import dataclass, field
+
+from ....options.dynamical_decoupling_options import DynamicalDecouplingOptions
+from ....options.twirling_options import TwirlingOptions
+from ....options.sampler_execution_options import SamplerExecutionOptionsV2
+from ....options.environment_options import EnvironmentOptions
+from ....options.simulator_options import SimulatorOptions
+
+
```
TsafrirA, line 59: I am not sure how much we should duplicate here.

On one extreme - we can choose to inherit the old `SamplerOptions` instead of duplicating it. This still gives extensibility, while removing the duplicate code.

On the other extreme, if we want to truly decouple the two, should we duplicate all sub-classes as well?

qiskit_ibm_runtime/executor/routines/sampler_v2/options.py
```diff
@@ -0,0 +1,63 @@
...
+from ....options.simulator_options import SimulatorOptions
+
+
+@dataclass
+class SamplerOptions:
+    """Options for the executor-based SamplerV2.
+
+    This class provides the same interface as the old SamplerOptions but is designed
+    to work with the executor-based implementation. It reuses the existing option
+    sub-classes from qiskit_ibm_runtime.options.
```
SamFerracin: I would avoid all mentions to "old" sampler VS "new" sampler, because when we'll remove the "old" sampler in the future, we'll need to remember to go through all the docstrings and it'll take time. I would suggest to just be minimal right now, and before releasing this, we can open a PR to add documentation, and we can ask help to the docs team
```suggestion
```

qiskit_ibm_runtime/executor/routines/sampler_v2/options.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Args:
+        default_shots: The default number of shots to use if none are specified in the
+            PUBs or in the run method. Default: 4096.
```
SamFerracin: Defaults are in the type-hint
```suggestion
            PUBs or in the run method.
```

qiskit_ibm_runtime/executor/routines/sampler_v2/options.py
```diff
@@ -0,0 +1,63 @@
...
+        execution: Execution time options. See :class:`SamplerExecutionOptionsV2`
+            for all available options.
+        twirling: Pauli twirling options. See :class:`TwirlingOptions` for all
+            available options.
+        experimental: Experimental options as a dictionary.
+        max_execution_time: Maximum execution time in seconds, based on system
+            execution time (not wall clock time). Inherited from OptionsV2.
+        environment: Options related to the execution environment. See
+            :class:`EnvironmentOptions` for all available options. Inherited from OptionsV2.
+        simulator: Simulator options. See :class:`SimulatorOptions` for all
```
SamFerracin: Consiider removing simulator options from the PR, and we can add them (+ the required fuctionality) in a future PR

qiskit_ibm_runtime/executor/routines/options/sampler_options.py
SamFerracin: It's not immediately clear to me what's the best thing to do here. However, I tend to prefer writing brand new options, which is what we did for the executor options (look at that file: so clean!). The reason is that the "old" options are more complex objects: They inherit from OptionsV2 and have a lot of methods (`update`, `_get_program_inputs`, ...) that we won't need in the future. Look at how clean the 

My preference:
- Creating a `qiskit_ibm_runtime/executor/routines/options/` subfolder where we can place options files, and reuse them across primitives (for example: one `twirling_option.py` file used by Sampler and Estimator).

Wdyt?

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -91,13 +101,25 @@ class SamplerV2(BaseSamplerV2):
...
+    - ``environment.*``: Environment options (log_level, job_tags, private)
+    - ``max_execution_time``: Maximum execution time
+    - ``experimental.image``: Runtime image
+
+    **Unsupported Options (will raise NotImplementedError):**
+
+    - ``dynamical_decoupling``: Dynamical decoupling sequences
+    - ``twirling``: Pauli twirling
+    - ``simulator.*``: Simulator options
+    - ``experimental.*``: Other experimental options
```
SamFerracin, line 143: I guess the plan is to remove this text when everything is supported?

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -182,16 +228,75 @@ def run(self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None) -> Ru
...
+        if self._options.dynamical_decoupling.enable:
+            raise NotImplementedError(
+                "Dynamical decoupling is not yet supported in the executor-based SamplerV2."
+            )
+
+        # Twirling - not supported yet
+        if self._options.twirling.enable_gates or self._options.twirling.enable_measure:
+            raise NotImplementedError(
+                "Twirling is not yet supported in the executor-based SamplerV2."
+            )
```
SamFerracin: I can imagine that `prepare` will take care of these options, and `_map_options_to_executor` will just ignore them, focusing on the executor options. Therefore, i would remove these errors from here and more them to `prepare`
```suggestion
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -130,47 +154,69 @@ class SamplerV2(BaseSamplerV2):
...
-        default_shots = (
-            shots if shots is not None else 4096
-        )  # TODO: Move to options once available.
-        quantum_program = prepare(coerced_pubs, default_shots=default_shots)
+        meas_level = (
+            self._options.execution.meas_type
+            if self._options.execution.meas_type is not Unset
+            else None
+        )
+        quantum_program = prepare(coerced_pubs, default_shots=default_shots, meas_level=meas_level)
```
SamFerracin: This is what we want, right (see proposal doc)? 
```suggestion
        quantum_program = prepare(coerced_pubs, default_shots, options)
```
Specifically, we want prepare to be a function from (pubs, sampler options) to (quantum program, executor options), and we want run to be:
``` python
def run(...):
   coerced_pubs = [SamplerPub.coerce(pub, shots) for pub in pubs]
   quantum_program, executor_options = prepare(coerced_pubs, options, default_shots)
   ... run the executor
```

In which case:
- Move `self._map_options_to_executor()` inside prepare, making it a public (no need for it to be private) function `sampler_to_executor_options` that we can test in isolation.
- `meas_level` is also handled inside `prepare`

qiskit_ibm_runtime/executor/routines/options/sampler_options.py
TsafrirA: Sounds good!

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
TsafrirA: I think so. But it might be redundant even now.

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
TsafrirA: You're right, this is better.
I just had to shift the options mapper function to the `SamplerOptions` class, to avoid passing the `SamplerV2` object itself to the `prepare` function.

qiskit_ibm_runtime/executor/routines/options/sampler_options.py
TsafrirA: @SamFerracin 
So, I duplicated the classes, but I tried to reduce some duplication by inheriting from the `Executor` classes. That works for the `ExecutionOptions`. For the `EnvironmentOptions`, it's a bit trickier, because some fields are defined in different locations compared to the legacy Sampler. So, I didn't inherit there (didn't want to run into situations where you can define the same field in multiple places).

qiskit_ibm_runtime/executor/routines/options/dynamical_decoupling_options.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Options for dynamical decoupling."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
```
TsafrirA: Should these be pydantic data classes?

ihincks: I would say "yes". You get some easy validation tools that save time for users who might put typos in string literals, etc etc. You also get the ability to do updates from nested dicts (see https://github.com/Qiskit/qiskit-ibm-runtime/pull/2616) which some users like.

qiskit_ibm_runtime/executor/routines/options/sampler_options.py
TsafrirA: (not sure about the pattern here. On the one hand it's a bit more future proof, but on the other hand it could end up breaking the SamplerV2 interface compared to the old implementation)

qiskit_ibm_runtime/executor/routines/options/dynamical_decoupling_options.py
```diff
@@ -0,0 +1,53 @@
...
+
+from pydantic.dataclasses import dataclass
+
+
+@dataclass
+class DynamicalDecouplingOptions:
+    """Options for dynamical decoupling (DD).
+
+    Args:
+        enable: Whether to enable DD as specified by the other options in this class.
```
SamFerracin, line 27: Could I ask you to move these args down where you define the type? As we do, for example, in `ExecutionOptions` and in the majority of datalasses we have. So this would become:

``` python
@dataclass
class DynamicalDecouplingOptions:
    """Options for dynamical decoupling (DD)."""

    enable: bool = False
    """Whether to enable DD as specified by the other options in this class.""""

    ...
```

qiskit_ibm_runtime/executor/routines/options/environment_options.py
```diff
@@ -0,0 +1,52 @@
...
+"""Options related to the execution environment."""
+
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import Field
+from pydantic.dataclasses import dataclass
+
+from ....options.executor_options import EnvironmentOptions as ExecutorEnvironemntOptions
```
SamFerracin, line 22:
```suggestion
from ....options.executor_options import EnvironmentOptions as ExecutorEnvironmentOptions
```

qiskit_ibm_runtime/executor/routines/options/environment_options.py
```diff
@@ -0,0 +1,52 @@
...
+            After the results are read, these are also deleted from the service.
+            When set to ``False``, the input parameters and results follow the
+            standard retention behavior of the API.
+    """
+
+    log_level: Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] = "WARNING"
+    job_tags: list[str] = Field(default_factory=list)
+    private: bool = False
+
+    def to_executor_environment_options(self) -> ExecutorEnvironemntOptions:
```
SamFerracin, line 46: Nit, non blocking: Consider if it's better in this form (more concise and convenient, with the "environment" part in the type-hint):
```suggestion
    def to_executor_options(self) -> ExecutorEnvironemntOptions:
```

If yes, the same would apply to SamplerExecutionOptions

qiskit_ibm_runtime/executor/routines/options/sampler_options.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    execution: SamplerExecutionOptions = Field(default_factory=SamplerExecutionOptions)
+    twirling: TwirlingOptions = Field(default_factory=TwirlingOptions)
+    experimental: dict | None = None
+
+    max_execution_time: int | None = None
+    environment: EnvironmentOptions = Field(default_factory=EnvironmentOptions)
+
+    def to_executor_options(self) -> ExecutorOptions:
+        """Map SamplerOptions to ExecutorOptions.
```
SamFerracin, line 92:
```suggestion
        """Map SamplerOptions to ExecutorOptions, ignoring all the irrelevant fields.
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -81,7 +106,18 @@ def prepare(pubs: list[SamplerPub], default_shots: int | None = None) -> Quantum
...
+    # Create QuantumProgram
+    quantum_program = QuantumProgram(
+        shots=shots,
+        items=items,
+        passthrough_data=passthrough_data,
+        meas_level=options.execution.meas_type,
+    )
+
+    # Map options to executor options
+    executor_options = options.to_executor_options()
```
SamFerracin, line 118: Nice!

PR #2619, opened 2026-02-18
Add `experimental` to `ExecutorOptions`

### LINKED ISSUE #2618: `Executor` should accept experimental options
The [0.2 schema](https://github.com/Qiskit/ibm-quantum-schemas/blob/main/ibm_quantum_schemas/models/executor/version_0_2/models.py) of the `Executor` added an experimental features field (as is available with other primitives). However, the field was not added to the `ExecutorOptions` and thus the user cannot use it.

**Acceptance criteria**
The user can set experimental features via the `ExecutorOptions` and pass them to the server side.

TsafrirA:
### Related Issues
Closes #2618 

### Details and comments
This PR adds an `experimental` field to the `ExecutorOptions` so that it can pass experimental options to the server.

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/converters/converters_0_2.py
+++ b/qiskit_ibm_runtime/quantum_program/converters/converters_0_2.py
@@ -75,6 +75,13 @@ def quantum_program_to_0_2(program: QuantumProgram, options: ExecutorOptions) ->
             raise ValueError(f"Item {item} is not valid.")
         model_items.append(model_item)
 
+    # Build options dict starting with execution options
+    options_dict = asdict(options.execution)  # type: ignore[call-overload]
+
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/quantum_program/converters/converters_0_2.py
```diff
@@ -75,14 +75,21 @@ def quantum_program_to_0_2(program: QuantumProgram, options: ExecutorOptions) ->
             raise ValueError(f"Item {item} is not valid.")
         model_items.append(model_item)
 
+    # Build options dict starting with execution options
```
TsafrirA, line 78: We could avoid this by putting the experimental inside the execution options.

PR #2623, opened 2026-02-23
Include ibm-quantum-schemas documentation

### LINKED ISSUE #2622: Include models documentation as part of the documentation

In preparing for the models for `ibm-quantum-schemas` being used by this package in subsequent releases, we should document them as part of this package's documentation. The goal is to be able to keep using the same rendering mechanisms, and just include the API documentation for that package as entries in this one. 

**Acceptance criteria**

The sphinx output of this package contains the public interface of `ibm-quantum-schemas`.

diego-plan9:
### Summary

Add a new entry to the `apidocs/` section containing the `ibm-quantum-schemas` API documentation.

The documentation for the models is rendered using [autodoc_pydantic](https://github.com/mansenfranzen/autodoc_pydantic), as the default sphinx `autodoc` does not produce readable/usable outputs for Pydantic models. The new page relies on the docstrings of the moduls in `ibm-quantum-schemas`.

### Details and comments

Fixes #2622

See https://github.com/Qiskit/ibm-quantum-schemas/pull/86 for more details.

### DIFF
--- a/test/integration/test_fake_backends.py
+++ b/test/integration/test_fake_backends.py
@@ -13,13 +13,74 @@
 # pylint: disable=missing-class-docstring,missing-function-docstring
 # pylint: disable=missing-module-docstring
 
+import operator
+
+from ddt import ddt, data
+
+from qiskit.circuit import QuantumCircuit
+from qiskit.utils import optionals
+from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
 from qiskit.providers.exceptions import QiskitBackendNotFoundError
 
+from qiskit_ibm_runtime import SamplerV2 as Sampler
 from qiskit_ibm_runtime import QiskitRuntimeService
-from qiskit_ibm_runtime.fake_provider import FakeSherbrooke
+from qiskit_ibm_runtime.fake_provider import FakeVigoV2, FakeSherbrooke, FakeProviderForBackendV2
 
-from ..ibm_test_case import IBMIntegrationTestCase
+from ..ibm_test_case import IBMTestCase, IBMIntegrationTestCase
 from ..decorators import production_only
+
+
+FAKE_PROVIDER_FOR_BACKEND_V2 = FakeProviderForBackendV2()
+
+
+@ddt
+class TestFakeBackends(IBMTestCase):
+    """Test runnning circuits in fake backends.
+
+    These tests are considered integration tests, even if they do not require network access, as
+    they depend on `aer` and are time consuming.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.circuit = QuantumCircuit(2)
+        cls.circuit.h(0)
+        cls.circuit.h(1)
+        cls.circuit.h(0)
+        cls.circuit.h(1)
+        cls.circuit.x(0)
+        cls.circuit.x(1)
+        cls.circuit.measure_active()
+
+    @data(*[be for be in FAKE_PROVIDER_FOR_BACKEND_V2.backends() if be.num_qubits > 1])
+    def test_circuit_on_fake_backend_v2(self, backend):
+        """Test running a circuit in fake backends."""
+        if not optionals.HAS_AER and backend.num_qubits > 20:
+            self.skipTest(f"Unable to run fake_backend {backend.backend_name} without qiskit-aer")
+        backend.set_options(seed_simulator=42)
+        pm = generate_preset_pass_manager(backend=backend, optimization_level=0)
+        isa_circuit = pm.run(self.circuit)
+        sampler = Sampler(backend)
+        job = sampler.run([isa_circuit])
+        pub_result = job.result()[0]
+        counts = pub_result.data.meas.get_counts()
+        max_count = max(counts.items(), key=operator.itemgetter(1))[0]
+        self.assertEqual(max_count, "11")
+
+    @data(0, 1, 2, 3)
+    def test_circuit_on_fake_backend_v2_with_optimization_level(self, optimization_level):
+        """Test running a circuit in a fake backend with different optimization level."""
+        backend = FakeVigoV2()
+        backend.set_options(seed_simulator=42)
+        pm = generate_preset_pass_manager(backend=backend, optimization_level=optimization_level)
+        isa_circuit = pm.run(self.circuit)
+        sampler = Sampler(backend)
+        job = sampler.run([isa_circuit])
+        pub_result = job.result()[0]
+        counts = pub_result.data.meas.get_counts()
+        max_count = max(counts.items(), key=operator.itemgetter(1))[0]
+        self.assertEqual(max_count, "11")
 
 
 class TestRefreshFakeBackends(IBMIntegrationTestCase):
--- a/test/unit/test_fake_backends.py
+++ b/test/unit/test_fake_backends.py
@@ -13,21 +13,14 @@
 # pylint: disable=missing-class-docstring,missing-function-docstring
 # pylint: disable=missing-module-docstring
 
-import itertools
-import operator
-
-from ddt import ddt, data, idata, unpack
+from ddt import ddt, data
 
 from qiskit.circuit import QuantumCircuit
 from qiskit.compiler import transpile
-from qiskit.utils import optionals
 from qiskit.circuit.library import (
     CZGate,
     ECRGate,
 )
-from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
-
-from qiskit_ibm_runtime import SamplerV2 as Sampler
 
 from qiskit_ibm_runtime.fake_provider import (
     FakeProviderForBackendV2,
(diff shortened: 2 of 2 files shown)


diego-plan9:
Marking as WIP as:
* this PR currently depends on a commit in the schemas repo (a new release will be made for proper rendering).
* the changes related to GitHub links have not been ported yet to this PR
* ideally, I'd like to mention the version of `ibm-quantum-schemas` in the rendered documentation

docs/conf.py
```diff
@@ -97,10 +103,10 @@ class Self:
 
 autosummary_generate = True
 autosummary_generate_overwrite = False
-autoclass_content = "both"
+autoclass_content = "class"
 autodoc_typehints = "description"
 autodoc_default_options = {
-    "inherited-members": None,
+    "inherited-members": "BaseModel",
```
diego-plan9, line 109: A bit of a dance with the config:
* `apidoc-pydantic` in Sphinx 8.x tries to document the inherited members, and emits warning as `BaseModel` is not prepared for sphinx rendering. Sphinx 9.x does not need the `BaseModel` explicit exclusion.
* `apidoc-pydantic` appends by default `BaseModel.__init__` docstring to the class docstring. I resorted to tweaking our template (so in practice, the behavior remains the same as prior to this change for this package's documentation).

docs/conf.py
```diff
@@ -209,19 +232,26 @@ def is_valid_code_object(obj):
...
     else:
         ending_lineno = lineno + len(source) - 1
         linespec = f"#L{lineno}-L{ending_lineno}"
-    return f"https://github.com/Qiskit/qiskit-ibm-runtime/tree/{GITHUB_BRANCH}/qiskit_ibm_runtime/{file_name}{linespec}"
+
+    # Build and return the GitHub URL.
+    if package_name == "ibm_quantum_schemas":
+        prefix = f"https://github.com/Qiskit/ibm-quantum-schemas/tree/{ibm_quantum_schemas_release}"
+    else:
+        prefix = f"https://github.com/Qiskit/qiskit-ibm-runtime/tree/{GITHUB_BRANCH}"
```
diego-plan9, line 255: Changes to this function are aimed at:
* for `qiskit_ibm_runtime` members, keep the previous behavior
* for `ibm-quantum-schemas` members, point to the schemas repository (specifically, to the used version). The repository does not use `stable/` branches, and the tag name are the same as the published version numbers.

pyproject.toml
```diff
@@ -183,11 +183,12 @@ common = [
 
 documentation = [
     "nbsphinx",
-    "Sphinx>=6",
+    "Sphinx>=8,<9",
```
diego-plan9, line 186: Some experimentation led me to refine the Sphinx versions used:
* Sphinx 9.x is only available for Python 3.12+. Without restricting, in local environment there was a chance a higher version than the one used in CI was used (CI is using Python 3.10)
* Sphinx 9 seems to have rewritten `autodoc`, which is used heavily here:
   > #14089: autodoc: Fix default option parsing. Patch by Adam Turner

   and seems to cause some inconsistent behaviors

Ultimately, this aims to ensure that we use a more homogeneous Sphinx version in CI and locally (most of the runs have been using `8.x` for quite a while), and intends to be conservative and minimize potential changes due to a new Sphinx version.

docs/_templates/autosummary/class.rst
```diff
@@ -7,12 +7,18 @@
...
+{% else %}
 .. autoclass:: {{ objname }}
    :no-members:
    :no-inherited-members:
    :no-special-members:
    :show-inheritance:
+   :class-doc-from: both
 
+{% if fullname not in ['qiskit_ibm_runtime.Estimator', 'qiskit_ibm_runtime.Sampler'] %}
+{# Skip problematic re-imports, as autodoc cannot resolve the references to attributes / methods #}
```
diego-plan9, line 21: In Sphinx 9.x, this results in several:
```
don't know which module to import for autodocumenting 'mode' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name) [autodoc]
```
warnings for the members in `Sampler` and `Estimator`.

The culprit seems to be the re-imports with alias:
```
from .sampler import SamplerV2, SamplerV2 as Sampler  # pylint: disable=reimported
```

Although I could not trace why this changed between 8.x and 9.x. Still, I preferred to keep this in place to cover for a future Sphinx 9.x transition, and also since `Estimator` and `Sampler` documentation just displays the alias and points to the actual classes.

diego-plan9:
Ok, not taking formally out of WIP - but the last remaining piece is making this depend on a `ibm-quantum-schemas` release rather than a commit. CC @arnaucasau

qiskit_ibm_runtime/options/noise_learner_v3_options.py
```diff
@@ -111,7 +111,8 @@ def _nonnegative_list(cls, value: list[int], info: ValidationInfo) -> list[int]:
 
     def to_options_model(self, schema_version: str) -> BaseModel:
         """Turn these options into an ``OptionsModel`` object.
-        Filters out every irrelevant field (i.e., those that are not fields of :class:`.OptionsModel`).
+
+        Filters out every irrelevant field (i.e., those that are not fields of ``OptionsModel``).
```
diego-plan9, line 115: Small tweak as Sphinx cannot resolve if it is `v0_1.OptionsModel` or `v0_2.OptionsModel`. So I went with avoiding the link to the class instead.

docs/apidocs/schema_models.rst
```diff
@@ -0,0 +1,4 @@
+.. automodule:: ibm_quantum_schemas.models
+   :no-members:
+   :no-inherited-members:
```
rainerschoe, line 3: Does this only apply to BaseModel? Asking, as we sometimes use inheritance deliberately e.g. for `QpyModel`

docs/apidocs/noise_learner_v3_0_2.rst
```diff
@@ -0,0 +1,4 @@
+.. automodule:: ibm_quantum_schemas.models.noise_learner_v3.version_0_2
```
rainerschoe, line 1: Maybe in the future we could model the whole params model as a pydantic model listing all the supported versions as possible types with schema_version as a discriminator.

This way we would only need to add the root module and not all versions manually.

docs/apidocs/index.rst
```diff
@@ -20,3 +20,4 @@
    debug_tools
    visualization
    models
+   ibm_quantum_schemas
```
rainerschoe, line 23: I find it a bit confusing: the above is documenting python API of the qiskit-ibm-runtime client.
But in the same list we see `ibm_quantum_schemas` documentation which is not bundled with the client.
Why is this?

docs/apidocs/schema_models.rst
diego-plan9: It is a bit convoluted, and keep in mind I might be missing something. Prior to this PR (and this has not changed in this PR):
* the [`autodoc`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) plugin uses templates, and we are using a custom template for [classes](https://github.com/diego-plan9/qiskit-ibm-runtime/blob/6c97d05abd1ea0fca0ba5ab3e7914228f31da43a/docs/_templates/autosummary/class.rst).
* for modules that are not `ibm-quantum-schemas`, the custom template seems to make `:no-inherited-members:` ignored (probably due to using `all-attributes` / `all-methods` and not taking into account `inherited_members`, compared to the [default template](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/templates/autosummary/class.rst)). This makes inherited members to actually show up in the documentation (see [this transpiler pass](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/transpiler-passes-convert-id-to-delay), for example).

With this PR, a couple of things are added on top:
* `"inherited-members": "BaseModel",` in `conf.py` instructs sphinx to not include members inherited by `BaseModel`, but to include the rest
* `autodoc-pydantic` does not use the templates for [autosummary](https://autodoc-pydantic.readthedocs.io/en/stable/users/usage.html#autosummary) - at least in Sphinx 8.x. It seems to be using them in Sphinx 9.x, that is why I left the `if` statement in the template.
* so for `ibm-quantum-schema` models, it should follow the directive: and it seems to be doing so (`QpyModelV13ToV17` does include methods and attributes from `QpyModel` in the rendered doc).

docs/apidocs/noise_learner_v3_0_2.rst
diego-plan9: Seems like a good thing - I explicitly tried for this PR not to cause changes in the actual models (ie. just in the `__init__.py`), as I believe that (and others such as re-importing higher up) is a design decision of `ibm-quantum-schemas`.

docs/apidocs/index.rst
diego-plan9: Because we want to re-use the same documentation site (ie. https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime) instead of a dedicate site for the schemas package.

docs/apidocs/schema_models.rst
rainerschoe: Understand. thanks. I was missing the part with:
> "inherited-members": "BaseModel", in conf.py instructs sphinx to not include members inherited by BaseModel, but to include the rest

PR #2626, opened 2026-02-25
Reintroduce executor / nlv3 docs

diego-plan9:
### Summary

Make the docs for executor and noise learner v3 part of the documentation. This is basically a partial revert of #2600.

### Details and comments

Fixes N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,6 +110,7 @@ jobs:
           flag-name: unit-tests_python${{ matrix.python-version }}-${{ matrix.os }}
           parallel: true
           file: coverage.lcov
+          fail-on-error: false
   integration-tests:
     if: github.event_name == 'push' && github.repository_owner == 'Qiskit'
     # only kick-off resource intensive integration tests if unit tests and all basic checks succeeded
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/options/__init__.py
```diff
@@ -71,12 +72,14 @@
    :nosignatures:
 
    NoiseLearnerOptions
+   NoiseLearnerV3Options
```
yaelbh, line 75: We don't have in qiskit-ibm-runtime a formal notion of `Base primitive options` and `Suboptions`. Informally, I'd say that `NoiseLearnerOptions` and `NoiseLearnerV3Options` should live together with the important options, i.e. those of the sampler, estimator, and executor.

diego-plan9: I partially agree: as a counter, the documentation [tends to consider](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/__init__.py#L29) `Sampler` and `Estimator` the first-class primitives, and the implications carry on (including this file). I'm leaning towards not changing conventions in this PR, and respect the original documentation changes that were performed in the preview release of executor.

PR #2627, opened 2026-02-26
Avoid CI failure when coveralls uploads fail

diego-plan9:
### Summary

`coveralls.io` is having an outage, which is preventing CI to pass. From the [status page](https://status.coveralls.io/):

> To avoid further disruption to your CI workflows, we recommend employing the fail-on-error: false input option available with all official coveralls integrations. 
See: https://docs.coveralls.io/integrations#official-integrations.

This PR applies that flag, so the job can continue even if the coveralls upload fails.

### Details and comments
Fixes N/A


TsafrirA (approved):
Thank you.

PR #2629, opened 2026-02-26
[sampler-preview] Add minimal options for the executor-based SamplerV2

### LINKED ISSUE #2586: Support general options in the new SamplerV2 implementation (execution, experimental)

TsafrirA:
### Related Issues
Closes #2586 
#2582
Replaces #2617

### Summary
This PR adds minimal options support for the new `SamplerV2`. Basic options are mapped to the `Executor`, while others raise a `NotImplementedError`.

### DIFF
--- a/qiskit_ibm_runtime/executor/routines/options/sampler_options.py
+++ b/qiskit_ibm_runtime/executor/routines/options/sampler_options.py
@@ -16,7 +16,6 @@ from __future__ import annotations
 
 from typing import Literal
 
-from dataclasses import asdict
 from pydantic import Field
 from pydantic.dataclasses import dataclass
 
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/executor/routines/options/sampler_options.py
```diff
@@ -0,0 +1,111 @@
...
+from .dynamical_decoupling_options import DynamicalDecouplingOptions
+from .twirling_options import TwirlingOptions
+from .environment_options import EnvironmentOptions
+from ....options.executor_options import ExecutorOptions
+
+from ....options.executor_options import ExecutionOptions
+
+
+@dataclass
+class SamplerExecutionOptions(ExecutionOptions):
```
TsafrirA, line 31: Not sure about the inheritance pattern here. On the one hand it's a bit more future proof, but on the other hand it could end up breaking the SamplerV2 interface compared to the old implementation.

SamFerracin: not sure, but we can leave it as is and fix later if needed

PR #2625, opened 2026-02-25
[sampler preview] add twirling support for the new SamplerV2

TsafrirA:
### Related Issues
#2582 
Closes #2587

### Summary
This PR adds twirling support for the new SamplerV2. When twirling is enabled, Samplomatic is used to box and build the circuit, and a SamplexItem is created.

Results are not flattened like legacy sampler. This will be done in a separate PR once flattening strategy is finalized.

### Test Evidence
Running the following circuits 
```python
from qiskit.circuit import QuantumCircuit, Parameter
from qiskit.transpiler import generate_preset_pass_manager
from qiskit_ibm_runtime import QiskitRuntimeService

lengths = [1, 51, 101, 201, 301]

circs = []

for x in lengths:
    circ = QuantumCircuit(2)
    circ.x(0)
    for _ in range(x):
        circ.cx(0,1)
    circ.measure_all()
    circs.append(circ)
```
on `ibm_fez` with twirling enabled for both gates and measurements yields
image

PR #2634, opened 2026-03-02
[sampler preview] Flatten results when doing twirling with the new SamplerV2

TsafrirA:
### Related Issues
~~Depends on #2625~~

### Summary
#2625 adds twirling support to the new SamplerV2. Here we flatten the results to match existing behavior of the legacy SamplerV2.


qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -133,12 +140,16 @@ def prepare(
...
             )
+            pub_shapes.append(list(param_shape))
 
-    passthrough_data = {
-        "post_processor": {
+        post_processor_data = {
             "context": "sampler_v2",
             "version": "v1",
-        },
+            "pub_shapes": pub_shapes,
```
TsafrirA, line 150: We end up not really using the shapes. We could go for something lighter, like number of dimensions in pub shape.

SamFerracin (dismissed):
I'll approve this so that we get to a working version.

One comment for the future: I think that we should not include `pub_shapes` in the passthrough data, as this can be inferred in post-processing from the shapes of the results + the twirling options. 
Benefits:
- The prepare function becomes simpler, to explain (".. and attach the options to the pass through data" VS ".. and attach the options to the pass through data, and if twirling is ON, also add the pub shapes, otherwise don't") and to re-implement (if users decide to go with custom prepare but default post-processing, they have less things to worry about)

Of course, this will add logic in post-processing, but if we can capture that logic in util functions (e.g., in this case, a function that maps results and options to shapes or None), then users can use it, and the post-processing fn remains readable.

But let's think about this later, when we have all the code in place :)

TsafrirA:
@SamFerracin needs a re-approval due to minor conflicts.

SamFerracin (approved):

PR #2635, opened 2026-03-02
Remove types from docstring args and returns

diego-plan9:
### Summary

Remove the types specified inside the docstrings, so we can rely on the actual type annotations for sphinx needs and other tooling. This ensures we have a single source of truth for the types (ie. the type annotations), reducing the risk for the docstring comments going out of sync. 

### Details and comments

Fixes N/A

Most of our documentation did not include the types in the docstrings - these are the remaining items I could find (grepping over parenthesis in the `Arguments:` block or `:` in `Returns:` blocks).

### DIFF
--- a/qiskit_ibm_runtime/models/backend_configuration.py
+++ b/qiskit_ibm_runtime/models/backend_configuration.py
@@ -153,7 +153,7 @@ class UchannelLO:
 
         Args:
             data: A dictionary representing the UChannelLO to create. It will be in the same
-            format as output by :func:`to_dict`.
+                format as output by :func:`to_dict`.
 
         Returns:
             The UchannelLO from the input dictionary.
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -290,22 +289,20 @@ def run(self, run_input, **options):  # type: ignore
         FakeBackendV2.
 
         Args:
-            run_input (QuantumCircuit or list): An
-                individual or a list of
-                :class:`~qiskit.circuit.QuantumCircuit`
+            run_input: An individual or a list of :class:`~qiskit.circuit.QuantumCircuit`
```
yaelbh, line 292: Even after the change the comment pretty much repeats what we already know from the type.

diego-plan9: Yes, some content is a bit redundant here and in other places - I did not focus on rewriting, just removing the types.

PR #2637, opened 2026-03-03
[sampler preview] Prepare function injection for the new SamplerV2

TsafrirA:
### Summary
This PR adds the ability to inject a custom prepare function for the new SamplerV2.

### DIFF
--- a/qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
+++ b/qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
@@ -184,7 +184,9 @@ class SamplerV2(BaseSamplerV2):
 
     You can inject a custom prepare function to replace the default conversion logic
     from SamplerPub objects to QuantumProgram. The custom function must have the
-    following signature::
+    following signature:
+    
+    ```python
 
         def my_prepare(
             pubs: list[SamplerPub],
@@ -192,6 +194,7 @@ class SamplerV2(BaseSamplerV2):
             default_shots: int | None = None,
         ) -> tuple[QuantumProgram, ExecutorOptions]:
             ...
+    ```
 
     The custom function can be provided either at initialization via the ``prepare_fn``
     parameter or later via the ``prepare_fn`` property. Set to ``None`` to restore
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -180,6 +180,23 @@ class SamplerV2(BaseSamplerV2):
...
+    You can inject a custom prepare function to replace the default conversion logic
+    from SamplerPub objects to QuantumProgram. The custom function must have the
+    following signature::
+
+        def my_prepare(
+            pubs: list[SamplerPub],
+            options: SamplerOptions,
+            default_shots: int | None = None,
+        ) -> tuple[QuantumProgram, ExecutorOptions]:
+            ...
```
```suggestion
    following signature:
    
    ```python

        def my_prepare(
            pubs: list[SamplerPub],
            options: SamplerOptions,
            default_shots: int | None = None,
        ) -> tuple[QuantumProgram, ExecutorOptions]:
            ...
    ```
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -215,18 +242,30 @@ class SamplerV2(BaseSamplerV2):
             for more information about execution modes.
 
         options: Sampler options. See :class:`SamplerOptions` for all available options.
+        prepare_fn: Optional custom prepare function to replace the default conversion
+            logic. Must have signature ``(pubs, options, default_shots) ->
+            (QuantumProgram, ExecutorOptions)``. Pass ``None`` to use the default prepare
+            function.
```
SamFerracin: Redundant with type-hints
```suggestion
            logic.
```

SamFerracin (commented):
Minor comments, but looks good to go in.

I am not sure that I like `prepare_fn` because of the `function` abbreviated into `fn`, I wonder if it could just be `prepare` or `custom_prepare` (and keeping `_prepare` in the internals). But we can leave it as is and think about it some more in future iterations

SamFerracin (approved):

PR #2654, opened 2026-03-11
Do not emit deprecation warning on RuntimeOptions

diego-plan9:
### Summary

Instead of fully removing `RuntimeOptions`, avoid the deprecation warning and rely on docstrings (and previous release notes). This prevents the warning to be emitted during normal flow of operations (as in the class is still used internally), and does not change the deprecation status of `RuntimeOptions`: it still is pending removal and users should not depend on it.

### Details and comments
Related to #2642, #2300

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -140,12 +140,12 @@ dependencies = [
     "urllib3>=1.21.1",
     "python-dateutil>=2.8.0",
     "ibm-platform-services>=0.22.6",
-    "ibm-quantum-schemas>=0.3.20260226",
+    "ibm-quantum-schemas>=0.4.20260309,<1",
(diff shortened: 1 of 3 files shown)


test/unit/test_options.py
```diff
@@ -20,7 +20,7 @@
 from qiskit.transpiler import CouplingMap
 from qiskit_aer.noise import NoiseModel
 
-from qiskit_ibm_runtime import RuntimeOptions
+from qiskit_ibm_runtime.runtime_options import RuntimeOptions
```
diego-plan9, line 23: This way we are a bit closer to removing the top-level import, as it is the only place in the package there it was imported directly from root. I kept the top-level import for consistency until the class is finally removed.

PR #2655, opened 2026-03-11
[sampler preview] Fix some sphinx references

TsafrirA:
### Summary
Fix some sphinx references.

### DIFF
--- a/qiskit_ibm_runtime/executor/routines/sampler_v2/post_processors/v0_1/post_processor.py
+++ b/qiskit_ibm_runtime/executor/routines/sampler_v2/post_processors/v0_1/post_processor.py
@@ -40,6 +40,8 @@ def sampler_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveRes
     Returns:
         Primitive result for :class:`~qiskit_ibm_runtime.executor.routines.sampler_v2.SamplerV2`.
     """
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/executor/routines/options/environment_options.py
```diff
@@ -43,7 +43,12 @@ class EnvironmentOptions:
     standard retention behavior of the API."""
 
     def to_executor_options(self) -> ExecutorEnvironmentOptions:
-        """Converts the environment options to executor environment options."""
+        """Converts the environment options to executor environment options.
+
+        Returns:
+            :class:`qiskit_ibm_runtime.options.executor_options.EnvironmentOptions`:
+                The converted executor environment options.
```
SamFerracin: In this repo, we don't normally include a type hint in the returns, only a description. There are a few docstrings to change this way
```suggestion
        Returns:
            The converted executor environment options.
```

TsafrirA: With this particular instance I think it's worthwhile, because we want to highlight the difference between the different `EnvironmentOptions` classes. But this is not a strong preference.

TsafrirA: (will check to see if there are other erroneous instances)

PR #2661, opened 2026-03-12
Add Embeddings for 156 and 133 qubits

ihincks:
### Summary

Along with adding the new embeddings, this PR also refactors the tests to avoid invoking the private method. This is done by passing in a fake backend with the correct number of qubits in each case. I also stopped testing with list-pastes, because they are impractical to manually construct and verify, so that in practice people are probably just copy-pasting the lists after running the function.

### Details and comments
Fixes #

### DIFF
--- a/test/unit/test_embeddings.py
+++ b/test/unit/test_embeddings.py
@@ -31,6 +31,7 @@ from qiskit_ibm_runtime.fake_provider import (
     FakePerth,
     FakeRochesterV2,
     FakeMarrakesh,
+    FakeTorino,
 )
 from qiskit_ibm_runtime.utils.embeddings import Embedding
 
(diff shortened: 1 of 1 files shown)


test/unit/test_embeddings.py
```diff
@@ -57,21 +102,30 @@ def test_init_error(self):
...
+        self.assertListEqual(embedding.coordinates, exp)
 
     def test_7(self):
         r"""Test for 7-qubit lattices."""
-        exp = [(0, 0), (0, 1), (0, 2), (1, 1), (2, 0), (2, 1), (2, 2)]
-        self.assertListEqual(_get_qubits_coordinates(7), exp)
+        embedding = Embedding.from_backend(FakePerth())
+        exp = ascii_to_coords(
+            """
+        xxx
```
SamFerracin, line 119: wow!

PR #2664, opened 2026-03-12
[sampler-preview] Remove `options` from passthrough data

SamFerracin:
This PR removes options as they add redundant info


TsafrirA (approved):
LGTM.

There are many other tests that set the options field. Not urgent to remove in this PR, as they will pop up when the schema is introduced.

PR #2670, opened 2026-03-15
[sampler preview] Simplify flattening function

TsafrirA:
### Summary
This PR follows up #2662 and simplifies the logic of the flattening function. Now that the PUBs are inferred instead of passed, there is no sense in testing things against the pub shapes. Also, the function is only called when twirling is on, so that too can be simplified.


SamFerracin (approved):
Ty, looks nice!

Issue #2682, opened 2026-03-18
Add support for `meas_type` to wrapper sampler


SamFerracin:
The existing sampler supports three values of `meas_type`:
```
meas_type: UnsetType | Literal["classified", "kerneled", "avg_kerneled"]
```

On the contrary, while wrapper sampler allows specifying the same three values in the options, a job with `meas_type != "classified"` will trigger a server-side error, since `Executor` does not currently support it.

TODOs:
- [ ] Add support for `meas_type` to Executor
- [ ] Ensure that wrapper sampler can run jobs with `meas_type != "classified"`

SamFerracin:
@yaelbh can we close this?

yaelbh:
#2687 addresses this issue for `meas_type="kerneled"`.

However the executor is not ready yet for `meas_type="avg_kerneled"`. Once it becomes supported, we will have to also update the wrapper sampler.

Therefore this issue remains open for now.

yaelbh:
The executor now supports `avg_kerneled`. I propose the next steps:
1. Add support for `avg_kerneled` in the wrapper sampler.
2. Verify that everything plays together by writing integration tests, for the both the executor and the wrapper sampler.

PR #2687, opened 2026-03-20
[Sampler-preview] Add support for level 1 measurements

SamFerracin:
Addresses #2682, for `meas_type="kerneled"`.


TsafrirA (commented):
I think we can use the data type of the array to infer the measurement level instead of adding another field to the passthrough data?

yaelbh:
> I think we can use the data type of the array to infer the measurement level instead of adding another field to the passthrough data?

I actually think it's generally, and also here, more robust to specify explicitly, while adding some redundancy, rather than inferring one thing from another. It will be possible to distinguish between `kerneled` and `kerneled_avg`, but  will involve adding more code that will bug-prone. For both `kerneled` and `kerneled_avg` the data type is complex numbers, and the dispatch can be done via the array shape.

Any reason to refrain from adding a field? We can exploit the redundancy by asserting that the specified measurement level matches the array's shape and data type.

yaelbh:
I fixed the existing tests so they should pass now, however it still remains to add new tests (currently `meas_type` is `classified` in all the tests). The code itself looks correct.

TsafrirA (approved):
Thank you!

Issue #2691, opened 2026-03-23
`setuptools_scm` wrongly (?) pinned in last release?


iyanmv:
**Describe the bug**
In the last release, the version of `setuptools_scm` is pinned to `~=8.0`.

This was changed in this commit https://github.com/Qiskit/qiskit-ibm-runtime/commit/08bc8a3fc2b1a6959ad0d28765fc030e7390d76a without any justification:

```
commit 08bc8a3fc2b1a6959ad0d28765fc030e7390d76a
Author: SamFerracin <sam.ferracin@ibm.com>
Date:   Wed Oct 29 14:39:11 2025 -0400

    conflicts

 .whitesource                                                   |  3 +++
 README.md                                                      |  2 ++
 pyproject.toml                                                 | 10 +++++-----
 qiskit_ibm_runtime/ibm_backend.py                              |  2 +-
 qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3_result.py |  2 --
 5 files changed, 11 insertions(+), 8 deletions(-)
 create mode 100644 .whitesource
```

This change breaks creating a wheel package using any distro with up to date setuptools-scm.

**Suggested solutions**

Do not pinned the version, unless it is known to be incompatible. If that's the case, please document it so that distros can track it properly.

- **qiskit-ibm-runtime version**: 0.46.0
- **Python version**: 3.14.3
- **Operating system**: Arch Linux

PR #2692, opened 2026-03-23
Unpin setuptools_scm

SamFerracin:
Addresses #2691 

In the last release, the version of setuptools_scm was incorrectly pinned to ~=8.0. This PR reverts it, restroring the original `>=6.2`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools >= 65.0.2", "setuptools_scm[toml]~=8.0"]
+requires = ["setuptools >= 65.0.2", "setuptools_scm[toml]>=6.2"]
 build-backend = "setuptools.build_meta"
 
 [tool.black]


ihincks:
It looks like the PR/commit that introduced the pin did so accidentally, possibly as a part of some sort of local testing workflow. This is a faulty dependency spec and should be considered a bug, warranting a patch release.

diego-plan9:
> It looks like the PR/commit that introduced the pin did so accidentally, possibly as a part of some sort of local testing workflow. This is a faulty dependency spec and should be considered a bug, warranting a patch release.

That is correct - indeed, I unintendedly introduced it in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2689, as the flow for identifying the minimal dependencies involved pinning to those minimal dependencies (and later on relaxing the pin manually, which did not happen for `setuptools_scm`).

Thanks for the speedy resolution and for the new release!

PR #2699, opened 2026-03-24
Deprecate max_execution_time and introduce max_usage

### LINKED ISSUE #2638: Update max_execution_time to max_usage
We should migrate away from max_execution_time in favour of max_usage as a more accurate name for the field

diego-plan9:
### Summary

Replace `max_execution_time` with `max_usage`:
* in the user-facing public Option classes, keep the two attributes (with some logic to ensuring `max_usage` ends up being populated), emitting deprecation warnings if `max_execution_time` is populated via Pydantic validators (whenever possible) or `__post_init__`
* in the internal non public members, replacing `max_execution_time` with `max_usage` directly

### Details and comments

Fixes #2638


qiskit_ibm_runtime/options_models/environment_options.py
```diff
@@ -73,9 +73,22 @@ class EnvironmentOptions:
     this time limit, it is forcibly cancelled.
     """
 
+    max_usage: int | None = None
+    """Maximum usage in seconds.
+
+    This value bounds system usage (not wall clock time). System usage is the amount of time that
+    the system is dedicated to processing your job. If a job exceeds this time limit, it is
+    forcibly cancelled.
```
jyu00, line 74: There is a [docs PR](https://github.com/Qiskit/documentation/pull/4787) with the new wording for usage definition. I was going to suggest a change here based on that, but there's now a new debate on whether the one used in the docs PR is correct 😅.... so stay tuned!

diego-plan9: Thanks, a proper wording would be welcome!

qiskit_ibm_runtime/options/noise_learner_v3_options.py
```diff
@@ -153,3 +163,7 @@ def _get_program_inputs(options: dict) -> dict:
             Inputs acceptable by primitives.
         """
         raise NotImplementedError("Not implemented by `NoiseLearnerV3Options`.")
+
+    def __post_init__(self) -> None:
```
yaelbh: Is it a model validation in one class and a post init function in another class because one class is a primitives dataclass and the other is just a dataclass?

qiskit_ibm_runtime/options/options.py
```diff
@@ -116,33 +117,40 @@ def _repr_html_(self) -> str:
...
     """
 
-    _VERSION: int = Field(2, frozen=True)
+    max_usage: UnsetType | int = Unset
+    """Maximum usage time in seconds.
+
+    Maximum usage in seconds, which is based on system execution time (not wall clock time). System
+    execution time is the amount of time that the system is dedicated to processing your job. If a
+    job exceeds this time limit, it is forcibly cancelled. Simulator jobs continue to use wall
+    clock time.
```
yaelbh, line 143: What's the default value?

qiskit_ibm_runtime/options/utils.py
```diff
@@ -32,6 +33,12 @@
     from ..options.options import BaseOptions
 
 
+MAX_EXECUTION_TIME_DEPRECATION_MSG = (
+    "`max_execution_time` is deprecated as of qiskit_ibm_runtime v0.47.0 and will "
+    "be removed in a future release. Use `max_usage` instead."
```
yaelbh, line 38: We usually specify a lower bound of time where we guarantee to keep supporting the old stuff.

qiskit_ibm_runtime/options/utils.py
```diff
@@ -293,3 +300,36 @@ def validator(cls: Any, value: Any, validation_info: ValidationInfo) -> Any:
...
+
+    This is a convenience function that should be removed once `max_execution_time` is
+    removed.
+    """
+    max_execution_time = instance.max_execution_time
+    max_usage = instance.max_usage
+
+    if max_usage is not comparator:
+        if max_execution_time is not comparator:
+            warnings.warn(
```
yaelbh, line 324: I prefer an error if they were both set but to different values.

qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -250,7 +250,7 @@ def submit(
             log_level=options.environment.log_level,
             session_id=None,
             job_tags=options.environment.job_tags,
-            max_execution_time=None,
+            max_usage=None,
```
yaelbh: This is a general comment for the PR: you deprecate for users of the option classes, but is it enough?
- For example here, `submit` is not a private method - can we modify its interface just like that?
- What about users of REST API? Is it going to break for them?

qiskit_ibm_runtime/options/noise_learner_v3_options.py
diego-plan9: Yes, reused as much as I could by making `match_max_execution_time_and_max_usage` flexible, but ultimately we need to use the parent class facilities that they provide (ie. sometimes `__post_init__`, sometimes  validator).

qiskit_ibm_runtime/options/options.py
diego-plan9: `Unset`, this is pretty much the same as existing `max_execution_time` docstrings all over the PR. Final docstrings are pending evolution of the above conversation at https://github.com/Qiskit/qiskit-ibm-runtime/pull/2699#discussion_r2982978452

qiskit_ibm_runtime/options/utils.py
diego-plan9: The formula "`xxx is deprecated as of yyy and will be removed in a future release`" is used in all the current deprecation messages existing in `main`, and I think it is a sane one -  the deprecation policy document describing the 3-months policy overall is the source of truth.

qiskit_ibm_runtime/ibm_backend.py
diego-plan9: > For example here, `submit` is not a private method - can we modify its interface just like that?

Can you elaborate on "modify its interface" when it comes to `submit()`? The signature has not been touched.

> What about users of REST API? Is it going to break for them?

The `qiskit_ibm_runtime.api` package is not part of the public interface (as in, not documented at https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime), and does not need to adhere to the deprecation policy. So while the change would potentially cause breakage, the assumption is that users that depend on non-public functionality are able to assume the cost.

The overall approach is described in the PR's description: please let me know if there are other items where changes to the public interface without deprecation might have crept!

mtreinish (commented):
Per the deprecation policy you must introduce the alternative in a release prior to marking the old path as deprecated see: https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/DEPRECATION.md

diego-plan9:
> Per the deprecation policy you must introduce the alternative in a release prior to marking the old path as deprecated see: https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/DEPRECATION.md

Considering this PR a feature removal (as in, the `max_execution_time` attribute is removed), I think this is the most relevant bit of the deprecation policy:
> * A deprecation warning must be issued prior to any removal. The warning must indicate what the alternative path is, and the alternative path must be in place when the warning is issued. [...]

where the alternative path is the `max_usage` attribute (introduced at the same time). All in all, the goal is to allow users to use the new attribute and transition (the parent issue is time-sensitive), while still allowing ample time for using the old attribute (ie. no rush in removing the deprecation), in line with the guiding principles.

This package's deprecation policy doesn't include a mention of introducing the alternative in a prior release, as far as I can see: can you point to the relevant bit? I suspect you might be referring to Qiskit's deprecation policy, where I do see an explicit entry.

diego-plan9:
Marking on hold pending internal conversations.

diego-plan9:
Closing as this became stale, with the intent to re-issue it without conflicts once the issue is again on scope.

Issue #2563, opened 2026-01-13
Automate `if TYPE_CHECKING` usage


diego-plan9:
**What is the expected feature or enhancement?**

From Yael's suggestion at https://github.com/Qiskit/qiskit-ibm-runtime/pull/2558#issuecomment-3740237694

> Another question: can we enable ruff to automatically add `from typing import TYPE_CHECKING` and `if TYPE_CHECKING:` whenever a name is used only in type hints?

This seems possible via the `TC*` rules. We should check if it is an option - if so, it can potentially make imports leaner (as in, just importing when needed by type checkers), and also potentially help untangling a bit some of the imports.

**Acceptance criteria**

The `TYPE_CHECKING` usage is enforced by the tooling.

PR #2700, opened 2026-03-25
Introduce pre-commit and pytest, and drop pylint

### LINKED ISSUE #2698: Upgrade the developer experience (pre-commit, testing, ruff)

The current flow for developers has been in place for a number of years, and in the meantime alternatives have appeared that provide a more convenient experience for developers. This issue is about taking advantage of those and improving the developer experience, via:

1. Using `pre-commit` as an alternative to `Makefile`, at least for style and quality checks. It allows for unifying a number of checks (and automate them as part of the `git` flow if the developer prefers), as well as including additional QA checks seamlessly.
2. Rely solely in `ruff` instead of `pylint` for the linting needs. We have already been using it for a few months, and provides good customizability as well as performance (plus automating a good number of common fixes).
3. Use `pytest` as the test runner, while keeping the test suite agnostic to the runner (so developers can chose the runner of their choice). It provides saner log output, as well as more features (by itself and through plugins) than the standard test runner which we can take advantage of. 

**Acceptance criteria**

`pre-commit` can be run in the rerpo for at least QA checks, `pytest` is run in CI, and `pylint` is no longer used.

diego-plan9:
### Summary

Revise the developer experience:
* introduce `pre-commit` as a dev and CI tool, running `black`, `ruff check`, `mypy`, our local `tools/`, and some general hooks (trailing whitespace, end of file, check yaml)
* drop `pylint` from the tooling in favor of just `ruff`
* use `pytest` as the default runner for local development and CI

### Details and comments

Other than the changes in the developer flow (now `pre-commit` centric), this includes other changes:
* the `make` target for linting (`make lint`, `make mypy`, `make style`, `make black`, `make ruff`) are removed, in favor of `pre-commit run`
* the CI output configuration has been tweaked:
    * the `ci.yml` stage is less verbose than before, and includes a listing of the 10 slowest tests. 
    * the integration test runs are as verbose as before, additionally listing the duration of all tests.
    * other test runs (smoke, terra latest integration) are as verbose as before, without listing the duration.
    * the rationale is that the unit tests are easy to reproduce and debug locally, and this brings the attention more easily to the failures. Overall, the config will likely be evolving: this aims to set some sane starting point.

Fixes #2698

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -262,8 +262,14 @@ as readable as possible. We use:
 * [mypy] as a static type checker for type hinting
 * [pre-commit] for automating running the above tools
 
-To ensure your changes respect the style guidelines, you can run the following command, which will
-also be run by our CI when pushing to the repository:
+To ensure your changes respect the style guidelines, you can install the `pre-commit` hooks:
+
+```sh
+pre-commit install
+```
+
+In this way, they will run automatically with every commit. You can also run them manually with the
+following command:
 
 ```sh
 pre-commit run


pyproject.toml
```diff
@@ -181,6 +181,11 @@ common = [
     "setuptools",
 ]
 
+qa = [
+    "mypy==1.19.0",
+    "pre-commit>=4",
+]
```
diego-plan9: This allows for saving some CI time (for running `pre-commit`, just these dependencies need to be installed, as opposed to the full `dev`).

We could use with a cleanup of the extra (`common` contains items that are only needed for `documentation` or for `test`), but I'm leaving it to another PR (and keeping the duplication of `mypy` until then).

tox.ini
```diff
@@ -47,7 +30,3 @@ skip_install = true
 deps =
 allowlist_externals = rm
 commands = rm -rf {toxinidir}/docs/stubs/ {toxinidir}/docs/_build
-
-[doc8]
-max-line-length=100
-ignore-path=docs/_build
```
diego-plan9, line 53: These are unused, so I'm taking the chance to remove them.

pyproject.toml
```diff
@@ -13,30 +13,30 @@ directory = "release-notes/unreleased"
...
-    [[tool.towncrier.type]]
-    directory = "bug"
-    name = "Bug Fixes"
-    showcontent = true
-
-    [[tool.towncrier.type]]
-    directory = "other"
-    name = "Other Notes"
-    showcontent = true
+[[tool.towncrier.type]]
```
diego-plan9, line 16: Some fixing of the indenting, while we are at it.

Makefile
```diff
@@ -11,38 +11,20 @@
...
-mypy:
-	mypy --module qiskit_ibm_runtime --package test
-
-style:
-	black --check qiskit_ibm_runtime test
+.PHONY: unit-test integration-test smoke-test docs-test unit-test-coverage
 
 unit-test:
-	python -m unittest discover --verbose --top-level-directory . --start-directory test/unit
+	pytest test/unit
```
diego-plan9, line 17: In general, I'm keeping the `xxx-test` targets agnostic to arguments (and taking advantage of `PYTEST_ADDOPTS` in the workflows), in order to be able to use `make xxx-test` in the workflows and still customize them. 

This hinders a bit the default output for developers using the targets, but I think it is fair to assume that developers will resort to their invocation of choice (without using the target) tuned to their needs (ie. more verbose, a different test runner, etc). This line of thinking also makes me thing of removing the targets entirely, but can be decided and tackled later on.

pyproject.toml
```diff
@@ -193,11 +198,15 @@ documentation = [
 ]
 
 visualization = ["plotly>=5.23.0"]
-test = ["ddt>=1.2.0,!=1.4.0,!=1.4.3"]
+test = [
+    "ddt>=1.2.0,!=1.4.0,!=1.4.3",
+    "pytest",
```
SamFerracin, line 203: nice, finally!

.github/workflows/ci.yml
```diff
@@ -33,13 +33,10 @@ jobs:
...
         run: |
           python -m pip install --upgrade pip
-          pip install -e '.[dev]'
-      - name: Run black
-        run: make style
-      - name: Run lint
-        run: make lint
-      - name: Run mypy
-        run: make mypy
+          pip install -e '.[qa]'
```
SamFerracin: This is not a breaking change, right? Users will still be able to `pip install -e '.[dev]'` and don't have to switch to `'.[qa]'`

I am pretty sure that this is the case, based on my looking at the pyproject file, but doesn't hurt to double check

pyproject.toml
```diff
@@ -181,6 +181,11 @@ common = [
     "setuptools",
 ]
 
+qa = [
```
SamFerracin: What does `qa` stand for? Is it standard nomenclature? If not, consider perhaps switching to a more descriptive name, for example `style`

qiskit_ibm_runtime/options/simulator_options.py
```diff
@@ -99,7 +99,7 @@ def set_backend(self, backend: BackendV2) -> None:
                 "qiskit-aer", "Aer provider", "pip install qiskit-aer"
             )
 
-        from qiskit_aer.noise import (  # pylint:disable=import-outside-toplevel
```
SamFerracin, line 102: It would still be nice to have this warning (for Ruff), to make sure that we keep feeling guilty about importing outside of top level

CONTRIBUTING.md
```diff
@@ -187,90 +188,103 @@ and the release notes in particular will be located at
...
-  and functionality checks
+* [ruff] as a linter providing deeper analysis for potential style and functionality issues, as
+  well as conformance to good practices
 * [mypy] as a static type checker for type hinting
+* [pre-commit] for automating running the above tools
 
-To ensure your changes respect the style guidelines, you can run the following
-commands:
+To ensure your changes respect the style guidelines, you can run the following command, which will
+also be run by our CI when pushing to the repository:
```
SamFerracin: To ensure your changes respect the style guidelines, you can install pre-commit hooks:

```sh
pip install pre-commit
pre-commit install
```
In this way, they will run automatically with every commit. You can also run them manually with the following command:

SamFerracin (commented):
Ty, Diego!

Took me a while, I was glad to see the final congratulatory note that GH adds at the end:
Image

But it looks good!

Just a few questions/suggestions

.github/workflows/ci.yml
diego-plan9: Yes, this is a bit related to https://github.com/Qiskit/qiskit-ibm-runtime/pull/2700#discussion_r2989103837

`dev`  brings `qa` in, and the recommendation for developers is still to install `[dev]`. We can do with a bit of re-vamp of the extras in general, but is a bit out of scope and deserves looking into it more carefully.

pyproject.toml
diego-plan9: It stands for quality assurance - `style`, `lint` felt like not capturing the full extent of this section, but I'll give it a re-thought as I agree `qa` might not be too descriptive.

qiskit_ibm_runtime/options/simulator_options.py
diego-plan9: I agree (and it [can](https://docs.astral.sh/ruff/rules/import-outside-top-level/) be done). There are other `ruff` and `pre-commit` rules that are arguably good to have (for example, unify file line endings, isort, some docstring rules, #2563, etc). I prefer to tackle them as a separate PR, and keep this one without touching the `ruff` configuration.

pyproject.toml
diego-plan9: Going with style at 55ccfe6b2ae84d482d6321ae7a0aaf088270820c

CONTRIBUTING.md
diego-plan9: Tweaked at 901603f7e6022c8db9595dfcc08c07a99ce2bd04

PR #2543, opened 2026-01-07
Bump qiskit minimal version to >2

diego-plan9:
### Summary

Bump the minimal Qiskit version to higher or equal to `2.0.0`, as the `1.x` series are no longer supported.

### Details and comments
Fixes #2517

PR #2707, opened 2026-03-27
Refactor package extras

diego-plan9:
### Summary

The optional dependencies declared for the package became a bit coarse over the years, preventing for easily installing a subset of them for a particular task. This PR:
* provides finer-grained extras, making use of them in CI
* keeps the `dev` extra as the main entry point for developers
* deprecates but preserves the existing `common` and `documentation` extras (additionally wrapping them at a `dev_legacy` extra), as technically this can be considered an user-facing feature (albeit unlikely to be consumed by non-developers).

Some likely unused dependencies have been dropped.

### Details and comments

From https://github.com/Qiskit/qiskit-ibm-runtime/pull/2700#discussion_r2989103837

Overall, this should save a couple of minutes of CI time, mostly by not paying the cost of installing all dependencies.


qiskit_ibm_runtime/utils/options.py
```diff
@@ -14,14 +14,10 @@
 
 from dataclasses import asdict, dataclass
 from typing import Any
-from packaging.version import Version
```
diego-plan9, line 17: These changes take advantage of the package requiring Qiskit 2+ in order to avoid the dependency on `packaging`, and remove if from the _main_ dependencies.

.github/workflows/ci.yml
```diff
@@ -98,7 +97,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e '.[dev]'
+          pip install -e '.[test,performance,visualization]'
```
TsafrirA, line 100: Why do we need the editable flag?

qiskit_ibm_runtime/utils/options.py
SamFerracin: Because qiskit 1 is no longer supported, eh?

.github/workflows/ci.yml
diego-plan9: Consistency (with the rest of entries where we install the package in CI so we always the same resulted installation; and arguably with project historical practices), and convenience (it avoids the building of the package and roughly adds the root of the repo to `PYTHONPATH`, which simplifies the final result).

qiskit_ibm_runtime/utils/options.py
diego-plan9: Yes, technically we have been able to remove them since around #2543.

.github/workflows/ci.yml
TsafrirA: Thanks! I assumed that `-e` makes things more complicated and not the other way around.

SamFerracin (approved):

Issue #2423, opened 2025-09-26
Performance regression of `QiskitRuntimeService()` in 0.42.0


t-imamichi:
**Describe the bug**

The initialization `QiskitRuntimeService()` becomes noticeably slower in 0.42.0 compared to 0.41.1.
I'm not sure whether this change is intentional.

**Steps to reproduce**

```python
from timeit import timeit
from qiskit_ibm_runtime import QiskitRuntimeService, __version__

print(f"runtime_version={__version__}")
print(f"QiskitRuntimeService(): {timeit(lambda: QiskitRuntimeService(), number=1)} sec")
service = QiskitRuntimeService()
print(f"backends(): {timeit(lambda: service.backends(), number=1)} sec")
```

```
runtime_version=0.42.0
QiskitRuntimeService(): 2.605221999983769 sec
backends(): 4.723045916995034 sec
```

```
runtime_version=0.41.1
QiskitRuntimeService(): 0.0004281250003259629 sec
backends(): 4.443328083987581 sec
```

**Expected behavior**

0.42.0 is as fast as 0.41.1

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.42.0
- **Python version**: 3.13.6
- **Operating system**: macOS 15.7

t-imamichi:
Thank your for the information. It makes sense to validate token at the initialization.

t-imamichi:
I close this because it is an intentional behavior.

Issue #2213, opened 2025-04-11
[Runtime 1.0] Investigate how to achieve faster response times


ElePT:
**What is the expected feature or enhancement?**
There are several directions that can be explored to speed up the response time, including:

- [ ] Providing alternative endpoints for querying more limited data sets 
- [ ] Adding response caching (especially for job results and other things which are mostly static)

Note: There have been multiple reports of a slow service initialization step, this seems to be faster with the new cloud API. Is there room for improvement?

**Acceptance criteria**

kt474:
> Note: There have been multiple reports of a slow service initialization step, this seems to be faster with the new cloud API. Is there room for improvement?

Based on heavy use over the last few months, a cloud account initialization takes about 1.5 seconds while an IQP (classic) account takes about 8. Maybe we could look into some benchmarking suggested in https://github.com/Qiskit/qiskit-ibm-runtime/issues/1286

Issue #1286, opened 2023-12-19
Automated Benchmarking for Provider Creation Time


drew-distefano:
**What is the expected feature or enhancement?**

Write an automated benchmarking test to measure provider creation time. Creation time is currently at 7-8 seconds (for `qiskit-ibm-provider`) which is below the optimal 2-3 second creation time defined by @blakejohnson in [Qiskit/qiskit-ibm-provider/issues/555](https://github.com/Qiskit/qiskit-ibm-provider/issues/555#issuecomment-1836269725), so we want to introduce automation to ensure we notice regressions.

**Acceptance criteria**

1. Tool accurately measures provider creation time
2. Fail test if creation time falls below a certain threshold
3. Integrate with current cloud staging tests that are run prior to release

drew-distefano:
@blakejohnson per your suggestion. Please let me know if you have any edits.

blakejohnson:
> Fail test if creation time falls below a certain threshold

Thanks for opening this. If you are defining pass/fail criteria, then you would fail if you **exceed** a threshold.

However, what I really want is something that records the outcome of the benchmark so we can track this metric over time. You might consider using ASV to produce a [dashboard like what we have for Qiskit](https://qiskit.github.io/qiskit/#/). Then it isn't so much about rejecting changes the fail the test once, but to observe that we are actively driving this time lower.

drew-distefano:
@blakejohnson Thanks for the feedback. I updated the second acceptance criteria to reflect the goal of tracking over time.

PR #2708, opened 2026-03-28
Setup minimal benchmarking facilities

diego-plan9:
### Summary

Introduce two `pytest-benchmark` benchmarks, for `import` time and for instantiating the arguably main class of the package (`QiskitRuntimeService`). 

This is intended just for developers' needs at this stage, providing a convenient way for comparing the amount of time to run use cases that are significant. In order to provide some baseline, the benchmarks are run in the daily workflow that runs the integration tests, so we can start gathering some data and better tackle some performance-related issues - and at the same time, explore if this relatively simple approach can be expanded later.

### Details and comments

Related to #1286 and related discussions (such as #2213).

No attempt at storing benchmark results or providing a consistent environment is part of this PR (some alternatives such as [this action](https://github.com/benchmark-action/github-action-benchmark) if the GitHub runners provide a consistent environment, which is to be explored). The benchmarks included are specifically for #2423, #1286, and evaluating the impact of #2563 (which might uncover additional work for reducing import time).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,9 @@ integration-test:
 smoke-test:
 	pytest test/smoke
 
+benchmark:
+	pytest test/benchmark
+
 docs-test:
 	./test/docs/vale.sh
 
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@
 # that they have been altered from the originals.
 
 
-.PHONY: unit-test integration-test smoke-test docs-test unit-test-coverage
+.PHONY: unit-test integration-test smoke-test benchmark docs-test unit-test-coverage
 
 unit-test:
 	pytest test/unit
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ smoke-test:
 	pytest test/smoke
 
 benchmark:
-	pytest test/benchmark
+	pytest test/benchmarks
 
 docs-test:
 	./test/docs/vale.sh
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -50,7 +50,39 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -e '.[dev]'
-      - name: Run benchmarks
-        run: make benchmark
       - name: Run integration tests
         run: make integration-test
+  benchmarks:
+    if: github.event_name == 'workflow_dispatch' || github.repository_owner == 'Qiskit'
+    name: Run benchmarks - ${{ matrix.environment }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      # avoid cancellation of in-progress jobs if any matrix job fails
+      fail-fast: false
+      matrix:
+        python-version: [ '3.10' ]
+        os: [ "ubuntu-latest" ]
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -22,6 +22,9 @@ integration-test:
 smoke-test:
 	pytest test/smoke
 
+benchmark:
+	pytest test/benchmark
```
TsafrirA:
```suggestion
	pytest test/benchmarks
```

TsafrirA (commented):
There is a typo in the make command path.

If we don't store the results, what's the point of running the benchmarks regularly?

Makefile
diego-plan9: Thanks!

diego-plan9:
> If we don't store the results, what's the point of running the benchmarks regularly?

Even if the output file is not stored, the output (containing the stats, similar to the first screenshot [here](https://pytest-benchmark.readthedocs.io/en/latest/#screenshots) ) will be present in the logs for the action. This allows for a degree of manual comparison and to collect a small baseline for deciding on next steps.

TsafrirA (commented):
Thanks for the clarification @diego-plan9!

Overall, I think this is a really good start. I am a little scared to have `command = [sys.executable, "-c", cmd]` in our codebase, even though this is internal. 

I am also a little worried about the consistency and meaningfulness of these tests. But I guess those are things we can play with and learn along the way. For example:

- The import test measures not only the import time but also the interpreter spawning time.
- How susceptible to caching and cold\warm starts?

diego-plan9:
Thanks - indeed, the last question is an example of a question to be answered as this approach evolves.

TsafrirA (dismissed):
LGTM

Makefile
```diff
@@ -22,6 +22,9 @@ integration-test:
 smoke-test:
 	pytest test/smoke
 
+benchmark:
+	pytest test/benchmarks
+
```
rainerschoe, line 27: What are we using this Makefile for?

I recently read a bit about `uv`'s venv management. maybe we could in the future embrace this a bit more to get more easily reproducible test runs both in CI and on dev setups.

rainerschoe (dismissed):
Thanks. This is an interesting starting point.

I think it is important to run this in GHA for now not for documenting the benchmark, but mostly to prove that the benchmark infrastructure still works.

TsafrirA (dismissed):
LGTM

Makefile
diego-plan9: Historically, mostly as a convenience for invoking commands relevants to developers and CI. Since https://github.com/Qiskit/qiskit-ibm-runtime/issues/2698, it only contains targets relevant for running the tests (which are a bit thin since it mostly changes the `pytest` test folder), and an extra for testing the docs (which is also a single command too).

I don't see a direct relevance to `uv` vs `pip`, but evaluating the need of this file is likely a good thing.

diego-plan9: (This actually reminded me to keep the file up to date via [32899d7](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2708/commits/32899d77fe0b4488fac98d757617fa76767f0b64))

TsafrirA (approved):
LGTM

Issue #2475, opened 2025-11-13
Add fake backends for the new QPUs


jyu00:
Many of the new backends don't have their fake counterparts - e.g. kingston, pittsburgh, boston, etc. Note that we should avoid adding fake backends for dedicated systems.

Issue #1796, opened 2024-07-09
Allow retrieving all backend data


jyu00:
**What is the expected feature or enhancement?**

Today on IQP website, one can download the data for all backends, even if they don't have access to actually run jobs on that backend. In `qiskit-ibm-runtime`, however, an exception is raised if I try to instantiate a backend (to get its data) that my instance doesn't have access to.

**Acceptance criteria**

One can instantiate a backend object even if their instance doesn't have access to it. An exception should only be raised if they try to run jobs on that.

jyu00:
The `backends()` should include a new flag to indicate if people want just backends they can access (default), or all public backends.

kt474:
After looking into this I don't think this currently possible. In order to instantiate a backend ee need the backend configuration and users don't have access to `/backend/{id}/configuration` if they don't have access to the backend. 

I thought we could use https://api.quantum.ibm.com/api/users/backends but this endpoint does not give the configuration

PR #2711, opened 2026-03-30
Add script to fetch conf and props of real backends + Add FakeKingston

### LINKED ISSUE #2710: FakeKingston is missing
**Describe the bug**
IBM Kingston was made available to all Open Plan users two weeks ago:
https://quantum.cloud.ibm.com/announcements/en/product-updates/2026-03-16-open-plan-news

However, the corresponding snapshot is missing here in the fake backends.

- **qiskit-ibm-runtime version**: 0.46.1
- **Python version**: 3.14.3

iyanmv:
### Summary
This adds a script to fetch the configuration and properties of a real backend and save them as the JSON files in the right path. The script is then used to obtain a snapshot of IBM Kingston to current fake backends.

Script was based on [this gist](https://gist.github.com/HuangJunye/fdc4c855e9260515f6a5450377ae19c4) by @HuangJunye with some minor modifications and fixes.

### Details and comments
Fixes #2710
Partially addresses #2475


diego-plan9:
Thanks: let's give some courtesy time for #2475 to land before deciding on merging this PR.

iyanmv:
> Thanks: let's give some courtesy time for #2475 to land before deciding on merging this PR.

Oh, I missed that issue. Yeah, sure, let's wait ;)

iyanmv:
Btw, if @HuangJunye is busy with other tasks, I can clean a bit and include in this PR the script that I used to fetch the configuration and properties.

HuangJunye:
@diego-plan9 Sorry for the delay in getting back. Unfortunately I don't have time to work on this at the moment.

@iyanmv Thanks for contributing! I have created [a script](https://gist.github.com/HuangJunye/fdc4c855e9260515f6a5450377ae19c4) to pull the config and properties as well in case you want to refer to it. Do you mind adding other fake backends and the script to this PR as well? Thanks a lot!

diego-plan9:
Thanks @HuangJunye ! It is fine.

@iyanmv, if you still have the interest and bandwidth, happy to see this PR include a script (yours, Junye's, or a combination of both) and extend the scope so it also closes #2475!

iyanmv:
> @iyanmv, if you still have the interest and bandwidth, happy to see this PR include a script (yours, Junye's, or a combination of both) and extend the scope so it also closes #2475!

Alright, yes, let me have a look at the script and I will put everything together. I don't think my PR will be able to close #2475 completely since I only have access to the Open Plan's backends. But you can add some commits for the missing backends in this PR, or maybe do it after this is merged with the script.

iyanmv:
I cannot reproduce locally the complains from pre-commit with `uv run pre-commit run --all-files`. Perhaps something changed while the PR was open. I'll rebase from the current main.

diego-plan9:
> I cannot reproduce locally the complains from pre-commit with `uv run pre-commit run --all-files`. Perhaps something changed while the PR was open. I'll rebase from the current main.

They seem to all come from #2726 - it did tweak `ruff` rules (and minor `pre-commit` hooks too).

iyanmv:
I mean... I fetch the latest main, run the things locally and I got different complains than the CI :sweat_smile: Don't know what's going on Exactly that weird line with the docstring that CI is complaining now I had to fix to have the ending `"""` in a new line

iyanmv:
Looks happy now :)

iyanmv:
A different topic but perhaps interesting to discuss here a bit: would it make sense to relax a bit the restrictions to fetch information about backends? In the website, one can see the backends that one does not have access. I think it should be similar with the python API. One cannot submit jobs to a backend that is not included on that account, but it should be okay to fetch configuration, properties, noise models, etc. What do you think?

diego-plan9:
> Don't know what's going on

Me neither - if you are locally working off a branch that contains #2726, `pre-commit` should pick up the new rules 🤔 . I would suggest checking that your local contents of `pyproject.toml` and `.pre-commit-config.yaml` match, or explore if it also happens in a fresh clone of `main`, as I cannot imagine an scenario where `pre-commit run --all-files` _does_ get executed and runs `ruff`, but only enforces the pre-2726 rules. We are not using `uv` in CI or locally, which might be a source of differing behavior? If you find out, please ping back, specially if this might be affecting other users.

> A different topic but perhaps interesting to discuss here a bit: would it make sense to relax a bit the restrictions to fetch information about backends? In the website, one can see the backends that one does not have access. I think it should be similar with the python API. One cannot submit jobs to a backend that is not included on that account, but it should be okay to fetch configuration, properties, noise models, etc. What do you think?

I think this is actually #1796? It is a sensible request, mostly needs prioritization - let's follow up conversation there if needed.

tools/fetch_config_properties_backend.py
```diff
@@ -0,0 +1,81 @@
...
+    The configuration and properties and then saved as valid JSON files in the backends directory.
+    """
+    try:
+        service = QiskitRuntimeService()
+    except AccountNotFoundError:
+        print("This script requires a saved IBM account")
+        sys.exit(1)
+
+    backend_city = backend_name.split("_")[-1]
+    use_fractional_gates = False
```
diego-plan9: Not sure if this is a byproduct: is the intent to make this configurable via an arg (or other means)? If not, can the variable be removed?

tools/fetch_config_properties_backend.py
```diff
@@ -0,0 +1,81 @@
...
+tools_dir = Path(__file__).parent
+backends_dir = tools_dir.parent / "qiskit_ibm_runtime" / "fake_provider" / "backends"
+
+parser = argparse.ArgumentParser(description="Retrieve latest config and properties from a backend")
+parser.add_argument(
+    "-b",
+    "--backend",
+    required=True,
+    help="backend name (required), e.g. 'ibm_boston'",
+)
```
diego-plan9, line 41: As a small improvement, would you consider using `append` or a similar approach in this argument for allowing the user to specify multiple backends in the same invocation (by repeating the argument)? It would be quite convenient for refreshing / creating several backends at once, and it seems like a small change.

tools/fetch_config_properties_backend.py
```diff
@@ -0,0 +1,81 @@
...
+
+def main(backend_name: str) -> None:
+    """Retrieves latest configuration and properties from a backend.
+
+    The configuration and properties and then saved as valid JSON files in the backends directory.
+    """
+    try:
+        service = QiskitRuntimeService()
+    except AccountNotFoundError:
+        print("This script requires a saved IBM account")
```
diego-plan9: Can you catch `AccountsError` (a bit more generic) and print the exception, along with a slightly more generic text ("Unable to create service"-ish)? A bit of an edge case in normal circumstances, but this way we ensure that other kind of user errors with credentials don't mislead the user of the script.

diego-plan9 (commented):
Thanks @iyanmv ! I put the script to use by testing with `ibm_kingston`, and it's looking great. Some minor suggestions, and I would like to get @HuangJunye to take a look as well - and we should be good to go, it's very nice to have both the new fake backend and a programmatic way for retrieving further backends in place.

As a small ask: would you be fine not rebasing force-pushing now that the review has started, for easier time following up conversations pointing to lines of code in the interface?

tools/fetch_config_properties_backend.py
iyanmv: Sure, we can add another argument, and perhaps with a default value? Should it be `False` or `True`?

tools/fetch_config_properties_backend.py
iyanmv: Sure, I can do that.

tools/fetch_config_properties_backend.py
iyanmv: Okay, thinking a bit more, I will remove it since we want the control flow gates in the fake backend, the default value is `False`, and it will probably be removed in the future from the `backend(s)` method anyways.

diego-plan9: Removing sounds like a sensible choice (and if it is needed, we can introduce it later).

tools/fetch_config_properties_backend.py
diego-plan9: Thanks! And the `verbose` argument is nice to have too.

diego-plan9 (commented):
Thanks for the changes! It's looking good to me - pending a final review from Junye.

HuangJunye (approved):
This is great. I tested with ibm_boston and it worked perfectly. Thanks a lot for working it. I will open another PR to add the missing backends since you don't have access to them.

diego-plan9 (approved):

PR #2713, opened 2026-04-01
Fix integration tests and benchmarks

diego-plan9:
### Summary

Fixes for a couple of oversights from #2607 and #2708:
* pass the `url` to one of the proxy tests, as otherwise the staging run would fail due to using production's IAM
* revise the skipping of the benchmarks (reusing some decorator pieces), to ensure they are not always skipped in CI, as well as using seconds always in the output and take a bit of time off conforming to #2707

### Details and comments
Fixes N/A

### DIFF
--- a/test/benchmarks/test_benchmarks.py
+++ b/test/benchmarks/test_benchmarks.py
@@ -18,7 +18,8 @@ import sys
 from unittest import SkipTest
 
 from qiskit_ibm_runtime import QiskitRuntimeService
-from qiskit_ibm_runtime.accounts import AccountManager
+
+from ..decorators import _get_integration_test_config
 
 
 def run_in_subprocess(cmd: str) -> None:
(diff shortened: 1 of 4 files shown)


test/benchmarks/test_benchmarks.py
```diff
@@ -40,8 +41,17 @@ def test_import_qiskit_ibm_runtime(benchmark):
 
 def test_instantiate_qiskit_runtime_service(benchmark):
     """Benchmark the instantating of `QiskitRuntimeService`."""
-    account_manager = AccountManager()
-    if len(account_manager.list()) == 0:
+
+    channel, token, url, instance, _ = _get_integration_test_config()
+    if not all([channel, token, url]):
```
diego-plan9: The plan was to reuse the `integration_test_setup` [decorator](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/decorators.py#L82), but conflicts a bit with the usage of the `benchmark` fixture, so opted for this more direct solution.

rainerschoe: Makes sense.
Should we then remove the `_` prefix from `_get_integration_test_config()`, since we are using it outside of the file?

diego-plan9: Yes, does not hurt - [e304da7](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2713/commits/e304da73fc418755e9a84db1be18b3c69d9408cc)

rainerschoe (approved):
LGTM. Thanks for the explanations.

PR #2715, opened 2026-04-09
Bump actions to Node.js 24 supported actions

diego-plan9:
### Summary

For the past weeks, we have been having some warnings in the Actions, such as
https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/24174231141

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: [...]

This bumps the versions to more recent versions that don't require additional changes.

### Details and comments

Fixes N/A


rainerschoe (approved):
LGTM, thanks.
If the workflows are greed it should be fine :-)

PR #2717, opened 2026-04-09
Add UTC timezone info to executor result timing

ihincks:
### Summary

This PR fixes a shortcoming of the 0.1, 0.2 schemas by adding timezone information to the converted models to mark them as being UTC (which is what we promise). Without this information, most tooling will assume they're in the local timezone which is incorrect.

### Details and comments

image


diego-plan9 (approved):
Thanks Ian, good catch!

PR #2720, opened 2026-04-10
Improve executor.options validation configuration

ihincks:
### Summary

image

This did not solve the issue, but I think it's still worth merging since it improves the options experience of validation in executor options.

Both of the new tests fail without the changes to the dataclass config.

### Details and comments

This PR also fixes some minor docstring spacing issues.


diego-plan9:
> I'm writing here because I can't figure out how to get it to not auto-close this issue.

I wonder if editing so it does not have `closes #xxx` in the description would work? Maybe GitHub is not taking the strike-through into account. Could you also edit the title to reflect the current scope?

diego-plan9 (approved):
Thanks! While I think `pydantic` default approach is sensible in most use cases (ie. ignoring extra fields if present), I agree it makes sense to enforce things a bit more in this particular case and be more protective (and explicitly re-validate on changes), being an user-facing class.

PR #2725, opened 2026-04-13
Add logging instructions to Sampler

### LINKED ISSUE #2685: Log progress in wrapper sampler

Since wrapper sampler does more client-side processing, there was a concern that people would think it's stuck when `sampler.run` doesn't return immediately. One proposal is to log progress information - e.g. "identifying unique layers", "inserting twirling annotations", etc, to show that things are being done. 

**Acceptance criteria**

Log progress messages in `sampler.run`

diego-plan9:
### Summary

Add logging instructions to `SamplerV2`, emitted at  `INFO` level.

### Details and comments
Fixes #2685

### DIFF
--- a/qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
+++ b/qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 from collections.abc import Callable, Iterable
 import logging
 from typing import Any
+from os import getenv
 
 from qiskit.primitives.base import BaseSamplerV2
 from qiskit.primitives.containers.sampler_pub import SamplerPub, SamplerPubLike
@@ -38,6 +39,10 @@ from ..utils import validate_no_boxes, extract_shots_from_pubs, calculate_twirli
 from ..options.sampler_options import SamplerOptions
 
 logger = logging.getLogger(__name__)
+
+if not getenv("QISKIT_IBM_RUNTIME_LOG_LEVEL", ""):
+    # If the user has not specified a logging configuration, use `INFO` for this logger.
+    logger.setLevel(logging.INFO)
 
 
 def prepare(
@@ -96,7 +101,8 @@ def prepare(
 
     if not twirling_enabled:
         # No twirling path: validate no boxes, create CircuitItem objects
-        for pub in pubs:
+        for i, pub in enumerate(pubs):
+            logger.info("Procesing pub %d/%d", i, len(pubs))
             validate_no_boxes(pub.circuit)
 
             # Apply DD if enabled
@@ -132,7 +138,8 @@ def prepare(
             twirling_strategy=options.twirling.strategy.replace("-", "_"),
         )
 
-        for pub in pubs:
+        for i, pub in enumerate(pubs):
+            logger.info("Procesing pub %d/%d", i, len(pubs))
             boxed_circuit = boxing_pm.run(pub.circuit)
             template_circuit, samplex = build(boxed_circuit)
 
@@ -348,8 +355,9 @@ class SamplerV2(BaseSamplerV2):
 
         # Submit to executor
         logger.info(
-            "Submitting %d pub(s) to executor with %d shots",
+            "Submitting %d pub%s to executor with %d shots",
             len(coerced_pubs),
+            "s" if len(coerced_pubs) > 1 else "",
             quantum_program.shots,
         )
 
--- a/qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
+++ b/qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
@@ -97,7 +97,7 @@ def prepare(
     if not twirling_enabled:
         # No twirling path: validate no boxes, create CircuitItem objects
         for i, pub in enumerate(pubs):
-            logger.info("Procesing pub %d/%d", i, len(pubs))
+            logger.info("Processing pub %d/%d", i, len(pubs))
             validate_no_boxes(pub.circuit)
 
             # Apply DD if enabled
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -39,6 +40,10 @@
 
 logger = logging.getLogger(__name__)
 
+if not getenv("QISKIT_IBM_RUNTIME_LOG_LEVEL", ""):
+    # If the user has not specified a logging configuration, use `INFO` for this logger.
+    logger.setLevel(logging.INFO)
```
diego-plan9: By default (as in, if using the library in a regular Python environment), users will be running with `WARNING` debug level. I am assuming that the goal of the issue is to show progress information _in the default case_.

The package's way of setting up the log level for the library is `QISKIT_IBM_RUNTIME_LOG_LEVEL` explicitly. These lines assume that if the user has set up the log level via environment variable, that user wants full control of the log level, and in that case the module's `logger` level will not be touched.

We can probably do better (via a dedicated `handler` or a similar one), but this likely requires touching the logging approach in `main`. I would like this to be tackled eventually, although outside the scope of this PR.

diego-plan9:
> I would consider adding a `progress_bar` argument to `SamplerV2.run`[...]

This PR is the low-tech approach, using just logging facilities, on purpose. I prefer to think of `qiskit-ibm-runtime` as a library first, and keep the spirit as close to it as possible (with visualization and advanced UI belonging to higher levels such as the previous qiskit jupyter widgets, or to user land). 

If we want to move into the direction of a more rich user interface, I would also prefer to introduce it in a more planned way (for example, looking at other instances where long-running operations might benefit from visual indications to the user) instead of a single occurrence for an experimental feature. Implementation-wise, I do agree with the concern (and believe that a non-invasive way of incorporating it can be achieved - with some debate over touching the `run()` signature as technically we are constrained by the base class).

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
SamFerracin: Thanks for putting this together?

> I am assuming that the goal of the issue is to show progress information in the default case.

There is no set requirement for this, we can propose and see what product has to say. Personally, I don't like using loggers unless I really really need to use them. On top of that, I assume that many jobs will be small-sized, and for them logging is not a requirement. So, my recommendation would be:
- Logging OFF by default
- Explaining (in the docstring) that "`sampler.run` may take a while because ... and if you are curious to know at what stage of `sampler.run` we're at, you can consider using the logger by ..." (of course reworded)

What do you think?

jyu00: `QISKIT_IBM_RUNTIME_LOG_LEVEL` is the convenient way to set log level, but users can also use the standard Python `setLevel`. I would respect whatever level the user sets, or use the Python default if none is set, rather than changing it to INFO.

diego-plan9: I'm 100% happy just having the user rely on `QISKIT_IBM_RUNTIME_LOG_LEVEL` or on setting the level for this logger manually, this was mostly a concession for ensuring these entries are shown if a user does not act on logging configuration. I'm all-in for empowering users to rely on standard mechanisms.

@jyu00 just for verifying, are you fine with users having to opt-in for having the logs displayed?

jyu00: > @jyu00 just for verifying, are you fine with users having to opt-in for having the logs displayed?

Yes. I think that's more user-friendly than surprising them with a non-standard configuration 🙂. We should certainly doc this expectation like Sam mentioned though.

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -96,7 +96,8 @@ def prepare(
 
     if not twirling_enabled:
         # No twirling path: validate no boxes, create CircuitItem objects
-        for pub in pubs:
+        for i, pub in enumerate(pubs):
+            logger.info("Procesing pub %d/%d", i, len(pubs))
```
```suggestion
            logger.info("Processing pub %d/%d", i, len(pubs))
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -132,7 +133,8 @@ def prepare(
             twirling_strategy=options.twirling.strategy.replace("-", "_"),
         )
 
-        for pub in pubs:
+        for i, pub in enumerate(pubs):
+            logger.info("Procesing pub %d/%d", i, len(pubs))
```
```suggestion
            logger.info("Processing pub %d/%d", i, len(pubs))
```

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -309,6 +311,13 @@ def __init__(
     def run(self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None) -> RuntimeJobV2:
         """Submit a request to the sampler primitive.
 
+        For moderate and complex workloads, the client-side processing can be resource intensive
```
```suggestion
        For moderate and complex workloads, the client-side processing done to map sampler inputs to executor inputs can be resource intensive
```

SamFerracin (commented):
Some minor comments (a few typos that I had missed before) and this is good to go!

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
diego-plan9: Oh my! Thanks for catching this

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
diego-plan9: Thanks - [1f52ea4](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2725/commits/1f52ea45cb5106e998b42e986d8bc0f096b175a6) instead of commititing suggestion, for proper wrapping.

SamFerracin (approved):

PR #2726, opened 2026-04-13
Docstrings lint (length, copyright year, style)

diego-plan9:
### Summary

Linting centered on docstrings:
* ensure the codebase respects the 100-char limit.
    * switch from `black` to `ruff` for enforcing it
* enable the `pydocstyle` ruff [rules](https://docs.astral.sh/ruff/rules/#pydocstyle-d)
* update the copyright script (credit to https://github.com/Qiskit/samplomatic/pull/313)
* moves `Args` section of `__init__` to the class-level docstring
* revise raw-string usage in docstrings (removing them, letting `ruff` re-introduce them)
* enable some extra `pre-commit` hooks

### Details and comments

AI usage: 55363f2b2b918df07d75e78316a2ef0b20e2120f has been created using `Bob` version `1.0.1`, for wrapping docstrings to 100 chars (not fully successfully).

When fixing the `pydocstyle` errors, manual modifications have been made. In general:
* tried to re-use arguments docstrings from other functions, in case of missing arguments
* best-effort for keeping the docstring first line within the char limit by omitting articles, removing sphinx syntax, and condensing.
* when the first-line summary seemed to omit important information, resorted to including the original docstring as part of the body (most notably in tests)
* `D102` will not take into account parent classes docstrings. In those cases, resorted to copy-pasting the parent class docstring (excluding args), as ignoring the rule would not allow us to catch legitimate undocumented public functions.


pyproject.toml
```diff
@@ -85,25 +85,32 @@ future-annotations = true
...
+    "RUF013", # implicit-optional
     "T10",    # flake8-debugger
     "UP",     # pyupgrade
-    "RUF013", # implicit-optional
 ]
+
 ignore = [
-    "E501",  # line-too-long: black handles line length
-    "UP032", # f-string: requires manual revision of non f-string formatting
+    "D105",   # undocumented-magic-method
```
diego-plan9, line 98: It felt a bit of overkill to document these (`__eq__`, `__repr__`, `__str__`, etc).

pyproject.toml
```diff
@@ -85,25 +85,32 @@ future-annotations = true
...
+    "D107",   # undocumented-public-init
```
diego-plan9, line 99: We had roughly half of the codebase declaring `Args:` in the class docstring, and half of it declaring them in `__init__`. Declaring them in the class docstring seems better inline with the documentation rendering, so homogenized in that direction.

pyproject.toml
```diff
@@ -85,25 +85,32 @@ future-annotations = true
...
 ]
 
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = [
+    "D205",   # missing-blank-line-after-summary - allows for sphinx headers to remain
+    "D212",   # multi-line-summary-first-line - allows for sphinx headers to remain
+    "D415",   # missing-terminal-punctuation - allows for sphinx headers to remain
```
diego-plan9, line 106: Prevent `ruff` from touching the `=====` sphinx blocks in selected locations.

qiskit_ibm_runtime/api/clients/runtime.py
```diff
@@ -217,6 +217,9 @@ def create_session(
         """Create a session.
 
         Args:
+            backend: IBM backend to use for the session.
```
yaelbh: It's confusing. For the Primitives we have a `mode` parameter, which has another meaning than here, and can take a backend (a `backend` parameter does not exist anymore for the Primitives).

For example
```python
```

yaelbh (dismissed):

qiskit_ibm_runtime/api/clients/runtime.py
diego-plan9: Thanks (for the review as a whole!).

I tweaked it slightly in 4c72eb41. This arg ends up being passed as-is to the API (https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/2025-05-01/tags/sessions). Can you suggest something that you don't find confusing, it the latest change does not seem right?

yaelbh: This change seems right.

What I'm trying to point on is that `create_session` has two parameters named `backend` and `mode`, and `SamplerV2` (and the rest of the primitives) has a parameter named `mode`. The sampler's `mode` and the `create_session`'s mode are probably different things, that happen to be given the same name by coincidence. Moreover, the sampler's mode is most often assigned a backend, which is another parameter of `create_session`.

I don't see a way to resolve this overloading of `mode` and `backend` without an API-breaking change, and anyway it's just something that I spotted when reviewing the PR, but is not related to the PR.

yaelbh (approved):

PR #2730, opened 2026-04-14
Add git blame ignore file, revise contribution guide

diego-plan9:
### Summary

As we had a number of noisy lint/style PRs, add a `.git-blame-ignore-revs` for more convenient tracing of changes in the GitHub interface (and optionally locally). Some of the commits in the list _do_ contain non-style changes (for example, updates to `pyproject.toml` as part of the `ruff` rules enabled): even if this will hide those changes in the interface, the amount of style changes contained in them is significant enough to prefer including them (putting the priority in frequent diffing over the Python files).

Additionally, this PR also fixes some lingering references to `black` that are no longer needed/accurate since #2726.

### Details and comments

Small follow-up to #2726 (taking care of references to `black` that should have been dropped during that PR)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -262,7 +262,6 @@ make benchmark
 Please submit clean code and please make effort to follow existing conventions in order to keep it
 as readable as possible. We use:
 
-* [black] as the tool for ensuring consistent code formatting
 * [ruff] as a linter providing deeper analysis for potential style and functionality issues, as
   well as conformance to good practices
 * [mypy] as a static type checker for type hinting
@@ -282,7 +281,7 @@ pre-commit run
 ```
 
 This will run the tools against your staged changes, and will automatically update your code to
-conform to the style in some instances (for the `black` and a subset of the `ruff` issues). For
+conform to the style in some instances (for example, a subset of the `ruff` issues). For
 other errors, you will have to fix the issues manually by updating your code, and re-run again.
 Please refer to the tools' documentation for more information on running them manually and
 customizing your workflow if needed.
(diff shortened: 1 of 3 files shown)


yaelbh:
As a user of `git blame`, I prefer without `.git-blame-ignore-revs`:
- I perceive the hindering of non-style changes as a serious concern.
- When I run `git blame` (I do it from terminal - I'm not familiar with the UI) I easily identify that the shown commit is not what I'm looking for. Then I run `git blame <commit hash>^`. In fact I usually have to perform several such iterations until I find the commit (and the person haha) that's really to blame, skipping code refactoring and small edits. In short, the new file is not going to solve a problem that bothers me much, because I know how to bypass it easily, and will probably not get the correct commit immediately anyway.

Let's see the opinion of other major developers of the repo.

diego-plan9:
> I perceive the hindering of non-style changes as a serious concern.

Please elaborate. I don't understand if the concern is about non-style changes included in the current list, or a different concern.

> When I run git blame [...]

The file will not be used locally unless you opt-in (see https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view): this is mostly for GitHub interface diffing, and developers are in control whether to use it or not locally (so if I'm reading it correctly, your scenario does not need to change unless you want to. You can also [bypass the ignores in the GitHub interface](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#bypassing-git-blame-ignore-revs-in-the-blame-view) if needed.

yaelbh:
That's encouraging. So I remove my concern, and there's no need to elaborate about it... I'll review the changes.

CONTRIBUTING.md
```diff
@@ -298,6 +297,15 @@ make docs-test
...
 
+### Using the `git blame` ignored revisions list
+
+This repository contains a list of commits for `git blame` to ignore: commits that are style or
+linting changes that affect formatting, and thus tend to be disruptive when comparing changes in
+the repository history. When issuing a PR that contains a significant amount of style changes,
+please update the `.git-blame-ignore-revs` file adding an entry to the end of the file, and a
+comment pointing to the pull request. Please check the [Ignore commits in the blame view]
+GitHub documentation for instructions on using the ignore list locally and for general information.
+
```
yaelbh, line 308: Very clear!

yaelbh (approved):

CONTRIBUTING.md
TsafrirA: I don't have a strong opinion about this PR, but I think it's a little overly optimistic to think people (including us) will remember to follow this guideline.

diego-plan9: It is something to take care of by reviewers too, same as other guidelines we can't automate. The number of such commits should not be as high as these couple of months, and it is something that is easy to address retroactively (it needs to be a separate commit anyway, as we don't know the hash until a PR is merged).

PR #2731, opened 2026-04-14
For Executor jobs, make job.inputs return quantum programs 

### LINKED ISSUE #2728: For Executor jobs, `job.inputs` should return quantum programs

Currently, for an Executor job, `job.inputs["quantum_program"]` returns a `dict`:
```python
# an executor job
job = service.job(<job_id>)

print(job.inputs["quantum_program"])
-----
{'items': [{'shape': [2, 4], 'chunk_size': 'auto', 'circuit': {'circuit_b64': 'UUlTS0lUEAIDAAA ...}
```
This dict can be used to instantiate a `QuantumProgram` object, but this requires users to:
- Understand the relationship between qiskit-ibm-runtime and schemas
- Import and use the `ParamModel_X_Y` classes and the `quantum_program_from_X_Y` function to convert the dict to a quantum program. 

As a user of executor, I would like `job.inputs` to simply return a quantum program, as in the snippet below:
```python
# an executor job
job = service.job(<job_id>)

print(job.inputs["quantum_program"])
-----
QuantumProgram(shots=100, items=[QuantumProgramSamplexItem(...), ....]
```

**Acceptance criteria**
- [ ] For executor jobs, `job.inputs` returns `QuantumProgram` objects.

SamFerracin:
Closes #2728 

Changes the behavior of `job.inputs` for executor jobs as shown in the following snippet:
```python
# an executor job
job = service.job(<job_id>)

print(job.inputs["quantum_program"])
-----
QuantumProgram(shots=100, items=[QuantumProgramSamplexItem(...), ....]
```


TsafrirA (commented):
I think this is a bit better than the alternative. Keeps things separated in a more consistent way.

qiskit_ibm_runtime/utils/json.py
```diff
@@ -431,6 +434,26 @@ def __init__(self, *args: Any, **kwargs: Any):
             kwargs.pop("encoding")
         super().__init__(object_hook=self.object_hook, *args, **kwargs)
 
+    def decode(self, s):  # type: ignore[no-untyped-def]
+        """Overrides the built-in method."""
```
diego-plan9:
```suggestion
    def decode(self, s: str) -> Any:  # type: ignore[override]
        """Return the Python representation of a ``str`` instance containing a JSON document.

        Applies additional conversion for ``executor`` program ``params``, preserving the
        superclass `decode()` output in all other cases.

        Args:
            s: a string containing a JSON document.
        """
```

Mostly for avoiding a full type ignore, and clarifying intent. The remaining ignore is due to the [docs](https://docs.python.org/3/library/json.html#json.JSONDecoder.decode) actually declaring just `s` as an argument, but the implementation containing an extra `_w` parameter (with a default). The implementation of `loads()` seems to not pass the `w`parameter, so it seems safe to consider that just `s` is the actual interface.

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
```diff
@@ -0,0 +1,80 @@
...
+    "v0.1": (ParamsModel_0_1, quantum_program_from_0_1, quantum_program_to_0_1),
+    "v0.2": (ParamsModel_0_2, quantum_program_from_0_2, quantum_program_to_0_2),
+}
+
+
+class QuantumProgramParamsConverter:
+    """Converter to/from schema model for the inputs of executor."""
+
+    @classmethod
+    def get_converters(cls, schema_version: str):  # type: ignore[no-untyped-def]
```
diego-plan9:
```suggestion
    def get_converters(cls, schema_version: str) -> tuple[BaseParamsModel, Callable, Callable]:
```

Small bit also for avoiding a type ignore. This also requires `from collections.abc import Callable`.

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
```diff
@@ -0,0 +1,80 @@
...
+from .quantum_program import QuantumProgram
+from ..options import ExecutorOptions
+
+AVAILABLE_CONVERTERS = {
+    "v0.1": (ParamsModel_0_1, quantum_program_from_0_1, quantum_program_to_0_1),
+    "v0.2": (ParamsModel_0_2, quantum_program_from_0_2, quantum_program_to_0_2),
+}
+
+
+class QuantumProgramParamsConverter:
```
diego-plan9: Minor and more of an implementation choice discussion, but I wonder if we really get many advantages by providing a singleton-ish `QuantumProgramParamsConverter` compared to exposing the "mapping" directly, such as:

```python
class ConverterThingie(NamedTuple):
    model: Type[BaseParamsModel]
    encoder: Callable
    decoder: Callable

AVAILABLE_CONVERTERS = {
    "v0.1": ConverterThingie(ParamsModel_0_1, quantum_program_from_0_1, quantum_program_to_0_1),
    "v0.2": ConverterThingie(ParamsModel_0_2, quantum_program_from_0_2, quantum_program_to_0_2),
}

# Consumers code in other file.
AVAILABLE_CONVERTERS[schema_version].decoder(...)
```

I see some usability benefits of introducing `QuantumProgramParamsConverter` (detection of invalid `schema_version` and exception massaging, a bit more centralized interface, probably eases up scaling), although ultimately it feels like the goal is to provide a mapping and the user still needs to take care of the Exceptions being raised, and the benefits feel a bit diluted by having just 2 converters and 2 usages of `QuantumProgramParamsConverter` at the moment. All in all, I'm wondering it the extra complexity/indirection pays off (comparing to just having users access the dict naturally).

I also imagine this is inspired by `QuantumProgramResultDecoder` (which has an extra constraint of being a subclass of `ResultDecoder` which does change the scenario from my point of view). Actually I also wonder if (longer-term!) it might be worth thinking about reconciling the bits about "how to provide mapping to version" somehow (as in centralize in a module, and/or bubble up, specially in light of noise learner too).

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
```diff
@@ -0,0 +1,80 @@
...
+
+from .converters import quantum_program_from_0_1, quantum_program_to_0_1
+from .converters import quantum_program_from_0_2, quantum_program_to_0_2
+from .quantum_program import QuantumProgram
+from ..options import ExecutorOptions
+
+AVAILABLE_CONVERTERS = {
+    "v0.1": (ParamsModel_0_1, quantum_program_from_0_1, quantum_program_to_0_1),
+    "v0.2": (ParamsModel_0_2, quantum_program_from_0_2, quantum_program_to_0_2),
+}
```
diego-plan9: Can you document and type hint this too?

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
```diff
@@ -0,0 +1,80 @@
...
+
+from ibm_quantum_schemas.common import BaseParamsModel
+from ibm_quantum_schemas.executor.version_0_1 import ParamsModel as ParamsModel_0_1
+from ibm_quantum_schemas.executor.version_0_2 import ParamsModel as ParamsModel_0_2
+
+
+from .converters import quantum_program_from_0_1, quantum_program_to_0_1
+from .converters import quantum_program_from_0_2, quantum_program_to_0_2
+from .quantum_program import QuantumProgram
+from ..options import ExecutorOptions
```
diego-plan9: Can these be put into a `if TYPE_CHECKING:` block? If so, for alleviating a bit the circular imports (and making it easier to move the code around).

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
```diff
@@ -0,0 +1,80 @@
+# This code is part of Qiskit.
```
diego-plan9: (On filename and location)

With this change we have `.quantum_program.converters.*` ... and `.quantum_program.quantum_program_params_converters` . Can we:
* unify/simplify things?
* avoid such a long module name?

Without looking at the implications it would have regarding circular imports, some options might be:
* end up in `.quantum.programs.converter` (if the suggestion of just using `AVAILABLE_CONVERTER` goes through, can be placed in the `__init__.py` directly)
* come up with a good module name (`conversion`? `tranformers`? abuse `utils`? `mappings`?) and keep it at `.quantum_program` level
* the above, and take the chance to merge `quantum_program_decoders.py` content into this too, since both are related and quite small

I see this module as a solution "how to provide mapping", so I have a preference for 2 (optionally +3), from the point of view that the maps are a feature of `QuantumProgram` (and it makes sense to keep them high in the hierarchy for easier usage from other places).

diego-plan9 (commented):
Thanks for trying this out, @SamFerracin !

I agree with the benefits and drawbacks. For me, the main benefit of this approach is that we are consistent with where the responsibility for converting raw `dict`s into domain-rich `dict`s stays the same, and we don't diverge. Whether the current place (ie. `api.rest.program_job.ProgramJob.get`) is the proper place is a different debate: the edge I see about this approach is such consistency.

I also agree the detection of the right conditions for doing the conversion is a bit tacky - but I share that concern also with the other approach to an extent. I see the root cause as that  neither `api.rest.program_job.ProgramJob.get` nor `BaseJob` know that kind of job they are dealing with by design, and I don't think we can fully solve it without a deeper re-design. It does feel a bit more natural in `BaseJob` (from the point of view that is closer to the concept of a `Job`), but not too natural either (from the point of view that should be agnostic to the kind of job).

If going with `RuntimeDecoder`, we should be super cautious that the extra conversion only takes place in the right circumstances, and fails gracefully (and I think we are being cautious with the current state in the PR). `ResultDecoder` is the only place where it used explicitly, and it does not seem to be affected.

TL;DR: I'm actually good with moving either PR forward, despite the bunch of comments. All in all, they are internal implementation details and decisions, and can be tweaked later: thanks for putting the feature in place!

qiskit_ibm_runtime/utils/json.py
SamFerracin: thanks. yeah i've been sloppy on typehints (didn't want to do the full work until we agreed), will fix them now

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
SamFerracin: https://github.com/Qiskit/qiskit-ibm-runtime/pull/2731/changes/5b1ee7b271cd0a952558b0fab06bda969c43ecef

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
SamFerracin: Thanks. I have removed the class, we can introduce it later if we see a need: https://github.com/Qiskit/qiskit-ibm-runtime/pull/2731/changes/ad289b633e89495919db3d6636ea536a80b4c7a2

> I also imagine this is inspired by QuantumProgramResultDecoder

yes, and i've been thinking hard about how we could just provide a `QuantumProgramConverter` class, but as you say, until this are `ResultDecoder`s, there is nothing to do. Long term (surely before 1.0), this is something that we should revisit

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
SamFerracin: [b8b5f0d](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2731/commits/b8b5f0d6f2bb1ffab753470af96dfa60101075cf)

qiskit_ibm_runtime/quantum_program/quantum_program_params_converters.py
SamFerracin: I've tried placing everything in the same file, but ended up stuck with circular dependencies. I did take a step towards avoiding long module names (renamed modules to remove unnecessary prefixes: [c8638b1](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2731/commits/c8638b19b09fbd5a0845777467b98ced5ece768e)), but i'd leave the other part ("unify things") for when we have more time, if you're okay with it

diego-plan9: Thanks! And ...

> I've tried placing everything in the same file, but ended up stuck with circular dependencies

Sounds like a reasonable argument (an unfortunate one, but very understandable, we are not in a good shape in that regard). Revisiting the unifying later sounds good!

qiskit_ibm_runtime/utils/json.py
```diff
@@ -431,6 +434,34 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+                try:
+                    converter = QUANTUM_PROGRAM_PARAMS_CONVERTERS[params["schema_version"]]
+                    quantum_program, options = converter.decoder(converter.model(**params))
+                    decoded["params"]["quantum_program"] = quantum_program
+                    decoded["params"]["options"] = options
+                except Exception:
+                    warnings.warn(
+                        "Unable to convert executor 'params' to a pair of quantum program and "
+                        "options."
+                    )
```
diego-plan9: Minor: would it be worth having the exception as part of the warning or logged? I imagine this should be a rare case - and if so, it makes sense to make it visible instead of silently swallowing it (so it allows us to debug better).

diego-plan9 (dismissed):
Thanks Sam!

qiskit_ibm_runtime/utils/json.py
SamFerracin: [e33e4f1](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2731/commits/e33e4f1bd7a39b17a53590ac9a99852f2457772c)

diego-plan9 (approved):

PR #2733, opened 2026-04-16
Docstrings lint for sampler-preview branch

diego-plan9:
### Summary

Apply linting to docstrings, using the same rules as #2726, over the files that change between current `main` (74b7cc33a488b3760f0e1d18ca02f37ae5616ecf) and current `sampler-preview`.

This is mostly for easier merging from `main` to `sampler-preview` (so the `pre-commit` after the merge passes).

### Details and comments

Same rationale as in #2726 applied when manual modifying of docstring was needed (most notably, liberal removing of sphinx formatting for line-length purposes).


test/unit/executor/routines/sampler_v2/post_processors/v0_1/test_flatten_twirling_axes.py
```diff
@@ -1,148 +1,149 @@
-# This code is part of Qiskit.
```
diego-plan9, line 1: These are due to CRLF/LF endings, this file seemed to be with CRLF.

qiskit_ibm_runtime/executor/routines/options/sampler_options.py
```diff
@@ -51,7 +51,8 @@ class SamplerExecutionOptions(ExecutionOptions):
     def to_executor_options(self) -> ExecutionOptions:
         """Convert to execution options.
 
-        This drops the `meas_type` field, which is passed as part of the QuantumProgram."""
+        This drops the `meas_type` field, which is passed as part of the QuantumProgram.
```
TsafrirA:
```suggestion
        This drops the ``meas_type`` field, which is passed as part of the :class:`~.QuantumProgramResult`.
```

TsafrirA (dismissed):
All in all, this looks good to me. After an offline discussion I realized the references were carefully placed, despite looking a little inconsistent. I wonder, however, if such a level of fine-tuning is to be expected later on, or should we agree on a simpler set of guidelines for developers to follow ("always add references in the body of the doc string, but never in the first line"?)

qiskit_ibm_runtime/executor/routines/options/sampler_options.py
diego-plan9: Thanks!

diego-plan9:
> All in all, this looks good to me. After an offline discussion I realized the references were carefully placed, despite looking a little inconsistent. I wonder, however, if such a level of fine-tuning is to be expected later on, or should we agree on a simpler set of guidelines for developers to follow ("always add references in the body of the doc string, but never in the first line"?)

Thanks! I like the effort put on references (they are very useful for consumers), and I would personally love to see them more frequently used, and ending up in a guideline.

The implied guideline with #2726 became "Ensure the first line of the docstring is a single line, and fits into the 100 char limit" (with references or not).

qiskit_ibm_runtime/executor/routines/sampler_v2/post_processors/v0_1/post_processor.py
```diff
@@ -26,7 +26,9 @@
 
 @register_post_processor("v0.1")
 def sampler_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveResult:
-    """Convert :class:`~.QuantumProgramResult` to a :class:`~qiskit.primitives.PrimitiveResult`,
+    """Convert a ``QuantumProgramResult`` to a ``PrimitiveResult``, for ``SamplerV2``.
```
SamFerracin: This change seems off

qiskit_ibm_runtime/executor/routines/sampler_v2/sampler.py
```diff
@@ -298,16 +295,6 @@ def __init__(
...
-
-        Args:
-            mode: The execution mode (:class:`~qiskit.providers.BackendV2`,
-                :class:`~.Session`, or :class:`~.Batch`).
-            options: Options for the sampler. Can be a
-                :class:`~qiskit_ibm_runtime.executor.routines.options.sampler_options.SamplerOptions`
-                instance or a dict.
-            custom_prepare: Optional custom prepare function. Pass None to use the default.
-        """
         BaseSamplerV2.__init__(self)
```
```suggestion
        super().__init__(self)
```

qiskit_ibm_runtime/executor/routines/utils.py
```diff
@@ -76,10 +75,12 @@ def calculate_twirling_shots(
     """Calculate num_randomizations and shots_per_randomization for twirling.
 
     Implements the logic from TwirlingOptions documentation:
+
     - If both "auto": shots_per_randomization = max(64, ceil(shots/32))
                      num_randomizations = ceil(shots/shots_per_randomization)
     - If only num_randomizations "auto": num_randomizations = ceil(shots/shots_per_randomization)
-    - If only shots_per_randomization "auto": shots_per_randomization = ceil(shots/num_randomizations)
+    - If only shots_per_randomization "auto":
```
SamFerracin: This will be rendered as italic, I think.
```suggestion
    - If only ``shots_per_randomization`` "auto":
```

Happy to leave these kinds of changes for a future PR, but we should go through the docs and fix this kinds of things before we merge into main

qiskit_ibm_runtime/executor/routines/sampler_v2/post_processors/v0_1/post_processor.py
diego-plan9: Any suggestion? The way I see it, it goes from:

```
    """Convert :class:`~.QuantumProgramResult` to a :class:`~qiskit.primitives.PrimitiveResult`,
    for :class:`~qiskit_ibm_runtime.executor.routines.sampler_v2.SamplerV2`.

    This function transforms the raw quantum program execution results into the
```

to:

```
    """Convert a ``QuantumProgramResult`` to a ``PrimitiveResult``, for ``SamplerV2``.

    Convert :class:`~.QuantumProgramResult` to a :class:`~qiskit.primitives.PrimitiveResult`,
    for :class:`~qiskit_ibm_runtime.executor.routines.sampler_v2.SamplerV2`.
    
    This function transforms the raw quantum program execution results into the
```

Preserving a bit the references work (I think it was the only top-level docstring with 3 refs). Would you recommend removing the 2nd paragraph?

qiskit_ibm_runtime/executor/routines/utils.py
diego-plan9: Yes, it ties a bit with the above conversation with Tsafrir (in the sense of homogenizing and going over them). These files are not picked up by `make docs` yet, but when the time comes, we will be forced to go through them anyway and fix the warnings  (but I also encourage just setting up some time before that for a general revision and checking how the render looks like, even if not publishing docs).

qiskit_ibm_runtime/executor/routines/sampler_v2/post_processors/v0_1/post_processor.py
SamFerracin: >Would you recommend removing the 2nd paragraph?

Yes, I would recommend that. I don't think that the second line buys us much, as the links are available in the type-hint of the function. 

I would even suggest a simpler format:
```
"""Convert a quantum program result to a primitives result, for a V2 sampler.
```

WDYT?

diego-plan9: Sounds good! 72688db

SamFerracin (approved):

PR #2734, opened 2026-04-16
Add converters to/from Executor 1.0 schema models

SamFerracin:
The new model adds support for:
- specifying "semantic role" of a quantum program
- setting meas level to "both" to request L1+L2 (not currently supported)

In addition, it stores circuits in a list instead of inside circuit/samplex items


test/integration/executor/test_executor.py
```diff
@@ -64,10 +64,13 @@ def test_executor_with_circuit_item(self):
         program.append_circuit_item(isa_circuit, circuit_arguments=circuit_arguments)
 
         executor = Executor(self.backend)
+        executor.options.environment.image = (
```
SamFerracin: remove all changes in this file

qiskit_ibm_runtime/quantum_program/converters/converters_0_2.py
```diff
@@ -85,7 +85,12 @@ def quantum_program_from_0_2(model: ParamsModel) -> tuple[QuantumProgram, Execut
         else:
             raise ValueError("Unexpected model item type.")
 
-    quantum_program = QuantumProgram(shots=program_model.shots, items=items)
+    quantum_program = QuantumProgram(
+        shots=program_model.shots,
+        items=items,
+        meas_level=program_model.meas_level,
```
SamFerracin, line 91: these two had been forgotten in previous PRs

SamFerracin (commented):
@ihincks two FYIs:
- This PR does not modify ``ChunkPart``s.
- Please note how I've left ``semantic_role`` outside of the inputs to ``__init__``. I prefer it this way, as it does not advertise it too much

qiskit_ibm_runtime/quantum_program/converters/converters_1_0.py
```diff
@@ -0,0 +1,173 @@
...
+        items=items,
+        meas_level=program_model.meas_level,
+        passthrough_data=program_model.passthrough_data,
+    )
+    quantum_program.semantic_role = program_model.semantic_role
+
+    options = ExecutorOptions()
+    options.execution.init_qubits = model.options.init_qubits
+    options.execution.rep_delay = model.options.rep_delay
+    options.experimental = model.options.experimental
```
ihincks: nit: It's probably good practice to copy here, so that mutating the model after conversion doesn't mutate the `QuantumProgram`. Unlikely to surface in the wild.

qiskit_ibm_runtime/quantum_program/quantum_program.py
```diff
@@ -237,6 +237,11 @@ def __init__(
         self.meas_level = meas_level
         self.passthrough_data = passthrough_data
 
+        # Semantic role indicating how execution results may be post-processed by runtime clients.
+        # Reserved system values include 'sampler-v2' and 'estimator-v2', and are subject to change
+        # without notice. Third party clients should not set or depend on this value.
+        self.semantic_role: str | None = None
```
ihincks: What's the advantage of keeping this public? It's more API surface, but without any advantages I can see? I'm proposing that anything inside of qiskit-ibm-runtime that needs to know about it should break the private access rule.

```suggestion
        # Semantic role indicating how execution results may be post-processed by runtime clients.
        # Reserved system values include 'sampler-v2' and 'estimator-v2', and are subject to change
        # without notice. Third party clients should not set or depend on this value.
        self._semantic_role: str | None = None
```

qiskit_ibm_runtime/quantum_program/quantum_program_result.py
```diff
@@ -195,6 +195,11 @@ def __init__(
         self.metadata = metadata or Metadata()
         self.passthrough_data = passthrough_data
 
+        # Semantic role indicating how execution results may be post-processed by runtime clients.
+        # Reserved system values include 'sampler-v2' and 'estimator-v2', and are subject to change
+        # without notice. Third party clients should not set or depend on this value.
+        self.semantic_role: str | None = None
```
ihincks: Same comment as `QuantumProgram`.

qiskit_ibm_runtime/quantum_program/quantum_program.py
SamFerracin: I am fully in favour of this, makes it even more "internal only use"

SamFerracin: [262a33e](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2734/commits/262a33e7434eea30d87aa32da170947be24f623c)

qiskit_ibm_runtime/quantum_program/converters/converters_1_0.py
SamFerracin: see if you like this: [f402f39](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2734/commits/f402f39c89eebeb8699b5f6c4f5d58ff2b0ce613)

ihincks (approved):

PR #2736, opened 2026-04-17
Add all missing fake backends + modifying script for adding fake backends.

### LINKED ISSUE #2475: Add fake backends for the new QPUs

HuangJunye:
### Summary

NOTE please merge #2711 first before merging this one.

Building on top of https://github.com/Qiskit/qiskit-ibm-runtime/pull/2711, I modified the script to add `__init__.py` and `fake_[backend_name].py` files and also accepts a new `-a` /`--account` argument for passing named account. The latter is needed to pull fake backend data from different cloud region (US and EU).

Using the new script, I added all missing fake backends that are currently online

- aache
- berlin (new nighthawk in the eu which just released yesterday!)
- boston
- brussels
- miami
- pittburgh
- strasbourgh

The following backends are still missing but they have been retired

- nazca
- ithaca
- seattle
- dublin

### Details and comments
Fixes #2475


diego-plan9:
Thanks - and glad to see @iyanmv script put to good use! 

While waiting for the merging of #2711, could you also please add a release note to this PR, mentioning the new backends?

tools/generate_fake_backend_files.py
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env python3
```
diego-plan9, line 1: Comment on the file - might be missing a `git rm` over the original script, so changes are picked up instead of resulting in two files? The name change seems sensible to me.

HuangJunye: @diego-plan9 I think it's because I created the PR before the previous one was merged (and squashed) so the changes here include those changes instead of the diff. Can you advise what I should do?

diego-plan9: Ah that might explain it! I think `git rm tools/fetch_config_properties_backend.py` will do the trick, and `git` will be smart enough to consider it a rename with diff of the changes. Currently 1ae92ad shows both files in the `tools/` directory.

HuangJunye: It's not showing as rename with diff but at least the old script is removed now. I didn't notice it wasn't. I hope it's good now.

diego-plan9: It is fine - I played a bit locally for trying to get `git` to handle it as a rename+change, but I believe `git` considers the amount of changes above the threshold and will end up considering a delete+create no matter what. I'll compare manually the files for a final review.

diego-plan9 (approved):
Thanks @HuangJunye ! The updates to the script make generating fake backends even more streamlined - and having the new set of backends is also much welcome.

PR #2739, opened 2026-04-18
Avoid @property in Executor and NoiseLearnerV3, and other cleanups

### LINKED ISSUE #2719: Executor options auto complete doesn't work

I expect Executor options auto-complete to work the same way as Sampler and Estimator - e.g. if I do `executor.options.execution.<tab>`, I expect to see all the execution related options. Instead it shows me a long list of attributes I can't actually use:

Image

And actual options like `init_qubits` doesn't show up

Image

**Acceptance criteria**

Executor options auto complete should show actual available options.

diego-plan9:
### Summary

Prefer `__setattr__` as the way to enforce specific checks when setting `Executor.options`, and directly expose `NoiseLearnerV3.options` as a regular field (instead of via a `@property`). This hopefully fix the issues with auto-complete in `ipython` (they seem to currently working in a regular `python` terminal, though), and makes the usage of `.options` a bit more homogeneous (both can be replaced directly by setting the attribute, and both can be mutated by the user).

Additionally, this PR removes a couple of executor-related leftovers in `constant.py`, and makes active use of `if TYPE_CHECKING` (related to #2563) in both files to better behave when it comes to imports.

### Details and comments

Fixes #2719

Tested the auto-completion bits with `ipython`.

### DIFF
--- a/qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
+++ b/qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
@@ -71,27 +71,25 @@ class NoiseLearnerV3:
     _PROGRAM_ID = "noise-learner"
     _DECODER = NoiseLearnerV3ResultDecoder
 
+    options: NoiseLearnerV3Options
+    """The options in this noise learner."""
+
     def __init__(
         self,
         mode: BackendV2 | Session | Batch | None = None,
         options: NoiseLearnerV3Options | None = None,
     ):
-        self._options = options or NoiseLearnerV3Options()
+        self.options = options or NoiseLearnerV3Options()
         if (
-            isinstance(self._options.experimental, UnsetType)
-            or self._options.experimental.get("image") is None
+            isinstance(self.options.experimental, UnsetType)
+            or self.options.experimental.get("image") is None
         ):
-            self._options.experimental = {}
+            self.options.experimental = {}
 
         self._session, self._service, self._backend = get_mode_service_backend(mode)  # type: ignore[assignment]
 
         if isinstance(self._service, QiskitRuntimeLocalService):  # type: ignore[unreachable]
             raise ValueError("``NoiseLearnerV3`` is currently not supported in local mode.")
-
-    @property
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -71,36 +70,34 @@ class NoiseLearnerV3:
...
+        self._session, self._service, self._backend = get_mode_service_backend(mode)
 
-        if isinstance(self._service, QiskitRuntimeLocalService):  # type: ignore[unreachable]
+        if isinstance(self._service, QiskitRuntimeLocalService):
             raise ValueError("``NoiseLearnerV3`` is currently not supported in local mode.")
 
-    @property
-    def options(self) -> NoiseLearnerV3Options:
-        """The options in this noise learner."""
-        return self._options
```
diego-plan9, line 97: Changing `options` to a regular attribute (as in line 73) changes our approach to `options`. Previously, users were only able to set them during constructor, and not able to replace them. However, they were able to mutate them (ie. `noise_learner_v3.options.foo = 2`).

Now the user is able to replace them, and the change is intentional, to have parity with `Executor` (introduced in #2616). Actually I would like to also include a similar `__setattr__` (so it restricts the possible values to the option class, or a valid `dict`) for even more parity, but I want to get some feedback on this direction before doing so.

diego-plan9:
@jyu00 can you check if this works autocomplete-wise in your setup?

SamFerracin (approved):
Thanks, looks good!

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
SamFerracin: > but I want to get some feedback on this direction before doing so.

I agree with including a similar `__setattr__`, it solidifies the intended behavior. I can't imagine a scenario where a use may want to do something different than that and manage it to make it work (eg: setting `executor.options = nlv3_options`)

PR #2741, opened 2026-04-20
Fix typo in qiskit_ibm_runtime/options/zne_options.py

schrodinteq:
### Summary
Fix typo in qiskit_ibm_runtime/options/zne_options.py

### Details and comments
Fixes #
at which the ``extrapolator``\\s are evaluated
-> at which the ``extrapolator``s are evaluated


qiskit_ibm_runtime/options/zne_options.py
```diff
@@ -142,7 +142,7 @@ class ZneOptions:
 
     If unset, this will default to ``[0, *noise_factors]``. This
     option does not affect execution or model fitting in any way, it only determines the
-    points at which the ``extrapolator``\\s are evaluated to be returned in the data
+    points at which the ``extrapolator``s are evaluated to be returned in the data
```
diego-plan9:
```suggestion
    points at which the ``extrapolator``\s are evaluated to be returned in the data
```

diego-plan9 (changes requested):
Thanks @schrodinteq - while the docstrings might look a bit funky purely from the Python side, a number of them (the ones linked from the `docs/` indices) are being used (via `sphinx`) for [rendering the documentation](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/options-zne-options#methods), so they will have some Sphinx syntax on them intentionally.

In general, the escaping sequences are required after several constructs in order to preserve the formatting (you will also find `\\s` after class references, for example) while pluralizing or append characters directly afterwards. The instance you are changing is actually the only rendered place (there are two more instances in a non-rendered docstring) where we are appending characters to something that is not a reference, and it is not rendering fully correctly indeed, thanks for catching it!

I left a small suggestion for actually fixing the rendering bug (removing the escape characters entirely does lead to rendering artifacts) and playing nicely with sphinx. Actually, can you also apply the same fix to the `Unset` and `None` mentions at https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/options/noise_learner_v3_options.py#L134?

schrodinteq:
@diego-plan9 
Thank you for your detailed review. I have addressed the points you mentioned.
In addition, I have also applied the same fix to the `Unset` and `None` references within this PR.

diego-plan9 (approved):
Thanks!

Issue #2673, opened 2026-03-16
Jobs that trigger max_execution_time are not in final state


emilkovacev:
**Describe the bug**

When a job (RuntimeJobV2) is run against a device using `SamplerV2` or `EstimatorV2`, and mx_execution_time is specified ([via options](https://quantum.cloud.ibm.com/docs/en/guides/max-execution-time)), the resulting job returns the status `CANCELLED - RAN TOO LONG`.

Because `CANCELLED - RAN TOO LONG` is not in `JOB_FINAL_STATES` (defined [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/4bb3ddaf4a628d7679d0fc5a244e3b4cd2fa0f07/qiskit_ibm_runtime/runtime_job_v2.py#L56)), calling `job.in_final_state()` or `job.wait_for_final_state()` or `job.result()` hangs indefinitely.

**Steps to reproduce**

Run this snippet against a backend, `job.wait_for_final_state()` will hang forever.

```
from qiskit.circuit.random import random_circuit
from qiskit import transpile

from qiskit_ibm_runtime import QiskitRuntimeService
from qiskit_ibm_runtime import SamplerV2

service = QiskitRuntimeService()
backend = service.backend("<backend>")

sampler = SamplerV2(mode=backend)

sampler.options.max_execution_time = 10

qc = random_circuit(100, 100, measure=True)
qc = transpile(qc, backend)

job = sampler.run([qc])

job.wait_for_final_state()
```

**Expected behavior**

I expect that `wait_for_final_state` would return once the job reaches max_execution_time limit.

In reality, it hangs forever.

**Suggested solutions**

* `CANCELLED - RAN TOO LONG` can be added to `JOB_FINAL_STATES` (defined [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/4bb3ddaf4a628d7679d0fc5a244e3b4cd2fa0f07/qiskit_ibm_runtime/runtime_job_v2.py#L56))
* Maybe there is a fix that needs to be made so that the status returns `CANCELLED`?

**Additional Information**

- **qiskit-ibm-runtime version**: 0.45.1
- **Python version**: 3.11.9
- **Operating system**: MacOS

diego-plan9:
Closing as it has been fixed API-side.

PR #2746, opened 2026-04-20
Cleanup api-related variables in constants

diego-plan9:
### Summary

Cleanup a couple of api-related variables in `constants.py`:
* delete unused `API_TO_JOB_STATUS`
* move `API_TO_JOB_ERROR_MESSAGE`  closer to its only consumer

### Details and comments

Originally related to #2673 
Related to #2749


diego-plan9:
CC @jyu00 and @emilkovacev : Emil, can you check if this allows solving the `.wait_for_final_state()` block in your scenario?

jyu00:
`CANCELLED - RAN TOO LONG` is not a documented status. I feel we should push the server to fix it instead of trying to work around their mistake.

diego-plan9:
> `CANCELLED - RAN TOO LONG` is not a documented status. I feel we should push the server to fix it instead of trying to work around their mistake.

I agree a server-side fix is be ultimate fix. This PR allows the client to handle the odd status correctly in the meantime, restoring user's functionality and leaving us in a better state (and the changes are easy to remove once the real fix is up).

emilkovacev:
I can confirm on my end that this PR does resolve the issue, `job.status()` now returns `ERROR`, and `job.result()` now fails with this error:
```
RuntimeJobMaxTimeoutError: 'Error code 1305; Ran too long Increase [max execution time](https://quantum.cloud.ibm.com/docs/guides/max-execution-time) or contact support.'
```

qiskit_ibm_runtime/constants.py
```diff
@@ -22,20 +20,6 @@
 from .utils.runner_result import RunnerResult
 
 
-API_TO_JOB_STATUS = {
```
SamFerracin, line 25: this was a duplicate, eh?

SamFerracin (commented):
I am not opposed to this change as a stepping stone as we wait for server-side changes. Happy to approve, unless there are strong oppositions from the other reviewers

qiskit_ibm_runtime/constants.py
diego-plan9: Yes, it seems it became unused at some point (and instead `qiskit_ibm_runtime.runtime_job_v2.API_TO_JOB_STATUS` is the map really used).

diego-plan9:
An API fix is on its way - on hold until verifying changes.

diego-plan9:
After verifying the API fix worked, I'm re-targeting this PR for mostly some cleanups.

For future reference, we still handle "jobs cancelled due to exceeding specified max time" as a particular case, considering them [ERROR instead of CANCELLED](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.46.1/qiskit_ibm_runtime/runtime_job_v2.py#L179-L180) and having a dedicated `RuntimeJobMaxTimeoutError` exception. The first bit is inconsistent with how the API signals them - and it can be argued that the current behavior was an explicit decision by the client (`.error_message()` can be interpreted as only available for jobs that error, and changing the status might be a breaking change in some scenarios).

So I'm considering tackling that out of the scope of this PR (original for catching a bug, now for some non-breaking cleanup), and prefer to revise this particular case as part of a bigger revision that can be included in 1.0.

yaelbh (approved):

Issue #2749, opened 2026-04-21
Revise modularization and reduce circular imports


diego-plan9:
**What is the expected feature or enhancement?**

Currently, it is very easy to fall into circular imports. This seems to be a combination of existing policies and conventions:
* preferring using intermediate `__init__.py` files for re-exporting members
* combination of either using too fine-grained modules (for example, for containing just one class) or too wide modules (in particular, `utils` and similar modules accumulating disparate functionality)
* using top-level `__init__.py` (seemingly) for facilitating user imports

As a side effect, most of the package is imported by `import qiskit_ibm_runtime`, and the risk of circular import trickles into a number of changes into the codebase. This story is about:
* refactoring existing code, aiming to improve maintainability and reduce import cycles
* evaluate the practices in the process, and eventually add some guidelines that help us produce more cohesive modules and a better balance between user convenience / developer overhead / scalability

**Acceptance criteria**

PRs contributing to this issue directly reduce the risk of circular imports and/or provide better modularization

PR #2752, opened 2026-04-21
Update schema version for NLV3

### LINKED ISSUE #2747: In NLV3 use schema version 0.2
**What is expected?**

We are currently still using Version 0.1 for NLV3. Upgrade to schema version 0.2.

**Acceptance criteria**

Encoders and decoder of NLV3 use schema version 0.2.

SamFerracin:
Closes #2747 

Also moves the global variables to the dedicated `constants.py` file

### DIFF
--- a/qiskit_ibm_runtime/constants.py
+++ b/qiskit_ibm_runtime/constants.py
@@ -44,9 +44,3 @@ DEFAULT_DECODERS: dict[str, type[ResultDecoder] | list[type[ResultDecoder]]] = {
     "circuit-runner": RunnerResult,
     "qasm3-runner": RunnerResult,
 }
-
-SCHEMA_VERSION_EXECUTOR = "v0.2"
-"""The schema version used by default by executor to encode the input params."""
-
-SCHEMA_VERSION_NLV3 = "v0.2"
-"""The schema version used by default by noise learner V3 to encode the input params."""
--- a/qiskit_ibm_runtime/constants.py
+++ b/qiskit_ibm_runtime/constants.py
@@ -44,3 +44,9 @@ DEFAULT_DECODERS: dict[str, type[ResultDecoder] | list[type[ResultDecoder]]] = {
     "circuit-runner": RunnerResult,
     "qasm3-runner": RunnerResult,
 }
+
+SCHEMA_VERSION_EXECUTOR = "v0.2"
+"""The schema version used by default by executor to encode the input params."""
+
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/constants.py
```diff
@@ -44,3 +44,9 @@
     "circuit-runner": RunnerResult,
     "qasm3-runner": RunnerResult,
 }
+
+SCHEMA_VERSION_EXECUTOR = "v0.2"
+"""The schema version used by default by executor to encode the input params."""
+
+SCHEMA_VERSION_NLV3 = "v0.2"
+"""The schema version used by default by noise learner V3 to encode the input params."""
```
diego-plan9: I agree they are constants (as in, variables that contain values that don't change during the package life time). However, they are very specific to a particular feature (currently each only used once, in `NoiseLearnerV3` and `Executor` respectively), and they feel closer to a specific implementation detail of each of the classes (ie. you could argue they would be class attribute if we want to avoid a global) and also closer semantically (the class and the variable do not exist without each other).

There is some gain in moving it to `constants` (easier to change, more visibility, and easier to use from multiple places if/when that happens) but also some drawbacks (a bit of cognitive overhead in this case where they are only used once as developers need to jump between files, and some relation to #2749 as `constants` imports go through `utils` currently).

If those are the only benefits, can you move them outside `constants`, and either keep them as variables in the existing modules (I do like the renaming!) or even move them inside their classes (no strong preference on one over the other)? I have applied the same approach in #2746 consciously (and with the same arguments), as I feel the trade-off makes sense.

diego-plan9 (commented):
Thanks Sam! A small question about using `constants`, and we should be good to go.

qiskit_ibm_runtime/constants.py
SamFerracin: Yes, that sounds good. I was partial to it myself. I moved them inside their classes in [6e543e5](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2752/commits/6e543e5cd1bfddf059d75618f993e49aec24ded6)

SamFerracin: Yes, that sounds good. I was partial to it myself. I moved them inside their classes in [6e543e5](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2752/commits/6e543e5cd1bfddf059d75618f993e49aec24ded6)

diego-plan9 (approved):

Issue #2748, opened 2026-04-21
NLV3 has options that can be of type UnsetType


yaelbh:
**What is expected?**

We are not happy with the `UnsetType`, and aim to stop using it in Executor (where this is indeed the case) and NLV3. For NLV3, currently some options need to be fixed: `experimental`, `max_execution_time`, maybe more.

**Acceptance criteria**

None of the NLV3 options can be of type `UnsetType`.

jyu00:
> We are not happy with the UnsetType, and aim to stop using it in Executor 

What is the alternative you plan to use? It was first introduced to differentiate `None` (a valid value) and "use server default".

SamFerracin:
@jyu00 
> What is the alternative you plan to use? It was first introduced to differentiate None (a valid value) and "use server default".

With Executor and NLV3, we have shifted to a new paradigm where there are no more server defaults, for transparency. Every default is set in the client and/or in the schemas. Therefore, `Unset` is no longer useful in these newer option classes

ihincks:
> Every default is set in the client and/or in the schemas.

Small subtlety: while I agree what you write is true, it also uses a particular definition of what a default is that may not initially be obvious. Here, a default is the literal value assigned to a field in the schema, and all fields in the input schema have a default value. Therefore if you omit a field, its eventual value is not a property of the service version, but rather of the schema.

_However_, maybe the best example to focus on here is chunk size. The default value is the string literal `"auto"`. So in this sense, the default value of the field means that you opt-in to whatever the service version wants you to do that day. I like this because it's very explicit. It's rather different than, say, a field ` foo: bool` with no default value, that gets assigned by the service when omitted. Rather, we would need to do `foo: bool | Literal["choose_for_me"]` if we wanted this behaviour.

jyu00:
> [@jyu00](https://github.com/jyu00)
> 
> > What is the alternative you plan to use? It was first introduced to differentiate None (a valid value) and "use server default".
> 
> With Executor and NLV3, we have shifted to a new paradigm where there are no more server defaults, for transparency. Every default is set in the client and/or in the schemas. Therefore, `Unset` is no longer useful in these newer option classes

[...]

Issue #2743, opened 2026-04-20
Finalize executor and noise learner v3 release


diego-plan9:
### Description

As `executor` and `noise learner v3` are close to completion, it is time for ensuring they are ready for consumption:
* [ ] Add executor/nlv3 public documentation to the rendered docs
* [x] Depend on a `ibm-quantum-schemas` release that contains `executor` `1.0`
* [x] Ensure `job.inputs()` returns contain rich objects
  * https://github.com/Qiskit/qiskit-ibm-runtime/issues/2728
  * https://github.com/Qiskit/qiskit-ibm-runtime/issues/2740
* [ ] Revise public interface 
  * [x] Revise constructor arguments (make signature identical / revise allowing optional `mode`)
  * [x] Revise `NoiseLearnerV3.backend()` method need
  * [x] Revise `NoiseLearnerV3.run()` `raises` clause
  * [x] https://github.com/Qiskit/qiskit-ibm-runtime/issues/2748
  * [x] https://github.com/Qiskit/qiskit-ibm-runtime/issues/2747
  * [x] https://github.com/Qiskit/qiskit-ibm-runtime/issues/2756
* [x] Usability
   * https://github.com/Qiskit/qiskit-ibm-runtime/issues/2719
* [ ] Add release note 

### Acceptance criteria

`executor` and `noise learner v3` are ready to be consumed and documented for the next release.

diego-plan9:
Completed in https://github.com/Qiskit/qiskit-ibm-runtime/releases/tag/0.47.0

PR #2753, opened 2026-04-22
Make Executor and NoiseLearnerV3 interfaces more consistent

diego-plan9:
### Summary

Changes in order to make `Executor` and `NoiseLearnerV3` interfaces more consistent (mostly by getting `Executor` closer to `NoiseLearnerV3`):
* match constructor signatures
* `Executor`:
  * add `.backend()` method
  * avoid `_run()` method (mostly for reducing levels of indirection, as `run()` was quite small)
* `NoiseLearnerV3`:
  *  allow setting options after construction with validation, and allow dicts (ie. #2616 +  #2739 for nlv3)
  *  more strict validation (#2720, leaving touching `NoiseLearnerV3Result` to #2748)
  * add tests over options creation and validation (same approach as executor tests)

Additionally, remove `IBMBackend.submit()`, as it was introduced in relation to `executor` unintendedly.

### Details and comments

Related to #2743

The goal is to be consistent between the two interfaces, and as secondary concern be similar to `BasePrimitiveV2`interface. Some of the decisions are due to the latter and have some degree of subjectiveness (for example, I like the `*` in the signature, but depending on how consistent we want to be with the old primitives, had to go away).


qiskit_ibm_runtime/options/utils.py
```diff
@@ -32,6 +32,14 @@
     from ..options.options import BaseOptions
 
 
+PRIMITIVES_CONFIG = ConfigDict(validate_assignment=True, extra="forbid")
+"""Custom ``ConfigDict`` for pydantic dataclasses.
+
+These config settings ensure we get validation on attribute mutation, not just at construction
+time, and also that we get a validation error if someone spells an attribute name wrong.
+"""
```
diego-plan9, line 40: This is a continuation of #2720 - so we can reuse the custom `ConfigDict` in the nlv3 option classes.

There is also the option of making it a custom decorator (similar to existing `@primitive_dataclass`). However, the current implementation of `@primitive_dataclass` has issues with pylance currently:
```
Cannot access attribute "experimental" for class "PydanticDataclass"
  Attribute "experimental" is unknownPylancereportAttributeAccessIssue
```

which causes auto-complete and warnings in IDEs, as they cannot infer the actual type. I think it can be solved by some type-hinting craftiness, but I was not able to fully solve it during this round.

Also, I could not apply this to `NoiseLearnerV3Options` (as it is using `Unset`, and that requires one of the properties used by the existing `@primitive_dataclass` configdict). I am counting in this to be tackled by #2748

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -79,20 +79,34 @@ class NoiseLearnerV3:
...
 
         self._session, self._service, self._backend = get_mode_service_backend(mode)
 
         if isinstance(self._service, QiskitRuntimeLocalService):
             raise ValueError("``NoiseLearnerV3`` is currently not supported in local mode.")
 
+    def __setattr__(self, name: str, value: Any) -> None:
+        """Set attribute ``name`` to ``value``.
+
+        Handle ``options`` as a special case, ensuring it is set to an ``ExecutorOptions`` instance.
```
SamFerracin:
```suggestion
        Handle ``options`` as a special case, ensuring it is set to a ``NoiseLearnerV3Options`` instance.
```

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -103,14 +117,14 @@ def run(self, instructions: Iterable[CircuitInstruction]) -> RuntimeJobV2:
...
-            IBMInputValueError: If an instruction a box with non-ISA gates.
-            IBMInputValueError: If an instruction cannot be learned by any of the supported
-                learning protocols.
+            IBMInputValueError: If the instructions cannot be used with the noise learner, such as:
+
+                * If an instruction contains a box without twirl annotation.
+                * If an instruction contains unphysical qubits, i.e., qubits that do not belong to
+                  the "physical" register ``QuantumRegister(backend.num_qubits, 'q')``
+                  for the backend in use.
+                * If an instruction a box with non-ISA gates.
```
SamFerracin:
```suggestion
                * If an instruction contains a box with non-ISA gates.
```

qiskit_ibm_runtime/options/utils.py
SamFerracin: And I guess that this won't be used for options other than executor/NLV3, unless we go through the effort of removing unsets everywhere?

qiskit_ibm_runtime/executor.py
```diff
@@ -63,7 +62,7 @@ class Executor:
     Args:
         mode: The execution mode used to make the query. It can be:
 
-            * A :class:`IBMBackend` if you are using job mode.
+            * A :class:`BackendV2` if you are using job mode.
```
SamFerracin: Btw I suspect that the hyperlink only works if we do this, but I may be wrong:
```suggestion
            * A :class:`~.BackendV2` if you are using job mode.
```

test/unit/noise_learner_v3/test_noise_learner_v3.py
```diff
@@ -12,15 +12,145 @@
...
+    PostSelectionOptions,
+)
 
 from ...ibm_test_case import IBMTestCase
 
 
+class TestNoiseLearnerV3Options(IBMTestCase):
+    """Tests option setting on the ``NoiseLearnerV3`` class."""
+
+    @expectedFailure
```
SamFerracin, line 35: thanks! will save work in the followup PR to remove unsets

SamFerracin (commented):
Looks good, thanks Diego! Small comments!

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -79,20 +79,34 @@ class NoiseLearnerV3:
...
+        autocomplete features.
+        """
+        if name == "options":
+            if isinstance(value, dict):
+                value = NoiseLearnerV3Options(**value)
+            elif not isinstance(value, NoiseLearnerV3Options):
+                raise TypeError(f"Expected NoiseLearnerV3Options or dict, got {type(value)}")
+
+            if isinstance(value.experimental, UnsetType) or value.experimental.get("image") is None:
+                value.experimental = {}
```
diego-plan9, line 104: This was previously present in `__init__` (I moved it here for consistency, as otherwise `experimental` would not be chopped of when setting outside constructor but would when setting in constructor) - and the tests contain an `image` field.

Is this intended behavior, and needed at this stage?

qiskit_ibm_runtime/options/utils.py
diego-plan9: > And I guess that this won't be used for options other than executor/NLV3, unless we go through the effort of removing unsets everywhere?

Yes, using `Unset` requires the `arbitrary_types_allowed=True` in `ConfigDict`. My suggestion would be to just make this mechanism (either as-is, with passing the `PRIMITIVES_CONFIG` to the decorator; or creating a new decorator with proper typing) the one to use with new features, and keep the old decorator for the old classes.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
SamFerracin: I haven't had time to investigate it yet, but i think what we want is simply:
```
if isinstance(value.experimental, UnsetType):
    value.experimental = {}
```

Now, the devil will be in the details here. When we make `NoiseLearnerOptionsV3` be a dataclass and not be a child of `BaseOptions`, we'll need to figure out how to create the runtime options and make sure that the image ends up where it should. I imagine we can follow the blueprint of `ExecutorOptions`, but again, let me try doing it before I actually say something that turns out to be incorrect

Does this make sense to you?

SamFerracin: Anyway, I would say, leave it as is right now. Afaik, NLV3 doesn't currently use any exeperimental option server-side, so the `or value.experimental.get("image") is None` is harmless

diego-plan9: ~If you can look at it in detail, would be appreciated. I interpret this as "disallow `.experimental` entirely unless an `image` is specified" currently.~

I did not read your previous message when typing the above. So keeping as-is 👍

SamFerracin (approved):

PR #2757, opened 2026-04-23
Reformat NLV3 Options in the style of Executor options

### LINKED ISSUE #2748: NLV3 has options that can be of type UnsetType

### LINKED ISSUE #2756: NLV3 Options should be reformatted in the style of Executor options.
NLV3 Options should be reformatted in the style of Executor options.

**Acceptance criteria**
- [ ] NLV3 Options should not inherit from `BaseOptions`.
- [ ] They should not be decorated with `@primitive_dataclass`, but with `@dataclass(config=PRIMITIVES_CONFIG)`.
- [ ] They should not have unset types.

SamFerracin:
Closes #2748 
Closes #2756 

This PR leaves a few more discrepancies between NLV3 options and Executor options that will be tackled in future PRs, following discussion. These discrepancies include:
- Executor has "its own" environment options (a dataclass), whereas in NLV3 we use the previous ones (a child of `BaseOptions` and a `primitive_dataclass`). What's the best way forward to address this discrepancy?
- There are small inconsistencies in where one sets certain options, for example the custom image.


test/unit/noise_learner_v3/test_noise_learner_v3.py
```diff
@@ -175,3 +180,19 @@ def test_run_of_service_is_selected(self):
             noise_learner = NoiseLearnerV3(mode=backend)
             selected_run = noise_learner.run([])
             self.assertEqual(selected_run, "service")
+
+    def test_runtime_options(self):
+        """Test the ``_runtime_options`` method."""
```
SamFerracin, line 178: The fact that I want to test this makes me wonder if we should make this method public, both here and in executor

qiskit_ibm_runtime/options/noise_learner_v3_options.py
```diff
@@ -85,16 +70,20 @@ class NoiseLearnerV3Options(BaseOptions):
...
-    experimental: UnsetType | dict = Unset
+    experimental: dict = Field(default_factory=dict)
     """Experimental options.
 
     These options are subject to change without notification, and stability is not guaranteed.
     """
 
+    max_execution_time: int | None = None
+    environment: EnvironmentOptions | dict = Field(default_factory=EnvironmentOptions)
+    simulator: SimulatorOptions | dict = Field(default_factory=SimulatorOptions)
```
diego-plan9: I wonder if here (and in `post_selection`) we can declare them just as `FooOptions`, not allowing `dict` (ie. `ExecutorOptions` style)? 

We do have a conversion happening at `NoiseLearnerV3` level (at init at and setter) at the root level, plus pydantic features overall, and I like the idea of not supporting variants at a fine-grained level (as in, either all the `NoiseLearnerV3Options` comes from a dict, or comes as an instance directly). It reduces complexity and helps maintainability, and I think we can trust such "railguards" at the program level (plus piggy-back on pydantic usage and conversation for user consumption).

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -99,12 +98,18 @@ def __setattr__(self, name: str, value: Any) -> None:
                 value = NoiseLearnerV3Options(**value)
             elif not isinstance(value, NoiseLearnerV3Options):
                 raise TypeError(f"Expected NoiseLearnerV3Options or dict, got {type(value)}")
-
-            if isinstance(value.experimental, UnsetType) or value.experimental.get("image") is None:
-                value.experimental = {}
```
diego-plan9, line 104: With this being removed: can you take a pass at `test/unit/noise_learner_v3/test_noise_learner_v3.py` and remove the `"image": "foo` entries from the tests that contain it, along with the: 

```
        This test requires `image` to be present in experimental options to avoid them being
        removed.
```

docstring comments?

diego-plan9 (commented):
Thanks Sam!

> This PR leaves a few more discrepancies between NLV3 options and Executor options that will be tackled in future PRs, following discussion.

Sounds good.

test/unit/noise_learner_v3/test_noise_learner_v3.py
diego-plan9: I would lean towards not making it public. Technically `RuntimeOptions` has been deprecated for a while and is a candidate for removal (#2300) - it is a bit messy currently (some primitives do not use it already, others still do) and requires some work. Keeping it non-public in the meantime aligns better with the current state of `RuntimeOptions` (as in, internal-only until fully removed).

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
SamFerracin: [8e7c645](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2757/commits/8e7c64530accf520d7070021f86082f294889d52)

qiskit_ibm_runtime/options/noise_learner_v3_options.py
SamFerracin: [07009ce](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2757/commits/07009ce5bea9bb726da8b74d30ef982b22f2b1a7)

SamFerracin: i also strengthened a bit the tests, to make sure that we can support dicts properly

diego-plan9: Thanks!

diego-plan9 (approved):

PR #2759, opened 2026-04-25
Fix iterator exhaustion issue in NoiseLearner.run()

rajnisht7:
### Summary

whenever a user pass generator,map,filter or any single pass iterator in `NoiseLearner.run()`, `all(isinstance(...))` check consumes all the circuits due to which job used to get submit with 0 circuits

this PR resolves this issue

the bug can be generated using 
```
from qiskit_ibm_runtime.noise_learner import NoiseLearner

backend = service.least_busy(operational=True, simulator=False, min_num_qubits=2)

print("backend:", backend.name)

pm = generate_preset_pass_manager(backend=backend, optimization_level=1)

learner = NoiseLearner(backend)


def circuit_generator(n):
    for _ in range(n):
        qc = QuantumCircuit(2)
        qc.h(0)
        qc.cx(0, 1)
        transpiled_qc = pm.run(qc)
        yield transpiled_qc

job = learner.run(circuit_generator(3))

print("job",len(job.inputs["circuits"]))
print("job id:", job.job_id)
```

### Details and comments
Fixes #
replaced the previous iterator handling logic in `NoiseLearner.run()` with a single list comprehension:

```python
circuits = [
    circuit if isinstance(circuit, QuantumCircuit)
    else EstimatorPub.coerce(circuit).circuit
    for circuit in circuits
]

### DIFF
--- a/qiskit_ibm_runtime/noise_learner/noise_learner.py
+++ b/qiskit_ibm_runtime/noise_learner/noise_learner.py
@@ -167,8 +167,10 @@ class NoiseLearner:
         runtime_options = NoiseLearnerOptions._get_runtime_options(options_dict)
 
         # Define the program inputs
-        inputs = {"circuits": circuits}
-        inputs.update(learner_options)
+        inputs: dict[str, Any] = {
+            "circuits": circuits,
+            **learner_options,
+        }
 
         calibration_id = None
         if self._backend:
--- a/qiskit_ibm_runtime/noise_learner/noise_learner.py
+++ b/qiskit_ibm_runtime/noise_learner/noise_learner.py
@@ -159,8 +159,7 @@ class NoiseLearner:
         # Convert all inputs (QuantumCircuit or EstimatorPubLike) to a list of circuits
         # in a single pass and also supports mixed inputs.
         circuits = [
-            circuit if isinstance(circuit, QuantumCircuit)
-            else EstimatorPub.coerce(circuit).circuit
+            circuit if isinstance(circuit, QuantumCircuit) else EstimatorPub.coerce(circuit).circuit
             for circuit in circuits
         ]
 
--- a/qiskit_ibm_runtime/noise_learner/noise_learner.py
+++ b/qiskit_ibm_runtime/noise_learner/noise_learner.py
@@ -156,10 +156,13 @@ class NoiseLearner:
             The submitted job.
 
         """
-        circuits = circuits if isinstance(circuits, list) else list(circuits)
-        if not all(isinstance(t, QuantumCircuit) for t in circuits):
-            coerced_pubs = [EstimatorPub.coerce(pub) for pub in circuits]
-            circuits = [p.circuit for p in coerced_pubs]
+        # Convert all inputs (QuantumCircuit or EstimatorPubLike) to a list of circuits
+        # in a single pass and also supports mixed inputs.
+        circuits = [
+            circuit if isinstance(circuit, QuantumCircuit)
+            else EstimatorPub.coerce(circuit).circuit
+            for circuit in circuits
+        ]
 
         # Store learner-specific and runtime options in different dictionaries
         options_dict = asdict(self.options)
--- a/qiskit_ibm_runtime/noise_learner/noise_learner.py
+++ b/qiskit_ibm_runtime/noise_learner/noise_learner.py
@@ -156,7 +156,7 @@ class NoiseLearner:
             The submitted job.
 
         """
-        circuits = list(circuits)
+        circuits = circuits if isinstance(circuits, list) else list(circuits)
         if not all(isinstance(t, QuantumCircuit) for t in circuits):
             coerced_pubs = [EstimatorPub.coerce(pub) for pub in circuits]
             circuits = [p.circuit for p in coerced_pubs]
--- a/qiskit_ibm_runtime/noise_learner/noise_learner.py
+++ b/qiskit_ibm_runtime/noise_learner/noise_learner.py
@@ -156,6 +156,7 @@ class NoiseLearner:
             The submitted job.
 
         """
+        circuits = list(circuits)
         if not all(isinstance(t, QuantumCircuit) for t in circuits):
             coerced_pubs = [EstimatorPub.coerce(pub) for pub in circuits]
             circuits = [p.circuit for p in coerced_pubs]
--- a/test/unit/test_noise_learner.py
+++ b/test/unit/test_noise_learner.py
@@ -83,13 +83,27 @@ class TestNoiseLearner(IBMTestCase):
 
     @combine(task_type=["circs", "pubs"])
     def test_run_program_inputs_with_default_options(self, task_type):
-        """Test a circuit with default options."""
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: qiskit_ibm_runtime/noise_learner/noise_learner.py [hunk 1]
```diff
@@ -156,6 +156,7 @@ def run(self, circuits: Iterable[QuantumCircuit | EstimatorPubLike]) -> RuntimeJ
             The submitted job.
 
         """
+        circuits = circuits if isinstance(circuits, list) else list(circuits)
         if not all(isinstance(t, QuantumCircuit) for t in circuits):
             coerced_pubs = [EstimatorPub.coerce(pub) for pub in circuits]
             circuits = [p.circuit for p in coerced_pubs]
```
diego-plan9: @SamFerracin (as the original author) - do you see a sensible way of performing an equivalent check and fulfilling the docstring description _without_ consuming `circuits` in the process? I don't think it is (perhaps peeking at the first item and deciding based on that) - and actually might not be worth it.

It is mostly for trying to avoid blindly converting to list before the check, which is potentially costly. However, in practice the lower layers (at around `self._service.run()` expect a _sequence_ anyway, so it might not be a real cost impact anyway.

Looking from another angle: currently, the method would fail ungracefully if `circuits` contain a mixture of circuits and pub-like objects due to this check (by entering, and attempting `EstimatorPub.coerce(pub)` in a `QuantumCircuit`). I don't know if that is a valid use case, but if so, an alternative might be iterating over the `circuit` args and performing the check (and conversion if needed) inside? Similar to:

```python
        # Some meaningful comment ...
        actual_circuits = [
            circuit if isinstance(circuit, QuantumCircuit) else EstimatorPub.coerce(circuit).circuit
            for circuit in circuits
        ]
```

test/unit/test_noise_learner.py
```diff
@@ -132,3 +134,23 @@ def test_get_backend(self):
...
+
+        inst = NoiseLearner(backend)
+        inst.run(circuit_generator(4))
+
+        input_params = backend.service._run.call_args.kwargs["inputs"]
+        self.assertEqual(
+            len(input_params["circuits"]),
+            4,
+            "Generator should not be exhausted - all 4 circuits must be present",
+        )
```
diego-plan9: Can you check if this test can be incorporated into `test_run_program_inputs_with_default_options()`? 

They seem to be quite similar (running estimator with a specific input, and asserting over the `inputs` received by the service) - and rather than a regression specific test, I see it as exercising a feature. By keeping them in the same test, it is easier to maintain.

For example, you could try something similar to (pseudo-code):

```python
    @combine(task_type=["circs", "circs_iterator", "pubs", "pubs_iterator"])
    def test_run_program_inputs_with_default_options(self, task_type)
        [...]
        if task_type == "circs":
            tasks = [transpile(c) for c in self.circuits]
        elif task_type == "circs_iterable":
            tasks = circuit_generator(4)  # or similar - ideally reusing `self.circuits` in the custom function
        elif task_type == "pubs:
            tasks = [(transpile(c), "Z" * c.num_qubits) for c in self.circuits]
        else:  # similar generator for pubs
        [...]
```

diego-plan9 (commented):
Thanks @ rajnisht7, this is a good catch!

We are indeed declaring that `NoiseLearner.run()` accepts an `Iterable`, but not dealing correctly with non-sequences. While waiting for a bit more of input on how the arguments are intended to be validated, a small ask regarding the tests for consistency.

Also, can you add a [release note](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/CONTRIBUTING.md#adding-a-new-release-note) (at `release-notes/unreleased/2759.bug.rst`) describing the fix (mostly mentioning that ``NoiseLearner.run()`` now handles single pass iterators correctly)?

test/unit/test_noise_learner.py
rajnisht7: updated the test as per your suggestion

qiskit_ibm_runtime/noise_learner/noise_learner.py
rajnisht7: Thank you for the feedback, have updated the implementation to use single list comprehension as you suggested. It now handles mixed inputs correctly.

diego-plan9: Thanks - the suggestions were aimed at providing more context to @SamFerracin on the check overall, so deferring on his input on whether to keep the change and how to best tackle the topic overall.

test/unit/test_noise_learner.py
```diff
@@ -81,13 +81,27 @@ def test_run_program_inputs(self, task_type, options_type):
 
     @combine(task_type=["circs", "pubs"])
```
diego-plan9:
```suggestion
    @combine(task_type=["circs", "pubs", "circs_iterator", "pubs_iterator"])
```

As otherwise the new conditions will not be reached (`combine` is a custom decorator defined [here](https://github.com/rajnisht7/qiskit-ibm-runtime/blob/c106e30bb067135cf57741b6245ff88a22d66408/test/utils.py#L376), which will create a separate test per each `task_type` in this case).

SamFerracin (dismissed):
Thanks, @rajnisht7! 
This looks like the correct behaviuor. Thank you for fixing this!

test/unit/test_noise_learner.py
```diff
@@ -79,15 +79,29 @@ def test_run_program_inputs(self, task_type, options_type):
         expected["support_qiskit"] = True
         self.assertEqual(input_params["options"], expected)
 
-    @combine(task_type=["circs", "pubs"])
+    @combine(task_type=["circs", "pubs", "circs_iterator", "pubs_iterator"])
     def test_run_program_inputs_with_default_options(self, task_type):
-        """Test a circuit with default options."""
+        """Test a circuit with default options.
+        Also tests support for generators/iterators
```
diego-plan9:
```suggestion
        """Test a circuit with default options.

        Also tests support for generators/iterators.
```

diego-plan9 (changes requested):
Thanks both! It seems nearly ready - I added a couple of suggestions that are a result of the [pre-commit](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/CONTRIBUTING.md#style-guide) check (which is run in CI, and it is recommended to run locally).

test/unit/test_noise_learner.py
```diff
@@ -79,15 +79,29 @@ def test_run_program_inputs(self, task_type, options_type):
...
+        """
         backend = get_mocked_backend()
 
+        def circuit_generator(circuits):
+            for c in circuits:
+                yield c
```
diego-plan9:
```suggestion
            yield from circuits
```

rajnisht7:
@diego-plan9 thank you for the feedback had made the changes and also verified locally using `ruff check test_noise_learner.py` all checks passed

diego-plan9 (approved):
Great, thanks for the contribution! This was a non-obvious bug and having this nice fix in place will help `NoiseLearner` users.

Issue #2738, opened 2026-04-17
Tracking of Deprecation removals status


diego-plan9:
**What is the expected feature or enhancement?**

This issue acts as a tracker for deprecations that have been started and are awaiting full removal:

| Deprecation | PR | Since | Minimal removal date | Minimal removal version |
| --- | --- | --- | --- | --- |
| `RuntimeOptions` class | #2435 | 0.43.0 | 2026-01-14 | 0.45.0 |
| `IBMFractionalTranslationPlugin` | #2366 | 0.42.0 | 2025-12-25 | 0.44.0 |
| `DynamicCircuitInstructionDurations` class (and `duration` args) | #2403 | 0.43.0 | 2026-01-14 | 0.45.0 |

The goal is to have an up to date convenient way for keeping track and making decisions on full removal, subject to the [deprecation policy](https://github.com/Qiskit/qiskit-ibm-runtime/blob/sampler-preview/DEPRECATION.md) (in particular, being conservative when removing). 

**Acceptance criteria**

PRs pointing to this issue remove at least one entry from the list.

PR #2761, opened 2026-04-27
Move executor/nlv3 options to own module, revise nlv3 options

diego-plan9:
### Summary

Move the options for the new programs (`executor`, `nlv3`) to a separate `qiskit_ibm_runtime.options_models` module:
* introduce and document the new module, which contains just the options needed for this program (and avoids the `@primitive_dataclass`-based approach
* prefer pydantic `Annotation`-based approach for validation 
* avoid using `RuntimeOptions` in these programs, as `EnvironmentOptions` is a subset of the attributes actually used (except `backend`)
* in `NoiseLearnerV3`:
  * use `ExecutionOptions` for `init_qubits` and `rep_delay`, adjusting the converters accordingly
  * remove `SimulatorOptions`, as it does not seem to be supported currently (and the class requires more massaging)
  * use `EnvironmentOptions` from `options_v3` (same as `ExecutorOptions` is using)

Additionally (and contributing a bit to the noise):
* use `if TYPE_CHECKING` and run `isort` in the nlv3 and options_v3 modules
* remove some obsolete files and code 

### Details and comments

This strives for making the differentiation between current options and the options for new programs more evident. In `executor` and `nlv3`, we steer away from `Unset` and some other mechanisms - and the plan is to do the same in upcoming features (`samplerwrapper`, `samplerestimator`). The new module allows for getting that separation via module naming (instead of via prefixes in the class names, which in this particular case is not possible without jumping to `V3` due to `options.ExecutionOptionsV2` and `options.EnvironmentOptions` already existing and in use).

Related to #2743 (preparing of nlv3/executor releases), #2738 (related to `RuntimeOptions`)


qiskit_ibm_runtime/options_models/noise_learner_v3_options.py
```diff
@@ -80,20 +66,8 @@ class NoiseLearnerV3Options:
     These options are subject to change without notification, and stability is not guaranteed.
     """
 
-    max_execution_time: int | None = None
+    execution: ExecutionOptions = Field(default_factory=ExecutionOptions)
+    """Low-level execution options."""
+
     environment: EnvironmentOptions = Field(default_factory=EnvironmentOptions)
-    simulator: SimulatorOptions = Field(default_factory=SimulatorOptions)
```
diego-plan9, line 85: The removal of `.simulator` is due to:
* it [seems](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py#L86-L87) they cannot actually be used
* existing `options.SimulationOptions` is a bit different than the others (in particular, using `CouplingMap` as one of the type hints, which seems to require `arbitrary_types_allowed=True` (defeating a bit the purpose of the new classes) and requires a bit of time

If the first assumption is correct, I'm arguing for introducing it later (if/when local mode is supported), as an additive change.

qiskit_ibm_runtime/options_models/environment_options.py
```diff
@@ -74,19 +62,13 @@ class EnvironmentOptions:
     this time limit, it is forcibly cancelled.
     """
 
-    image: str | None = None
+    image: (
+        Annotated[
```
diego-plan9, line 66: This is the same validation that is at https://github.com/Qiskit/qiskit-ibm-runtime/blob/7a23ec5926a171b5c7b47dc2c4571f4293475c5d/qiskit_ibm_runtime/runtime_options.py#L96

qiskit_ibm_runtime/executor.py
```diff
@@ -112,16 +111,6 @@ def __setattr__(self, name: str, value: Any) -> None:
...
 
-    def _runtime_options(self) -> RuntimeOptions:
-        return RuntimeOptions(
-            backend=self._backend.name,
-            image=self.options.environment.image,
-            job_tags=self.options.environment.job_tags,
-            log_level=self.options.environment.log_level,
-            private=self.options.environment.private,
-            max_execution_time=self.options.environment.max_execution_time,
-        )
```
diego-plan9, line 123: This (and the equivalent at `NoiseLearner`) takes advantage that all the fields are actually in `EnvironmentOptions` (except for `backend`), so no need for a dedicated transformation to `RuntimeOptions` is needed. 

The PR actually started as an attempt to remove `RuntimeOptions` (ie. #2738 ). This changes allows to get closer (and have the new programs not depend on `RuntimeOptions`). I added some notes to the `RuntimeOptions` class to facilitate continue tackling the problem in a separate PR.

qiskit_ibm_runtime/noise_learner_v3/converters/version_0_2.py
```diff
@@ -60,9 +74,14 @@ def noise_learner_v3_inputs_from_0_2(
...
     instructions = list(model.instructions.to_quantum_circuit())
 
-    options = NoiseLearnerV3Options(
-        **{key: val for key, val in model.options.model_dump().items() if val is not None}
+    # Convert `model.options` to dict, moving the fields that are part of `options.execution` from
+    # top-level.
+    top_level_dump = model.options.model_dump(exclude_none=True, exclude=EXECUTION_FIELDS)
+    top_level_dump["execution"] = model.options.model_dump(
+        exclude_none=True, include=EXECUTION_FIELDS
     )
```
diego-plan9, line 82: Would appreciate some double-checking of this logic compared to the previous one (here I'm using `model_dump` arguments for seemingly achieving the same as before, plus some separation of specific fields).

qiskit_ibm_runtime/runtime_options.py
```diff
@@ -55,13 +55,13 @@ class RuntimeOptions:
     """
 
     backend: str | Backend | None = None
-    image: str | None = None
+    image: str | None = None  # TODO: can come from `options` or `options.experimental`
```
SamFerracin, line 58: these TODOs sound about right!

qiskit_ibm_runtime/noise_learner_v3/converters/version_0_2.py
SamFerracin: Yes, this looks fine

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -141,15 +133,14 @@ def run(self, instructions: Iterable[CircuitInstruction]) -> RuntimeJobV2:
...
+        params = converter.encoder(instructions, self.options)
+        runtime_options = asdict(self.options.environment)  # type: ignore[call-overload]
+        runtime_options["backend"] = self._backend.name
 
         if self._session:
-            run = self._session._run
+            _run = self._session._run
         else:
-            run = self._service._run
-            runtime_options.instance = self._backend._instance
```
SamFerracin, line 152: is this line gone intentionally?

```
runtime_options.instance = self._backend._instance
```

qiskit_ibm_runtime/executor.py
```diff
@@ -137,13 +126,13 @@ def run(self, program: QuantumProgram) -> RuntimeJobV2:
...
-        runtime_options = self._runtime_options()
+        runtime_options = asdict(self.options.environment)  # type: ignore[call-overload]
+        runtime_options["backend"] = self._backend.name
 
         if self._session:
-            run = self._session._run
+            _run = self._session._run
         else:
-            run = self._service._run
-            runtime_options.instance = self._backend._instance
```
SamFerracin, line 146: same comment as above

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
diego-plan9: It was in the original scope of the PR (which was centered around removing `RuntimeOptions`). I am quite certain `.instance` it is not used, but until being 100%, it indeed makes more sense to keep it and tackle it when revising `RuntimeOptions` fully at #2738 .

So reintroduced at f7b9f947

qiskit_ibm_runtime/executor.py
diego-plan9: https://github.com/Qiskit/qiskit-ibm-runtime/pull/2761#discussion_r3152628391

qiskit_ibm_runtime/noise_learner_v3/converters/version_0_1.py
```diff
@@ -14,24 +14,30 @@
 
 from __future__ import annotations
 
-from collections.abc import Iterable
 from dataclasses import asdict
+from typing import TYPE_CHECKING
```
SamFerracin, line 18: Why all these changes to import statements? Is it the that, for example, placed
```
from ibm_quantum_schemas.common import F64TensorModel, QpyModelV13ToV16
```
in line 20?

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -141,15 +133,15 @@ def run(self, instructions: Iterable[CircuitInstruction]) -> RuntimeJobV2:
         except KeyError:
             raise ValueError(f"No converters for schema version {self._SCHEMA_VERSION}.")
 
-        inputs = converter.encoder(instructions, self.options).model_dump()
-        inputs["version"] = 3
-        runtime_options = self._runtime_options()
+        params = converter.encoder(instructions, self.options)
+        runtime_options = asdict(self.options.environment)  # type: ignore[call-overload]
```
SamFerracin, line 137: Slight preference to keep these inside a `_runtime_options` (with or without underscore) function that returns a dict, so that we can add dedicated tests

qiskit_ibm_runtime/options_models/environment_options.py
SamFerracin: Nice!

qiskit_ibm_runtime/options_models/noise_learner_v3_options.py
SamFerracin: 👍

qiskit_ibm_runtime/options_models/post_selection_options.py
```diff
@@ -61,12 +60,3 @@ class PostSelectionOptions:
 
     Defaults to ``node``.
     """
-
-    @staticmethod
-    def _get_program_inputs(options: dict) -> dict:
```
SamFerracin, line 66: Ugh.. This was ugly

SamFerracin (approved):
Thanks! Took me a while, but it seems correct!

I left some small comments here and there, but nothing blocking.

One thing that this PR doesn't do is adding tests for options, which by the way were not present before either. There are some small little things that we could think of testing, for example:
- Are the environment options willing to accept a custom image of the correct format, or did we screw up the regex part?
- `_runtime_options` if we decide to leave it as a method

But we can think about this and have a followup PR to add coverage, if necessary

diego-plan9:
> One thing that this PR doesn't do is adding tests for options, which by the way were not present before either:

Actually there was some testing for options (most of them added in the past couple of weeks) - and it has been kept that way, via the `test_executor.py` and `test_noise_learner_v3.py` (which actually provide non-explicit test for the `image` validation, as in [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2761/changes#diff-9eb0c87a0ca210245d0103ca67b147f4e56bdf1378ce5685630dfc7e9e9ec308R74)). The "shared" options (`ExecutionOptions` and `EnvironmentOptions`) are indeed less tested, I'll evaluate how well covered and issue a follow-up accordingly.

qiskit_ibm_runtime/noise_learner_v3/converters/version_0_1.py
diego-plan9: I ran ruff's checker for `TYPE_CHECKING` and `isort` over the `noise_learner_v3` and `option_models` folders:
> Additionally (and contributing a bit to the noise):
> *  use if TYPE_CHECKING and run isort in the nlv3 and options_v3 module 

This is related to #2563 and #2749 - we cannot apply both checkers yet to all files automatically as it will result in errors unless some things are cleaned up first, so I took the opportunity to at least apply them to the modules and saves us a tiny bit of potential import cycles.

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
diego-plan9: I'll evaluate while working on the PR for revising `RuntimeOptions`. For new-style options, `RuntimeOption` is arguably `EnvironmentOptions` + `backend` (pending evaluating all fields to fully see if they are used). If that is the case, it might be that tests over `EnvironmentOptions` suffice.

PR #2762, opened 2026-04-28
Configurable poll_interval parameter in RuntimeJobV2

### LINKED ISSUE #2755: Reduce default RuntimeJob status polling interval

The `wait_for_final_status` method polls the `GET /v1/jobs/:id` API every 100ms waiting for a terminal status for a job:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/runtime_job_v2.py#L244

That puts a lot of load on multiple API services in the Quantum Cloud stack, and additional infra resources, e.g. tracing, monitoring, logging.

We'd like to see that hard polling lessened by default for non-session jobs but with a way to opt into faster polling (previous behavior) if desired, like for automated integration tests.

**Acceptance criteria**

- Change the default for non-session jobs to be 500ms or 1s for the polling interval
- Allow the polling interval to be overridden (kwarg on the QiskitRuntimeService instance and/or RuntimeJob and/or env var) with a **floor** value of 100ms
- For Session jobs, leave the polling interval default to be 100ms because we don't want longer polling times to impact wall lock usage calculations on sessions

diego-plan9:
### Summary

Add a `poll_interval` argument to `RuntimeJobV2` `.result()` and `.wait_for_final_state()`, which allows configuring the interval between calls to the API for getting the status of a job. The defaults have been tweaked to allow for sensible defaults, and enforce lower bounds.

### Details and comments

Fixes #2755 

This changes the default poll interval for non-session jobs, but I lean towards considering the specific default value to be an implementation detail (rather than a user-facing feature) - hence no need to go through a full deprecation cycle.


qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -107,12 +107,18 @@ def result(
...
         timeout: float | None = None,
         decoder: type[ResultDecoder] | None = None,
+        poll_interval: float | None = None,
     ) -> Any:
         """Return the results of the job.
 
         Args:
             timeout: Number of seconds to wait for job.
             decoder: A :class:`ResultDecoder` subclass used to decode job results.
+            poll_interval: Number of seconds to wait between querying the service for the status
```
yaelbh: Since we talk all the time in milliseconds, consider changing it to an integer and make it the number of milliseconds to wait between queries. It may be more natural for the user.

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -107,12 +107,18 @@ def result(
...
+                of the job.
+
+                * For non-session jobs, the default is ``500ms``, and the floor value is ``100ms``.
+                * For session jobs, the default and the floor value is ``100ms``.
```
yaelbh:
```suggestion
                * For session jobs, the default and the floor value are ``100ms``.
```

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -223,15 +229,31 @@ def logs(self) -> str:
...
 
         Raises:
             RuntimeJobTimeoutError: If the job does not complete within given timeout.
         """
+        # Calculate the poll interval.
+        min_poll_interval = 0.1 if self._session_id else 0.5
```
yaelbh: Not sure, but such hard-coded constants (0.1 and 0.5) are often defined separately, in some file dedicated for constant values, or at least with capital letters.

release-notes/unreleased/2762.upgrade.rst
```diff
@@ -0,0 +1,4 @@
+The ``RuntimeJobV2.result()`` and ``RuntimeJobV2.wait_for_final_state()`` methods now accept an
+optional ``poll_interval`` parameter that allows customizing the interval between request to the
```
yaelbh:
```suggestion
optional ``poll_interval`` parameter, that allows customizing the interval between requests to the
```

yaelbh (changes requested):
1. I suspect that there's a bug: for non-session jobs, `min_poll_interval` is `0.5`, so intervals from 0.1 to 0.5 are not allowed (i.e., are converted to `0.5` and emit a warning).
2. Tests are missing.
3. It will be good to obtain Matt's confirmation about the interface, that it matches his expectations as the author of the issue (the interface looks very good to me).

qiskit_ibm_runtime/runtime_job_v2.py
diego-plan9: Can you define "all the time"? In the codebase we seem to use seconds as the most common unit (as a close example, in the `timeout` argument itself), but I did not measure scientifically the number of times we use seconds vs milliseconds. 

If seconds is indeed more popular, might make sense to adjust the docstring instead, so it uses seconds?

qiskit_ibm_runtime/runtime_job_v2.py
diego-plan9: Those are constants, but only used in the function (and `min_poll_interval` is not a constant itself). Keeping them localized and close to the code tends to reduce cognitive overhead.

diego-plan9:
> 1. I suspect that there's a bug: for non-session jobs, `min_poll_interval` is `0.5`, so intervals from 0.1 to 0.5 are not allowed (i.e., are converted to `0.5` and emit a warning).

This would be a feature rather than a bug, #2755? We do want to not allow the user to specify lower values.

>     2. Tests are missing.

Ultimately, this PR is changing the value passed to `time.sleep()` in a function. I will see if I can include some tests related to 1 at least.

>     3. It will be good to obtain Matt's confirmation about the interface, that it matches his expectations as the author of the issue (the interface looks very good to me).

Indeed - cc @mriedem

qiskit_ibm_runtime/runtime_job_v2.py
yaelbh: :smile: By "all the time" I was thinking about the issue, where the discussion was about 100ms, 500ms, etc., and the docstrings indeed. On the other hand, I agree that consistency is important, and also there's a value in working with standard units (this was the policy applied in qiskit-experiments, where all the very short times are nevertheless measured in seconds). So I'm convinced to leave things as they are, I wouldn't modify the docstrings either.

diego-plan9:
> > 1. I suspect that there's a bug: for non-session jobs, `min_poll_interval` is `0.5`, so intervals from 0.1 to 0.5 are not allowed (i.e., are converted to `0.5` and emit a warning).
> 
> This would be a feature rather than a bug, #2755? We do want to not allow the user to specify lower values.

@yaelbh , thanks for bringing awareness as I misinterpreted your comment - indeed, the PR is not using the right floor and it is a bug - I will fix!

yaelbh (dismissed):
For completeness, it will be nice to test also the case of a non-default interval which is greater than `0.1`. It should be very easy, given the existing tests, something like:
```python
def test_wait_for_final_state_non_default_poll_interval(self, service):
        """some docstring"""
        job = run_program(service)
        with patch.object(RuntimeJobV2, "status", side_effect=["WAITING", "DONE"]):
            with patch("time.sleep", return_value=None) as patched_sleep:
                with self.assertLogs("qiskit_ibm_runtime", level="WARNING") as logs:
                    job.wait_for_final_state(poll_interval=1.1)
                    patched_sleep.assert_called_with(1.1)
```

Anyway I'm approving, this missing test is not important enough to block approval.

diego-plan9:
Thanks! Revised tests in 0a6f481 - and revised so a `warnings.warn` is emitted instead of a log entry with the warning level, to make it more visible and such usage is more of an user error.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1040,6 +1040,7 @@ def _run(
             program_id=program_id,
             result_decoder=result_decoder,
             image=qrt_options.image,
+            session_id=session_id,
```
SamFerracin, line 1049: had this been forgotten in another PR?

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -107,12 +108,18 @@ def result(
...
         timeout: float | None = None,
         decoder: type[ResultDecoder] | None = None,
+        poll_interval: float | None = None,
     ) -> Any:
         """Return the results of the job.
 
         Args:
             timeout: Number of seconds to wait for job.
             decoder: A :class:`ResultDecoder` subclass used to decode job results.
+            poll_interval: Number of seconds to wait between querying the service for the status
```
SamFerracin: I find this a bit more clear:
```suggestion
            poll_interval: Number of seconds to wait between successive queries of the job's status.
```

qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -107,12 +108,18 @@ def result(
...
+                of the job.
+
+                * For non-session jobs, the default is ``500ms``, and the floor value is ``100ms``.
```
SamFerracin, line 122: Why this difference? Is this because session jobs normally take longer, and so 500ms isn't a big deal?

SamFerracin (commented):
Looks good. Some minor comments and questions

qiskit_ibm_runtime/qiskit_runtime_service.py
diego-plan9: Actually this is on purpose - and I should have mentioned either as a note in the review, or in the code itself.

This (combined with using `self._session_id`, ie. accessing the attribute directly) aims to prevent an extra API query to all jobs originating here, at the `Job.session_id()` call: https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.46.1/qiskit_ibm_runtime/base_runtime_job.py#L339.

The goal is that the `job.session_id` is populated right after being created by `Session._run()`. As far as I could tell, the `session_id` is propagated correctly, it was just not added to the `Job` constructor.

qiskit_ibm_runtime/runtime_job_v2.py
diego-plan9: This is per the specs in the issue https://github.com/Qiskit/qiskit-ibm-runtime/issues/2755:
> For Session jobs, leave the polling interval default to be 100ms because we don't want longer polling times to impact wall lock usage calculations on sessions

qiskit_ibm_runtime/qiskit_runtime_service.py
SamFerracin: tested, it seems to work correctly

SamFerracin (approved):

mriedem:
Sorry I'm late to this, but I was wondering if it'd be possible to provide an environment variable to set the value?

I'm thinking about an integration test environment (argo workflows) where tests are calling `job.wait_for_final_state()` and we don't want to have to re-write all of those tests to override the polling interval to set it back to the 100ms value that was being used before.

edit: before making any changes, we're discussing if we actually need this or if we just see how our integration tests work after this change and then see if we need to adjust some tests on a per-test basis.

PR #2766, opened 2026-04-28
[Sampler-preview] Update SamplerV2's prepare and post-process so that they use semantic value

SamFerracin:


qiskit_ibm_runtime/quantum_program/converters/converters_1_0.py
```diff
@@ -147,6 +147,7 @@ def quantum_program_to_1_0(program: QuantumProgram, options: ExecutorOptions) ->
             items=model_items,
             meas_level=program.meas_level,
             passthrough_data=program.passthrough_data,
+            semantic_role=program._semantic_role,
```
SamFerracin, line 150: this is a bug in main, I'll fix it

TsafrirA (approved):
LGTM.

PR #2767, opened 2026-04-28
Fix a bug in executor's converters to "v1.0" models

SamFerracin:
This PR fixes a bug where the executor's converters to "v1.0" models forget to set the `_semantic_role` of the quantum program

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/converters/converters_1_0.py
+++ b/qiskit_ibm_runtime/quantum_program/converters/converters_1_0.py
@@ -147,6 +147,7 @@ def quantum_program_to_1_0(program: QuantumProgram, options: ExecutorOptions) ->
             items=model_items,
             meas_level=program.meas_level,
             passthrough_data=program.passthrough_data,
+            semantic_role=program._semantic_role,
         ),
         options=options_dict,
     )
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/quantum_program/converters/converters_1_0.py
```diff
@@ -147,6 +147,7 @@ def quantum_program_to_1_0(program: QuantumProgram, options: ExecutorOptions) ->
             items=model_items,
             meas_level=program.meas_level,
             passthrough_data=program.passthrough_data,
+            semantic_role=program._semantic_role,
```
ihincks, line 150: will private access cause lint error?

ihincks (approved):

qiskit_ibm_runtime/quantum_program/converters/converters_1_0.py
SamFerracin: No, which surprised me too.
cc @diego-plan9, I wonder if we forgot a rule in Ruff

diego-plan9: There is https://docs.astral.sh/ruff/rules/#flake8-self-slf - but we make heavy use of accessing private members:

```sh
$ ruff check --select SLF qiskit_ibm_runtime
[...]
Found 69 errors.
```

So we are not in a good position for enabling it without some refactorings, and I lean towards just considering we don't follow the private vs public underscore convention at this point.

PR #2769, opened 2026-04-29
Add workflow for testing minimal versions of dependencies

### LINKED ISSUE #2575: Run CI tests against minimal version of Qiskit supported

Currently, CI runs the tests against a combination of options (Python version x platform), and against the head of the Qiskit repo. It would be good to also run the tests against the minimal version of Qiskit, to ensure compatibility and catch potential issues.

Open questions:
* does it make sense to extend the criteria and use the minimal versions of _all_ dependencies?
* what is the right frequency (on all PRs, daily)?

**Acceptance criteria**

A workflow exists that uses the minimal version of Qiskit supported (as declared by the package).

diego-plan9:
### Summary

Revise the lower bound of our dependencies and update the nightly workflow to run the unit tests against such versions (by making use of `extremal-python-dependencies` for modifying `pyproject.toml` files in place before installing the package).

### Details and comments

Having them run in the nightly cron job would not allow us to catch issues during regular PR flow, but it should be infrequent enough - this way the regular PR flow is not hindered.

Fixes #2575


pyproject.toml
```diff
@@ -140,14 +140,14 @@ requires-python=">=3.10"
...
     "requests>=2.32.4",
     "requests-ntlm>=1.1.0",
-    "numpy>=1.26",
+    "numpy>=1.26.2",
     "urllib3>=2.4.0",
     "python-dateutil>=2.9.0",
     "ibm-platform-services>=0.61.1",
     "ibm-quantum-schemas>=0.7.20260419,<1",
-    "pydantic>=2.7.0",
+    "pydantic>=2.11.0",
```
diego-plan9, line 148: This is forced by the `ibm-quantum-schemas` dependency (https://github.com/Qiskit/ibm-quantum-schemas/pull/123).

pyproject.toml
```diff
@@ -211,7 +211,7 @@ visualization = [
 
 performance = [
     "qiskit-aer>=0.17.0",
-    "scipy>=1.0"
+    "scipy>=1.15",
```
diego-plan9, line 214: This is due to `scipy` not being compatible with Python 3.13 until `1.14` - and not including `cp313` wheels until `1.15`. 
https://docs.scipy.org/doc/scipy/dev/toolchain.html#numpy
https://pypi.org/project/scipy/1.15.0/#files

pyproject.toml
```diff
@@ -227,18 +227,18 @@ style_local = [
...
 doc = [
     "autodoc_pydantic==2.2.0",
-    "matplotlib>=2.1",
-    "nbconvert>=5.3.1",
-    "nbformat>=4.4.0",
+    "matplotlib>=3.10",
+    "nbconvert>=6.4.3",
+    "nbformat>=5.0.2",
     "nbqa==1.5.3",
     "nbsphinx",
```
diego-plan9, line 243: These are quite related - took a bit of trial and error. It might be possible to go lower, but since there are mostly for docs rendering, it seemed sensible not to dig further.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools >= 65.0.2", "setuptools_scm[toml]>=6.2"]
+requires = ["setuptools >= 66.1.0", "setuptools_scm[toml]>=8.1.0"]
```
diego-plan9, line 2: For `setuptools_scm`, we need a version that supports `version_file` (8.0.0) plus some fixes that landed later.
For `setuptools`, we need the `#3685` fix at https://setuptools.pypa.io/en/stable/history.html#v66-1-0

pyproject.toml
```diff
@@ -140,14 +140,14 @@ requires-python=">=3.10"
 dependencies = [
     "requests>=2.32.4",
     "requests-ntlm>=1.1.0",
-    "numpy>=1.26",
+    "numpy>=1.26.2",
```
diego-plan9, line 143: Small bump as `1.26.0` and `1.26.1` are not compatible with Python 3.13
```
1.26.0 Requires-Python >=3.9,<3.13; 1.26.1 Requires-Python >=3.9,<3.13
```

.github/workflows/ci.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           pip install -U tox
           pip install nbqa docutils
           sudo apt install -y graphviz pandoc
-          wget https://github.com/errata-ai/vale/releases/download/v2.23.0/vale_2.23.0_Linux_64-bit.tar.gz
+          wget https://github.com/vale-cli/vale/releases/download/v2.23.0/vale_2.23.0_Linux_64-bit.tar.gz
```
diego-plan9, line 58: I picked this along the way - basically we were being redirected all the time as it seems the `vale` org changed at some point.

.github/workflows/integration-tests.yml
```diff
@@ -87,3 +87,47 @@ jobs:
...
+      matrix:
+        python-version: ['3.10', '3.11', '3.12', '3.13']
+        os: [ "macos-latest", "ubuntu-latest", "windows-latest" ]
+        exclude:
+          - os: "macos-latest"
+            python-version: '3.10'
+            # numpy 1.26.x does not provide cp313 wheels, and mingw-w64 compiled numpy raises
+            # RuntimeWarning at test/unit/test_data_serialization::TestDataSerialization
+          - os: "windows-latest"
+            python-version: '3.13'
```
diego-plan9, line 104: Due to https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/25108542831/job/73576278596

diego-plan9:
An example of a successful run is at https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/25110976410 (that one was run in the CI workflow for testing purposes, as of the latest commit it will be run in the nightly run).

pyproject.toml
```diff
@@ -227,18 +227,18 @@ style_local = [
 ]
 
 test = [
-    "ddt>=1.2.0,!=1.4.0,!=1.4.3",
```
SamFerracin, line 230: wow!

SamFerracin (approved):
This makes sense to me. One possible improvement may be to add a mechanism to notify us when the build fails, for example via Slack notifications

Nightly builds can be quite useful. We can think if there are other usecases for them that would be helpful to us in qiskit-ibm-runtime!

diego-plan9:
Thanks! This actually just add a stage to the existing integration tests workflow (I referred to it as "nightly" as it is run daily via a cron job). But we could give some thought to the idea of a nightly build indeed in the future!

PR #2770, opened 2026-04-29
fix: return type annotation of `_get_inputs_options` from `dict[str, str]` to `dict[str, Any]`

rajnisht7:
### Summary

this PR fixes incorrect return type annotation in `NoiseLearner._get_inputs_options`, currently it is returning `dict[str, str]` but  actual returned values include `bool`, `int`, `list`, and `dict` never exclusively `str`.

### Details and comments
Fixes # 

current
```
def _get_inputs_options(options_dict: dict[str, Any]) -> dict[str, str]:
```
fix
```
def _get_inputs_options(options_dict: dict[str, Any]) -> dict[str, Any]:
```


test/unit/test_noise_learner.py
```diff
@@ -146,3 +148,18 @@ def test_get_backend(self):
         backend = get_mocked_backend()
         inst = NoiseLearner(backend)
         self.assertEqual(inst.backend().name, backend.name)
+
+    def test_get_inputs_options_return_annotation(self):
```
diego-plan9: Thanks for the effort on the test - however, for type annotations, it is sufficient that `mypy` (which is run by CI and by the recommended contributor setup) passes. If `mypy` passes (combined with getting reviewer approval), we have enough assurances that the hints are working correctly, and the extra tests don't contribute to those assurances (plus, they need to be maintained). Could you please remove the test?

Also, can you comment on whether you used an LLM or other AI tool as part of this PR?

rajnisht7: havent used any LLM

rajnisht7: also removed the test

rajnisht7:
@diego-plan9 also wanted to ask one thing that in PR #2759, missed to add validation for empty circuits list, so should i add that in this PR or file a new PR for that

test/unit/test_noise_learner.py
diego-plan9: Thanks on both fronts - the question about usage of AI is due to, if used, they need to be disclosed in the PR description. All good!

diego-plan9:
> @diego-plan9 also wanted to ask one thing that in PR #2759, missed to add validation for empty circuits list, so should i add that in this PR or file a new PR for that

In general, the recommended flow is to open an issue describing the bug / feature / enhancement so the maintainers can comment on the need, prioritize/plan if needed, and provide feedback before starting work (through a PR). This flow is mainly a communication / coordination vehicle, for maximizing alignment and preventing duplicated (or extra) work, and is to be taken as a general rule (it is fine to deviate in specific occasions such as smaller bits as this PR or your previous one!).

Regarding adding validation for empty circuits - if the intent is to have a check inside `NoiseLearner.run()` raising an exception if no circuits are passed, I'd actually suggest _not_ going for it.

It might seem counter-intuitive (and we are not too consistent in the codebase in that regard) when it comes to usability, but the approach is placing trust in the users and guiding through type hints and documentation for the function's behavior and meaning of arguments in general. Fine-grained and abundant validation has a maintenance cost (ie. it can be argued that if there is a check for empty pubs, we should also check for the validity of every item, for extra conditions, to extend it to all primitives, etc) that has a chance of restricting users' freedom and extensibility without providing too much value _if_ we are able to guide the user correctly through other means.

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -246,7 +246,7 @@ def _set_options(
             self._options = NoiseLearnerOptions(**options)
 
     @staticmethod
-    def _get_inputs_options(options_dict: dict[str, Any]) -> dict[str, str]:
+    def _get_inputs_options(options_dict: dict[str, Any]) -> dict[str, Any]:
```
diego-plan9, line 249: Thanks! It is a bit unfortunate that we lose specificity, but this was indeed incorrect and ...

> but actual returned values include bool, int, list, and dict never exclusively str.

the list of allowed types extends actually further, so `Any` is a good choice (coming up with the full list would not help too much when it comes to typing tools as they are not able to dig that deep, and it would be brittle to maintain as it depends on changes in other files).

Thanks!

diego-plan9 (approved):
Thanks!

As a suggestion, and only if interested in pursuing this line of contributions to the extent of your choosing, there is also #2514 (with different degrees of involvement and meant to picked up collaboratively).

test/unit/test_noise_learner.py
rajnisht7: sure, if used will mention in description thank you

rajnisht7:
> > @diego-plan9 also wanted to ask one thing that in PR #2759, missed to add validation for empty circuits list, so should i add that in this PR or file a new PR for that
> 
> In general, the recommended flow is to open an issue describing the bug / feature / enhancement so the maintainers can comment on the need, prioritize/plan if needed, and provide feedback before starting work (through a PR). This flow is mainly a communication / coordination vehicle, for maximizing alignment and preventing duplicated (or extra) work, and is to be taken as a general rule (it is fine to deviate in specific occasions such as smaller bits as this PR or your previous one!).
> 
> Regarding adding validation for empty circuits - if the intent is to have a check inside `NoiseLearner.run()` raising an exception if no circuits are passed, I'd actually suggest _not_ going for it.
> 
> It might seem counter-intuitive (and we are not too consistent in the codebase in that regard) when it comes to usability, but the approach is placing trust in the users and guiding through type hints and documentation for the function's behavior and meaning of arguments in general. Fine-grained and abundant validation has a maintenance cost (ie. it can be argued that if there is a check for empty pubs, we should also check for the validity of every item, for extra conditions, to extend it to all primitives, etc) that has a chance of restricting users' freedom and extensibility without providing too much value _if_ we are able to guide the user correctly through other means.

thank you for the info and clarification and also will raise issue for significant changes.

rajnisht7:
> Thanks!
> 
> As a suggestion, and only if interested in pursuing this line of contributions to the extent of your choosing, there is also #2514 (with different degrees of involvement and meant to picked up collaboratively).

thank you, will check it out

PR #2776, opened 2026-05-04
Add `QuantumProgramItemResult` class

SamFerracin:
Currently, the results of a quantum program item are stored in a dictionary:
```python
result = job.result()
result[0] = {"meas": np.array(...)}
```

This format is not sufficient to store execution metadata for individual items.

This PR adds a `QuantumProgramItemResult` class that can store both the dictionary of results and the metadata. This class is an implementation of `MutableMappings`, which ensures full backward compatibility: `result[0]` still behaves like a dictionary, no script needs to change.

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/quantum_program_result.py
+++ b/qiskit_ibm_runtime/quantum_program/quantum_program_result.py
@@ -229,11 +229,16 @@ class QuantumProgramResult:
 
     def __init__(
         self,
-        data: list[dict[str, np.ndarray]],
+        data: list[dict[str, np.ndarray] | QuantumProgramItemResult],
         metadata: Metadata | None = None,
         passthrough_data: DataTree | None = None,
     ):
-        self._data = data
+        self._data = [
+            datum
+            if isinstance(datum, QuantumProgramItemResult)
+            else QuantumProgramItemResult(datum)
+            for datum in data
+        ]
         self.metadata = metadata or Metadata()
         self.passthrough_data = passthrough_data
 
@@ -242,10 +247,10 @@ class QuantumProgramResult:
         # without notice. Third party clients should not set or depend on this value.
         self._semantic_role: str | None = None
 
-    def __iter__(self) -> Iterator[dict[str, np.ndarray]]:
+    def __iter__(self) -> Iterator[QuantumProgramItemResult]:
         yield from self._data
 
-    def __getitem__(self, idx: int) -> dict[str, np.ndarray]:
+    def __getitem__(self, idx: int) -> QuantumProgramItemResult:
         return self._data[idx]
 
     def __len__(self) -> int:
--- a/qiskit_ibm_runtime/quantum_program/quantum_program_result.py
+++ b/qiskit_ibm_runtime/quantum_program/quantum_program_result.py
@@ -208,7 +208,7 @@ class QuantumProgramItemResult(MutableMapping):
     def __delitem__(self, key: str) -> None:
         del self._result[key]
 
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/quantum_program/quantum_program_result.py
```diff
@@ -176,6 +183,41 @@ def draw(
...
+    def __getitem__(self, key: str) -> np.ndarray:
+        return self._result[key]
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        self._result[key] = value
+
+    def __delitem__(self, key: str) -> None:
+        del self._result[key]
+
+    def __iter__(self) -> np.ndarray:
```
ihincks:
```suggestion
    def __iter__(self) -> str:
```

qiskit_ibm_runtime/quantum_program/quantum_program_result.py
```diff
@@ -176,6 +181,41 @@ def draw(
...
+        result: dict[str, np.ndarray],
+        metadata: ItemMetadata | None = None,
+    ):
+        self._result = result
+        self.metadata = metadata or ItemMetadata()
+
+    def __getitem__(self, key: str) -> np.ndarray:
+        return self._result[key]
+
+    def __setitem__(self, key: str, value: Any) -> None:
```
diego-plan9:
```suggestion
    def __setitem__(self, key: str, value: np.array) -> None:
```

For consistency with `__getitem__` and the declaration of `self._result`, right?

qiskit_ibm_runtime/quantum_program/quantum_program_result.py
```diff
@@ -187,11 +227,16 @@ class QuantumProgramResult:
 
     def __init__(
         self,
-        data: list[dict[str, np.ndarray]],
+        data: Sequence[dict[str, np.ndarray] | QuantumProgramItemResult],
```
diego-plan9, line 230: Do we need / want to still accept `Sequence[dict[str, np.ndarray]]`?

I think we are only instantiating `QuantumProgramResult` in the two converters. Accepting just a sequence of `QuantumProgramItemResult` would keep the `QuantumProgramResult` interface less ambiguous (and arguably a bit more clean), and move the responsibility of using the right objects to the instance creators. 

There are other places in the codebase where we make the constructor arguments lenient (converting them as we go), and this conversion is not too intrusive, so it is a valid angle for keeping this as-is though.

diego-plan9 (commented):
Thanks Sam! A couple of suggestions / questions, but should be almost ready.

qiskit_ibm_runtime/quantum_program/quantum_program_result.py
SamFerracin: [3e65a1e](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2776/commits/3e65a1e06161d1994cabe37a9df35c284b3e7019)

qiskit_ibm_runtime/quantum_program/quantum_program_result.py
SamFerracin: I have a slight preference for keeping the dual option, which 1) is non-breaking, 2) makes the constructor more convenient, and 3) is consistent with what we do in many parts of the repo (so, it's not a new addition). But happy to remove it if you have a strong preference

diego-plan9: I do have a preference, but not so strong - sounds good!

diego-plan9 (approved):

PR #2778, opened 2026-05-05
make local-mode ZNE (and other error mitigation) raise rather than warn-and-continue

### LINKED ISSUE #2777: Qiskit Aer Simulator with ZNE - notebook and script divergent behavior
### **Describe the bug**
I am trying to run ZNE Quantum Error Mitigation technique with Qiskit Aer Simulator via a noise model retrieved from a backend "ibm_kingston".

**Running the code as a python script normally on the terminal vs a python notebook has distinct behavior**:

Local mode silently ignores `resilience.zne_mitigation`. The script appears to terminate only because `job.result()` is never awaited and the interpreter exits before Aer's background workers finish; in a Jupyter kernel those workers persist and continue the (very large, 30-qubit, noisy, mirrored Trotter) simulation. So the bug is that `Estimator.run()` in local mode returns a job whose worker lifecycle isn't tied to the parent process, which makes the "script vs notebook" behavior diverge confusingly.

The python code for the same is added below:

[zne.py](https://github.com/user-attachments/files/27385887/zne.py)

The output of the above code:
```bash
API Key: <hidden>
qiskit_runtime_service._discover_account:WARNING:2026-05-05 00:33:43,541: Loading account with the given token. A saved account will not be used.
qiskit_runtime_service.__init__:WARNING:2026-05-05 00:33:45,794: Instance was not set at service instantiation. Free and trial plan instances will be prioritized. Based on the following filters: (tags: None, region: us-east, eu-de), and available plans: (open), the available account instances are: open-instance. If you need a specific instance set it explicitly either by using a saved account with a saved default instance or passing it in directly to QiskitRuntimeService().
qiskit_runtime_service.backends:WARNING:2026-05-05 00:33:45,794: Using instance: open-instance, plan: open
/home/ayush268/.pyenv/versions/qem_profiler-0.2/lib/python3.12/site-packages/qiskit_ibm_runtime/fake_provider/local_service.py:274: UserWarning: Options {'resilience': {'zne_mitigation': True, 'zne': {'noise_factors': [1.0, 3.0, 5.0], 'extrapolator': ('exponential', 'linear')}}} have no effect in local testing mode.
  warnings.warn(f"Options {options_copy} have no effect in local testing mode.")
Job ID: <hidden>
```

---

### **Steps to reproduce**

Please run the above python code as the following python notebook:
(Note: Replace token with your api-key for ibm quantum cloud)

[zne_ibm_with_bug.ipynb](https://gist.github.com/ayush268/2ed66963adcd30aa83a0784829f0a5be)

The above notebook will seemingly terminate, but still it will spawn a lot of threads running the simulation unlike graceful termination of the python script. Please see attached screenshot:

Image

I have to manually restart the notebook (kill everything associated) to stop this. Interrupting the notebook also does not stop all the threads running the simulation.

---

### **Expected behavior**

Ideally, the notebook should terminate as the script does or both script and notebook should also start a simulation and block on it, the behavior should not be different.

---

### **Suggested solutions**

A clearer fix on the library side would be to either:

1. #2778 , or
2. ensure local-mode workers are daemonized / cancellable so they don't outlive the foreground task.

---

### **Additional Information**

- **qiskit-ibm-runtime version**:  `0.46.1`
- **Python version**: `3.12.13`
- **Operating system**: `Arch Linux`; `6.19.11-arch1-1`

ayush268:
### Summary

Right now, anything in `options` that local mode doesn't understand falls through to a generic warning at line 273–274:

```
if options_copy:
    warnings.warn(f"Options {options_copy} have no effect in local testing mode.")
```

For option like `dynamical_decoupling`, "warn and continue" is reasonable. For resilience options that change the meaning of the result (ZNE, PEC, PEA, measurement twirling, gate twirling), silently dropping them is dangerous — the user thinks they got a mitigated estimate but didn't, and the bug is invisible unless they read the warning.

### Details and comments
Fixes #2777

### DIFF
--- a/qiskit_ibm_runtime/fake_provider/local_service.py
+++ b/qiskit_ibm_runtime/fake_provider/local_service.py
@@ -218,6 +218,30 @@ class QiskitRuntimeLocalService:
         """
         options_copy = copy.deepcopy(options)
 
+        # Resilience options materially change the *meaning* of estimator results.
+        # Local mode does not implement them, so silently dropping them produces
+        # misleading output. Raise instead of warn for these.
+        if primitive == "estimator":
+            resilience = options_copy.get("resilience", {})
+            unsupported = [
+                name for name in (
+                    "zne_mitigation",
+                    "pec_mitigation",
+                    "measure_mitigation",
+                    "measure_noise_learning",
+                    "layer_noise_learning",
+                    "zne",
+                    "pec",
+                )
+                if resilience.get(name)
+            ]
+            if unsupported:
+                raise ValueError(
+                    f"The following resilience options are not supported in local "
+                    f"testing mode: {unsupported}. To use these mitigations, run on "
+                    f"a real backend via QiskitRuntimeService."
+                )
+
         prim_options = {}
         sim_options = options_copy.get("simulator", {})
         if seed_simulator := sim_options.pop("seed_simulator", None):
--- a/release-notes/notes/change.yaml
+++ b/release-notes/notes/change.yaml
@@ -1,0 +1,9 @@
+upgrade:
+  - |
+    :class:`~qiskit_ibm_runtime.EstimatorV2` now raises ``ValueError`` when
+    resilience options such as ``zne_mitigation`` or ``pec_mitigation`` are
+    set while running in local testing mode (i.e. with a non-cloud backend
+    such as :class:`~qiskit_aer.AerSimulator` or a fake backend). Previously
+    these options were silently dropped with only a generic warning, which
+    could lead users to believe their results were error-mitigated when they
+    were not.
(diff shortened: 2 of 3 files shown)


diego-plan9 (commented):
Thanks @ayush268 for the neat PR, including following the contribution guidelines.

[Local testing mode](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/runtime-service#local-testing-mode) (as in, using the simulators) is meant to be a way for validating a program before submitting it to a real quantum device (without expectations that the simulators will provide the same features as the real devices): in general, we strive for the user to only have to change the `mode` parameter to the program constructors (passing a fake device or a real device as needed without any extra changes) to switch between those.

As you noticed in the issue, there is a [warning](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.46.1/qiskit_ibm_runtime/fake_provider/local_service.py#L189-L190) being emitted, which provides a strong hint about the options not being taken into consideration. We count on users acting on those warnings as a way to communicate and highlight non-fatal messages that need user attention or intervention - and promoting the warning to an `Exception` hinders being able to switch transparently between real devices and simulators, which we strive to (and limits artificially the user's freedom).

Thanks again - and will follow-up in the issue for the simulator orphaning behavior.

PR #2779, opened 2026-05-05
Move WrapperSampler to top-level module

### LINKED ISSUE #2750: Promote WrapperSampler to higher up

In the current head of the branch, (wrapped) `SamplerV2` is nested as part of the `.executor.routines` hierarchy. This provides a piece of context (as in, a strong hint of the relation between this sampler an executor), but at the same time it produces arguably excessive nesting (mostly for developers, and for final import lengths).

As it will become arguably a first-class citizen, and the number of `routine`s is well defined and short (ie. 2) in the foreseeable future, this issue is about finding a place closer to the top-level for the functionality - making it easier to consume and to work with.

**Acceptance criteria**

`SamplerV2` is importable with less levels of nesting.

diego-plan9:
### Summary

Move `WrapperSampler` to `qiskit_ibm_runtime.sampler_v2`, along with revising other items:
* `.sampler_v2` becomes the module for "all items specific to wrappersampler that don't have a dedicated place"
   * `.sampler_v2.post_processors` contains the post_processor registry and the 0.1 post_processor, flat.
   * `.sampler_v2.converters` centralizes the converting of options and results
* `.executor` becomes the module for "the executor and items that are useful for programs based on executor"
* `.options_models` contains the WrapperSampler options (which now use `PRIMITIVES_CONFIG`)

Extra changes:
* options lose their `to_executor()` methods (now part of converters), to keep the option classes thin
* `Sampler` loses `quantum_program_result_to_primitive_result` (now part of converters) for avoiding having to import the class for a classmethod and reduce import cycles
* `Sampler` avoid `._options` and instead `options` (the attribute) becomes public, for parity with executor/nlv3
* `TYPE_HINTING` + `isort` linting for minimizing import cycles 

### Details and comments

Fixes #2750


qiskit_ibm_runtime/executor/dynamical_decoupling.py
```diff
@@ -13,12 +13,11 @@
 """Dynamical decoupling utilities for executor routines."""
```
diego-plan9, line 13: I'm keeping this in `.executor` based of the docstring: it seemed the only item that is not specific to `Sampler` (even though it is only used by sampler).

A potential other place (other than `.sampler_v2`) might be in `qiskit_ibm_runtime.transpiler`. If the file is not kept in `.executor`, we should convert back `/executor` into `executor.py`, as it would end up with just one file.

qiskit_ibm_runtime/options_models/environment_options.py
```diff
@@ -72,3 +72,30 @@ class EnvironmentOptions:
         | None
     ) = None
     """Runtime image used for this job."""
+
+
+@dataclass(config=PRIMITIVES_CONFIG)
+class SamplerEnvironmentOptions:
```
diego-plan9, line 78: This is the same as `EnvironmentOptions` with two less fields. I understand we _do_ want to keep them separate for compatibility with non-wrapper sampler environment options?

qiskit_ibm_runtime/executor_sampler/post_processors/utils.py
```diff
@@ -10,15 +10,21 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-"""Utility functions for executor-based SamplerV2."""
+"""Utility functions for executor-based SamplerV2 post-processors."""
```
diego-plan9, line 13: `executor_metadata_to_sampler_metadata.py` and `flatten_twirling_axes.py` felt like they contained auxiliary functions for the post_processor v1. I moved them both to this file, and I would like to come up with a name better than `utils.py`, but could not find it.

It might be also a good option to just move them inside `post_processor_v0_1.py` (the file would end up a bit too large, but still can cope).

qiskit_ibm_runtime/executor_sampler/converters.py
```diff
@@ -0,0 +1,121 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Conversions between different entities."""
```
diego-plan9, line 13: Added this file as:
* having `.to_executor_options()` in the `Option` classes prevented them from being "pure" containers (as in, with no methods), and forced them to depend on `ExecutorOptions`
* `Sampler.quantum_program_result_to_primitive_result()` forced users to import `Sampler` (which brings quite some imports) in order to use that classmethod.

This way, we decouple all these.

qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -275,14 +281,15 @@ def my_prepare(pubs, options, backend, default_shots=None):
...
-            See
-            :class:`~qiskit_ibm_runtime.executor.routines.options.sampler_options.SamplerOptions`
+        options: Sampler options. See :class:`~qiskit_ibm_runtime.model_options.SamplerOptions`
             for all available options.
         custom_prepare: Optional custom prepare function to replace the default conversion
             logic. If ``None``, the default function is used.
     """
 
+    options: SamplerOptions
+    """The options of this Sampler."""
```
diego-plan9, line 290: Same as https://github.com/Qiskit/qiskit-ibm-runtime/pull/2753

TsafrirA (commented):
All in all, I think this looks good.
I do think though that

> options lose their to_executor() methods (now part of converters), to keep the option classes thin

should be revisited. This class is very executor oriented. You might be decoupling the imports by removing the method, but I don't think there you can actually run a code which won't eventually import the executor options. Plus, I think it's more convenient this way.

Edit - is it not confusing that we have a sampler folder for wrapper sampler, and a sampler file for legacy sampler?

diego-plan9:
> All in all, I think this looks good. I do think though that
> 
> > options lose their to_executor() methods (now part of converters), to keep the option classes thin
> 
> should be revisited. This class is very executor oriented. You might be decoupling the imports by removing the method, but I don't think there you can actually run a code which won't eventually import the executor options. Plus, I think it's more convenient this way.

Some more context at https://github.com/Qiskit/qiskit-ibm-runtime/pull/2779#discussion_r3190104990, for easier debating.

> Edit - is it not confusing that we have a sampler folder for wrapper sampler, and a sampler file for legacy sampler?

Might be, but we did have a module for legacy sampler and a package for wrapper sampler already (just not at the same level), and we cannot escape the need of having 2 modules/packages one way or the other. Technically both modules/packages them are internal (the only public item is `qiskit_ibm_runtime.Sampler`, the re-exported class), although it might be non-obvious. If the major contributor to confusion is the name, and a better name for the package (`wrapper_sampler`?) is available, happy to change it!

qiskit_ibm_runtime/executor_sampler/converters.py
TsafrirA: About the first point, I prefer to have the method. I think you can't really use the new options without the `ExecutorOptions`, so the removed dependency is not very liberating. Also, the conversion to `ExecutorOptions` is an important use case for these options (arguably the only use case?) so having this conveniently available seems worthwhile to me.

qiskit_ibm_runtime/executor/dynamical_decoupling.py
SamFerracin: We also have a `transpiler` subfolder where this may potentially live. Eventually (hopefully before the wrappers are done), we get this from the addons

qiskit_ibm_runtime/executor_sampler/converters.py
SamFerracin: I will leave this to you guys. Just for the record, I have a slight preference for this converter function, but would not be opposed to maintaining the state of things.

Reason for the slight preference:
- Diego has been cleaning up the options quite a lot in a separate line of work, and he has reasons to keep them clean that he can elaborate
- With this change, we remove a method from the class that users should not really be worrying about. It makes it more explicit that they should just set the options, and we take care of the rest (including conversion)

SamFerracin:
I am also of the idea that the folder called `sampler_v2` is a bit ambiguous. A few options that I can think of:
- Changing name, as suggested in other comments
- Placing `sampler_v2` inside a subfolder `executor_routines` (name can be improved). This would give us a natural container to place functions that need to be used by both wrapper-sampler and wrapper-estimator, such as the DD pass (@TsafrirA , do we expect any other function to be "shared"?)

qiskit_ibm_runtime/executor_sampler/converters.py
diego-plan9: Ok, some middle ground at b29bdfb.

That reintroduces `SamplerOptions.to_executor_options()`, as it centralizes the responsibility for conversion (the others are arguably byproducts of the use case, which is converting `SamplerOptions` into `ExecutorOptions`). I still think it pays off to consider `options/` and `options_models/` as pure data transfer objects (minimal classes that store values), but I see the convenience angle.

diego-plan9:
Thanks for the input! Going with `executor_sampler` at e99e7cd

qiskit_ibm_runtime/executor/dynamical_decoupling.py
diego-plan9: Yes (`qiskit_ibm_runtime.transpiler.passes` is a bit tainted by import cycles, though). Let's re-evaluate as things evolve!

SamFerracin (commented):
LGTM, thanks!

qiskit_ibm_runtime/executor_sampler/converters.py
TsafrirA: Thanks @diego-plan9! We should revisit this later to see if the extra convenience is outweighed by the extra coupling.

TsafrirA:
> (@TsafrirA , do we expect any other function to be "shared"?)

Nothing jumps to mind.

@diego-plan9, this looks good to me. I can approve once the conflicts are resolved.

TsafrirA (approved):
Thank you!

PR #2783, opened 2026-05-06
Fix integration tests

SamFerracin:
The PR introducing `QuantumProgramItemResult` broke the integration tests. This PR fixes them back.

In addition, it adds `QuantumProgramItemResult` and `ItemMetadata` to the `__init__.py` file to include them in the docs


diego-plan9 (approved):
Thanks! And ...

> In addition, it adds QuantumProgramItemResult and ItemMetadata to the __init__.py file to include them in the docs

Good catch!

Issue #2781, opened 2026-05-05
[Stretch] Explore post-processor implementation options


diego-plan9:
**What is the expected enhancement?**

The post-processor feature in WrapperSampler started off as allowing custom post-processing, with the customizability being dropped off during #2716. Some of the machinery might not be needed if such customization is not needed (for example, we might not really need to "register" explicitly if the number of post-processors is limited and one per version), and it might be possible to avoid modifying the core `RuntimeJobV2` class.

This issue is about exploring such options, aiming at reducing complexity.

**Acceptance criteria**

The post-processor implementation is explored, and a solution that reduces complexity is proposed.

diego-plan9:
Marking as completed as per #2785 (and #2802 generalizing). There is still an improvement (chaining decoders), but I see this as an optimization, with the "post-processors as decoders" approach well in place now.

PR #2784, opened 2026-05-07
Force registering of sampler_v2_post_processor_v_0_1

diego-plan9:
### Summary

During #2779 , the reorganizing of imports led to `sampler_v2_post_processor_v0_1()` not being imported unless explicitly used - which implied that the `@register_post_processor` decorator did not kick in, leaving the registry empty (and causing integration tests to fail). This adds an import in a path that is traversed when the registry is used.

### Details and comments

Followup to #2779 (for #2750). A bit related to #2781


TsafrirA (approved):
LGTM. Thanks!

PR #2785, opened 2026-05-07
Move post-processing to the QuantumProgramResultDecoder

diego-plan9:
### Summary

Move `SamplerWrapper` post-processing to the existing `QuantumProgramResultDecoder`. The `ResultDecoder` mechanism is already present in the codebase for seemingly the same purpose (convert raw input into specific results objects), and taking advantage of it seems more natural.

This PR moves `_apply_post_processing` from `RuntimeJobV2` to the specific decoder for Quantum Program results, relaxing it so it fails gracefully if conversion is not possible (so users can at least get the Executor results). In the process, the dynamic "registry" is dropped (and instead becomes an item of `ResultDecoder`) as there is no need for customizing the post-processing.

### Details and comments

This still causes the `Executor` jobs to go through an extra-check. With some extra tweaks (by allowing a sequence of decoders, which has been partially possible in the past and [some bits remain](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.46.1/qiskit_ibm_runtime/base_runtime_job.py#L63)), we can make the extra-check more clean and provide the users a degree of customizability easily. I'm leaving that for a follow-up PR, as it will require some cleanup of import cycles.

Related to #2781


qiskit_ibm_runtime/quantum_program/result_decoders.py
```diff
@@ -60,4 +66,59 @@ def decode(cls, raw_result: str):  # type: ignore[no-untyped-def]
...
+
+        if semantic_role == "sampler_v2":
+            # TODO: Circular import issue. Consider changing file structure.
+            from ..executor_sampler.post_processors.post_processor_v0_1 import (
+                sampler_v2_post_processor_v0_1,
+            )
+
+            SAMPLER_POST_PROCESSORS = {
+                "v0.1": sampler_v2_post_processor_v0_1,
+            }
```
diego-plan9, line 98: Still not able to fix this. The idea is for `SAMPLER_POST_PROCESSORS` to bubble up to the top of the file after a follow-up.

qiskit_ibm_runtime/quantum_program/result_decoders.py
```diff
@@ -60,4 +66,59 @@ def decode(cls, raw_result: str):  # type: ignore[no-untyped-def]
...
+                except KeyError:
+                    raise ValueError("Could not determine a post-processor version.")
+
+                try:
+                    post_processor_fn = SAMPLER_POST_PROCESSORS[version]
+                except KeyError:
+                    raise ValueError(f"No post-processor found for version {version}.")
+
+                return post_processor_fn(result)
+            except Exception:
```
diego-plan9: The relaxing is for allowing users to at least have a result. If it is a hard-requirement to raise, it can be reverted.

test/unit/quantum_program/test_result_decoders.py
```diff
@@ -98,3 +106,72 @@ def test_unknown_schema_version(self):
         encoded_as_str = json.dumps(encoded_as_json)
         with self.assertRaisesRegex(ValueError, "No decoder found for schema version unknown."):
             QuantumProgramResultDecoder.decode(encoded_as_str)
+
+
+class TestDecoderPostProcessing(unittest.TestCase):
```
diego-plan9, line 111: This is a bit of a re-hash of `test/unit/test_runtime_job_post_processing.py`, counting that the functionality at the higher level (ie. the results for a samplerwrapper job being the right kind) is tackled elsewhere.

qiskit_ibm_runtime/quantum_program/result_decoders.py
```diff
@@ -60,4 +66,58 @@ def decode(cls, raw_result: str):  # type: ignore[no-untyped-def]
...
+
+    @staticmethod
+    def _apply_post_processing(
+        result: QuantumProgramResult,
+    ) -> QuantumProgramResult | PrimitiveResult:
+        """Apply post-processing to the decoded result.
+
+        Post-processing is only applied if the result is a QuantumProgramResult (from Executor
+        jobs) and ``result._semantic_role`` has a supported value. Otherwise, the result is returned
+        unchanged.
```
SamFerracin: Since `result: QuantumProgramResult` in the typehint,
```suggestion
        Post-processing is only applied if ``result._semantic_role`` has a supported value. Otherwise, the result is returned
        unchanged.
```

qiskit_ibm_runtime/quantum_program/result_decoders.py
SamFerracin: Two arguments to keep this as it was before:
- If we get to this stage, something went wrong, likely users messing with `_semantic_role`.
- More importantly, if for a wrapper Sampler job we return executor results, we break the SamplerV2 abstraction

diego-plan9: From another angle, if we prevent the user from getting results by raising, it is arguably a stronger breakage of the abstraction. 

I agree with considering this a very unusual case, so reverted at 1f4d4ca and easier evaluated when we are ready to finalize the wrappers.

qiskit_ibm_runtime/quantum_program/result_decoders.py
```diff
@@ -60,4 +66,50 @@ def decode(cls, raw_result: str):  # type: ignore[no-untyped-def]
...
+        Args:
+            result: The decoded result.
+
+        Returns:
+            Post-processed result or original result if no post-processing applies.
+        """
+        if not (semantic_role := result._semantic_role):
+            return result
+
+        if semantic_role == "sampler_v2":
```
SamFerracin, line 90: As an additional comment, in a future PR we can maybe move the content of this `if` in a dedicated function, to make `_apply_post_processing` more readable. Otherwise, with wrapper estimator it will become very long

SamFerracin (approved):

PR #2786, opened 2026-05-07
[wrapper estimator] Add prepare function

### LINKED ISSUE #2795: [wrapper estimator] Add a prepare function for Vanilla estimator
### Summary
We need to add a `prepare` function for the Vanilla estimator use-case (no error mitigation). The function will take in PUBs and relevant options (a small subset of the full options of the estimator) and return a `QuantumProgram`.

TsafrirA:
### Summary
Add prepare function for the wrapper estimator.
Closes #2795


qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+from qiskit.circuit import ClassicalRegister
+from qiskit.circuit.exceptions import CircuitError
+
+from ..quantum_program import QuantumProgram
+from ..quantum_program.quantum_program import SamplexItem
+from ..quantum_program.datatree import is_datatree_compatible
+from ..exceptions import IBMInputValueError
+from ..options_models.twirling_options import TwirlingOptions
+from .utils import get_bases, pauli_to_ints
+from ..executor_sampler.utils import calculate_twirling_shots
```
TsafrirA: Should move to a different location, but not sure where.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def prepare(
```
TsafrirA, line 42: I went for a non-standardized signature (compare, for example, to the custom prepare function signature). This is in line with what expect the addons to deliver us (especially now that import cycles prevent some shared classes).

qiskit_ibm_runtime/options_models/estimator_options.py
```diff
@@ -0,0 +1,96 @@
...
+
+    If twirling is enabled, the value of this option will be divided over circuit
+    randomizations, with a smaller number of shots per randomization. See the
+    :attr:`~twirling` options.
+    """
+
+    execution: ExecutionOptions = Field(default_factory=ExecutionOptions)
+    """Execution options.
+
+    See :class:`.EstimatorExecutionOptions` for all available options."""
```
diego-plan9: The docstring refers to `EstimatorExecutionOptions`, which is not available. For `WrapperSampler`, it ended up having a specific execution options class (in order to match the non-wrapper sampler interface) - can you double-check if a specific execution class is needed, or if the docstring needs to be adjusted?

qiskit_ibm_runtime/options_models/estimator_options.py
```diff
@@ -0,0 +1,96 @@
...
+    See :class:`.TwirlingOptions` for all available options."""
+
+    experimental: dict | None = None
+    """Experimental options."""
+
+    max_execution_time: int | None = None
+    """Maximum execution time in seconds, based on system execution time (not wall clock time)."""
+
+    environment: EnvironmentOptions = Field(default_factory=EnvironmentOptions)
+    """Options related to the execution environment."""
```
diego-plan9: Similar to the `ExecutionOptions` comment - can you double-check if using the new-style `EnvironmentOptions` is correct, or if some specific class needs to be introduced?

qiskit_ibm_runtime/executor_estimator/prepare.py
diego-plan9: I vote for `qiskit_ibm_runtime.executor`, touching that module's docstring so instead of saying `"""Executor."""` says something similar to "Executor program and functionality for executor-based programs."

We already have the dynamical decoupling there, and seems a sensible start (and we can better evaluate how crammed it is once more pieces of the estimator land).

diego-plan9:
Thanks @TsafrirA  - minor comments (not full review).

Can you also please open an epic for the wrapper estimator work, and break down the work?

qiskit_ibm_runtime/options_models/estimator_options.py
SamFerracin: As a matter of fact, this entire file is not needed by the `prepare` function. `prepare` only needs `TwirlingOptions`, which exists already. For the sake of reducing the PR size and make this easy to parse, I would remove this file and introduce it in a future PR, where we can address comments like this one in isolation. @TsafrirA wdyt?

qiskit_ibm_runtime/executor_estimator/prepare.py
SamFerracin: yep, i agree that this signature is good enough for now, and we can  come back to it when we see the other prepare functions. I expect the changes will be minimal, if any

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+    twirling_options: TwirlingOptions,
+    shots: int,
+) -> QuantumProgram:
+    """Convert a list of ``EstimatorPub`` objects to a ``QuantumProgram``.
+
+    Args:
+        pubs: List of estimator pubs to convert.
+        twirling_options: ``TwirlingOptions`` object.
+        shots: The number of shots to use. Will be overridden by
+            `num_randomizations * shots_per_randomization` when both are specified explicitly
```
SamFerracin: this is going to be rendered as italic
```suggestion
            ``num_randomizations * shots_per_randomization`` when both are specified explicitly
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+def prepare(
+    pubs: list[EstimatorPub],
+    twirling_options: TwirlingOptions,
+    shots: int,
+) -> QuantumProgram:
+    """Convert a list of ``EstimatorPub`` objects to a ``QuantumProgram``.
+
+    Args:
+        pubs: List of estimator pubs to convert.
+        twirling_options: ``TwirlingOptions`` object.
```
SamFerracin: I would go for something more lightweight like this, since we have type hints already
```suggestion
    """Convert estimator PUBs to a quantum program.

    Args:
        pubs: List of estimator pubs to convert.
        twirling_options: The twirling options.
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+from ..exceptions import IBMInputValueError
+from ..options_models.twirling_options import TwirlingOptions
+from .utils import get_bases, pauli_to_ints
+from ..executor_sampler.utils import calculate_twirling_shots
+
+logger = logging.getLogger(__name__)
+
+
+def prepare(
+    pubs: list[EstimatorPub],
```
```suggestion
    pubs: Sequence[EstimatorPub],
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+    Args:
+        pubs: List of estimator pubs to convert.
+        twirling_options: ``TwirlingOptions`` object.
+        shots: The number of shots to use. Will be overridden by
+            `num_randomizations * shots_per_randomization` when both are specified explicitly
+            and twirling is on.
+
+    Returns:
+        :class:`~.QuantumProgram` with :class:`~.SamplexItem` objects for each pub,
+        with passthrough_data configured for
```
```suggestion
        with passthrough data configured for
```
or
```suggestion
        with ``passthrough_data`` configured for
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+        :class:`~qiskit_ibm_runtime.executor_estimator.estimator.EstimatorV2` post-processing.
+
+    Raises:
+        IBMInputValueError: If pubs have mismatched precision,
+            if a circuit contains mid-circuit measurements, or if a circuit already uses the
+            reserved classical register name ``wrapper_estimator_data``.
```
SamFerracin: I would go for something shorter and with an underscore, to signal that it is reserved. How about this?
```suggestion
            reserved classical register name ``_meas``.
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+    for i, pub in enumerate(pubs):
+        logger.info("Processing pub %d/%d", i + 1, len(pubs))
+
+        # Determine measurement bases
+        measure_bases = get_bases(pub.observables)
+
+        # Remove any existing final measurements
+        prepared_circuit = pub.circuit.remove_final_measurements(inplace=False)
+
+        # TODO: Adjust so change basis is applied only to the last box.
```
SamFerracin: Could we please remove the TODOs, and turn them into issues instead? It will help us remember about them and to track remaining tasks

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,185 @@
...
+        if prepared_circuit.count_ops().get("measure", 0) > 0:
```
```suggestion
        if prepared_circuit.get_instructions("measure"):
```

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -0,0 +1,164 @@
...
+            non_identity_bases.append(basis)
+
+    # Handle identity case
+    if not non_identity_bases:
+        raise ValueError("No measurement bases found. Only identity in the observables.")
+
+    return PauliList(non_identity_bases)
+
+
+def resolve_precision(
```
SamFerracin: Are we using this function in `prepare`? If not, I'd introduce it when we need it

qiskit_ibm_runtime/executor_estimator/prepare.py
SamFerracin: Actually, one question: We will need to resolve conflicts between `shots` (specified in a lot of places) and `precision`. I guess the plan is to resolve them before we call `prepare`? Just to confirm, but I'm good with that. Also gives addons less to worry about

SamFerracin (commented):
Thanks! I left some comments, but looks good.

The `prepare` method is a bit long to read, so I'd suggest that in the future we try to encapsulate the main components into utility function. This will also help testing all the aspects of prepare without having to use mocks. But let's leave that for a future cleanup phase, when things are more stable

qiskit_ibm_runtime/executor_estimator/prepare.py
diego-plan9: +1 to this. If some of them are related or relatively minor, it is fine to capture more than one in a single issue - and I agree the key is capturing via issues, as that allows for easier tracking and coordinating.

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: Yes. I resolved the precision first. It doesn't really matter, and simplifies things. (The only complication is the precedence order when `num_randomizations` and `shots_per_randomization` are specified. As being discussed elsewhere).

qiskit_ibm_runtime/options_models/estimator_options.py
TsafrirA: @SamFerracin is right. I cherry picked a file which is still not needed.
To @diego-plan9's question, as far as I could tell the Estimator didn't use any extra options besides those available in the base class. But I'll verify again when the files go back in.

qiskit_ibm_runtime/executor_sampler/utils.py
```diff
@@ -69,46 +68,3 @@ def extract_shots_from_pubs(pubs: list[SamplerPub], default_shots: int | None =
     if len(pub_shots) != 1:
         raise IBMInputValueError(f"All pubs must have the same number of shots. Found: {pub_shots}")
     return next(iter(pub_shots))
-
-
-def calculate_twirling_shots(
```
TsafrirA, line 74: moved to executor folder.

qiskit_ibm_runtime/executor/calculate_twirling_shots.py
```diff
@@ -0,0 +1,58 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Calculate twirling shots for wrapper sampler/estimator."""
+
+import math
+
+
+def calculate_twirling_shots(
```
TsafrirA, line 18: Moved from wrapper sampler.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    if twirling_options.enable_gates or twirling_options.enable_measure:
+        num_randomizations, shots_per_randomization = calculate_twirling_shots(
+            shots,
+            twirling_options.num_randomizations,
+            twirling_options.shots_per_randomization,
+        )
+    else:
+        num_randomizations = 1
+        shots_per_randomization = int(shots)
```
SamFerracin: This is type-hinted as shots, so this should do
```suggestion
        shots_per_randomization = shots
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,184 @@
...
+            samplex_args = {
+                "parameter_values": param_array.reshape(
+                    param_shape + (1, pub.parameter_values.num_parameters)
+                )
+            }
+        else:
+            samplex_args = {}
+            param_shape = ()
+
+        # Item shape: (num_randomizations,) + param_shape + (num_bases,)
```
```suggestion
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,184 @@
...
+        item_shape = (num_randomizations,) + param_shape + (len(measure_bases),)
```
SamFerracin, line 127: Any implications for this VS what's in the suggested change?
```suggestion
        item_shape = (len(measure_bases),) + (num_randomizations,) + param_shape
```

Not saying that we need to change, just trying to understand if there are any implications

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,184 @@
...
+    circuits_metadata = [pub.circuit.metadata for pub in pubs]
+
+    # Validate that circuit metadata is compatible with DataTree format
+    for idx, metadata in enumerate(circuits_metadata):
+        if metadata is not None and not is_datatree_compatible(metadata):
+            raise IBMInputValueError(
+                f"Circuit metadata at index {idx} is not compatible with DataTree format. "
+                f"Metadata must be a nested structure of lists, dicts (with string keys), "
+                f"numpy arrays, or primitive types (str, int, float, bool, None)."
+            )
```
SamFerracin, line 162: This can probably end up in a utils, used for sampler wrapper and any other wrapper estimator class here and in the addons

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -0,0 +1,184 @@
...
+
+    passthrough_data = {
+        "post_processor": {
+            "version": "v0.1",
+            "circuits_metadata": circuits_metadata,
+        },
+        "observables": observables_list,
+        "measure_bases": measure_bases_list,
+    }
```
SamFerracin, line 171: Preference for this, to keep all the post-processing info inside `"post_processor"`:
```suggestion
    passthrough_data = {
        "post_processor": {
            "version": "v0.1",
            "circuits_metadata": circuits_metadata,
            "observables": observables_list,
            "measure_bases": measure_bases_list,
        },
    }
```

test/unit/estimator_v2/test_estimator_v2_prepare.py
```diff
@@ -0,0 +1,189 @@
...
+        basis_changes_spec = MagicMock()
+        basis_changes_spec.name = "basis_changes"
+        mock_samplex.inputs.return_value.get_specs.return_value = [basis_changes_spec]
+        mock_samplex.inputs.return_value.make_broadcastable.return_value = MagicMock()
+
+        with patch(
+            "qiskit_ibm_runtime.executor_estimator.prepare.build",
+            return_value=(circuit, mock_samplex),
+        ):
+            prepare([pub], twirling_options, 1024)
```
SamFerracin, line 116: Do we need this patch? I think we can do without
```suggestion
        prepare([pub], twirling_options, 1024)
```

test/unit/estimator_v2/test_estimator_v2_prepare.py
```diff
@@ -0,0 +1,189 @@
...
+        circuit = QuantumCircuit(2)
+        circuit.rx(0.1, 0)
+        circuit.ry(0.2, 1)
+
+        observables = ObservablesArray.coerce([{"ZI": 1}, {"IZ": 1}])
+        pub = EstimatorPub.coerce((circuit, observables))
+
+        quantum_program = prepare([pub], twirling_options, 2000)
+
+        item = cast(SamplexItem, quantum_program.items[0])
```
SamFerracin: This should also fix mypy
```suggestion
        self.assertIsInstance(quantum_program.items[0], SamplexItem)
```

Same for the other test

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: The order doesn't matter much. You just need to correctly pad the shapes of the Samplex arguments. Having the randomizations on the left makes it a bit easier because you don't have to pad any argument for them. Other than that, it's a pick your poison kind of decision - do you want to pad the parameters or the measurement basis.

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: Created #2805 to track and complete.

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: Good point. Done.

test/unit/estimator_v2/test_estimator_v2_prepare.py
TsafrirA: We are mocking the pass manager, so have nothing to build with.

SamFerracin (approved):
Thanks! LGTM!

PR #2787, opened 2026-05-08
Add payload size information to logger

SamFerracin:
This PR adds payload size information to the logger, for every level below `logging.INFO`


qiskit_ibm_runtime/executor/executor.py
```diff
@@ -35,7 +35,7 @@
     from ..session import Session
 
 
-logger = logging.getLogger()
+logger = logging.getLogger(__name__)
```
diego-plan9, line 38: Nice catch!

diego-plan9 (approved):

PR #2788, opened 2026-05-08
[WIP] Move results and decoder to own module, allow multiple decoders

### LINKED ISSUE #2781: [Stretch] Explore post-processor implementation options

diego-plan9:
### Summary

WIP

* Move the `ResultDecoder` subclasses to its own `decoders` module
   * `.decode()` is widened, with `data: Any`
   * logic needed by specific decoders (`executor_sampler` and `noise_learner_v3`) is moved to the module
* Move the classes that represent Results to its own `results` module
* Allow "chaining" of `ResultDecoder`s, for modelling the post-processor seamlessly (and enabling custom post-processors through existing mechanisms)
  *  `QuantumProgramResultDecoder` is split into two
  *  noise learner (v2 and v3) are revised so they are chained

### Details and comments

Fixes #2781, continuation of #2785

The main motivation for the new modules is to re-organize functionality that was spread around (most in `utils.py`, which is quite crammed up and disjoint) for providing better modularization. They are also distinct enough mechanisms (results as it is directly relevant for users; decoders as it is the under-the-hood mechanism for converting) and with enough members to merit their own packages.

Please note this is still WIP - moving some of the `results` in particular does require some deprecation / announcing of the new path (and a hack in `json.py`), and other loose ends are to be tackled.


diego-plan9:
Closing as most of the work was introduced incrementally via #2802. The last remaining bit is to be tackled at https://github.com/Qiskit/qiskit-ibm-runtime/issues/2852.

Issue #2802, opened 2026-05-11
Add modules for decoders and results


diego-plan9:
**What is the expected feature?**

Currently, Results and `ResultDecoder` class are defined mostly in the `utils` module, and in the minority of cases in the specific modules for the corresponding program.

Results are very user facing, as the way for users to interact with the output of a program. `ResultDecoder`s are more internal, but central to the responsibility of converting payloads from the API into usable objects (Results, actually), and are also relatively exposed to users (as optional args to `job.result()`, `service.run()` and other places). Decoders usually have self-contained logic for emitting a Result, and Results are self-contained.

This story is about creating new modules for each, for improving modularization and make exposing Results as part of the interface easy and consistent. For existing programs, it is likely a breaking change, which hopefully can be tackled via keeping re-exports during the deprecation window (and some special attention to the hard-coded noise learner modules in the payloads).

**Acceptance criteria**

Two new top-level modules, one for results and one for decoder, are created.

diego-plan9:
Closed via linked PR, finalized at #2835.

Issue #2801, opened 2026-05-10
Modify `QuantumProgramResultDecoder` to post process wrapper estimator jobs


TsafrirA:
### Summary
We need to modify the `QuantumProgramResultDecoder` to correctly dispatch `QuantumProgramResult` to the correct post processor. Namely, for vanilla estimator, use #2796.

PR #2803, opened 2026-05-11
Introduce decoders module (for new programs)

diego-plan9:
### Summary

Introduce a `qiskit_ibm_runtime.decoders` module, containing the decoders for the new programs:
* `executor_sampler`
* `noise_learner_v3`
* `quantum_program`

The logic specific to the decoders for each programs (ie. part of the converters, post_processor) is also moved to the new module accordingly.

### Details and comments

Related to #2802 (with just the new programs, leaving the others out of scope)
Related to #2788 (larger WIP PR, this PR extracts the parts for the decoders module and finalizes them)


qiskit_ibm_runtime/decoders/quantum_program/converters.py
```diff
@@ -0,0 +1,88 @@
...
+    ChunkSpan,
+    Metadata,
+    QuantumProgramResult,
+)
+
+if TYPE_CHECKING:
+    from ibm_quantum_schemas.executor.version_0_1 import QuantumProgramResultModel
+
+
+def quantum_program_result_from_0_1(model: QuantumProgramResultModel) -> QuantumProgramResult:
```
SamFerracin, line 31: slight and personal preference to make `converters` be a folder and have `version_x_y.py` files, for consistency

SamFerracin (dismissed):
Looks good, one minor comment, feel free to ignore it if your personal preference is opposite

qiskit_ibm_runtime/decoders/quantum_program/converters.py
diego-plan9: Taking advantage of "feel free to ignore if your personal preference is opposite" 🧌 

I find it natural for a "container of related functionality" to start as a module, then promote to a package with the same name once it grows. This allows for less nesting and easier imports (ergonomic-wise) and to just have similar functionality closer (as in, just a few lines ahead), and prevent too much granularity. In this case, the three functions are very similar, small, and importing the same subset of items, plus they are only consumed by the `decoder` - which for me is an ideal candidate for keeping them in the same module and decide if promoting to a package later on.

qiskit_ibm_runtime/decoders/executor_sampler/converters.py
```diff
@@ -14,41 +14,12 @@
...
 from qiskit.primitives.containers import BitArray, DataBin, SamplerPubResult
 
-from ..options_models import EnvironmentOptions, ExecutionOptions, ExecutorOptions
-
 if TYPE_CHECKING:
-    from ..options_models import SamplerOptions
-    from ..results import QuantumProgramItemResult
-
-
-def sampler_options_to_executor_options(options: SamplerOptions) -> ExecutorOptions:
```
SamFerracin, line 29: Was wondering where this went, but it looks like it's just not used and not tested?

SamFerracin (approved):
LGTM. This is going to conflict one more time with  @TsafrirA's open PRs, since he's writing new post-processors for estimator. You guys may want to figure out what's the least painful way to go about it

qiskit_ibm_runtime/decoders/executor_sampler/converters.py
diego-plan9: It should have been removed in an earlier PR, around the two last commits of https://github.com/Qiskit/qiskit-ibm-runtime/pull/2779 - I forgot to remove it while re-introducing the `SamplerOptions` method, so it was left unused.

diego-plan9:
> This is going to conflict one more time with @TsafrirA's open PRs, since he's writing new post-processors for estimator. You guys may want to figure out what's the least painful way to go about it

It might not be as conflict-y as it seems, if I'm reading in-flight PRs correctly:
* #2810 seems might have minor conflicts
* #2801 is the one likely to have conflicts if it has been started already
* there is a bit from #2788 that would cause conflicts (the one about using multiple decoders), but that bit can be postponed
* actually doing something similar to the current PR _after_ 2810 merges would ease things up quite a bit

In any case, we will figure it out!

PR #2806, opened 2026-05-12
Update documentation for executor release

diego-plan9:
### Summary

Update documentation for executor release:
* add quantum program, noise learner v3, and results documentation to index
* include `executor_1_0` schemas documentation

### Details and comments

Related to #2743


docs/apidocs/index.rst
```diff
@@ -8,10 +8,13 @@
    :maxdepth: 1
```
SamFerracin, line 8: You need executor in this file, right?
I don't see it when i build

SamFerracin: Image

diego-plan9: It is referenced, but not in this file - it is referenced at https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/__init__.py#L192

Which is somewhat consistent with existing documentation (sampler and estimator are in that index; noise learner and noise learner v3 did make it to this index), I suspect it has been by design. Actually, for it being part of the `docs/apidocs/index.rst`, we would need to add some documentation to the corresponding `__init__.py`  file.

SamFerracin (approved):

PR #2810, opened 2026-05-13
[wrapper estimator] Add post processing function

### LINKED ISSUE #2796: [wrapper estimator] Add a post-processing function
### Summary
We need to add a post processing function which will take a `QuantumProgramResult` created by running the `QuantumProgram` of #2795, and return a `PrimitveResult` fulfiling the contract of `EstimatorV2`.

TsafrirA:
### Summary
Depends on #2819 

Closes #2796


qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,160 @@
...
+        raise ValueError(
+            f"Number of observables lists ({len(observables_lists)}) does not match "
+            f"number of pubs ({len(result)})."
+        )
+
+    if len(measure_bases_lists) != len(result):
+        raise ValueError(
+            f"Number of measure bases ({len(measure_bases_lists)}) does not match "
+            f"number of pubs ({len(result)})."
+        )
```
SamFerracin: Consider this to reduce the number of lines:
```suggestion
    if len({len(circuits_metadata), len(observables_lists), len(measure_bases_lists), len(result)}) != 1:
        raise ValueError(
            f"Number of circuit metadata items ({len(circuits_metadata)}) , observables ({len(observables_lists)}), "
            f"basis ({len(measure_bases_lists)}), and results ({len(result)}) are not equal."
        )
```

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,160 @@
...
+
+    shots = result[0]["_meas"].shape[0] * result[0]["_meas"].shape[-2]
```
SamFerracin, line 96:
```suggestion
    if any("_meas" not in item_result for item_result in result):
        # feel free to choose a different error msg
        raise ValueError("Unable to find standard creg.")
    
    shots = result[0]["_meas"].shape[0] * result[0]["_meas"].shape[-2]
```

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # Broadcast expectation values and standard deviations to output shape
+        exp_vals_array, stds_array = broadcast_expectation_values(
+            exp_vals_array, stds_array, param_shape, obs_shape
+        )
+
+        data_bin = DataBin(
+            evs=exp_vals_array,
+            stds=stds_array,
+        )
```
SamFerracin: Should fit in a line, and perhaps move it above pub result (line 157)?
```suggestion
        data_bin = DataBin(evs=exp_vals_array, stds=stds_array)
```

SamFerracin (commented):
Looks good! Some very minor comments

I'd like more time to make sure that I understand it fully and that I don't spot any bugs. Are you okay if I re-review it this afternoon?

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -129,3 +129,159 @@ def get_bases(observables: ObservablesArray) -> PauliList:
...
+    """
+    pauli_support = np.logical_or(pauli.z, pauli.x)
+
+    for basis_idx, basis in enumerate(measure_bases):
+        # On all non-identity positions of ``pauli``, the measurement basis
+        # must match both the z/x symplectic components exactly.
+        if np.array_equal(pauli.z[pauli_support], basis.z[pauli_support]) and np.array_equal(
+            pauli.x[pauli_support], basis.x[pauli_support]
+        ):
+            return basis_idx
```
SamFerracin: Room for improvement for the future, if we want to make optimal use of data: We can yield instead of returning. There may be more than one basis able to measure a given observable

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -129,3 +129,159 @@ def get_bases(observables: ObservablesArray) -> PauliList:
         raise ValueError("No measurement bases found. Only identity in the observables.")
 
     return PauliList(non_identity_bases)
+
+
+def project_to_z(term: str) -> np.ndarray:
+    """Project observable term to Z computational basis.
+
+    Maps X,Y,Z → "Z", projectors 0,1,+,-,r,l → "0"/"1", I → "I"
```
SamFerracin, line 121: Just a heads up that when I wrote these on my notebook, I randomly chose to map r -> 0 and l -> 1. We should check if this is correct

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -129,3 +129,159 @@ def get_bases(observables: ObservablesArray) -> PauliList:
         raise ValueError("No measurement bases found. Only identity in the observables.")
 
     return PauliList(non_identity_bases)
+
+
+def project_to_z(term: str) -> np.ndarray:
```
SamFerracin:
```suggestion
def project_to_z(term: str) -> np.ndarray[int]:
```

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -129,3 +129,159 @@ def get_bases(observables: ObservablesArray) -> PauliList:
...
+        datum: Boolean array of measurement outcomes, shape
+            (num_randomizations, ..., shots_per_randomization, num_qubits)
+
+    Returns:
+        Tuple of (expectation_values, variance), each with shape (...,)
+    """
+    z_term = project_to_z(observable_term)
+
+    # Compute masks
+    # Reverse to match endian-ness
```
SamFerracin, line 176: Why is reversing required? Are executor results (and I guess results in the server, too) and Paulis following different conventions?

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,160 @@
...
+
+                # Accumulate with coefficient
+                exp_val = exp_val + coeff * term_exp_val
+                variance = variance + (coeff**2) * term_variance
+
+            exp_vals_array[obs_idx] = exp_val
+            stds_array[obs_idx] = np.sqrt(variance / shots)  # Standard error
+
+        # Broadcast expectation values and standard deviations to output shape
+        exp_vals_array, stds_array = broadcast_expectation_values(
```
SamFerracin: What is this broadcasting needed for?

When above you do: 
``` pytest
exp_vals_array = np.zeros(obs_shape + param_shape, dtype=float)
```
Doesn't this have the correct shape already?

test/unit/estimator_v2/test_estimator_v2_utils.py
```diff
@@ -152,3 +158,232 @@ def test_all_identity_observable_raises(self):
...
+
+    def test_multiple_randomizations(self):
+        """Test variance calculation with multiple randomizations.
+
+        With multiple randomizations, the variance should be computed across
+        all randomizations and shots combined.
+        """
+        # Shape: (num_randomizations=3, shots_per_randomization=10, num_qubits=1)
+        # Create 3 randomizations with different distributions
+        datum = np.array(
```
SamFerracin, line 188: The shape of this is: `(3, 1, 10, 1)`

Could you remind me what the left-most `1` is? Is it measurement basis?

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
TsafrirA: No, unfortunately it doesn't.
The output shape should be the broadcast of `obs_shape` and `param_shape`. For example, if observables shape is `(3,1)` and param shape is `(1,3)` the result needs to be `(3,3)` and not `(3,1,1,3)`.

We start by calculating the all observables for all parameter values, and then we pick the combinations needed according to the broadcasting rules. (we might be over computing as some combinations might not be needed, but I think numpy computes faster than we can determine which ones are needed).

qiskit_ibm_runtime/executor_estimator/utils.py
TsafrirA: Good point!

qiskit_ibm_runtime/executor_estimator/utils.py
TsafrirA: I think (though the endian-ness always has me confused) it's because it's a string. We read the string left to right, but the qubits are ordered right to left.

test/unit/estimator_v2/test_estimator_v2_utils.py
TsafrirA: That's the number of qubits. We pass to `compute_exp_val` a single measurement basis. (following your previous comment, maybe we need to change that to handle several measurement as once).

qiskit_ibm_runtime/executor_estimator/utils.py
TsafrirA: Tracked in #2814 .

qiskit_ibm_runtime/executor_estimator/utils.py
TsafrirA: I compared against [qiskit docs](https://qiskit.qotlabs.org/docs/api/qiskit/qiskit.quantum_info.SparseObservable#mathematics) and it is consistent.

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,152 @@
...
+        != 1
+    ):
+        raise ValueError(
+            f"Number of circuit metadata items ({len(circuits_metadata)}), "
+            f"observables ({len(observables_lists)}), "
+            f"basis ({len(measure_bases_lists)}), and results ({len(result)}) are not equal."
+        )
+
+    if any("_meas" not in item_result for item_result in result):
+        # feel free to choose a different error msg
```
SamFerracin: oops
```suggestion
```

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # Get measurement data
+        # Shape: (num_randomizations,) + param_shape + (num_bases,) + (shots, num_bits)
+        meas_data = item_data.pop("_meas")
+        # Apply measurement flips if present
+        if "measurement_flips._meas" in item_data:
+            meas_data ^= item_data.pop("measurement_flips._meas")
+
+        # Extract param_shape from measurement data
+        param_shape = meas_data.shape[1:-3] if meas_data.ndim > 4 else ()
```
SamFerracin: Given that, as described above,
```
# Shape: (num_randomizations,) + param_shape + (num_bases,) + (shots, num_bits)
```

What if the circuit had no randomization and twirling was on? In this case, `meas_data.ndim > 4` is true. Would we be producing an incorrect `param_shape`?

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
SamFerracin: I see, so the problem is the `+` between `obs_shape` and `param_shape`. And what if you defined this as:
```
exp_vals_array = np.zeros(np.broadcast_shapes(param_shape, obs_shape), dtype=float)
```
as you do below?

TsafrirA: When we calculate the expectation values we want to group together all of the parameter sets of a given observable so that the numpy calculation would be more efficient. If you broadcast first, it becomes tricky to figure out which parameter sets you need. So, the approach here is to calculate all combinations and then pick the ones you need.
You pay with memory and extra numpy time to reduce the python run time. I suspect, though haven't tested, that at least in terms of time this would beneficial (particularly in what I suspect to be the common case where you want all parameters with all observables).

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
TsafrirA: Probably rendered moot by recent events, but I will mention that we set `num_randomization=1` in the no twirling case to keep the shapes consistent. So, no the shape will not be incorrect.

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,193 @@
...
+        # Shape: (num_randomizations, num_configs, shots, num_bits)
+        # where num_configs is the total number of (param_index, basis) pairs
+        meas_data = item_data.pop("_meas")
+        # Apply measurement flips if present
+        if "measurement_flips._meas" in item_data:
+            meas_data ^= item_data.pop("measurement_flips._meas")
+
+        # Build efficient lookup: param_ndindex -> list of (measurement_basis, config_idx)
+        # This allows us to find all available measurement bases for a given parameter
+        config_lookup = defaultdict(list)
```
TsafrirA, line 117: @SamFerracin. This is the data structure I mentioned in #2819

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,181 @@
...
+    shots = result[0]["_meas"].shape[0] * result[0]["_meas"].shape[-2]
+
+    # Build EstimatorPubResult for each pub
+    pub_results = []
+    for idx, (item_data, observables, param_basis_pairs, param_shape) in enumerate(
+        zip(
+            result,
+            observables_lists,
+            param_basis_pairs_lists,
+            param_shapes_list,
```
SamFerracin: with a pre-commit run, it will fit in a single line
```suggestion
            param_shapes_list
```

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,181 @@
...
+
+            # Get the observable for this index
+            observable = observables[obs_index]
+
+            # Get the available (measurement_basis, config_idx) pairs for this parameter index
+            param_basis_list = config_lookup.get(param_index, [])
+            if not param_basis_list:
+                raise ValueError(
+                    f"No measurement basis configurations found for parameter index {param_index}"
+                )
```
SamFerracin: I guess we could switch to a more efficient try/except here:
```suggestion
            try:
                param_basis_list = config_lookup[param_index]
            except KeyError:
                raise ValueError(
                    f"No measurement basis configurations found for parameter index {param_index}"
                )
```

We don't expect to have an empty list here though, right?

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,181 @@
...
+                config_idx = identify_measure_basis(pauli_basis, param_basis_list)
+
+                # Get measurement data for this configuration
+                # Shape: (num_randomizations, shots, num_qubits)
+                datum = meas_data[:, config_idx, :, :]
+                term_exp_val, term_variance = compute_exp_val(observable_term, datum)
+
+                # Accumulate with coefficient
+                exp_val += coeff * term_exp_val
+                variance += (coeff**2) * term_variance
```
SamFerracin, line 162: It will be very easy now (in a followup PR) to do the optimization where we use all the basis that can give us info, as opposed to using only the first valid basis:
```suggestion
                for config_idx in identify_measure_bases(pauli_basis, param_basis_list):

                    # Get measurement data for this configuration
                    # Shape: (num_randomizations, shots, num_qubits)
                    datum = meas_data[:, config_idx, :, :]
                    term_exp_val, term_variance = compute_exp_val(observable_term, datum)

                    # Accumulate with coefficient
                    exp_val += coeff * term_exp_val
                    variance += (coeff**2) * term_variance
```

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Get circuit metadata for this pub if available
+        pub_metadata = {}
+        if (circuit_meta := circuits_metadata[idx]) is not None:
+            pub_metadata["circuit_metadata"] = circuit_meta
+
+        pub_result = EstimatorPubResult(data=data_bin, metadata=pub_metadata)
+        pub_results.append(pub_result)
+
+    return PrimitiveResult(pub_results, metadata=result.metadata or {})
```
SamFerracin: I suggest we do this, and then tackle the metadata as part of #2825. Otherwise, the printout of the job results can become overwhelmingly large in a notebook
```suggestion
    return PrimitiveResult(pub_results)
```

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Build EstimatorPubResult for each pub
+    pub_results = []
+    for idx, (item_data, observables, param_basis_pairs, param_shape) in enumerate(
+        zip(
+            result,
+            observables_lists,
+            param_basis_pairs_lists,
+            param_shapes_list,
+        )
```
SamFerracin: NIT: this should be able to fit in a single line now
```suggestion
        zip(result, observables_lists, param_basis_pairs_lists, param_shapes_list)
```

SamFerracin (commented):
Thanks, Tsafrir! Great job!

I'll keep testing it, but so far so good: the shapes match those of the previous estimator, and the results (tested on HW as well as using a simulator) are correct. I will keep testing it tomorrow, but so far, I cannot break it.

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
```diff
@@ -0,0 +1,181 @@
...
+        config_lookup = defaultdict(list)
+        for config_idx, (param_ndindex, basis_label) in enumerate(param_basis_pairs):
+            config_lookup[tuple(param_ndindex)].append((Pauli(basis_label), config_idx))
+
+        obs_shape = observables.shape
+        output_shape = np.broadcast_shapes(param_shape, obs_shape)
+
+        # Compute expectation values for all observables
+        exp_vals_array = np.zeros(output_shape, dtype=float)
+        stds_array = np.zeros(output_shape, dtype=float)
```
SamFerracin:
```suggestion
        # Compute expectation values for all observables
        exp_vals_array = np.empty(output_shape, dtype=float)
        stds_array = np.empty(output_shape, dtype=float)
```

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
TsafrirA: Yes, shouldn't ever happen.

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
TsafrirA: I prefer to wait with this. #2828 might make things a bit more complicated.

qiskit_ibm_runtime/executor_estimator/post_processors/post_processor_v0_1.py
TsafrirA: Done.

SamFerracin (approved):
LGTM!

PR #2813, opened 2026-05-14
[wrapper estimator] Add support for mid-circuit measurement

### LINKED ISSUE #2797: [wrapper estimator] Support mid-circuit measurements
### Summary
In the context of the estimator, mid-circuit measurements cause problems, as normal usage of Samplomatic's pass manager will add `ChangeBasis` annotations on them. We need to modify the use of the preset pass manger such that only end-of-circuit measurements are annotated with `ChangeBasis`, while all measurements have `Twirl` annotation if twirling is on.

SamFerracin:
Closes #2797


test/unit/estimator_v2/test_estimator_v2_prepare.py
```diff
@@ -85,43 +84,6 @@ def test_prepare_general_case(self):
         self.assertEqual(len(passthrough["post_processor"]["measure_bases"][0]), 3)
         self.assertEqual(len(passthrough["post_processor"]["measure_bases"][1]), 2)
 
-    @patch("qiskit_ibm_runtime.executor_estimator.prepare.generate_boxing_pass_manager")
-    def test_prepare_passes_twirling_values_to_boxing_pass_manager(self, mock_generate_boxing_pm):
```
SamFerracin, line 139: @TsafrirA this test isn't passing and I cannot fix it, due to how it's structured (specifically, due to `mock_boxing_pm.run.side_effect = lambda circuit: circuit`). I took the liberty to remove it, but can find a way to put it back if you prefer!

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -87,29 +89,42 @@ def prepare(
...
+            if (
+                isinstance(instr.operation, BoxOp)
+                and (a := get_annotation(instr.operation, ChangeBasis))
+            )
+        ]
+
+        # Remove change basis annotations from every box except the last one
+        change_basis_ref = change_basis_annotations[-1].ref
+        if len(change_basis_annotations):
+            boxed_circuit = replace_annotations(
```
SamFerracin: I managed to use the function provided by samplomatic

TsafrirA: Nice!

test/unit/estimator_v2/test_estimator_v2_prepare.py
TsafrirA: No worries, it was an ugly test. I'll think of a different way to test this.

test/unit/estimator_v2/test_estimator_v2_prepare.py
```diff
@@ -144,8 +106,8 @@ def test_prepare_with_twirling_enabled(self):
         self.assertEqual(quantum_program.items[0].shape, (4, 1))
         self.assertEqual(quantum_program.items[0].circuit.num_parameters, 3 * circuit.num_qubits)
 
-    def test_prepare_with_mid_circuit_measurements_raises(self):
-        """Test that prepare raises error for circuits with mid-circuit measurements."""
+    def test_prepare_with_mid_circuit_measurements(self):
```
TsafrirA: I think the test is sufficient, but I would have made it more explicit in asserting that only one `basis_changes` spec exists.

I would also add a test with twirling on, to make sure that we don't filter out other annotations.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -87,29 +89,42 @@ def prepare(
...
             enable_gates=twirling_options.enable_gates,
             enable_measures=True,
             twirling_strategy=twirling_options.strategy.replace("-", "_"),
             measure_annotations="all" if twirling_options.enable_measure else "change_basis",
         )
-        prepared_circuit = boxing_pm.run(prepared_circuit)
+        boxed_circuit = boxing_pm.run(prepared_circuit)
+
+        # Gather all change basis annotations
+        change_basis_annotations = [
```
TsafrirA: We could consider iterating in reversed order, and stopping at the first change basis annotation. We might need to unconditionally run the replacement logic, so I am not sure which option is better.

TsafrirA (dismissed):
This looks great!
I tested it end to end, and it works as expected.

I left a couple of comments for your consideration about unit tests and a possible optimization.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -87,29 +88,25 @@ def prepare(
...
-            )
-
+        # Add final measurements
         creg = ClassicalRegister(prepared_circuit.num_qubits, "_meas")
         try:
             prepared_circuit.add_register(creg)
         except CircuitError:
             raise IBMInputValueError("Name `_meas` is reserved for a dedicated classical register.")
-
+        prepared_circuit.barrier()
```
SamFerracin, line 99: Added a barrier here because otherwise, the pass manager will try to move them to previous boxes if possible, in which case we end up having a final box with less than `observable.qubits`

The passmanager will keep the EOC measurements together in the same box, but will take care of removing this barrier

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -126,21 +123,29 @@ def prepare(
...
         item_shape = (num_randomizations,) + param_shape + (len(measure_bases),)
 
-        # Add basis changes to samplex_arguments
-        basis_changes_specs = samplex.inputs().get_specs("basis_changes")
-        basis_changes_name = basis_changes_specs[0].name
-        # Create basis array with shape (num_bases, num_qubits)
-        measure_bases_int = np.array([pauli_to_ints(basis) for basis in measure_bases])
-
-        samplex_arguments = samplex.inputs().make_broadcastable()
-        samplex_arguments.bind(**{**samplex_args, basis_changes_name: measure_bases_int})
```
SamFerracin: These two lines are actually run by `SamplexItem` when it is constructed, so they can be removed

``` python
        samplex_arguments = samplex.inputs().make_broadcastable()
        samplex_arguments.bind(**{**samplex_args, basis_changes_name: measure_bases_int})
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -126,21 +123,29 @@ def prepare(
...
+        for spec in samplex.inputs().get_specs("basis_changes"):
```
SamFerracin, line 121: @TsafrirA removing the annotations turned out to be a bad idea. `ChangeBasis` acts as a collector, and it it gets removed, then the box may not be able to collect incoming virtual gates coming from the left (may = it can if there happen to be a Twirl annotation, it cannot if the ChangeBasis was alone).

I thought of a few ways, and the only one that seems sensible is to:
- Leave the ChangeBasis annotations
- Default to `np.zeros` to make sure that the change basis is trivial

This works fine now and (due to broadcasting) does not mess with the shape of the items.
Josh will try and fix the pass manager so that it supports "mcm_change_basis = True/False," and if he does, we can start using that here

test/unit/estimator_v2/test_estimator_v2_prepare.py
```diff
@@ -144,9 +110,15 @@ def test_prepare_with_twirling_enabled(self):
         self.assertEqual(quantum_program.items[0].shape, (4, 1))
         self.assertEqual(quantum_program.items[0].circuit.num_parameters, 3 * circuit.num_qubits)
 
-    def test_prepare_with_mid_circuit_measurements_raises(self):
-        """Test that prepare raises error for circuits with mid-circuit measurements."""
-        # Create a circuit with mid-circuit measurements
+    @combine(enable_gates=[True, False], enable_measure=[True, False])
+    def test_prepare_with_mid_circuit_measurements(self, enable_gates, enable_measure):
+        """Test the prepare function for circuits with mid-circuit measurements."""
+        if enable_measure:
```
SamFerracin, line 165: Turns out MCMs are not fully supported in Samplomatic. @joshuasn is on it

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: > Default to np.zeros to make sure that the change basis is trivial

As we are not compressing the data, this will have a negative effect on the payload size, and will increase RTT.

TsafrirA (approved):
LGTM.

I think we can come up with a test which will be boxable and buildable with MCM and measurement twirling. Shouldn't block this PR though.

PR #2816, opened 2026-05-14
Remove custom prepare functionality from wrapper sampler

### LINKED ISSUE #2815: Remove custom prepare functionality from wrapper sampler
### Summary
Current wrapper sampler has a custom prepare option, allowing to specify their own prepare function. It was decided that this customizability is misplaced, as it is not clear what value the wrapper sampler provides in this scenario (why not work directly with the Executor?).

TsafrirA:
### Summary
Closes #2815


qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -370,7 +327,7 @@ def run(self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None) -> Ru
 
         # Convert pubs to QuantumProgram and map options using the prepare function
         logger.info("Starting pre-processing")
-        quantum_program, executor_options = self._prepare(
+        quantum_program, executor_options = prepare(
```
TsafrirA, line 330: I think this means that `prepare` should be a private method of the class. No point in having it a separate function any more. I didn't want to blow up the diff of this PR, so leaving it for a follow up one.

SamFerracin: yes, i agree that we can leave it for another PR, thanks!

As for your other comment ("should be a private method of the class"), I think it should be a public method, so that people know it's there, and can customize it by creating a subclass and overwriting the prepare function

SamFerracin (approved):

PR #2818, opened 2026-05-14
[wrapper estimator] Add new EstimatorV2 class

### LINKED ISSUE #2800: Add a new `EstimatorV2` class
### Summary
We need to create a new class which will allow users to run estimator job. The class will dispatch to different prepare functions (for example, vanilla prepare of #2795) according to the options, and run the executor.

TsafrirA:
### Summary
Closes #2800 

This PR adds a new EstimatorV2 class that wraps around the Executor.


TsafrirA:
@diego-plan9 following your comment on #2786 I verified the structure of the option classes. (thanks @yaelbh for the help!)

TsafrirA:
@SamFerracin 
This has been waiting for review for a while.

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -0,0 +1,168 @@
...
+
+        resolved_precision = resolve_precision(coerced_pubs, precision)
+        if resolved_precision is not None:
+            shots = int(np.ceil(1.0 / (resolved_precision**2)))
+        elif self.options.default_shots is not None:
+            shots = int(self.options.default_shots)
+        else:
+            shots = int(np.ceil(1.0 / (self.options.default_precision**2)))
+
+        quantum_program = prepare(coerced_pubs, self.options.twirling, shots)
```
SamFerracin, line 145: An FY that this is not passing `self.options.measure_noise_learning`. Are you planning to add this later?

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -104,3 +110,37 @@ def _pad_broadcast_shape(shape: tuple[int, ...], ndims: int) -> tuple[int | slic
...
+
+    Raises:
+        IBMInputValueError: If pubs have different precision values.
+    """
+    # Extract precision from pubs, using fallback for pubs without explicit precision
+    pub_precisions = {pub.precision if pub.precision is not None else run_precision for pub in pubs}
+
+    # Remove None if it's still there (no precision specified anywhere)
+    pub_precisions = {p for p in pub_precisions if p is not None}
+
```
SamFerracin: Does something like this work?
```suggestion
    pub_precisions = {precision := (pub.precision or run_precision) if precision for pub in pubs}
```

Alternatively, you can use set built-ins:
```suggestion
    pub_precisions = {pub.precision if pub.precision is not None else run_precision for pub in pubs}
    
    # Remove None if it's still there (no precision specified anywhere)
    pub_precisions.discard(None)
```

qiskit_ibm_runtime/options_models/estimator_options.py
```diff
@@ -0,0 +1,96 @@
...
+class EstimatorOptions:
+    """Options for the executor-based EstimatorV2.
+
+    This is a minimal implementation without twirling, dynamical decoupling,
+    or error mitigation features.
+    """
+
+    default_precision: float = 0.015625
+    """The default precision for expectation value estimates if not specified in the PUBs
+    or in the run method."""
```
SamFerracin: Perhaps worth adding something like this (feel free to rephrase):
```suggestion
    or in the run method.
    
    The default value of ``0.015625``, equivalent to ``4096**-0.5``, represents the precision expected from ``4096`` shots in the presence of i.i.d. noise.
    """
```

SamFerracin (commented):
some small comments, then good to go in!

qiskit_ibm_runtime/executor_estimator/estimator.py
TsafrirA: Yes, that's the plan.

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -104,3 +110,36 @@ def _pad_broadcast_shape(shape: tuple[int, ...], ndims: int) -> tuple[int | slic
...
+        run_precision: Precision specified in run() method.
+
+    Returns:
+        The resolved precision value, or None if no precision is specified anywhere.
+
+    Raises:
+        IBMInputValueError: If pubs have different precision values.
+    """
+    # Extract precision from pubs, filtering out None values
+    pub_precisions = {pub.precision for pub in pubs if pub.precision is not None}
```
SamFerracin: Just checking, but isn't this what you want?
```suggestion
    pub_precisions = {precision := (pub.precision or run_precision) if precision for pub in pubs}
```

Imagine you have two pubs, and:
```python
run_precision = x

pub1.precision = y  # let y be different from x
pub2.precision = None
```
I would expect this to error: `pub2.precision` is assigned a value of `x`, and since `x` is not equal to `y`, you get `len(pub_precisions) = 2`. Instead, with your code, `pub2.precision` is never assigned, it is assumed that the users wanted it to be `y`

Likely you are right, but I want to make sure I understand what's going on

SamFerracin (dismissed):
Happy for you to merge, pending my comment about the precision

qiskit_ibm_runtime/executor_estimator/utils.py
SamFerracin: As a matter of fact, I think my example above would have errored with the code you had in place before I suggested a change: [link](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2818/changes/b62e291591e614f4a411da3446dd3a487721e747#diff-06f7af026545b7b5ca2011d8da3aa133917036c8fd977de1f996a2f49f3a1944L136)

TsafrirA: You are right. This is a bit different than it was before. In the legacy sampler you don't run into this issue, because you can specify precision at the pub level. So, the edge case you specified is something we need to decide how to handle - should we error or not? I think we errored in the sampler, so maybe I was hasty to make this change here.

TsafrirA: @SamFerracin 
So, I changed it again. The logic now forces you to have all pubs resolve to the same value, with `None` being a legitimate value (there is a shot override in the options if you set precision to `None`). Originally, you could have had one pub setting the precision to all pubs with `None`.

I think this is the way to go.

SamFerracin (approved):
Thanks for fixing the precision thing. LGTM now!

PR #2819, opened 2026-05-15
[wrapper estimator] Reduce the number of circuits run by wrapper estimator

SamFerracin:
In the current implementation or `prepare`, given an observable array and a set of params, we get the minimal set of measurement basis that can measure that observable, then we measure all these basis on all the subsets of parameters.
This turns out to be largely unnecessary, due to our own broadcasting rules ([link](https://quantum.cloud.ibm.com/docs/en/guides/primitive-input-output#broadcasting-rules))

### Example
Consider the following simple case:
```python
param.shape = (2, 2)

obs= ObservablesArray(PauliList(["ZZZ", "XXX"]))
assert obs.shape == (2,)
```
What the user is asking us (due to broadcasting rules) is to measure `ZZZ` for `params[0]` and `XXX` for `params[1]`. instead, we currently prepare a program that measures `ZZZ` and `XXX` on the entire set of params, in fact running twice as many circuits in the server as we should


test/unit/estimator_v2/test_estimator_v2_prepare.py
```diff
@@ -68,13 +68,10 @@ def test_prepare_general_case(self):
         self.assertIsInstance(item2, SamplexItem)
 
         self.assertEqual(item1.shape, (1, 3))
-        self.assertEqual(item2.shape, (1, 2, 2))
+        self.assertEqual(item2.shape, (1, 2))
```
SamFerracin, line 123: this test nicely highlights the issue that the PR solves

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -181,3 +171,93 @@ def prepare(
...
+    param_basis_map = {}
+    total_size = 0
+    for param_index, meas_groups in param_meas_groups_map.items():
+        meas_paulis = []
+        total_size += len(meas_groups)
+        meas_paulis = [
+            Pauli((np.logical_or.reduce(paulis.z), np.logical_or.reduce(paulis.x)))
+            for paulis in meas_groups
+        ]
+        param_basis_map[param_index] = meas_paulis
```
SamFerracin: @TsafrirA I haven't modified the passthrough data, too much going on in this PR.
But I figure that what will need to go in is:
- `observables.tolist()`, as before
- The original shape of the parameters, which can give us the result shape via `np.broadcast_shapes(observables.shape, params_shape)`
- This `param_basis_map`, which basically says: "For the first param, I measured in these two basis, for the second one in these three basis, ..." And then you know that data[0] and data[1] are the results for first param and two basis, data[2], data[3], data[4] are for the second param, etc

Definitely a tricky one to imagine without writing it.

For an example to play with:
```python
circuit = QuantumCircuit(3)
for idx in range(7):
    circuit.rz(Parameter(f"th_{idx}"), 0)
circuit.cx(0, 1)
circuit.measure_all()

param_shape = (2, 2, 1)  # replace with (2, 2), and `param_basis_map` changes
params = np.random.random(param_shape + (circuit.num_parameters,))

obs = ObservablesArray(PauliList(["ZZZ", "XXX", "YYY", "IYI"])).reshape((2, 2))

estimator_pub = EstimatorPub.coerce((circuit, obs, params))
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -181,3 +171,103 @@ def prepare(
     quantum_program._semantic_role = "estimator_v2"
 
     return quantum_program
+
+
+def compute_samplex_arguments(
+    pub: EstimatorPub,
+) -> tuple[np.array[float], np.array[int], list[tuple[tuple[int, ...], str]]]:
```
TsafrirA, line 178: I think the numpy type hints here are wrong, and that you need numpy.typing to specify the dtype of the arrays.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -181,3 +171,103 @@ def prepare(
...
+
+    basis_idx = 0
+    for ndindex, basis in param_basis_map.items():
+        for bases in basis:
+            change_basis[basis_idx] = pauli_to_ints(bases)
+            flat_parameter_values[basis_idx] = parameter_values.as_array()[ndindex]
+            basis_idx += 1
+
+    # Step 4. Log info.
+    param_basis_pairs: list[tuple[tuple[int, ...], str]] = [
```
TsafrirA, line 269: It will be very inefficient to post process based of this data structure. We would probably need something like a dictionary keyed on param indexes, and containing the pairs (meas_basis, index) where the index is the configuration index in the ItemResult.

However, I am not sure if we should compute the dictionary here and pass it with the passthrough data, or compute it later in post processing.

TsafrirA (approved):
Thanks for catching this oversight. The changes look good to me.
I left two non-blocking comments about the passthrough data and some type hints.

SamFerracin:
@TsafrirA I'll take advantage of the approval and merge this in, but will tackle these two comments in a clean way (one separate commit for each of them) in your post-process PR

qiskit_ibm_runtime/executor_estimator/prepare.py
SamFerracin: Addressed in your PR: [8c4b268](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2810/commits/8c4b268e0ff4d7b7917ba0c26b411a106ec8ea0b)

qiskit_ibm_runtime/executor_estimator/prepare.py
SamFerracin: > I am not sure if we should compute the dictionary here and pass it with the passthrough data, or compute it later in post processing.

I tried computing the lookup dict here and I run into an issue, namely that a dictionary of the form:
``` python
config_lookup = {some_tuple: ...}
```
cannot be stored in the passthrough data: `some_tuple` needs to be `some_list`, otherwise we get pydantic error. Of course though, `some_list` is not a valid key for a dict (lists are not hashable).

Shall we fry bigger fishes now, and come back to this when we have the post-processing merged in main?

PR #2821, opened 2026-05-18
Change wrapper sampler's prepare function to method

### LINKED ISSUE #2820: In the wrapper sampler change the prepare function to a public instance method
### Summary
Following the changes in #2816, we want to change the prepare function from a function to a public instance method. Now that you can't specify a custom function, it no longer makes sense to have it outside the class.

TsafrirA:
### Summary
Closes #2820

### DIFF
--- a/qiskit_ibm_runtime/executor_sampler/utils.py
+++ b/qiskit_ibm_runtime/executor_sampler/utils.py
@@ -21,6 +21,8 @@ from qiskit.circuit import BoxOp
 from ..exceptions import IBMInputValueError
 
 if TYPE_CHECKING:
+    from collections.abc import Sequence
+
     from qiskit.circuit import QuantumCircuit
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/executor_sampler/utils.py
```diff
@@ -43,15 +45,15 @@ def validate_no_boxes(circuit: QuantumCircuit) -> None:
             )
 
 
-def extract_shots_from_pubs(pubs: list[SamplerPub], default_shots: int | None = None) -> int:
-    """Extract and validate shots value from a list of ``SamplerPub``` objects.
+def extract_shots_from_pubs(pubs: Sequence[SamplerPub], default_shots: int | None = None) -> int:
```
TsafrirA, line 48: Changed the type hint in the prepare method, so just matching it here.

SamFerracin (approved):
Thanks!

PR #2822, opened 2026-05-18
broken link in release notes for 0.47.0

abbycross:
Fix rst in release notes so that the link to the Samplomatic repo will work properly now.

### DIFF
--- a/release-notes/0.47.0.rst
+++ b/release-notes/0.47.0.rst
@@ -5,7 +5,7 @@ New Features
 ------------
 
 - Added the `Executor` and `QuantumProgram` classes, which allow taking full advantage of the set of
-  capabilities offered by the `Samplomatic <https://github.com/Qiskit/samplomatic>`__ package.
+  capabilities offered by the `Samplomatic <https://github.com/Qiskit/samplomatic>`_ package.
 - Added the `NoiseLearnerV3` class to learn the noise of instructions that contain annotated boxes.
 - Added a script to fetch configuration and properties of a real backend.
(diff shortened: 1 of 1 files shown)


release-notes/0.47.0.rst
```diff
@@ -5,7 +5,7 @@ New Features
 ------------
 
 - Added the `Executor` and `QuantumProgram` classes, which allow taking full advantage of the set of
-  capabilities offered by the `Samplomatic <https://github.com/Qiskit/samplomatic>_` package.
+  capabilities offered by the `Samplomatic <https://github.com/Qiskit/samplomatic>`__ package.
```
jyu00: Isn't it just a single underscore in the end?

abbycross: I am not sure how rst works 😆 the other ones in this file use two, it seems, but also there are some with just one 🤷 I'll use one

abbycross:
```suggestion
  capabilities offered by the `Samplomatic <https://github.com/Qiskit/samplomatic>`_ package.
```

jyu00: I've always used 1, but apparently 1 underscore just allows the reference to be used elsewhere, and 2 doesn't. If the link is only used once (like in this case), then 1 vs 2 makes no difference.

jyu00 (approved):

abbycross:
A merge button doesn't display for me - can someone please merge @jyu00 @SamFerracin ?

PR #2830, opened 2026-05-25
[wrapper-estimator] Some more unit tests for post-processing

SamFerracin:
This PR simply moves a few lines around (without modifications) to add a helper to process individual `item_results`. 

The tests become easier to handle:
- We are left with only a few of those annoying tests where one needs to define passthrough data, lists of observables, lists of this and lists of that
- And we can have easier tests where we only focus on the essential part: the correctness of evs, the shapes, ...

Note:
This PR moves code around and it adds a few tests, but it does not change the code nor remove any tests


diego-plan9 (approved):
Thanks! Refactoring-wise, looks good to me

PR #2832, opened 2026-05-25
Move remaining decoders to decoders package

diego-plan9:
### Summary

Move the remaining decoders (ie. the ones for not-new programs) into the `qiskit_ibm_runtime.decoders` package, revising in the process:
* remove `EstimatorResultDecoder` and `SamplerResultDecoder`, as they were copies of `ResultDecoder`
* split `RunnerResult` in decoder and result
* move the base class and the mapping into the package

### Details and comments
Part of #2802


test/unit/test_results.py
```diff
@@ -0,0 +1,68 @@
...
+
+from qiskit_ibm_runtime.results.runner import RunnerResult
+from qiskit_ibm_runtime.runtime_job_v2 import RuntimeJobV2
+
+from ..ibm_test_case import IBMTestCase
+from .mock.fake_runtime_client import BaseFakeRuntimeClient
+
+
+@ddt
+class TestRunner(IBMTestCase):
```
diego-plan9, line 28: This is for extra-assurances about the splitting of `RunnerResult`, as it was:
```
class RunnerResult(Result, ResultDecoder):
```

and now it is split into two separate classes. I don't think we had an explicit test before, and could not find a way of reproducing `circuit-runner` results: so the test uses the output from the simulator (which still provide `backend.run()`. Support for `backend.run()` in the cloud and the circuit-runner programs has been sunset in Jan 2025: this aims to provide a bit of confidence (combine with reviewers inspecting the new classes).

qiskit_ibm_runtime/decoders/noise_learner.py
```diff
@@ -30,10 +31,6 @@ class NoiseLearnerResultDecoder(ResultDecoder):
     def decode(cls, raw_result: str) -> NoiseLearnerResult | NoiseLearnerV3Results:
         """Convert the result to NoiseLearnerResult."""
         if "schema_version" in raw_result:
-            from qiskit_ibm_runtime.decoders.noise_learner_v3.decoder import (
```
SamFerracin, line 33: Nice!

test/unit/test_results.py
```diff
@@ -0,0 +1,68 @@
...
+    @data("circuit-runner", "qasm3-runner")
+    def test_job_results(self, program):
+        """Results for a Runner job should be a RunnerResult."""
+        client = BaseFakeRuntimeClient()
+        job = RuntimeJobV2(
+            backend=None, api_client=client, job_id="123", program_id=program, service=None
+        )
+        job._status = "DONE"
+
+        # Excerpt from running a job in aer, and extracting the `results` key (without `metadata`)
```
SamFerracin, line 40: why aer? in theory, there is no relationship between this part of the client and aer. i would hard-code a result that comes from the API (and it may well be the same)

SamFerracin (approved):
This seems all fine, though not the easiest PR to review. A lot of subtle details, even in such few lines. I am happy to approve, and if anything's wrong anyway, the integration tests will catch it

test/unit/test_results.py
diego-plan9: The comment is mostly for reproducibility / traceability (ie. to understand where the dict came from). I have not been able to locate a response from the API for that kind of programs (as they are not supported), so I resorted to a simulated program (partly also assuming that the feature allowed to use a simulator at some point): not the best test ever, but I mostly wanted to have some confidence.

PR #2833, opened 2026-05-25
Move wrapper estimator post_process to decoders package

diego-plan9:
### Summary

Pave the way for #2801, tackling the last bullet point in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2803#issuecomment-4509711144.

Move the `post_processor` of WrapperEstimator into the `decoders` package, so it can be picked up easily when modifying `QuantumProgramDecoder`. This also applies the same rationale as when the ExecutorEstimator post processor was moved: the `utils` functions that are used solely by the decoder, are moved into the package as well (and the registration mechanism is removed).

### Details and comments

This is mostly for trying to advance work and prevent future conflicts, although might cause earlier conflicts (specially with #2830). Happy to merge in the order that is less painful, or keep this PR as reference.

Related to #2801 and #2802


SamFerracin (commented):
@dekelmeirom this is another PR that may cause conflicts in the code you're writing, as it moves a few files around. Nothing major, and together with #2830, it should be the only other PR that can bother you. Feel free to ask if it becomes unclear how some of these conflicts should be fixed :)

SamFerracin (dismissed):

SamFerracin (approved):

PR #2835, opened 2026-05-25
Move remaining results to results package

diego-plan9:
### Summary

Move the remaining results (ie. the ones for not-new programs) into the `qiskit_ibm_runtime.results` package. Imports from the previous modules will emit a deprecation warning. 

### Details and comments

Part of https://github.com/Qiskit/qiskit-ibm-runtime/issues/2802


qiskit_ibm_runtime/utils/estimator_pub_result.py
```diff
@@ -10,126 +10,16 @@
...
-            max_mag=max_mag,
-            max_std=max_std,
-            height=height,
-            width=width,
-            colorscale=colorscale,
-        )
+warnings.warn(
+    "The `EstimatorPubResult` class has been moved to the `qiskit_ibm_runtime.results` package as "
+    "of qiskit_ibm_runtime v0.48.0, and it will be its only location in a future release. Please "
+    "adjust your imports accordingly.",
```
SamFerracin: I think we need to give a date, in the format of "it will be removed no earlier than this day." But I'm going by memory here. Mostly so that people know when to expect this (three months from now, or the first release after three months from now)

qiskit_ibm_runtime/utils/json.py
```diff
@@ -520,7 +532,7 @@ def object_hook(self, obj: Any) -> Any:
                         settings=_cast_strings_keys_to_int(obj_val),
                     )
             if obj_type == "_json":
-                if obj["__module__"] == "qiskit_ibm_runtime.utils.noise_learner_result":
+                if obj["__module__"] in (LEGACY_NOISE_LEARNER_MODULE, CURRENT_NOISE_LEARNER_MODULE):
```
SamFerracin, line 535: This should work. Let me test it quickly tomorrow morning, to
Make sure

qiskit_ibm_runtime/utils/estimator_pub_result.py
diego-plan9: Tweaked - all the existing deprecation warnings refer to "in a future release", and my interpretation of the deprecation policy is that including the excerpt on specific dates in the deprecation warning is not really mandatory (I see the warning as complemented by the general deprecation documentation). I agree it does not hurt!

SamFerracin (approved):
LGTM. I've verified that the changes to `json.py` work, for results of new jobs as well as old ones (old = results collected prior to this change)

diego-plan9:
> LGTM. I've verified that the changes to `json.py` work, for results of new jobs as well as old ones (old = results collected prior to this change)

Thanks for verifying!

PR #2836, opened 2026-05-26
Introduce `backend.calibrate`

### LINKED ISSUE #2829: Support of a calibration program

Some users will be allowed to calibrate the backend. The interface will be either
```python
cal_job = Calibration(backend).run()
cal_job.result()
```
or
```python
cal_job = backend.calibrate()
cal_job.result()
```

**Acceptance criteria**

- [ ] The chosen interface is implemented
- [ ] Upon calibration, a request is submitted to the server with program id `calibrations`

yaelbh:
### Summary

Allows the following:
```python
cal_job = backend.calibrate()
cal_job.result()
```

### Details and comments
Fixes #2829

### DIFF
--- a/qiskit_ibm_runtime/ibm_backend.py
+++ b/qiskit_ibm_runtime/ibm_backend.py
@@ -506,6 +506,7 @@ class IBMBackend(Backend):
         return self._service._run(
             program_id="calibrations",
             inputs={},
+            options={"backend": self},
             result_decoder=ResultDecoder,
         )
 
--- a/qiskit_ibm_runtime/ibm_backend.py
+++ b/qiskit_ibm_runtime/ibm_backend.py
@@ -23,6 +23,7 @@ from qiskit.result import MeasLevel, MeasReturnType
 from qiskit.providers.backend import BackendV2 as Backend
 from qiskit.providers.options import Options
 from qiskit_ibm_runtime.utils.result_decoder import ResultDecoder
+from qiskit_ibm_runtime.runtime_job_v2 import RuntimeJobV2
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/ibm_backend.py
```diff
@@ -494,6 +497,29 @@ def run(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
...
+
+        Returns:
+            A job.
+        """
+        if self._session:
+            _run = self._session._run
+        else:
+            _run = self._service._run
+
+            if get_cm_session():
```
yaelbh: This is copied from primitives such as executor, and I'm not sure if it's relevant here at all

jyu00: This is to check if the call is invoked inside a Batch/Session context manager. 

For this project, calibration can only be invoked inside a Session, and presumably the syntax would be the same as session jobs:

```python
# using a context manager
with Session(backend):
    backend.calibrate()

# does not use a context manager
session = Session(backend)
backend.calibration(mode=session)
session.close()
```

jyu00:
Note that the initial release will be experimental and access is restricted. So we should document as such, or even go to the extend of making the method private (e.g. `backend._calibrate()`)

yaelbh:
We now tend more towards `Calibrator(backend).run()`, exactly in order not to touch `ibm_backend.py`: not to add new functions there, and new imports.

yaelbh:
Closing because we decided to do it in a different way.

PR #2839, opened 2026-05-26
add trex mitigation support for wrapper estimator

### LINKED ISSUE #2831: Add support for TREX
**Acceptance criteria**

- [ ] Following the style set by the [resilience options](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/options-resilience-options-v2), the `prepare` method needs to accept:
  - `measure_mitigation: bool`
  - `measure_noise_learning: MeasureNoiseLearningOptions`, where `MeasureNoiseLearningOptions` is a class that needs to be written anew in `qiskit-ibm-runtime.options_model` (similar to `TwirlingOptions`).
- [ ] The `post_process` step needs to be able to apply TREX mitigation to the evs and adjust the stds

dekelmeirom:
### Summary
Add a support for the case measure_mitigation is ``True`` in the WrapperEstimator.
In the estimator's prepare function - add a additional item to the QuantumProgram containing the TREX calibration circuit (measure all used qubits, which is all the backend qubits as for now), and mark in the ``passthourgh_data`` that this mitigation was used.
In the estimator's post_process function - check in the ``passthourgh_data`` if that mitigation was used, and if so - calculate the noise model based on the calibration circuit, and calculate for each observable term its rescale factors to mitigate the SPAM errors.

### Details and comments
Fixes #2831


qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -56,6 +60,9 @@ def prepare(
         shots: The number of shots to use. Will be overridden by
             ``num_randomizations * shots_per_randomization`` when both are specified explicitly
             and twirling is on.
+        measure_mitigation: If true, Twirled Readout Error eXtinction (TREX) mitigation method
```
```suggestion
        measure_mitigation: If``True``, Twirled Readout Error eXtinction (TREX) mitigation method
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -169,6 +178,12 @@ def prepare(
         passthrough_data=passthrough_data,
     )
 
+    # Add TREX calibration circuit
+    if measure_mitigation:
+        trex_item = create_trex_calibration_circuit(pubs, measure_noise_learning)
+        quantum_program.items.append(trex_item)
+        passthrough_data["post_processor"]["measure_mitigation"] = "True"
```
SamFerracin, line 200: NIT, non-blocking suggestion:
Slight preference for being explicit rather than implicit, and set `passthrough_data["post_processor"]["measure_mitigation"]` to `False` by default when we create the passthrough data, rather than not including it at all

qiskit_ibm_runtime/executor_estimator/trex_utils.py
```diff
@@ -0,0 +1,121 @@
...
+        pubs: List of estimator pubs to extract relevant qubits from.
+        measure_noise_learning: Measure noise learning options.
+
+    Returns:
+        Samplex item containing calibration circuit for TREX factors calculation.
+    """
+    # create the combined noise learning layer of all given inputs
+    max_num_qubits = 0
+    for pub in pubs:
+        max_num_qubits = max(max_num_qubits, pub.circuit.num_qubits)
```
SamFerracin:
```suggestion
    max_num_qubits = max(pub.circuit.num_qubits for pub in pubs)
```

qiskit_ibm_runtime/executor_estimator/trex_utils.py
```diff
@@ -0,0 +1,121 @@
...
+
+    classical_cal_reg = ClassicalRegister(max_num_qubits, name="_trex_cal")
```
SamFerracin, line 48: I think we need to be a bit careful here. Unfortunately, users may instantiate circuits in a very fancy way, such as:
```python
from qiskit.circuit import QuantumCircuit, QuantumRegister

qreg = QuantumRegister(5, name="q")

circuit1 = QuantumCircuit([qreg[0], qreg[1], qreg[2]])
circuit2 = QuantumCircuit([qreg[2], qreg[3], qreg[4]])
```
As long as they add ISA gates, they could submit these two circuits straight to an estimator, and then this logic wouldn't be correct anymore (you'd be creating a single TREX circuit with two qubits, and fail to mitigate the second circuit).

Looks like we may need some forms of qubit maps here?

qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
```diff
@@ -0,0 +1,37 @@
...
+class MeasureNoiseLearningOptions:
+    """Options for measurement noise learning. This is only used by V2 Estimator.
+
+    .. note::
+        These options are only used when the resilience level or options specify a
+        technique that requires measurement noise learning.
+
+    """
+
+    num_randomizations: UnsetType | int = Unset
```
SamFerracin: We've stopped using `Unset` in these new options. You can see some examples in TwirlingOptions (we have two versions of it)
```suggestion
    num_randomizations: int = 32
```

qiskit_ibm_runtime/executor_estimator/trex_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    template_trex_circuit, trex_samplex = build(annotated_trex_circuit)
+    if isinstance(measure_noise_learning.num_randomizations, int):
+        cal_randomizations = measure_noise_learning.num_randomizations
+    else:
+        cal_randomizations = 32
+    trex_calibration_item = SamplexItem(
+        circuit=template_trex_circuit,
+        samplex=trex_samplex,
+        shape=(cal_randomizations,),
+    )
```
SamFerracin: After my other comment about the options is applied, this becomes:
```suggestion
    trex_calibration_item = SamplexItem(
        circuit=template_trex_circuit,
        samplex=trex_samplex,
        shape=(measure_noise_learning.num_randomizations,),
    )
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -72,6 +73,24 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
+            readout_noise = calculate_trex_noise_model(calibration_result)
+        except ValueError as e:
+            raise ValueError(f"Failed calculating TREX noise model. Internal failure: {e}")
+
+        # create a result object without the calibration item
+        result = QuantumProgramResult(
+            data=list(result._data[:-1]),
+            metadata=result.metadata,
+            passthrough_data=result.passthrough_data,
+        )
```
SamFerracin, line 92: This should work, and if it doesn't, it's a bug and I'll fix it
```suggestion
        result = result[:-1]
```

SamFerracin (commented):
Nice! Fewer lines than I thought! 😄 

I left a comment about qregs that I think may be a bit tricky to wrap our heads around. Another thing is that we'll need to addd `shots_per_randomizations` in the options for backward compatibility, as well as an error that triggers if someone sets it to anything different than that of the pubs. But we can clean the options up in a follow up PR

qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
diego-plan9: +1 - please also revert the change to `PRIMITIVES_CONFIG` as it should be no longer needed after Sam's suggestion, as we consciously refrain from using `arbitrary_types_allowed=True` in the "new" models at `options_models`.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -47,6 +49,8 @@ def prepare(
     pubs: Sequence[EstimatorPub],
     twirling_options: TwirlingOptions,
     shots: int,
+    measure_mitigation: bool,
+    measure_noise_learning: MeasureNoiseLearningOptions,
```
TsafrirA: What about
```
```
And remove the bool flag?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: this was not working smoothly. I added an option to do `result[:-1]`, but it still returns a list of `QuantumProgramItemResult` and not a `QuantumProgramResult`. I am not sure if the expected output in such case should be `QuantumProgramResult`.
If you think it should, we can fix this and change these lines accordingly

qiskit_ibm_runtime/executor_estimator/trex_utils.py
TsafrirA: In your example, what layout\ordering do the observables obey?

qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
```diff
@@ -0,0 +1,37 @@
...
+    .. note::
+        These options are only used when the resilience level or options specify a
+        technique that requires measurement noise learning.
+
+    """
+
+    num_randomizations: int = 32
+    """The number of random circuits to draw for the measurement learning experiment.
+
+    Default: 32.
```
TsafrirA:
```suggestion
```

qiskit_ibm_runtime/options_models/utils.py
```diff
@@ -16,7 +16,9 @@
 
 from pydantic import ConfigDict
 
-PRIMITIVES_CONFIG = ConfigDict(validate_assignment=True, extra="forbid")
+PRIMITIVES_CONFIG = ConfigDict(
+    validate_assignment=True, extra="forbid", arbitrary_types_allowed=True
```
TsafrirA: Why is this needed?

TsafrirA (commented):
Some minor comments.

qiskit_ibm_runtime/options_models/utils.py
dekelmeirom: it was needed for `Unset` type, but not anymore, as I removed it :)

qiskit_ibm_runtime/executor_estimator/trex_utils.py
TsafrirA: After some discussions with @dekelmeirom and members of the Qiskit SDK team, I think this is not a real problem (and if it is, it should manifest in other places as well).

Both circuits in the above examples have `num_qubits = 3`. Without transpilation and layout selection, both circuits (as far as I understand) will be executed on the same 3 qubits - 0,1,2 on the QPU. So, by measuring just the first 3 qubits, you are respecting what's actually going to be measured.

The fact that the two circuits share \ don't share Qubit objects does not change the physical qubits they are using.

qiskit_ibm_runtime/options_models/utils.py
TsafrirA: Oh. I think the file changed as I was reviewing it :)

qiskit_ibm_runtime/executor_estimator/trex_utils.py
SamFerracin: Mmmh i want to think about this a bit carefully, in the big picture. For example, in a NLV3 TREX experiment, this works the way I said (if we pass a box that contains qubits `2, 3, 4`, then TREX is done on qubits `2, 3, 4`, regardless of the layout -- remember that these qubits need to come from the "physical register" `QuantumRegister(backend.num_qubits, name="q")`, for which there is no ambiguity)

I am not disagreeing with what you say, but this is something to reason carefully about, very subtle points here on which we can slip

dekelmeirom: as far as I understand it, as part of the ISA conditions, in order for a circuit to execute gates on qubit with physical index `i` the circuit must have at least i qubits (after transpilation, and if it is not transpiled, so the virtual circuit must have that amount of qubits defined)

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -20,15 +20,16 @@
...
 
 import numpy as np
 from qiskit.primitives import DataBin, PrimitiveResult
 from qiskit.primitives.containers.estimator_pub import ObservablesArray
-from qiskit.quantum_info import Pauli
+from qiskit.quantum_info import Pauli, PauliLindbladMap
 
 from ...executor_estimator.utils import get_pauli_basis, unbroadcast_index
 from ...results.estimator_pub import EstimatorPubResult
+from ...executor_estimator.trex_utils import calculate_trex_noise_model, calculate_trex_factor
```
diego-plan9: If these functions are only used by `post_processor_v0.1.py`, can you move them to `qiskit_ibm_runtime/decoders/executor_estimator/utils.py` (or perhaps `trex_utils.py`, if that is more consistent)?

This is a relatively new rationale, coming from #2802. The post-processing ends up being actually a `ResultDecoder`, and we benefit from having its logic encapsulated near the decoder (which avoids traversing through other packages and keeps our import logic a bit more sane). The same approach was applied in #2803 and #2833, as a reference (the rule of thumb being, if the function is used just by the decoder, it is to be placed in the decoders package).

qiskit_ibm_runtime/decoders/executor_estimator/trex_utils.py
```diff
@@ -0,0 +1,70 @@
...
+    trex_noise_calibration_data_flipped = np.logical_xor(
+        trex_noise_calibration_data, trex_calibration_measurement_flips
+    )
+    noise_list = []
+    num_qubits = trex_noise_calibration_data.shape[-1]
+    for qubit_index in range(num_qubits):
+        # the shape of the calibration data is (randomizations, shots, measured_qubit)
+        qubit_data = trex_noise_calibration_data_flipped[:, :, qubit_index]
+        excited_state_count = np.sum(qubit_data)
+        total_shots = len(qubit_data.flatten())
```
TsafrirA: This is the same for all qubits and can be calculated once.

qiskit_ibm_runtime/decoders/executor_estimator/trex_utils.py
```diff
@@ -0,0 +1,70 @@
...
+        observable_term: observable term to calculate TREX factor for.
+
+    Returns:
+        TREX factor for the observable term.
+    """
+    sparse_pauli = QubitSparsePauli(observable_term)
+    z_sparse_pauli = QubitSparsePauli(
+        ("Z" * len(sparse_pauli.indices), sparse_pauli.indices), num_qubits=sparse_pauli.num_qubits
+    )
+    return 1 / noise_model.pauli_fidelity(z_sparse_pauli)
```
TsafrirA: As we discussed, there is some discrepancy between the exponential form of the lindblad map and the 1/(1-2e) form of TREX. If we merge as is, let's make sure to create an issue for tracking.

qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
```diff
@@ -0,0 +1,42 @@
...
+        technique that requires measurement noise learning.
+
+    """
+
+    num_randomizations: int = 32
+    """The number of random circuits to draw for the measurement learning experiment.
+    """
+
+    shots_per_randomization: int | Literal["auto"] = "auto"
+    """The number of shots to use for the learning experiment per random circuit.
```
TsafrirA, line 39: I am not sure what's the best way to handle `shots_per_randomization` here. As it stands, it doesn't have any effect - you can set it to auto, or to match the global shots_per_randomization, in which case it doesn't do anything. If you set it to anything else, it errors. So what's the point?

Are we adding just so the user would get a nice error message if he migrates from legacy estimator to the wrapper one?

If we leave this in, I think the documentation should better describe the pointlessness of this parameter.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -203,9 +224,16 @@ def process_expectation_values(
...
+            term_scale_factor = (
+                calculate_trex_factor(measure_noise_model, observable_term)
+                if measure_noise_model
+                else 1
+            )
+
             # Accumulate with coefficient
-            exp_val += coeff * term_exp_val
-            variance += (coeff**2) * term_variance
+            exp_val += coeff * term_exp_val * term_scale_factor
```
TsafrirA, line 236: Let's add an issue for the error propagation.

test/unit/decoders/executor_estimator/test_post_processor.py
```diff
@@ -140,6 +140,168 @@ def test_post_processor_with_circuit_metadata(self):
...
+        )
+        result._semantic_role = "estimator_v2"
+
+        primitive_result = estimator_v2_post_processor_v0_1(result)
+
+        self.assertEqual(len(primitive_result), 1)
+        self.assertAlmostEqual(primitive_result[0].data.evs[0], 1.0)
+        self.assertAlmostEqual(primitive_result[0].data.stds[0], 0.0)
+
+    def test_measure_mitigation_changes_expectation_values(self):
```
TsafrirA: Can we be more explicit in this test, and test that the value is adjusted by the scale we expect?

test/unit/decoders/executor_estimator/test_trex_utils.py
```diff
@@ -0,0 +1,115 @@
...
+                ),
+            }
+        )
+
+        result = calculate_trex_noise_model(calibration_result)
+
+        self.assertIsInstance(result, PauliLindbladMap)
+        np.testing.assert_allclose(result.rates, [0.75, 0.0])
+
+    def test_calculates_different_flip_rates_per_qubit(self):
```
TsafrirA: I think this is not testing anything beyond what the previous test did.

TsafrirA (commented):
This looks great. Some minor comments for your considerations. I think a couple of tracking issues are in order. Also, I think the options model needs some more thought (maybe in a subsequent PR).

qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
dekelmeirom: I have opened an issue for that #2856

qiskit_ibm_runtime/decoders/executor_estimator/trex_utils.py
dekelmeirom: The lindbladmap values are the rates of the errors, while in the other form you refer to, the values are the probabilities. the relation between the two is `rates = -0.5 * np.log(1.0 - 2.0 * probabilities)`.
So if assuming "perfect twirling" (each qubit result is not correlated with the other qubits) - both methods are the same.
BUT - your comment made me understand that I have a bug - I have created PauliLindbladMap based on probabilities rather than rates! The straight forward conversion might be problematic in cases where the noise probability is larger than 0.5, so I have changed the code to calculate the TREX factor differently in case a noise model is calculated before execution (in this case the calculation will be based on the PauliLindbladMap - this parameter should be added in a future PR related to PEC to the passthrough_data) and in the case it is calculated with an additional calibration circuit during execution (in this case the calculation will be based on exp_val calculation of the calibration circuit).

test/unit/decoders/executor_estimator/test_trex_utils.py
dekelmeirom: The tests were changed

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: issue added #2858

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -203,9 +225,18 @@ def process_expectation_values(
             datum = data[:, config_idx, :, :]
             term_exp_val, term_variance = compute_exp_val(observable_term, datum)
 
+            # Calculate scale factor in case TREX mitigation is used
+            print(measure_noise_data)
```
TsafrirA:
```suggestion
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -203,9 +225,18 @@ def process_expectation_values(
...
+            term_scale_factor = (
+                calculate_trex_factor(measure_noise_data, observable_term)
+                if measure_noise_data is not None
+                else 1
+            )
+            print(term_scale_factor)
```
TsafrirA:
```suggestion
```

qiskit_ibm_runtime/decoders/executor_estimator/trex_utils.py
```diff
@@ -23,48 +23,47 @@
 from qiskit.quantum_info import Pauli, PauliLindbladMap, QubitSparsePauli
 
 
-def calculate_trex_noise_model(calibration_result: QuantumProgramItemResult) -> PauliLindbladMap:
+def calculate_trex_noise_model(calibration_result: QuantumProgramItemResult) -> np.ndarray:
```
TsafrirA: The function name seems inappropriate now. Maybe, `get_processed_calibration_data`? or something like that.

TsafrirA (changes requested):
The code changes look good to me.
There a couple of print statements to remove, and a possible function name change. For your consideration.

TsafrirA (approved):
Thanks!

PR #2842, opened 2026-05-26
Removed integration test that fails due to missing measurements

SamFerracin:
Recently, the server started raising validation errors when it finds circuits without measurements, as a way to ensure that customer jobs yield value


diego-plan9 (approved):
Thanks!

PR #2843, opened 2026-05-26
Implement gate folding functionality

### LINKED ISSUE #2847: Implement gate folding functionality

caleb-johnson:
Fixes #2847 

This PR adds a transpiler pass for performing gate folding on a circuit.

- [x] Discuss how to handle control flow blocks for gate folding. The legacy pass didn't handle them. Is that still fine?
- [x] Release note?

Differences in this pass and legacy pass:
- The new pass implements one noise scale on one circuit rather than handling many noise scales at once. It is a proper transpiler pass.
- This pass does not re-use randomly sampled gates for higher noise scales. Each noise scale will get an independently sampled subset of gates (i.e. no caching).
- The pass no longer accepts ``gates`` as a positional arg. We just work with a set of supported folding gates (the fully entangling gates). These are hard coded in the current pipeline, so the behavior will be the same wrt the estimator. Handling arbitrary entangling gates would be fairly tricky for an MVP I think.
- No longer passing anything in the metadata field of the output dag


qiskit_ibm_runtime/executor_estimator/zne/gate_folding.py
```diff
@@ -0,0 +1,229 @@
...
+from qiskit.transpiler.exceptions import TranspilerError
+
+LOG = logging.getLogger(__name__)
+
+
+DEFAULT_FOLDED_GATES: tuple[type, ...] = (ECRGate, CXGate, CYGate, CZGate, SwapGate)
+"""Default 2-qubit gate types to fold."""
+
+
+class GateFolding:
```
SamFerracin: If this is a transpiler pass, it should inherit from one of the base passes:
```suggestion
from qiskit.transpiler.basepasses import TransformationPass

class GateFolding(TransformationPass):
```

and have a run method. But I guess the catch is that the run method (here, `fold`) maps one circuit to many circuits, unlike transpiler passes, eh? In which case, I would refrain from referring to it as a transpiler pass in the docstrings, and just talk about it as a class that can transform a circuit into a set of circuit via folding

qiskit_ibm_runtime/executor_estimator/zne/extrapolation_error.py
```diff
@@ -0,0 +1,19 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Exception class for ZNE errors."""
+
+from qiskit import QiskitError
+
+
+class ExtrapolationError(QiskitError):
```
SamFerracin: We have an `execeptions.py` file for errors like this one

TsafrirA (commented):
I think it will be nice to split this into smaller PRs to make reviews easier.

qiskit_ibm_runtime/executor_estimator/zne/gate_folding.py
```diff
@@ -0,0 +1,229 @@
...
+    probabilistic sampling is applied to randomly fold a subset of gates one
+    extra time so that the noise factor averages to the desired value.
+    """
+
+    def __init__(
+        self,
+        gates: type | Sequence[type] = DEFAULT_FOLDED_GATES,
+        noise_factors: float | Sequence[float] = 1.0,
+        method: Literal["random", "front", "back"] = "random",
+        seed: np.random.BitGenerator | None = None,
```
SamFerracin: You could consider simplifying some of this, based on the [options](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/options-zne-options) that users can actually specify. In particular:
- There never will be a way for them to specify gate types
- Noise factors will always be a sequence
```suggestion
        self,
        noise_factors: Sequence[float] = [1.0],
        method: Literal["random", "front", "back"] = "random",
        seed: np.random.BitGenerator | None = None,
```

test/unit/estimator_v2/test_estimator_v2_zne_gate_folding.py
```diff
@@ -0,0 +1,192 @@
...
+        circuit = QuantumCircuit(2)
+        circuit.cx(0, 1)
+        dag = circuit_to_dag(circuit)
+        dags = GateFolding(CXGate, noise_factors=(1, 3, 5)).run_single(dag)
+        for d, expected_nf in zip(dags, (1, 3, 5)):
+            self.assertEqual(d.metadata["zne_noise_factor"], expected_nf)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
SamFerracin:
```suggestion
```

test/unit/estimator_v2/test_estimator_v2_zne_helpers.py
```diff
@@ -0,0 +1,295 @@
...
+        }
+        vals, fit_md = fit_extrapolation_models(
+            y, md, models=["linear"], extrapolated_noise_factor=0.0
+        )
+        sel_vals, _ = select_zne_extrapolated_result(vals, fit_md)
+        self.assertAlmostEqual(sel_vals[0], 0.3, places=4)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
SamFerracin:
```suggestion
```

qiskit_ibm_runtime/executor_estimator/zne/gate_folding.py
```diff
@@ -0,0 +1,110 @@
...
+"""Default 2-qubit gate types to fold."""
+
+
+def fold_gates(
+    circuit: QuantumCircuit,
+    noise_factor: float,
+    method: Literal["random", "front", "back"] = "random",
+    seed: int | np.random.BitGenerator | None = None,
+) -> QuantumCircuit:
+    r"""Fold 2-qubit gates in ``circuit`` to amplify their noise by ``noise_factor``.
```
caleb-johnson: Call out the set of supported gates here since we don't take it as input anymore. I believe default entangling gate set has been used in prod since `ZneOptions` doesn't have a place for users to specify a custom gate set to fold

qiskit_ibm_runtime/executor_estimator/zne/passes/gate_folding.py
```diff
@@ -0,0 +1,107 @@
...
+    ):
+        """Initialize the pass.
+
+        Args:
+            noise_factor: Target noise factor (``>= 1``).
+            method: How to choose a subset of gates to fold an additional time
+                to implement noise factors other than even integers. The gates
+                may be chosen from the ``front`` or ``back`` of the circuit, or
+                uniformly at ``random``.
+            seed: Seed or BitGenerator for selecting gates at random.
```
caleb-johnson:
```suggestion
            seed: Random seed/generator for selecting gates at random.
```

test/unit/estimator_v2/test_estimator_v2_zne.py
caleb-johnson, line 1: I can make a zne sub folder here if we dont want to clutter up estimator_v2 level of test suite

dekelmeirom (commented):
LGTM!

qiskit_ibm_runtime/executor_estimator/zne/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Zero-Noise Extrapolation helpers for the executor-based EstimatorV2."""
+
+from .passes import SUPPORTED_FOLDED_GATES, GateFolding
```
diego-plan9: Can you avoid re-exporting in the `__init__.py` files until the package is more populated (so it can better be decided which items merit to be present at the package level) and/or the members are being used in other places?

This applies to both this file and `passes/__init__.py`. We are stepping away from re-declaring everything in `__init__.py` as the default practice for new code, as it increases the complexity of the import graph (ie. if this is kept, and something from `qiskit_ibm_runtime.executor_estimator.zne` is included in the `executor_estimator/__init__.py`  in future PRs, all `import qiskit_ibm_runtime.executor_estimator` will import everything) - and defaulting to _not_ re-exporting and being selective later helps in that regard.

qiskit_ibm_runtime/executor_estimator/zne/gate_folding.py
```diff
@@ -0,0 +1,113 @@
...
+from qiskit.converters import circuit_to_dag
+from qiskit.dagcircuit import DAGCircuit
+from qiskit.transpiler.basepasses import TransformationPass
+
+
+SUPPORTED_FOLDED_GATES: tuple[type, ...] = (ECRGate, CXGate, CYGate, CZGate, SwapGate)
+"""2-qubit gate types supported for folding."""
+
+
+class GateFolding(TransformationPass):
```
diego-plan9, line 31: Do you envisage more passes being added to the `passes/` module as part of the epic?  

Mostly for avoiding too many nesting levels / granularity, and postpone the decision on when to create a new level once there is more than 1 item of a kind, as it tends to be easier to decide. For example, [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/executor/dynamical_decoupling.py#L88) - even if it is a pass manager, as it is currently the only item, we preferred to just keep it as a module (instead of a package).

qiskit_ibm_runtime/executor_estimator/zne/passes/gate_folding.py
```diff
@@ -0,0 +1,113 @@
...
+            noise_factor: Target noise factor (``>= 1``).
+            method: How to choose a subset of gates to fold an additional time
+                to implement noise factors other than even integers. The gates
+                may be chosen from the ``front`` or ``back`` of the circuit, or
+                uniformly at ``random``.
+            seed: Random seed/generator for selecting gates at random.
+
+        Raises:
+            ValueError: If ``noise_factor`` is less than 1.
+        """
```
diego-plan9: A bit nitpicky, but can you move the `Args:` and `Raises:` blocks to the class docstring (and avoid the `__init__` docstring entirely)? It plays a bit better with the rendering of documentation (and is consistent with the codebase since #2726 - although we don't have a way of enforcing it automatically).

test/unit/estimator_v2/test_estimator_v2_zne.py
diego-plan9: Can you actually rename it to `test_estimator_v2_zne.py`, and decide later how is it to be split as the `zne` package evolves? This way it is easy to see at a glance what package / module the test is exercising just by name and location (similar rationale to the question on the passes in the other thread - if there is just an item (one test case), we just stop at the highest level that contains it).

qiskit_ibm_runtime/executor_estimator/zne/passes/gate_folding.py
```diff
@@ -0,0 +1,113 @@
...
+        num_extra_target = fractional * num_nodes
+        if not np.isclose(num_extra_target, 0):
+            num_extra = max(1, round(num_extra_target))
+            if self.method == "front":
+                extra_indices = set(range(num_extra))
+            elif self.method == "back":
+                extra_indices = set(range(num_nodes - num_extra, num_nodes))
+            else:
+                rng = np.random.default_rng(self.seed)
+                extra_indices = set(rng.choice(num_nodes, num_extra, replace=False).tolist())
```
TsafrirA: This should work as well. The type will be numpy int instead of int, but it hashes the same.
```suggestion
                extra_indices = set(rng.choice(num_nodes, num_extra, replace=False))
```

test/unit/estimator_v2/test_estimator_v2_zne_gate_folding.py
```diff
@@ -0,0 +1,226 @@
...
+        circuit.cx(0, 1)
+        circuit.cx(0, 1)
+        folded = fold(circuit, 3.0)
+        self.assertEqual(folded.parameters, circuit.parameters)
+        self.assertEqual(folded.count_ops()["cx"], 6)
+        self.assertEqual(folded.count_ops()["rz"], 1)
+
+    @ddt.data(3.0, 5.0, 7.0)
+    def test_unitary_equivalence(self, noise_factor):
+        """Folded circuit is unitarily equivalent to the original for any odd-integer factor."""
```
TsafrirA: Why just integers?

test/unit/estimator_v2/test_estimator_v2_zne_gate_folding.py
```diff
@@ -0,0 +1,226 @@
...
+        circuit.cx(0, 1)
+        circuit.cx(0, 1)
+        folded = fold(circuit, 1.0)
+        self.assertEqual(folded, circuit)
+
+    def test_empty_circuit(self):
+        """Folding a circuit with no operations is a no-op."""
+        circuit = QuantumCircuit(5)
+        folded = fold(circuit, 3.0)
+        self.assertEqual(folded.count_ops(), circuit.count_ops())
```
TsafrirA: Some tests demand circuit equality while others compare just the counts. I think comparing the circuits is a bit better.

TsafrirA (commented):
I think this looks great. Some really minor comments for your consideration.

test/unit/estimator_v2/test_estimator_v2_zne_gate_folding.py
caleb-johnson: Ya let me expand this test to include an even integer and a float

qiskit_ibm_runtime/executor_estimator/zne/passes/gate_folding.py
caleb-johnson: Moved those blocks up to class docstring as-is and deleted the init docstring entirely. (Let me know if you want me to keep the trivial "Initialize the pass" docstring for init; otherwise, I'll resolve this comment

qiskit_ibm_runtime/executor_estimator/zne/gate_folding.py
caleb-johnson: I don't think there is another pass needed for ZNE nor PEA. I'll move this up into the `zne` level.

test/unit/estimator_v2/test_estimator_v2_zne_gate_folding.py
caleb-johnson: > Some tests demand circuit equality while others compare just the counts. I think comparing the circuits is a bit better.

Agreed for this test. For tests where testing circuit equivalence is appropriate, I've done that rather than counting gate instances. For some tests tests counting gate instances is still necessary

caleb-johnson: Resolving this, but let me know if there is more test tidying you'd like to see

qiskit_ibm_runtime/executor_estimator/zne/passes/gate_folding.py
diego-plan9: Deleting it entirely works wonders 👌 Thanks!

diego-plan9:
Thanks Caleb! Pending @TsafrirA 's approval, this looks good to me.

> Release note?

They are not needed in general for Wrapper-related additions (release notes are mostly for public items or user-facing changes, and we will add a generic one once the full work is completed). Thanks for noticing it though!

TsafrirA (approved):
LGTM!

PR #2854, opened 2026-05-28
Allow instance auto selection without warning

jyu00:
### Summary

The `instance` parameter in `QiskitRuntimeService()` is optional, but if you skip it, you get a warning. This was added by #2374 to emphasize that you should always specify an instance as where you run has implications in billing, priority and limits. But if it's an optional parameter, there should be a way to run it that way without getting a warning. 

### Details and comments

This allows you to specify `instance="auto"` when instantiating `QiskitRuntimeService` and when saving an account. It behaves the same way as specifying `None` but doesn't issue the warning. 

cc @francabrera as we discussed this late last year.

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -147,6 +147,10 @@ class QiskitRuntimeService:
             Name (CRN) or the service name. If set, it will define an instance for
             service instantiation, if not set, the service will fetch all instances accessible
             within the account following the specified filtering criteria.
+            Pass ``"auto"`` to explicitly request auto-selection without triggering the
+            "instance not set" warning. This value can also be saved to the account file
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -208,6 +212,9 @@ def __init__(
         super().__init__()
         self._all_instances: list[dict[str, Any]] = []
         self._saved_instances: list[str] = []
+        self._instance_auto = instance == "auto"
```
diego-plan9, line 215: I wish we could avoid yet another private attribute in `QiskitRuntimeService` that needs to be traced back and forth when working on the class . But I see this as an issue of the design outside this PR plus I also don't see a way around it (and we are at ~13 already anyway)!

diego-plan9 (approved):
Thanks!

PR #2855, opened 2026-05-29
Allow passing use_fractional_gates to least_busy

jyu00:
### Summary

Support `use_fractional_gates` flag in `least_busy`.

### Details and comments

Today `least_busy()` doesn't support the `use_fractional_gates` flag. You can add a filter to look for rzz in the basis gates, but the returned backend still wouldn't include fractional gates. This PR adds that parameter to `least_busy()` and, when set to True, return the least busy backend that supports FG. 

Note 
1. the parameter is better suited _before_ `filters`, but that'd break people using positional args. 
2. this PR also updates the `_get_services()` method to return the fake service directly instead of a list, since we only have 1 channel now (and I didn't want to write a useless loop).


TsafrirA (approved):
LGTM.

PR #2857, opened 2026-05-31
A new Calibrator program

### LINKED ISSUE #2829: Support of a calibration program

yaelbh:
### Summary

Allows
```python
calibrator = Calibrator(backend)
job = calibrator.run()
result = job.result()
```

### Details and comments
Fixes #2829.

Depends on the `program_id=calibarations` to be supported (not yet, currently). And on the server, of course.


qiskit_ibm_runtime/calibrator.py
```diff
@@ -0,0 +1,93 @@
...
+if TYPE_CHECKING:
+    from qiskit.providers import BackendV2
+    from ..runtime_job_v2 import RuntimeJobV2
+
+
+logger = logging.getLogger(__name__)
+
+
+class Calibrator:
+    r"""Class for rcalibrating a backend.
```
TsafrirA:
```suggestion
    r"""Class for recalibrating a backend.
```

qiskit_ibm_runtime/calibrator.py
```diff
@@ -0,0 +1,93 @@
...
+    Args:
+        backend: The backend to calibrate.
+    """
+
+    _PROGRAM_ID = "calibrator"
+    _DECODER = ResultDecoder
+
+    def __init__(
+        self,
+        backend: BackendV2,
```
TsafrirA: Typical signature is
`mode: BackendV2 | Session | Batch | None = None,`
It's not clear to me why this should be different (I would have guessed that if we picked one it would be `Session` and not `BackendV2`).

TsafrirA (commented):
Some minor comments for your consideration.

qiskit_ibm_runtime/__init__.py
```diff
@@ -216,6 +216,7 @@
 from .utils.utils import setup_logger
 from .version import __version__
 
+from .calibrator import Calibrator
```
diego-plan9: Can we _not_ expose it in the top-level package for the time being? 

My thinking is three-fold:
* if this is a preliminary interface, users should be fine using `from qiskit_ibm_runtime.calibrator import Calibrator`
* while placing it in the top-level does not make it public (by our definition of public - it is not automatically documented), it does make it public from other conventions (as in, it becomes part of the `from qiskit_ibm_runtime` interface). Even though we don't respect them, we should be picky about which items are exported top-level (and sure that they are on the path for stability).
* whenever we add something to the top-level, it contributes to #2749 (as _every_ import now ends up going through that file). With the current implementation, the overall import situation does not get really worse (as everything from `calibrator.py` is hit by other paths anyway), but it is one more item to take into account for the future.

qiskit_ibm_runtime/calibrator.py
yaelbh: it's actually just `calibrating`, I'll fix

qiskit_ibm_runtime/calibrator.py
yaelbh: I'll change it to be consistent with the other programs.

qiskit_ibm_runtime/options/calibrator_options.py
```diff
@@ -0,0 +1,36 @@
...
+from __future__ import annotations
+
+from typing import Annotated
+
+from pydantic import ConfigDict, Field
+from pydantic.dataclasses import dataclass
+
+
+@dataclass(config=ConfigDict(validate_assignment=True, extra="forbid"))
+class CalibratorOptions:
```
diego-plan9: Can you make this part of `options_models/`? We are consciously retaining `options/` for non-new programs, and `options_models/` for the new ones. In the process, you can also take advantage of `PRIMITIVES_CONFIG` as the config for the decorator.

qiskit_ibm_runtime/options/calibrator_options.py
```diff
@@ -0,0 +1,36 @@
...
+
+    image: (
+        Annotated[
+            str,
+            Field(
+                pattern="[a-zA-Z0-9]+([/.\\-_][a-zA-Z0-9]+)*:[a-zA-Z0-9]+([.\\-_][a-zA-Z0-9]+)*$",
+            ),
+        ]
+        | None
+    ) = None
```
diego-plan9: Seeing that there is already `EnvironmentOptions.image`, have you considered piggy-backing on it (for example, so `CalibrationOptions` have just an `environment` field)? It might make the interface quite unified (as in, the way for specifying an image is the same as with other programs).

diego-plan9 (commented):
Still acknowledging that this is a draft, but some notes and thoughts on the options.

qiskit_ibm_runtime/calibrator.py
```diff
@@ -0,0 +1,113 @@
...
+
+        calibrator = Calibrator(backend)
+        job = calibrator.run()
+
+    Args:
+        backend: The backend to calibrate.
+    """
+
+    _PROGRAM_ID = "calibrate"
+    _DECODER = ResultDecoder
```
diego-plan9: This (and importing and passing it to `_run()`) is likely not needed thank to you including it in `DEFAULT_DECODERS`. I have an unpublished PR that will likely remove it from the rest of the programs, mostly as a heads up (it is fine to keep it at least until the other PR fate evolves).

qiskit_ibm_runtime/calibrator.py
```diff
@@ -0,0 +1,113 @@
...
+        self,
+        mode: BackendV2 | Session | Batch | str | None = None,
+        options: CalibratorOptions | dict | None = None,
+    ):
+        if options is None:
+            self.options = CalibratorOptions()
+        elif isinstance(options, dict):
+            self.options = CalibratorOptions(**options)
+        else:
+            self.options = options
```
diego-plan9: `Executor` (and `NoiseLearnerV3`, and `ExecutorSampler`) are arguably a better example than `EstimatorSampler` when it comes to initializing options:

* https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/executor/executor.py#L85-L86
* https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/executor/executor.py#L100-L113

That mechanism provides a slightly stronger behavior when replacing options. It might be a bit overkill for this program in the sense of not expecting to be fully consumer-ready at this stage, but adopting that mechanism would keep the new program more similar to the rest of new programs (with `EstimatorSampler` to be also made consistent too).

diego-plan9: edf174f from #2901 illustrates it better!

qiskit_ibm_runtime/calibrator.py
diego-plan9: #2899 landed - and it seems `_DECODER` can be safely removed.

yaelbh:
This PR is essentially ready, but I suggest to merge it only when the server is ready too, because trying to calibrate now results in
```
qiskit_ibm_runtime.exceptions.RuntimeJobFailureError: 'Unable to retrieve job result. Error code 1251; Error preprocessing job.' 
```

yaelbh:
> This PR is essentially ready, but I suggest to merge it only when the server is ready too

Maybe merge it now nevertheless?  (@diego-plan9, @SamFerracin)

qiskit_ibm_runtime/calibrator.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class Calibrator:
+    r"""Class for calibrating a backend.
+
+    The :meth:`run` method can be used to submit a calibration request for a backend.
+
+    .. code-block:: python
+
+        from qiskit_ibm_runtime import QiskitRuntimeService, Calibrator
```
diego-plan9:
```suggestion
        from qiskit_ibm_runtime import QiskitRuntimeService
        from qiskit_ibm_runtime.calibrator import Calibrator
```

Due to https://github.com/Qiskit/qiskit-ibm-runtime/pull/2857#discussion_r3333683493

test/unit/test_calibrator.py
```diff
@@ -0,0 +1,84 @@
...
+        calibrator.options = new_opts
+        self.assertIs(calibrator.options, new_opts)
+
+
+class TestCalibrator(IBMTestCase):
+    """Tests the ``Calibrator`` class."""
+
+    def setUp(self) -> None:
+        """Test level setup."""
+        super().setUp()
```
diego-plan9: Is this needed?

yaelbh: No, it's probably archeological, I deleted it.

qiskit_ibm_runtime/options_models/calibrator_options.py
```diff
@@ -0,0 +1,32 @@
...
+from __future__ import annotations
+
+from pydantic import Field
+from pydantic.dataclasses import dataclass
+
+from .environment_options import EnvironmentOptions
+from .utils import PRIMITIVES_CONFIG
+
+
+@dataclass(config=PRIMITIVES_CONFIG)
```
SamFerracin, line 24: @diego-plan9 more options for you to change when this goes in main

SamFerracin (approved):

PR #2860, opened 2026-06-01
[wrapper estimator] PEC mitigation

### LINKED ISSUE #2844: Add support for PEC
**Acceptance criteria**

dekelmeirom:
### Summary
Add support for the case `pec_options` is given as input.
Create a new `prepare_pec` function - force gate twirling and add noise injection annotations to the circuit. In addition, calculate the gamma factor of the circuit and pass it using the passthourgh_data.
In the post-process, create a new `post_process_pec` function - calculate the mitigated exp_val using the `pauli_signs` attribute of the results and the calculated gamma factor of the noise model of the circuit.

### Details and comments
Fixes #2844


SamFerracin (commented):
Thanks, @dekelmeirom! This looks great!
I will spend some time this afternoon checking the correctness of all of the funcs, but at a first glance, it looks correct!

Here's a question about code organization, which came up several times but we never talked about. Namely, it looks like you're going for the "one big function" approach, where you have a `prepare` and `post_process` that handle everything.
Looking at the ZNE work that Caleb has started, I am a bit nervous that ZNE will add quite a lot of complexity, which will result in a large numbef if/elses, branching logic etc. For example, it will need to handle folding, both with PEA amplification (which requires boxing with annotation) and with gate folding (which has a different path and is probabilistic).

I would like to suggest a slightly different approach, and you (and everyone else) please feel free to say if it doesn't make sense. In the alternative approach, we could have a number of smaller `prepare` and `post_process` functions, with this signature:
```python
def prepare(
    pubs: Sequence[EstimatorPub],
    shots: int,
) -> QuantumProgram:
    """Basically, prepares a quantum program when users do not request PEC nor ZNE"""

def prepare_pec(
    pubs: Sequence[EstimatorPub],
    pec_options: PECOptions,  # notice that `None` is gone
    shots: int,
) -> QuantumProgram:
    """Prepares a quantum program when users request PEC."""

def prepare_zne(
    pubs: Sequence[EstimatorPub],
    zne_options: ZNEOptions,
    shots: int,
) -> QuantumProgram:
    """Prepares a quantum program when users request PEC."""
```
We can then make the wrapper estimator select which one to use depending on the options, such as `options.pec_mitigation` being `True` or `False`.

Benefits:
- We get rid of branching logic, which will make the prepare and post-process functions easier to read
- We consolidate the concepts of "preparing a program when no mitigation is required," "preparing a program when PEC is required," etc. We get a nice symmetry between the three cases ("vanilla", pea, and zne), because they can all support twirling ON and OFF and trex ON and OFF.

It is true that there will be a bit of code duplication between the various functions, but in a follow-up PR (say, after `prepare_pec` goes in), we can place the duplicate code in util functions, with dedicated tests and docs.

What do you think?
cc @TsafrirA and @caleb-johnson

dekelmeirom:
We can try it and see have much code duplication it creates. Another benefit of this approach is that it will be easier to later move those classes to another repo.
In the `post_process` part - we will need to have a function that will read the `passthorugh_data` and decide by its content to which post-process function to call.

SamFerracin:
@dekelmeirom can this one be closed?

dekelmeirom:
Was split into 2 separated PRs that have been merged.

PR #2861, opened 2026-06-01
Use runtime time imports to break some import cycles

diego-plan9:
### Summary

Move some imports to runtime imports in:
* `get_mode_service_backend()`
* `BasePrimitiveV2._run()`
* `FakeBackendV2.refresh()`

As they are quite central classes, the goal is to reduce the import surface among them. Runtime imports is not the best mechanism ever, but the benefit in the import graph pays off.  

### Details and comments

`get_mode_service_backend()` could use a revamp in general, to be tackled outside this PR.

The rest of the line changes come from `TC` / `I` ruff rules.

Related to #2749

### DIFF
--- a/qiskit_ibm_runtime/base_primitive.py
+++ b/qiskit_ibm_runtime/base_primitive.py
@@ -137,7 +137,7 @@ class BasePrimitiveV2(ABC, Generic[OptionsT]):
         Returns:
             Submitted job.
         """
-        # Use runtime imports, to prevent `BasePrimitiveV2` to depend on `QiskitRuntimeService`.
+        # Use runtime imports, to prevent `BasePrimitiveV2` to depend on several core objects.
         from .ibm_backend import IBMBackend
         from .qiskit_runtime_service import QiskitRuntimeService
 
--- a/qiskit_ibm_runtime/base_primitive.py
+++ b/qiskit_ibm_runtime/base_primitive.py
@@ -12,20 +12,18 @@
 
 """Base class for Qiskit Runtime primitives."""
 
-from abc import ABC, abstractmethod
-from typing import TypeVar, Generic
-import logging
-from dataclasses import asdict, replace
+from __future__ import annotations
 
-from qiskit.primitives.containers.estimator_pub import EstimatorPub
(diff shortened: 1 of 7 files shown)


TsafrirA (commented):
I am conflicted about this one. On the one hand, it does simplify the dependency tree. On the other hand, it seems like a hacky solution to an actual issue, with some base classes really depending on many other classes. This is not a rare code path or something like that, it's a real dependency and I am not sure this is the proper solution for that.

yaelbh:
I agree with the change in `refresh`: it's an esoteric method and it indeed makes sense to import `QiskitRuntimeService` only if it is called.

However about the changes in `base_primitives.py`:
- `get_mode_service_backend` is about to be refactored anyway
- `_run` is almost always called given that a primitive was initialized, so what do you expect to gain?

diego-plan9:
The dependencies tackled in this PR can be seen as artificial (byproducts of the implementation) or stretching the definitions of the core concepts:
* `base_primitive.py` related to `get_mode_service_backend()`: I just created #2874 for a longer insight on the problem and future revising of the function. I see these are mostly accidental complexity and requiring time for proper refactor.
*  `BasePrimitiveV2._run()`: mostly related to the above (ie. they are used by `isinstance()`, as the way we model "mode" does not provide anything more usable). There is an `backend.check_faulty()` which is a bit more debatable, but I'm leaning towards containing it and make `BasePrimitiveV2` conceptually agnostic to `BackendV2`.
* `FakeBackendV2.refresh()`:  conceptually tricky - a local fake backend should not be depending on the god object that accesses the remote service. I see `refresh()` as an usability concession (which we cannot change), so the change contains the import side effects. 

All in all, this aims to move towards keeping the core classes lean (and limit the extent of the extra complexity). I don't like runtime imports, but it is the best mechanism we have: the alternative is continue having a convoluted import graph.

diego-plan9:
> `get_mode_service_backend` is about to be refactored anyway

I challenge the "about to be" 😬 . Even with the issue up, it requires some effort - and in the meantime, we keep contributing to the import graph (for example, #2857 or with the recent new programs). This PR alleviates it in the meantime.

yaelbh (approved):

PR #2864, opened 2026-06-01
Move Embeddings and CircuitSchedule to visualization module

diego-plan9:
### Summary

`Embeddings` and `CircuitSchedule` are only used by visualization functions (in the case of `Embeddings`, it is also imported by `NoiseLearner` - but only for visualization purposes). This PR moves them to the `visualization` module, for better modularity, and explicitly outside of `utils` to continue avoiding disparate on it. 

### Details and comments

Related to #2749 and #2563 - there was actually an import cycle between `draw_circuit_schedule_timings` and `CircuitSchedule` (as it traversed `utils` that would surface if tackling #2563).


diego-plan9:
As far as I can tell, they are not documented (although mentioned in some release notes without the module name). This makes me think we don't need a deprecation for them, I'm open to more input.

SamFerracin (approved):

PR #2865, opened 2026-06-01
Fix FoldRzzAngle global phase at pi/3pi wrap boundary

quantova-inc:
### Summary

`FoldRzzAngle` folds out-of-range `Rzz(θ)` gates into the calibrated range `[0, π/2]` while preserving the **full** unitary, including global phase. Wrapping `θ` into `(-π, π]` drops whole `2π` windings, and each dropped winding flips the sign of the operator (`Rzz(θ + 2π) = -Rzz(θ)`), so the pass re-adds a `GlobalPhaseGate(π)` for an odd number of dropped windings.

That correction was gated on `pi < angle % (4*pi) < 3*pi` — a strict floating-point modulo window. Odd-winding angles that wrap **exactly** onto the `±π` boundary (e.g. `3π`, `-3π`, `7π`) fail the strict `<` comparison and silently skip the phase, producing a folded circuit equal to `-Rzz(θ)` instead of `Rzz(θ)`. Angles like `±5π, ±9π, ±11π` only pass by floating-point rounding luck, so the behavior is latent and fragile.

This is unobservable for a standalone circuit measured directly (hardware discards global phase), but becomes physically observable when the circuit is used as a controlled subroutine, or compared/verified at the unitary level (`Operator`, equivalence checks, process tomography).

### Fix

Derive the dropped winding parity directly from the wrap that was already performed, so it stays consistent with `wrap_angle` regardless of rounding:

```python
windings = round((angle - wrap_angle) / (2 * pi))
if windings % 2:
    replace.apply_operation_back(GlobalPhaseGate(pi))
```

### Verification

A sweep over 40,000+ angues across `[-8π, 8π]` plus every multiple of `π` and `π/2`:

- **Before:** 9 failures, worst element error `2.0` (a global phase of π), all at the `±3π, ±7π`-type boundary angles.
- **After:** 0 failures, worst error `3e-15`.

### Tests

Added boundary-angle cases (`±3π, ±5π, ±7π`) to `test_folding_rzz_angles`, which already asserts exact `Operator ==` equality (unlike `.equiv()`, this includes global phase and so guards the regression directly).


yaelbh (approved):
LGTM, thanks!

PR #2790, opened 2026-05-08
Add support for scheduler timing and stretch values information in Executor

SamFerracin:
Closes #2789

PR #2870, opened 2026-06-02
Support scheduler timing and stretch values in WrapperSampler

### LINKED ISSUE #2791: Add support for scheduler timing and stretch values information in Wrapper Sampler
- [ ] Users of wrapper sampler are able to request scheduler timings as in the existing SamplerV2 ([link](https://quantum.test.cloud.ibm.com/docs/en/guides/qiskit-runtime-circuit-timing)).
- [ ] Similarly, they are able to request stretch values as in SamplerV2 ([link](https://quantum.cloud.ibm.com/docs/en/guides/stretch#view-stretch-values-in-qiskit-runtime))

Blocked by #2790

diego-plan9:
### Summary

Make use of #2790 for allowing retrieving scheduler timing and stretch values in `WrapperSampler`:
* the options conversion makes use of the `experimental` fields, in the format for non-wrapper sampler, mapping them to the executor options
* `quantum_program_item_result_to_sampler_pub_result` populates the `metadata` field based on the existence of such members

### Details and comments

Fixes #2791

### DIFF
--- a/qiskit_ibm_runtime/decoders/executor_sampler/converters.py
+++ b/qiskit_ibm_runtime/decoders/executor_sampler/converters.py
@@ -14,12 +14,30 @@
 
 from __future__ import annotations
 
+from dataclasses import asdict
 from typing import Any, TYPE_CHECKING, Literal
+from collections.abc import Iterable
 
 from qiskit.primitives.containers import BitArray, DataBin, SamplerPubResult
 
+from ...utils.circuit_schedule import CircuitSchedule
+
 if TYPE_CHECKING:
     from ...results import QuantumProgramItemResult
+
+
+def expanded_values_to_lists(key_value_pairs: Iterable[tuple[str, Any]]) -> dict[str, Any]:
+    """Dict factory that converts `expanded_values` tuples to lists.
+
+    Args:
+        key_value_pairs: pairs of (key, value) items
+
+    Returns:
+        A dictionary built from `key_value_pairs`, with the key `expanded_values` containing lists.
+    """
+    stretch_value = dict(key_value_pairs)
+    stretch_value["expanded_values"] = [list(i) for i in stretch_value["expanded_values"]]
+    return stretch_value
 
 
 def quantum_program_item_result_to_sampler_pub_result(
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/decoders/executor_sampler/converters.py
```diff
@@ -55,11 +71,23 @@ def quantum_program_item_result_to_sampler_pub_result(
...
+    if item.metadata.scheduler_timing:
+        pub_metadata.setdefault("compilation", {})
+        pub_metadata["compilation"]["scheduler_timing"] = {
+            "timing": item.metadata.scheduler_timing.timing,
+            "circuit_duration": item.metadata.scheduler_timing.circuit_duration,
+        }
+    if item.metadata.stretch_values:
+        pub_metadata.setdefault("compilation", {})
+        pub_metadata["compilation"]["stretch_values"] = [
+            asdict(stretch_value, dict_factory=expanded_values_to_lists)
```
diego-plan9, line 89: This extra complexity is due to `QuantumProgramItemResult.metatada.scheduler_timing.stretch_values` containing tuples in `expanded_values` (likely due to `StretchValues.expanded_values` being defined as a `list[tuple[int, int]]` and populated as such). Non-executor Sampler seems to contain `list[list[int]]`, always with 2-elements in the inner lists.

I went with preserving compatibility.

TsafrirA (approved):
This looks good.

I think we could insert it to the options directly, instead of using `experimental`, but matching legacy sampler is good enough for now.

diego-plan9:
> I think we could insert it to the options directly, instead of using experimental, but matching legacy sampler is good enough for now.

I had similar doubts when writing the PR, and mostly prioritized compatibility with legacy sampler. As `SamplerExecutionOptions` inherits from `ExecutionOptions`, I believe the user can just use them directly already:

```python
...
options.execution.scheduler_timing = True
``` 

I ended up seeing the ability to use either that or `experimental` as a "feature" of sorts, and mostly a matter of deciding when it is released if we encourage one over the other (or if we want to avoid `experimental` and break compatibility). Happy to tweak as needed.

SamFerracin:
@diego-plan9 and @TsafrirA regarding your discussion about options, see [this](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2790#discussion_r3303791753). Looks like we can add these values to the execution options in sampler as well (and we probably want to warn when users specify them via experimental options, for three months?)

SamFerracin (approved):

diego-plan9:
Thanks - created https://github.com/Qiskit/qiskit-ibm-runtime/issues/2878 for tackling the potential unifying of interfaces separately.

PR #2872, opened 2026-06-02
Add AI/LLM section to PR template

diego-plan9:
### Summary

Update PR template for complying with Qiskit contribution guidelines (https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#use-of-ai-tools). 

### Details and comments

Thanks to the Qiskit team for the content (the addition comes from Qiskit's PR template).

Fixes N/A

### AI/LLM disclosure


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,7 +11,14 @@
...
+
 Fixes #
+
+### AI/LLM disclosure
+
+- [ ] I didn't use LLM tooling, or only used it privately.
+- [ ] I used the following tool to help write this PR description:
+- [ ] I used the following tool to generate or modify code:
+
+<!-- Any code generated by LLM or modified from LLM suggestions must commented inline too. -->
```
TsafrirA: This seems really aggressive. I see Qiskit has the same comment too, but I don't think this is a sensible requirement.

yaelbh: I agree with Tsafrir. As a developer who works with AI tools, I honestly don't know how to comply with this requirement. Also, I don't understand how it is enforced in Qiskit.

yaelbh (commented):
What benefit do we expect to gain from the disclosure? How do we expect to use the provided information? I understand that we want to avoid a deluge of AI-generated PRs - is it the motive behind the disclosure? Will it help to prevent it?

The first option (`I didn't use LLM tooling`) can be almost dropped, it's almost like `I came here on a horse`...

.github/PULL_REQUEST_TEMPLATE.md
diego-plan9: The goal is to be inline with the policy, and prioritize such alignment - and in practice, also get closer to the Qiskit template. I don't think the comment will have such an impact in PR handling (for reviewers and contributors): it does steer contributors in the right direction which to me is the main goal, and we can evaluate the level of enforcement and effort as we move forward.

diego-plan9:
This is separate from discussing the policy: this is an effort for aligning with such policy, the short version being:
> If you use any AI tool while preparing your code contribution, you must disclose the name of the tool and its version in the PR description.

.github/PULL_REQUEST_TEMPLATE.md
yaelbh: I understand that the template is not for us to decide, therefore I'm about to approve the PR, don't worry... But I'm still curious about
> I don't think the comment will have such an impact in PR handling

because I feel that there's something fundamental that I don't understand: if I want to be good and do what I'm asked, i.e., if I want to insert inline comments for all LLM generations, then, technically, how to do it? And, suppose that it was a realistic request, doesn't it mean that soon most of our code lines will become such inline comments?

TsafrirA: Additionally, it's not clear to me what we're supposed to do with the inline information - should I be more suspicious of AI generated code? Less suspicious? What are we using that information for?

diego-plan9: Actually, in my comments prior to this one I was interpreting the `inline` part of:

> Any code generated by LLM or modified from LLM suggestions must commented inline too. 

as "inline **PR** comments", not "inline **code** comments". Thanks for pursuing the discussion and my bad for not realizing it earlier, as I agree "code comments" is arguably the intended interpretation - and if so, now I also share the concerns about practicability.

I'll remove the comment (even if breaks a bit more the parity between templates, at least it does enforce the disclosure, which is the key aspect of the policy), and pursue further clarification separately. Thanks!

yaelbh (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,7 +11,12 @@
...
-
 ### Details and comments
+
 Fixes #
+
+### AI/LLM disclosure
+
+- [ ] I didn't use LLM tooling, or only used it privately.
+- [ ] I used the following tool to help write this PR description:
+- [ ] I used the following tool to generate or modify code:
```
rainerschoe, line 22: I would add a very short sentence with information on why we collect this information and what are the implications of each answer / how we use this information.

With just this selection it might convey the meaning, that AI-use is dis-encouraged and will result in inferior treatment, which in-turn might lead to withholding of information.

Something like

```suggestion
### AI/LLM disclosure
This project encourages responsible use of AI coding tools.
For book-keeping please provide some information on used AI tools:

- [ ] I didn't use LLM tooling, or only used it privately.
- [ ] I used the following tool to help write this PR description:
- [ ] I used the following tool to generate or modify code:
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,7 +11,12 @@
...
 
-
-
 ### Details and comments
+
 Fixes #
+
+### AI/LLM disclosure
+
+- [ ] I didn't use LLM tooling, or only used it privately.
```
rainerschoe, line 20: What does "used it privately" mean? Does it mean I used my private LLM license or e.g. codex / copilot compared to a business one?

.github/PULL_REQUEST_TEMPLATE.md
diego-plan9: Thanks - I don't think the PR notice would be interpreted as such. We are counting on contributors reading this repo's contributing guidelines, and the linked Qiskit guidelines, which I believe do a good job of conveying that usage of AI tools is allowed an on par with the other contributions, it just needs to be disclosed.

.github/PULL_REQUEST_TEMPLATE.md
diego-plan9: Privately as in "for bits other than generating or modifying code", as far as I understand.

jakelishman:
Since Ian pointed me at this:

Fwiw, that message was only supposed to be a first pass at our policy, and to try and make it sufficiently easy to disclose LLM use in the PR body without judgment.  Before the change, it was a complete slog trying to get people to mention what they'd used, and soem people would (I'm _pretty_ sure) just lie to us about it

For the comment:

> Any code generated by LLM or modified from LLM suggestions must commented inline too.

This _is_ intended to refer to code comments.  You guys on this repo aren't required to follow the same principles as Qiskit, but the intention behind the policy is that the licensing status of LLM-generated code is not 100% clear, so we were treating it similarly to the licensing requirements of porting/using code from another open-source codebase.  The way I wrote it to Ian is:

> rather than "Function XYZ was originally taken from `<url>` at commit `<sha1>` and used under the terms of Apache-2.0." or whatever, it's "Function XYZ was originally partially generated by LLM `<model>`."

We don't worry about it when it's trivial changes that wouldn't be copyrightable, and I won't sit here and tell you that we successfully enforce it on every PR, but that _was_ the intent.  When I'm the reviewer, I do roughly try and enforce it.

---

We were actually talking relatively recently within the Qiskit SDK that we should revisit the wording of the PR prompt, because it could be clearer, and when I wrote out those three options, I wasn't imaginative enough about how people use LLMs - the message was written a few months ago, just when the influx was starting.

.github/PULL_REQUEST_TEMPLATE.md
jakelishman: Yeah, this is what I meant.  Basically: "I talked to an LLM, but none of its output tokens in this patch, either verbatim or later modified by me.", but I was trying to fit it into as few words as possible so there'd be less excuse not to read it.

jakelishman:
btw, of the three options, the only one I _actually_ care about knowing is if they used generative AI for any of the code that made it into the patch.  The middle option about "I used it for the PR description" was there so that people who use LLMs to help write in English would have an option to tick that wasn't "I used it to generate code" and "I didn't use an LLM", without me needing to spell out in far too many words what constitutes "generated code [that I actually care about from a potential licensing problem situation]"

diego-plan9:
Thanks for the insight, Jake! This brings in more clarity into the topic, and while in this repository we are indeed at a different stage when it comes to LLM-assisted contributions (small percentage from outside contributors), with the template changes we should be in a better place for accommodating them (and make the disclosing easier and more streamline). 

For what is worth, I do like the simplicity of the template lines (thanks for those!), even if it is a first pass  - we'll be keeping an eye on how it evolves in Qiskit, as I also value staying on par to the extent possibles specially for easing contributor's life between repos.

PR #2875, opened 2026-06-02
[NFC] Move zne test file to correct folder

TsafrirA:
### Summary
#2843 misplaced a test file in the old test folder. The file is moved to the correct folder.


diego-plan9:
This comes from https://github.com/Qiskit/qiskit-ibm-runtime/pull/2843#discussion_r3314151188, and probably boils down to the concept of `correct`.

As there is only one file at the moment, I advocated for not creating an extra folder, and just place it in the parent folder. It is the same approach we use in other places (for example, `test_execution_span.py`): even if `foo` is a module that has `foo.bar` and `foo.baz`, if the test is reasonably small, we just start by `test_foo.py` and later decide if it merits converting to `foo/test_bar.py` and `foo/test_baz.py`. 

From my point of view, it saves us some nesting churn and is more evident / effective, so that is my _subjective_ definition of "correct". I'll refrain from approving / rejecting in the hopes of a more unified definition of "correct", as I'm biased 😄

TsafrirA:
Thanks @diego-plan9. I was hesitant about the `zne` sub folder and don't mind removing it, but note that the file is under `estimator_v2` which is incorrect.

diego-plan9:
> note that the file is under `estimator_v2` which is incorrect.

Indeed! On that part, the move to the `test/unit/executor_estimator` hierarchy looks good to me, we missed it in the review.

TsafrirA:
@diego-plan9 
Removed the ZNE subfolder.

SamFerracin (approved):

diego-plan9 (approved):

PR #2876, opened 2026-06-02
Add dynamical decoupling to wrapper estimator

### LINKED ISSUE #2799: Support dynamical decoupling
### Summary
We need to support DD in both vanilla estimator, and all other error-mitigating versions.

For vanilla estimator, the prepare function of #2795 needs to be modified.
For other versions imported from external sources, we need to negotiate the responsibility of adding DD.

TsafrirA:
### Summary
~~Depends on #2818~~
Closes #2799 

This PR adds dynamical decoupling support. DD pass manager runs after `prepare` so it will be applied to all error mitigation techniques (once introduced).


qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -148,7 +149,31 @@ def run(
         else:
             shots = int(np.ceil(1.0 / (self.options.default_precision**2)))
 
+        if self.options.dynamical_decoupling.enable:
```
SamFerracin, line 152: Three small comments regarding the changes in prepare:
1. Can we move all validation together? Meaning: That check that you do about backend being None can happen in this block
2. Can you remind me why we need to check if backend is None? Is that ever the case?
3. This logic seems to be the same (at least conceptually) the same as the logic needed by wrapper sampler. Would it make sense to encapsulate it in a util function with signature as below?

``` python
def apply_dynamical_decoupling(
    quantum_program, dd_options
) -> quantum_program
```

TsafrirA: Converted to a shared utility function for the estimator and sampler. Removed the unnecessary validation of backend from both.

SamFerracin (approved):

PR #2877, opened 2026-06-02
Automate TYPE_CHECKING and isort checking

### LINKED ISSUE #2563: Automate `if TYPE_CHECKING` usage

diego-plan9:
### Summary

Enable:
* https://docs.astral.sh/ruff/rules/#isort-i
* https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc

in the `ruff` configuration, so we get automatic `if TYPE_CHECKING` enforcing (helps with import cycles) and we also get consistent ordering of imports. 

### Details and comments

Fixes #2563

### AI/LLM disclosure


qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
```diff
@@ -47,6 +49,9 @@ def draw_circuit_schedule_timing(
     Returns:
         A plotly figure.
     """
+    # Use runtime imports, to prevent circular imports.
+    from ..utils.circuit_schedule import CircuitSchedule
+
```
diego-plan9: This is the only manual change in this PR. Related to #2864 (if that PR lands, this will likely not be needed).

pyproject.toml
```diff
@@ -105,6 +107,19 @@ ignore = [
     "F403",   # undefined-local-with-import-star
 ]
 
+"qiskit_ibm_runtime/__init__.py" = [
+    "I",      # isort -- requires specific import order and tweaks
+]
+
+"qiskit_ibm_runtime/utils/__init__.py" = [
+    "I",      # isort -- requires specific import order and tweaks
+]
```
diego-plan9, line 116: There is some import cycle "hidden" by the current ordering of this file (like related to the `.json` imports). Moving `.json` outside of `utils` will likely take care of it, but it is a bigger refactor - in the meantime, we refrain ruff from touching the file.

pyproject.toml
```diff
@@ -105,6 +107,19 @@ ignore = [
...
+
+"qiskit_ibm_runtime/options/*options.py" = [
+    "TC",     # flake8-type-checking -- ruff moves Unset and Dict to TYPE_CHECKING block
+]
```
diego-plan9, line 120: `ruff` attempts to move `Unset` and `Dict` (as it cannot detect they are not used only for type hints, but also for the definitions). So this avoids changes in the files that contain old-style models.

diego-plan9:
Noisy PR, that will cause annoying conflicts: happy to coordinate for merging at the right time (ie. with fewer PRs ongoing). For reviewers, all changes except one come from:

```
ruff check qiskit_ibm_runtime test tools docs --fix --unsafe-fixes
ruff format qiskit_ibm_runtime test tools docs
```

SamFerracin (dismissed):
Very nice!

qiskit_ibm_runtime/visualization/draw_circuit_schedule_timings.py
diego-plan9: No longer needed as #2864 merged - reverted at cda8479.

SamFerracin (approved):

PR #2883, opened 2026-06-04
Add wrapper sampler parity test decorator

leonapivato:
### Summary

Adds the logic for the new decorator that runs the decorates test once for each sampler implementation. Also inserted the decorator in relevant integration tests.

### Details and comments

A future todo is to add a flag to the decorator to control which implementation runs instead of running both all the time.

Fixes #

### AI/LLM disclosure


test/decorators.py
```diff
@@ -87,6 +83,39 @@ def _wrapper(self, *args, **kwargs):
...
+        ):
+            overrides = {}
+            if has_sampler_v2:
+                overrides["SamplerV2"] = sampler_cls
+            if has_sampler:
+                overrides["Sampler"] = sampler_cls
+
+            with self.subTest(sampler=label):
+                with patch.dict(test_func.__globals__, overrides, clear=False):
+                    test_func(self, *args, **kwargs)
```
diego-plan9: Thanks Leonardo - and nice usage of the the decorator pattern and `unittest` facilities, with creative and working usage of `__globals__` and patching.

From another perspective, using and modifying `__globals__` comes with some drawbacks, specially for readability and predictability. While the decorator substitutes the Sampler class to use in a very transparent way, that transparency involves a hit in reasoning for humans and tools (ie. casual and not so casual readers would expect `sampler = Sampler(...)` to use the class defined by the `import` of the file, forcing them to keep the decorator magic in their minds when maintaining the file) and arguably clashes quite a bit with the "Principle of Least Astonishment".

Can you explore a solution that sacrifices a bit of the magic for a more explicit and maintainable approach? Some ideas:
* [`run_integration_test()`](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/test/decorators.py#L68-L81) as an inspiration for a decorator that passes additional arguments to a test function (via `kwargs` modifying)
* [`combine()`](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/test/utils.py#L376-L388) as an inspiration for creating decorator based on existing [`ddt`](https://ddt.readthedocs.io/en/latest/index.html) decorators

Depending on how the exploration goes, a possible alternative would be to just resort to `ddt` facilities, similar to:
```python
@data(LegacySamplerV2, ExecutorSamplerV2)
def test_something(self, sampler_class):
    ...
    sampler = sampler_class(self.backend, ...)
```

The end goal is to have a better trade-off between magic and maintainability/explicity when we want to easily have a test use either of the sampler implementations.

leonapivato: Understood, the decorator has been reworked to pass down which class implementation through the function arguments.

Additionally I added support for a pytest flag to run against both sampler implementations with the default case, which the CI uses, to run against just the legacy.

test/integration/test_sampler_v2.py
```diff
@@ -14,24 +14,27 @@
...
+from qiskit.circuit.library import real_amplitudes, UnitaryGate
 from qiskit.primitives import PrimitiveResult, PubResult
 from qiskit.primitives.containers import BitArray
 from qiskit.primitives.containers.data_bin import DataBin
 from qiskit.primitives.containers.sampler_pub import SamplerPub
 from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
 
-from qiskit_ibm_runtime import SamplerV2 as Sampler
 from qiskit_ibm_runtime import Session
+from qiskit_ibm_runtime import SamplerV2 as Sampler
```
SamFerracin, line 28: I would have expected this import to go, but it's still here. Are there tests that we can only run with the legacy sampler? Probably some with n shots, eh?

test/decorators.py
```diff
@@ -87,6 +84,21 @@ def _wrapper(self, *args, **kwargs):
     return _wrapper
 
 
+def run_configured_sampler_implementations(
+    test_func: Callable[..., Any],
+) -> Callable[..., Any]:
+    """Parameterize sampler tests based on the configured implementations."""
```
SamFerracin: I'd add some docs on how to switch
```suggestion
    """Parameterize sampler tests based on the configured implementations.
    
    Supports running with the legacy sampler (default), as well as with the
    new implementation of Sampler around Executor. For the latter,
    please do this: ... 
    """
```

test/integration/test_sampler_v2.py
leonapivato: Yes, the per-pub shot counts and local mode are the two main ones.

leonapivato: Local mode matters here because the backend gets mocked in one of the tests.

leonapivato:
Without IQP access I cannot pass all the tests on my end since my open-instance can't use session mode. When I get access I plan to run through it and ensure everything passes.

test/decorators.py
```diff
@@ -87,6 +84,26 @@ def _wrapper(self, *args, **kwargs):
...
 
+def run_configured_sampler_implementations(
+    test_func: Callable[..., Any],
+) -> Callable[..., Any]:
+    """Parameterize sampler tests based on the configured implementations.
+
+    Set ``QISKIT_IBM_TEST_BOTH_SAMPLER_IMPLEMENTATIONS=1`` to expand the wrapped
+    test over both the legacy sampler and the executor-based sampler.
+    Otherwise by default, the wrapped test is expanded only for the legacy sampler.
+    """
```
diego-plan9, line 102:
```suggestion
    Otherwise by default, the wrapped test is expanded only for the legacy sampler.

    The decorated tests receive a new argument that contains the sampler class.
    """
```

Mostly for a bit of maintainer convenience so the new arg can be inferred just by reading the docstring.

test/integration/test_sampler_v2.py
diego-plan9:
```suggestion
"""Tests for Sampler V2 (non-wrapper and optionally also wrapper)."""
```

For a hint for maintainers / developers and prevent assuming it is just for legacy sampler.

test/decorators.py
```diff
@@ -87,6 +84,26 @@ def _wrapper(self, *args, **kwargs):
...
+    from qiskit_ibm_runtime import SamplerV2 as LegacySamplerV2
+    from qiskit_ibm_runtime.executor_sampler import SamplerV2 as ExecutorSamplerV2
+
+    implementations = (
+        [("legacy", LegacySamplerV2), ("executor", ExecutorSamplerV2)]
+        if os.getenv("QISKIT_IBM_TEST_SAMPLER_V2_IMPLEMENTATIONS") == "1"
+        else [("legacy", LegacySamplerV2)]
+    )
+    return named_data(*implementations)(test_func)
```
diego-plan9, line 111: Nice touch for having sensible names in the tests!

diego-plan9 (commented):
Thanks Leonardo! A very minor couple of suggestions for future developers convenience, otherwise it looks really good, great work.

Please note that you also need to pull from main to resolve the conflicts (seem to be just minor import issues), and in order for CI to get triggered.

diego-plan9 (approved):
Thanks Leonardo! 🎉

PR #2889, opened 2026-06-04
Add ZNE extrapolation module

### LINKED ISSUE #2848: Implement ZNE extrapolation functionality
Create an `extrapolation` module which takes an `EstimatorResult` containing noise-amplified experiments, a list of models to fit to the experiment data, and a list of points to approximate with extrapolation.

- Ensure all previously supported models are supported
- Ensure output values and metadata are formatted the same as current EstimatorV2 implementation
- Perform all extrapolation using NumPy and SciPy
- Unit tests fully covering the module

caleb-johnson:
### Summary

This PR introduces an extrapolation module used for ZNE. It introduces one public function, `process_extrapolated_expectation_values`, which is identical in name and signature to the legacy implementation. We may need to expose more functions from this module when we add PEA -- TBD.

To reduce software complexity, all non-fallback models will now be implemented with NumPy and piped through the same `scipy.curve_fit` call to fit to the data. Additionally, the variance for each extrapolated point will be calculated in the same way, regardless of model, using the method described [here](https://www.astro.rug.nl/software/kapteyn/kmpfittutorial.html#confidence-and-prediction-intervals) (same method used for non-linear variance estimation by legacy fitting software internally). Piping all fits/extraps through the same workflow allowed for a significant reduction in branches for different models, and thus a reduction in code complexity.

This function is backward compatible with the legacy function of the same name; however, different behaviors may be noticed in degenerate cases. For example, using `exponential` on a curve with a few (~5) linear points may now lead to the fit being (correctly) rejected; whereas before, the extrapolated value may have collapsed to zero with an erroneously small variance (will show this in unit tests).

### Details and comments
Fixes #2848

This implementation uses `numpy/scipy` to perform curve fitting and extrapolation.

- [x] Add test suite for new module
- [x] Beef up public docstring. Include references to the heuristic variance calculation technique.

### AI/LLM disclosure

- [X] I used the following tool to generate or modify code: claude opus 4.6


qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -0,0 +1,477 @@
...
+
+    For every result the selected ("best") extrapolation is stacked on top of every
+    model's extrapolation, and the raw un-extrapolated noise-factor data is concatenated
+    onto the result. Callers decide whether extrapolation runs; this function always
+    extrapolates.
+
+    Args:
+        result: An estimator result optionally containing multiple ZNE noise factors
+            for each indexed result.
+        extrapolator: A builtin model name or sequence of names to use for extrapolation.
```
caleb-johnson: Itemize the extrapolators here and in the type hint above

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+def process_extrapolated_expectation_values(
+    result: EstimatorResult,
+    extrapolator: str | Sequence[str],
+    extrapolated_noise_factors: float | ArrayLike = 0,
+) -> EstimatorResult:
+    """Fit extrapolation models to an estimator result and evaluate at the target noise factors.
+
+    For every result the selected ("best") extrapolation is stacked on top of every
```
caleb-johnson: Give a better explanation of the structure of the data coming out (this is pretty good but could be better). Add link to docstring to explanation of how variance is calculated for extrapolated points.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -0,0 +1,499 @@
...
+from qiskit.primitives import EstimatorResult
+from scipy.optimize import curve_fit
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Sequence
+
+    from numpy.typing import ArrayLike
+
+
+_VALID_NAMES = [
```
SamFerracin, line 32: A few of comments here (which we can tackle in future PRs, to keep this lightweight):
1. I would make this public and add a docstring. Likely they'll need to appear in the public documentation, not sure how though
2. I would adopt the same strategy as in samplomatic:
  - Have literals, with all the values listed (included all the polynomial degrees from 1 to 7) -- example [here](https://github.com/Qiskit/samplomatic/blob/main/samplomatic/transpiler/passes/add_inject_noise.py#L76)
  - Add a decorator to validate literals -- example [here](https://github.com/Qiskit/samplomatic/blob/main/samplomatic/transpiler/passes/add_inject_noise.py#L66)

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -0,0 +1,499 @@
...
+    the result's noise factors and evaluated at the target noise factor(s) (``0`` for zero
+    noise). Models are tried in priority order: each point takes the result of the
+    first model whose extrapolation is valid. A valid extrapolation results in a finite
+    value and standard error, with the :math:`value \pm stderr` plausible with respect to
+    the measurement basis. If no models produce a valid value, the measured input value with
+    the smallest standard error will be returned.
+
+    The standard errors reported for the extrapolated values are first-order estimates
+    propagated from the fit covariance. For details see the `confidence and prediction intervals
+    section of this kapteyn tutorial`
```
SamFerracin: Slight cosmetic preference for this
```suggestion
    propagated from the fit covariance. For details see the confidence and prediction intervals
    section of this kapteyn tutorial, `link`
```

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -0,0 +1,499 @@
...
+    "linear",
+    "exponential",
+    "double_exponential",
+    "polynomial_degree_(1 <= k <= 7)",
+    "fallback",
+]
+
+
+def process_extrapolated_expectation_values(
+    result: EstimatorResult,
```
SamFerracin, line 44: What's the flow here, to make sure that we reach this function with an `EstimatorResult`?

I guess:
- The post-process function uses the v01 function already in main to map quantum program results to estimator results, neglecting the fact that the user requested ZNE.
- Do the estimator results have the right metadata (the ones you use below)? Or do we need a step to manipulate the metadata?
- Then we use this function, that maps "raw" estimator results to "zne-fitted" estimator results.

Does this make sense?
Tagging @TsafrirA and @dekelmeirom here, to see if they have thoughts

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -0,0 +1,499 @@
...
+            and _poly_degree(name) is None
+            and name
+            not in (
+                "exponential",
+                "double_exponential",
+            )
+        ):
+            raise ValueError(
+                f"Unsupported extrapolator name: {name}, must be one of {_VALID_NAMES}"
+            )
```
SamFerracin, line 163: With literals + decorators, this all dies

SamFerracin (commented):
A few high-level questions/comments :)

test/unit/executor_estimator/test_zne_extrapolation.py
```diff
@@ -0,0 +1,170 @@
...
+import numpy as np
+from ddt import data, ddt, unpack
+
+from qiskit_ibm_runtime.executor_estimator.zne.extrapolation import (
+    _as_noise_factors,
+    _clamp_degenerate_stds,
+    _copy_metadata,
+    _multi_exp,
+    _poly,
+    _poly_degree,
```
SamFerracin: I would remove the underscore from all these funcs, especially if they are so important that we feel like testing them individually. In general, in qiskit-ibm-runtime everything is considered private except what is included in the public-facing docs

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
caleb-johnson: Yes, agreed. I have them listed in the public function's docstring for now. Will leave this open until we decide how to document the valid extrapolators

test/unit/executor_estimator/test_zne_extrapolation.py
caleb-johnson: I can remove the underscores if things are hidden in the docs by default, although I do like using these prefixes to indicate a function was written with the understanding it wouldn't be used outside the module, This lets anyone who is considering making a function public know that they should look closely for any hidden assumptions in the function.

None of the private functions stood out to me as needing to be a part of the API, but that might change as we implement more of this QEM software.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
caleb-johnson: Leaving this here in this PR. Will consider how to clean this up in future PRs

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
caleb-johnson: We have some options when we port the ZNE options functionality. `options/zne_options.py` already has an `ExtrapolatorType` class we could consider porting. Resolving this here. Can revisit this when porting the options

pyproject.toml
```diff
@@ -156,6 +156,7 @@ dependencies = [
     "requests>=2.32.4",
     "requests-ntlm>=1.1.0",
     "numpy>=1.26.2",
+    "scipy>=1.5.0",
```
caleb-johnson, line 158: @SamFerracin @diego-plan9 , as discussed I'm making scipy a core dependency here and removing it from the environments below. I used v1.5 to match the Qiskit minimum.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
caleb-johnson: Yes, this function supports the same flow as the legacy estimator. `EstimatorResult` in, `EstimatorResult` out. We might want these functions to handle the ZNE information directly and let the caller handle packing/unpacking results/metadata objects.

pyproject.toml
diego-plan9: Sounds good on both fronts (making it a core dependency, and matching Qiskit from `2.2.0`, which is indeed `1.5`). Thanks!

test/unit/executor_estimator/test_zne_extrapolation.py
diego-plan9: I also vouch for removing the underscores. I agree they do convey some sense of the item not being public (and implicitly caution users about depending on them), but it is not a convention we follow strongly in this package. Instead, as Sam mentioned, we just consider that the public items are the ones documented at https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime, and everything else is considered private.

Removing the underscore would keep the codebase more consistent (not fully, as there are still functions with underscores, but overall fits better with the convention).

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
dekelmeirom: There is no correct or wrong answer here, but I personally like that parameters are passed as inputs for a function rather than inside the metadata of an object.
The current post process flow (for vanilla and PEC) - 
* there is a wrapper function `estimator_v2_post_processor_v0_1` that extracts the needed data from the job result object, loops over the different pubs' results and calls the relevant processing method.
* The expectation value (and std) is calculated for each pub separately in the relevant function (`process_expectation_values` or `process_expectation_values_pec`).
* `estimator_v2_post_processor_v0_1` wraps the results from each pub into a EstimatorResult object.

It seems that the function here loops over the results and calculates the expectation value for each pub separately, so I think it can also follow the flow of the other methods - 
* Create a function named something like `process_expectation_values_zne` and make it process the expectation values of a single pub (out from an item_result and observables).
* The function will calculate the raw expectation values and raw stds, and call this function, which will have the same functionality, but without the need for the loop, and the`zne_noise_factors` and  `standard_error` will be direct input and not extracted from metadata. In addition, the output should be just a tuple of (exp_vals, stds, ensemble_stds) for a single pub.
* The `estimator_v2_post_processor_v0_1` should wrap the results from all of the pubs into EstimatorResult, like it does for other methods.

dekelmeirom:
As we talked in one of the meetings, we might want to merge this PR, and do changes as needed in separated PRs

dekelmeirom (dismissed):
We can merge it and do input/output changes in another PR

dekelmeirom (approved):

diego-plan9 (approved):

PR #2890, opened 2026-06-04
Add converters for executor models version 1.1

SamFerracin:
## Summary
This PR adds converters to/from the newly-released executor models version 1.1. It does not yet bump the default schema version to 1.1 for jobs submitted by executor, since 1.1 are not yet accepted server-side.

## AI/LLM disclosure


pyproject.toml
```diff
@@ -159,7 +159,7 @@ dependencies = [
     "urllib3>=2.4.0",
     "python-dateutil>=2.9.0",
     "ibm-platform-services>=0.61.1",
-    "ibm-quantum-schemas>=0.7.20260419,<1",
+    "ibm-quantum-schemas>=0.8.20260605,<1.0",
```
SamFerracin:
```suggestion
    "ibm-quantum-schemas>=0.8.20260605,<1",
```

qiskit_ibm_runtime/executor/executor.py
```diff
@@ -92,7 +92,6 @@ def __init__(
     ):
         # Coerced to `ExecutorOptions` via `__setattr__()`.
         self.options = options if options is not None else ExecutorOptions()  # type: ignore[assignment]
```
SamFerracin, line 94:
```suggestion
        self.options = options if options is not None else ExecutorOptions()  # type: ignore[assignment]

```

qiskit_ibm_runtime/decoders/quantum_program/decoder.py
```diff
@@ -67,6 +71,7 @@
     "v0.1": (quantum_program_result_from_0_1, QuantumProgramResultModel_0_1),
     "v0.2": (quantum_program_result_from_0_2, QuantumProgramResultModel_0_2),
     "v1.0": (quantum_program_result_from_1_0, QuantumProgramResultModel_1_0),
+    "v1.1": (quantum_program_result_from_1_1, QuantumProgramResultModel_1_1),
```
diego-plan9, line 74: As I think 99% of the new code is a copy (rightfully so due to the nature of `1.1` vs `1.0`), part of me thinks we might better off with just:

```
"v1.1": (quantum_program_result_from_1_0, QuantumProgramResultModel_1_0),
```

It might be premature optimization and the saving in duplication might not compensate for bending the unwritten conventions (one converter per version, files, etc), though. So I'm comfortable reevaluating when the number of versions grows and it is easier to see if this was a bit of an exception.

diego-plan9 (commented):
Thanks Sam: with a bit of love to the PR description, should be good to go.

We might benefit from a release note, as users will likely welcome some visibility on the new feature supported: although perhaps it is better tackled closer to the release and once more pieces at the API level are in place.

qiskit_ibm_runtime/decoders/quantum_program/decoder.py
SamFerracin: I agree that we need to find a better strategy, this is on my radars. However, I am not sure that I want to tackle it in this PR, as I would like to fine-tune the strategy, particularly around the fact that it may be misleading/confusing for the reader to see that "v1.1" uses 1.0 converters

SamFerracin:
> We might benefit from a release note, as users will likely welcome some visibility on the new feature supported: although perhaps it is better tackled closer to the release and once more pieces at the API level are in place.

I added a release note. I am not sure if we want to keep it as is in the final changelog, perhaps we want a combined statement of the form "added converters to/from 1.1 [see this PR] and changed default schema version for executor [see this other PR]"

qiskit_ibm_runtime/decoders/quantum_program/decoder.py
diego-plan9: Sounds good to reevaluate in the future!

diego-plan9:
> I added a release note. I am not sure if we want to keep it as is in the final changelog, perhaps we want a combined statement of the form "added converters to/from 1.1 [see this PR] and changed default schema version for executor [see this other PR]"

Thanks, I was thinking more from the point of features it enables (ie. taking advantage of ssv 4). Either way having the release note acts as a reminder for preventing release until all the pieces are there, and we can reshuffle the content and pointers to PRs later.

diego-plan9 (approved):

PR #2894, opened 2026-06-06
Re-organize / clean parts of utils package

diego-plan9:
### Summary

Revise the `utils` package, aiming to reduce the imports traversing through it from other packages and remove unused code, as well as focusing on reducing the amount of functionality in the really diverse `utils.utils`:
* remove unused members
* for functions used by just one other package, move them to such package:
   * `account.utils`
   * `api.utils`
   * moving `backend_encoder.py` to `fake_provider`
* moved the logging functionality to its own module

### Details and comments

Related to #2749

This is not meant to tackle all the possible revamps of `utils`. After this step, `utils` roughly contains:
* utils used by the backends (backend converter, backend decoder) - prone to be combined in a module
* utils for validation (validations and most of the items in `utils.utils`) - prone to be combined together
* utils for a well defined purpose (deprecation, logging, default session) - prone to stay or bubble up in some cases
* smaller functions for time conversion and other functionality used in several places - requires more evaluation
* `json.py` - prone to be promoted to top level, requires some finesse as it is public

Import-wise, tackling `json.py` remains the biggest source of import complexity, and moving it out would alleviate most of the issues for keeping a well-behaved `utils` package. 

### AI/LLM disclosure


SamFerracin (approved):
Another one where it's hard to make sure, but as far as I can see, all good!

PR #2897, opened 2026-06-08
add pec prepare functionality

### LINKED ISSUE #2844: Add support for PEC

dekelmeirom:
### Summary
Add support for the case pec_options is given as input.
Create a new prepare_pec function - force gate twirling and add noise injection annotations to the circuit. In addition, calculate the gamma factor of the circuit and pass it using the passthourgh_data.

### Details and comments
split #2860 PR for easier review.

partially fixes #2844


dekelmeirom:
I consider moving the `prepare_pec` function into the `pec_utils.py` file. It will make all the pec preparation related functions into the same file (but on the other hand, it will cause the different prepare functions to be in different files).
Would like to hear your comments about it.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -303,3 +446,124 @@ def compute_samplex_arguments(
...
+        raise ValueError("Could not find a change basis annotation.")
+
+    return samplex_arguments
+
+
+def box_circuit(
+    circuit: QuantumCircuit,
+    twirling_options: TwirlingOptions,
+    measure_noise_learning: MeasureNoiseLearningOptions | None = None,
+    pec_options: PecOptions | None = None,
```
SamFerracin: You are passing `pec_options`, but you are only using it to check if it is None or not. Is that correct? If so, I would suggest passing a bool:
```suggestion
    inject_noise: bool = False,
```
In this way, we can easily use this for PEA as well, without having to make it become `pec_or_pea_options`

SamFerracin: And for consistency, you may want to replace `measure_noise_learning: MeasureNoiseLearningOptions | None = None` with `twirl_measurements: bool`, since also that one is only used in an `is None or not?` clause

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -303,3 +446,124 @@ def compute_samplex_arguments(
     ]
 
     return flat_parameter_values, change_basis, param_basis_pairs
+
+
+def build_basic_samplex_args(
```
SamFerracin: NIT, non-blocking suggestion:
```suggestion
def make_samplex_arguments(
```
Mostly because I was wondering what "basic" means here

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -303,3 +446,124 @@ def compute_samplex_arguments(
...
+    samplex: Samplex,
+    boxed_circuit: QuantumCircuit,
+    flat_parameter_values: npt.NDArray[float],
+    change_basis: npt.NDArray[int],
+) -> dict[str, Any]:
+    """Build a samplex args dictionary consisting change_basis and parameters data.
```
```suggestion
    """Build a samplex args dictionary consisting of ``change_basis`` and parameters data.
```

The change to `change_basis` is due to the fact that sphinx will otherwise render underscores as italic when we publish the documentation

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -303,3 +446,124 @@ def compute_samplex_arguments(
...
+
+    Args:
+        samplex: a samplex object to create args to.
+        boxed_circuit: a boxed circuit related to the samplex.
+        flat_parameter_values: a flattened array of parameter values.
+        change_basis: an array of bases to change.
+
+    Returns:
+        a samplex args dictionary.
```
SamFerracin: Minor edits to make sure that we're consistent with the rest of the repo
```suggestion
    Args:
        samplex: A samplex object to create args to.
        boxed_circuit: A boxed circuit related to the samplex.
        flat_parameter_values: A flattened array of parameter values.
        change_basis: An array of bases to change.

    Returns:
        A samplex args dictionary.
```

qiskit_ibm_runtime/executor_estimator/prepare.py
SamFerracin: Oh I see, there already is a `compute_samplex_arguments` function.. We could consider changing name of that other one? Even in a follow-up PR, to keep the diff of this PR small

qiskit_ibm_runtime/executor_estimator/prepare.py
SamFerracin, line 52: I went through `prepare` in detail, as it had more changes than I expected (I expected no changes). However, I see now that all you did was moving duplicate logic in dedicated functions. Could you confirm that I'm not missing any actual change in functionality?

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -92,53 +99,188 @@ def prepare(
...
+            twirling_options.shots_per_randomization,
+        )
+    else:
+        num_randomizations = 1
+        shots_per_randomization = shots
+
+    # validate noise_model_mapping length
+    if noise_model_mapping is None or (
+        noise_model_mapping is not None and len(noise_model_mapping) != len(pubs)
+    ):
```
```suggestion
    # validate noise_model_mapping length
    if noise_model_mapping is None or len(noise_model_mapping) != len(pubs):
```

qiskit_ibm_runtime/options_models/pec_options.py
```diff
@@ -0,0 +1,43 @@
...
+
+    * A value of 0 corresponds to removing the full learned noise.
+    * A value of 1 corresponds to no removal of the learned noise.
+    * A value between 0 and 1 corresponds to partially removing the learned noise.
+    * A value greater than one corresponds to amplifying the learned noise.
+
+    If "auto", the value in the range ``[0, 1]`` will be chosen automatically
+    for each input PUB based on the learned noise strength, ``max_overhead``,
+    and the depth of the PUB.
+    """
```
SamFerracin, line 49: Another suggestion to stick to the conventions
```suggestion
    """The amount by which to scale the noise, where:

    * A value of ``0`` corresponds to removing the full learned noise.
    * A value of ``1`` corresponds to no removal of the learned noise.
    * A value between ``0`` and ``1`` corresponds to partially removing the learned noise.
    * A value greater than one corresponds to amplifying the learned noise.

    If ``"auto"``, the value in the range ``[0, 1]`` will be chosen automatically
    for each input PUB based on the learned noise strength, ``max_overhead``,
    and the depth of the PUB.
    """
```

qiskit_ibm_runtime/options_models/pec_options.py
```diff
@@ -0,0 +1,43 @@
...
+
+@dataclass(config=PRIMITIVES_CONFIG)
+class PecOptions:
+    """Probabalistic error cancellation mitigation options. This is only used by V2 Estimator."""
+
+    max_overhead: float | None = 100
+    """The maximum circuit sampling overhead allowed, or ``None`` for no maximum.
+    """
+
+    noise_gain: float | Literal["auto"] = "auto"
```
SamFerracin: We should add validation (possibly inside this dataclass) that `noise_gain` is either "auto" or greater than 0. We have the same kind of validation in the legacy options

SamFerracin: Same for `max_overhead`--the legacy options validate that it is larger than 0

qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -0,0 +1,52 @@
...
+    Returns:
+        The PEC gamma factor.
+    """
+    gamma = 1.0
+    for instr in boxed_circuit:
+        if annot := get_annotation(instr.operation, InjectNoise):
+            plm = noise_model_mapping[annot.ref]
+            # scale the noise by noise_factor
+            plm = plm.scale_rates(noise_factor)
+            gamma *= plm.inverse().gamma()
```
SamFerracin, line 71: Is this doing, essentially, `float(np.exp(2 * np.sum(rates)))`?

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -92,53 +99,188 @@ def prepare(
...
+        samplex_arguments = build_basic_samplex_args(
+            samplex, boxed_circuit, flat_parameter_values, change_basis
+        )
+
+        # add samplex_arguments related to noise injection
+        if pec_options.noise_gain == "auto":
+            scaleless_gamma = calculate_gamma(boxed_circuit, noise_model_mapping[i], 1)
+            noise_gain = 1 - np.log(max_overhead) / np.log(scaleless_gamma**2)
+            # Truncate noise_gain to [0, 1]
+            noise_gain = min(1, max(0, noise_gain))
```
SamFerracin: Personal preference, feel free to ignore:
```suggestion
            gamma = calculate_gamma(boxed_circuit, noise_model_mapping[i], 1)
            noise_gain = 1 - np.log(max_overhead) / np.log(gamma**2)
            # Truncate noise_gain to [0, 1]
            noise_gain = min(1, max(0, noise_gain))
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -92,53 +99,188 @@ def prepare(
...
-            op = instr.operation
-            if op.name == "box" and (change_basis_annot := get_annotation(op, ChangeBasis)):
-                samplex_arguments[f"basis_changes.{change_basis_annot.ref}"] = change_basis
-                break
+        samplex_arguments = build_basic_samplex_args(
+            samplex, boxed_circuit, flat_parameter_values, change_basis
+        )
+
+        # add samplex_arguments related to noise injection
+        if pec_options.noise_gain == "auto":
```
SamFerracin: We should document what happens here, namely that we calculate the noise gain as `noise_gain = 1 - np.log(max_overhead) / np.log(gamma**2)`, and why we do it

> Where should this be documented?

I guess in the options? Wdyt? Instead of saying that _If "auto", [...] will be chosen automatically for each input PUB based on the learned noise strength, ``max_overhead``, and the depth of the PUB_, which is kind of vague

> Why do we do this?

My best guess by looking at this code is that `scaleless_gamma**2` is what you'd call overhead? If that's correct, we can improve the doctring of `max_overhead` as well, and explain that by overhead we mean ``gamma**2``, and with ``max_overhead`` we limit it

SamFerracin (commented):
Some comments, mostly minor. It looks good though 😄 I struggled to follow the part where you compute noise factors, noise scalings, and while it looks correct, it'd be great if @TsafrirA could also double-check it--it's EOD for me, easy to miss small details

qiskit_ibm_runtime/executor_estimator/pec_utils.py
dekelmeirom: Yes, it is essentially doing `float(np.exp(2 * np.sum(-rates * noise_factor)))` (note the minus sign). This multiplication is done for each layer where noise will be injected.

qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        pubs: List of estimator pubs to convert.
+        twirling_options: The twirling options.
+        shots: The number of shots to use. Will be overridden by
+            ``num_randomizations * shots_per_randomization`` when both are specified explicitly
+            and twirling is on.
+        measure_noise_learning: The measure noise learning options. If provided, Twirled Readout
+            Error eXtinction (TREX) mitigation method will be used.
+        pec_options: The options for PEC mitigation.
+        noise_model_mapping: List of mapping between layer ref to a noise model to use for PEC or
+            PEA mitigation methods. The list must contain a map for each pub. Assumes that the
```
TsafrirA: I see where you're going with this, but I think it's better to leave PEA out (might require a function name change later?).

qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    pec_options: PecOptions,
+    noise_model_mapping: Sequence[dict[str, PauliLindbladMap]],
+    measure_noise_learning: MeasureNoiseLearningOptions | None = None,
+) -> QuantumProgram:
+    """Convert estimator PUBs to a quantum program with PEC mitigation.
+
+    Args:
+        pubs: List of estimator pubs to convert.
+        twirling_options: The twirling options.
+        shots: The number of shots to use. Will be overridden by
```
TsafrirA: I would clarify that this is the pre-overhead value.

qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        # The sampling scaling is proportional to 1 - noise_gain, as 0 is full PEC and 1 is no PEC
+        noise_factor = 1 - noise_gain
+        for ref in noise_model_mapping[i]:
+            samplex_arguments[f"noise_scales.{ref}"] = noise_scale
+        samplex_arguments["pauli_lindblad_maps"] = noise_model_mapping[i]
+        scaled_gamma = calculate_gamma(boxed_circuit, noise_model_mapping[i], noise_factor)
+        pec_gamma_list.append(scaled_gamma)
+        # Scale the amount of randomizations by gamma**2
+        sampling_overhead = scaled_gamma**2
+        num_randomizations = 1 if num_randomizations == 0 else num_randomizations
```
TsafrirA: I think you can't have `num_randomizations == 0`

qiskit_ibm_runtime/executor_estimator/prepare.py
dekelmeirom: Yes, you are correct.
I have moved the `prepare_pec` function into another file so now it will be easier to see the changes in `prepare`.

qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+        # Store data for passthrough
+        observables_list.append(pub.observables.tolist())
+        param_basis_pairs_list.append(param_basis_pairs)
+        param_shapes_list.append(pub.parameter_values.shape)
+
+    # Collect circuit metadata from each pub
+    circuits_metadata = [pub.circuit.metadata for pub in pubs]
+
+    # Validate that circuit metadata is compatible with DataTree format
```
TsafrirA: We recently added a utility function for that.

qiskit_ibm_runtime/options_models/utils.py
```diff
@@ -14,11 +14,67 @@
...
 
 PRIMITIVES_CONFIG = ConfigDict(validate_assignment=True, extra="forbid")
 """Custom ``ConfigDict`` for pydantic dataclasses.
 
 These config settings ensure we get validation on attribute mutation, not just at construction
 time, and also that we get a validation error if someone spells an attribute name wrong.
 """
+
+
+def make_constraint_validator(
```
TsafrirA: This seems like an over kill to me. There are simpler ways to handle this.

For example (not necessarily my favorite option, but probably the smallest line count)

```
from typing import Literal, Annotated
from pydantic import BaseModel, Field

NonNegativeInt = Annotated[int, Field(ge=0)]

class MyModel(BaseModel):
    value: Literal["auto"] | NonNegativeInt

```

TsafrirA (commented):
This looks great!
As discussed elsewhere I think there is an issue with the shots per randomization and number of randomizations selection.

TsafrirA (approved):
This looks great!
Thanks @dekelmeirom.

There is a slight issue with the calculation of `num_randomizations` and `shots_per_randomization` but it goes beyond the scope of this PR, and can be addressed separately.

PR #2898, opened 2026-06-08
add pec post-process functionality

### LINKED ISSUE #2844: Add support for PEC

dekelmeirom:
### Summary
Add support for the case pec_options is given as input.
Create a new post_process_pec function - calculate the mitigated exp_val using the pauli_signs attribute of the results and the calculated gamma factor of the noise model of the circuit.

### Details and comments
split https://github.com/Qiskit/qiskit-ibm-runtime/pull/2860 PR for easier review.

partially fixes https://github.com/Qiskit/qiskit-ibm-runtime/issues/2844


qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -95,32 +99,51 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
 
     # Validate circuits_metadata length if provided
     circuits_metadata = circuits_metadata or [None] * len(result)
+    pec_gammas = pec_gammas or [None] * len(result)
```
SamFerracin: I'm not entirely sure that we want to:
- Instantiate `pec_gammas` to a list of `None`s, if the user didn't request PEC
- Discriminate between PEC/no PEC based on whether those gammas are None or a float

Especially because then we'll need to also deal with PEA and ZNE here.

Alternatively to this approach, we could restrict this PR to introducing `process_expectation_values_pec`, and let @TsafrirA figure out how to discriminate when he incorporates PEC into the wrapper estimator. For example, one thing that we could do at that point (which may be cleaner that what's in this PR) is to let the wrapper estimator add a `"mitigation": "pec"` field to the passthrough data, and use that to do the discrimination in post-processing, in a single line

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
```diff
@@ -75,21 +75,35 @@ def project_to_z(term: str) -> np.ndarray[int]:
...
+            generators in circuit, remaining shape must be `signs.shape[:-1] == datum.shape[:-2]`,
+            as signs array does not have a shots or qubits axes.
 
     Returns:
         Tuple of (expectation_value, ensemble_variance, twirl_variance):
             - expectation_value: Mean expectation value across all shots
             - ensemble_variance: Variance treating all shots as a single ensemble
             - twirl_variance: Variance of expectation values across twirls
     """
+    net_signs_bc = np.zeros(datum.shape[:-1], dtype=bool)
```
SamFerracin: Would you mind adding a few inline comments? In particular:

- It was not immediately clear to me what "net signs" means here
- I found it a bit hard to follow what you do with the shape of signs

SamFerracin: I would default this to ones, since that's effectively the sign you'd get if you didn't do PEC
```suggestion
    net_signs_bc = np.ones(datum.shape[:-1], dtype=bool)
```

Alternatively, you could go for a more performant path and do:
```suggestion
    net_signs_bc = np.empty(datum.shape[:-1], dtype=bool)
```

dekelmeirom: I added some comments and made the shaping of signs shorter and clearer

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: I restricted this PR into only introducing `process_expectation_values_pec`, and left the extracting of the `pec_gamma` from the passthrough_data and passing it into the expectation values function to future PR

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
SamFerracin: Thanks! Looks much cleaner!

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
```diff
@@ -107,6 +111,18 @@ def compute_exp_val(
     else:
         evals = np.ones(datum.shape[:-1])
 
+    # in case signs is provided - multiply the evals by -1 if the parity of the signs is odd
+    if signs is not None:
+        # sum all the indicators of the error generators for each randomization
+        signs_per_rand = np.asarray(np.sum(signs, axis=-1) % 2, dtype=bool)
+        # transform the bool array into an array consisting -1 (for True values)
+        # and 1 (for False values)
+        signs_per_rand = np.where(signs_per_rand, -1, 1)
```
TsafrirA: You can replace this check with
`1-2*signs_per_rand`

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
```diff
@@ -107,6 +111,18 @@ def compute_exp_val(
...
+        # signs do not have a `shots` axis, so need to broadcast to the shape of the
+        # data (without the qubits axis)
+        signs_per_rand_bc = np.broadcast_to(signs_per_rand[..., np.newaxis], shape=datum.shape[:-1])
```
TsafrirA: I think you can avoid broadcasting, and just reshape to have a dummy trailing axis? `[...] -> [..., 1]`

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
```diff
@@ -75,14 +75,18 @@ def project_to_z(term: str) -> np.ndarray[int]:
...
     """Compute expectation value and variances of an observable term from measurement data.
 
     Args:
         observable_term: Observable term string (e.g., "ZZZ", "0X1", "IXI")
         datum: Boolean array of measurement outcomes, shape
             (num_randomizations, shots_per_randomization, num_qubits)
+        signs: Optional boolean array used with probabilistic error cancellation (PEC). Indicates
+            which errors were inserted in each circuit randomization. The final axis index all error
+            generators in circuit, remaining shape must be `signs.shape[:-1] == datum.shape[:-2]`,
+            as signs array does not have a shots or qubits axes.
```
TsafrirA: Can we explicitly write down the shape like we do for datum?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -254,3 +254,140 @@ def process_expectation_values(
...
+
+            # Calculate scale factor in case TREX mitigation is used
+            term_scale_factor = (
+                calculate_trex_factor(measure_noise_data, observable_term)
+                if measure_noise_data is not None
+                else 1
+            )
+
+            # Accumulate with coefficient
+            exp_val += coeff * term_exp_val * term_scale_factor * pec_gamma
```
TsafrirA: The scaling with `pec_gamma` is uniform for all observables and parameters, so we can do it outside the loop.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -254,3 +254,140 @@ def process_expectation_values(
...
+            ensemble_variance += (
+                (coeff**2) * term_ensemble_variance * (pec_gamma * term_scale_factor) ** 2
+            )
+            twirl_variance += (
+                (coeff**2) * term_twirl_variance * (pec_gamma * term_scale_factor) ** 2
+            )
+
+        exp_vals[bcast_index] = exp_val
+        ensemble_stds[bcast_index] = np.sqrt(ensemble_variance / total_shots)
+        # When twirling is off (num_randomizations=1), stds equals ensemble_standard_error
```
TsafrirA: Following our discussion on the prepare function, we need to verify what happens in this scenario. We no longer have a case with `num_randomizations=1`.

TsafrirA (commented):
Thanks @dekelmeirom for putting this together!
I left some minor suggestions, and one thing we need to look into (the no twirling case).

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: in an extreme case `num_randomizations` might be 1 (if the user set it to 1 explicitly and the noise is really low). but even in that case we can skip the branching and just calculate the `twirl_variance`, and it will result to be equal to the `ensemble_std`

TsafrirA (approved):
Thanks @dekelmeirom!
LGTM.

Issue #2852, opened 2026-05-28
Allow using multiple decoders


diego-plan9:
**What is the expected enhancement?**

In the past, more than one `ResultDecoder` was in use (seemingly for the websocket functionality), and [some remains](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/qiskit_ibm_runtime/base_runtime_job.py#L84) where multiple decoders are accepted are in place. In practice only one decoder is used, and other functions expect a single item to be passed.

This is story is about restoring the ability of passing multiple decoders. Through the recent work in the wrappers (where post-processors become decoders), there is a valid use case for using multiple decoders, allowing potentially to split the logic specific to a kind (or subkind) of program more neatly (this would also allow for noiselearnerv3 to potentially be modeled more naturally). At the same time, we restore the user's ability of using multiple decoders (even if it does not seem to be a widely used feature).

**Acceptance criteria**

More than one decoder can be passed to user-facing functions, and the wrappers and nlv3 decoders are modeled as separate decoders.

PR #2899, opened 2026-06-09
Reintroduce support for multiple result decoders

diego-plan9:
### Summary

Revise the methods that accept `result_decoder` to ensure they also support sequences of decoders (ie. `type[ResultDecoder] | Sequence[type[ResultDecoder]] | None = None`) consistently, and update the logic so the decoders are called in chain in that case.

In practice, it re-introduces the ability of using multiple decoders, with provides for some degree of easy customizability by end users. It was introduced around #1450 for dealing with the websocket mechanism, and some remains remained at the interface level although in practice [it was restricted to one decoder](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/qiskit_ibm_runtime/base_runtime_job.py#L110-L113) out of two. 

Additionally, it avoids passing the decoder explicitly from the Primitive classes, as `DEFAULT_DECODERS` is used under the hood for selecting a default (making declaring and passing them from Primitives unneeded, and saving some import pressure).

### Details and comments

Related #2852 - not marking as fixed as this does not revise existing decoders.

### AI/LLM disclosure


qiskit_ibm_runtime/runtime_job_v2.py
```diff
@@ -147,7 +156,8 @@ def result(
             )
 
         result_raw = self._api_client.job_results(job_id=self.job_id())
-        return _decoder.decode(result_raw) if result_raw else None
+        # Invoke all decoders, chaining them (one decoders output becomes the next's input).
```
SamFerracin, line 159: nice!

qiskit_ibm_runtime/executor/executor.py
```diff
@@ -79,7 +78,6 @@ class Executor:
     """
 
     _PROGRAM_ID = "executor"
-    _DECODER = QuantumProgramResultDecoder
```
SamFerracin, line 82: are these just not needed? were they just forgotten here in previous PRs?

SamFerracin (approved):

qiskit_ibm_runtime/executor/executor.py
diego-plan9: This is:

> Additionally, it avoids passing the decoder explicitly from the Primitive classes, as `DEFAULT_DECODERS` is used under the hood for selecting a default (making declaring and passing them from Primitives unneeded, and saving some import pressure).

I did not trace history fully, but I imagine it was actually in use by `BasePrimitiveV2` in an earlier iteration, and made redundant by further changes to codebase. And as the rest of the programs used it as an inspiration, we carried them over. The removal is actually independent from the support for multiple decoders, it could have happened against `main` (just made sense to tackle it while touching decoder things).

Issue #2794, opened 2026-05-10
Add an `EstimatorV2` implementation wrapped around `Executor`


TsafrirA:
**What is the expected enhancement?**
We want to create a new `EstimatorV2` implementation that wraps around the `Executor`. The new estimator should support all functionality of the legacy estimator, unless technically not available via the `Executor`. Error mitigation techniques will be imported from an external source, and used seamlessly.

**Acceptance criteria**
A user can replace the legacy estimator with the new implementation as a drop-in replacement.

**Steps**

- **Vanilla estimator** - We start with a vanilla version that doesn't do any error mitigation. This version will live in `qiskit-ibm-runtime`.
- **Error mitigation** - We dispatch according to the options to prepare and postprocess functions which are imported from external sources.

diego-plan9:
After syncing, I moved the remaining issues outside the epic, as they are either non-blocking items or they have their own entity and are on track for merging (guides).

Issue #2681, opened 2026-03-18
[Wrapper sampler MVP2] Write an implementation of SamplerV2 around Executor


SamFerracin:
This is an epic to achieve the next MVP (MVP2) for Wrapper Sampler

TsafrirA:
@SamFerracin @diego-plan9 
I think this issue could be closed as completed?
There is one stretch goal which hasn't been completed, but its merit could be debated, and even if we choose to pursue it, I don't think it should block the epic.

diego-plan9:
Sounds reasonable, let's tackle the stretch goal as a separate issue and focus efforts in next stages (ie. https://github.com/Qiskit/qiskit-ibm-runtime/issues/2908 and QA).

PR #2901, opened 2026-06-10
Avoid options as property in WrapperEstimator

diego-plan9:
### Summary

Apply part of #2739 to `WrapperEstimator`, so it handles `options` in the same way as the rest of "new" programs (instead of a decorator, uses `__setattr__` for better auto-completing behavior and for more enforcing of validation when the attribute is set, not only as part of initializing during the constructor).

Tests have been copy-pasted (and adapted to use sampler / estimator option classes). In the process an inconsistency arose related to `experimental`, which defaults to a dict in `Executor` and allows `None` in `Estimator` and `Sampler`. I favored Executor-style handling of it, for consistency.

### Details and comments

Some tweaks of the docstrings in `option_models/` included.

Related to #2794 / #2681 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/options_models/estimator_options.py
+++ b/qiskit_ibm_runtime/options_models/estimator_options.py
@@ -61,14 +61,16 @@ class EstimatorOptions:
     execution: ExecutionOptions = Field(default_factory=ExecutionOptions)
     """Execution options.
 
-    See :class:`.ExecutionOptions` for all available options."""
+    See :class:`.ExecutionOptions` for all available options.
+    """
 
     twirling: TwirlingOptions = Field(default_factory=TwirlingOptions)
     """Twirling options.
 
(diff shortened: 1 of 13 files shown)


qiskit_ibm_runtime/options_models/estimator_options.py
```diff
@@ -78,7 +80,7 @@ class EstimatorOptions:
     See :class:`~.DynamicalDecouplingOptions` for all available options.
     """
 
-    experimental: dict | None = None
+    experimental: dict = Field(default_factory=dict)
```
diego-plan9, line 83: This (and the same change for `Sampler`) is the bit that felt inconsistent. I'm unsure if supporting `None` was by design - the change is contained in 36b022a in case the agreement is to not tackle the inconsistency.

TsafrirA (approved):
LGTM

Issue #2825, opened 2026-05-21
Write a method to convert executor metadata to estimator metadata


SamFerracin:
**Acceptance criteria**
- [ ] The results returned by the wrapper estimator contain metadata formatted in the same way as those of current estimator's jobs

TsafrirA:
Replaced by #3046.

PR #2902, opened 2026-06-10
Add options to passthrough data and result metadata in wrapper estimator

TsafrirA:
### Summary
Related to #2825
This PR adds the estimator options to the pass through data, and eventually to the result metadata. This is meant to replicate the legacy estimator's behavior. This does not close the issue completely, as we are still missing some option fields, and option combinations.

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
+++ b/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
@@ -76,7 +76,7 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
     circuits_metadata = post_processor_data.get("circuits_metadata", None)
 
     # Extract options if present
-    options_metadata = post_processor_data.get("options", None)
+    options_metadata = post_processor_data.get("options", {})
 
     # Check if measure_mitigation was used
     measure_mitigation = post_processor_data.get("measure_mitigation", None)
@@ -134,7 +134,7 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
         pub_result = EstimatorPubResult(data=data_bin, metadata=pub_metadata)
         pub_results.append(pub_result)
 
-    return PrimitiveResult(pub_results, metadata=options_metadata or {})
+    return PrimitiveResult(pub_results, metadata=options_metadata)
 
 
 def process_expectation_values(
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -173,6 +174,13 @@ def run(
                 quantum_program=quantum_program,
             )
 
+        # Serialize options (assuming passthrough is correctly configured)
+        quantum_program.passthrough_data["post_processor"]["options"] = {  # type: ignore[index, call-overload]
+            "twirling": asdict(self.options.twirling),  # type: ignore[call-overload]
+            "dynamical_decoupling": asdict(self.options.dynamical_decoupling),  # type: ignore[call-overload]
+            "resilience": asdict(self.options.resilience),  # type: ignore[call-overload]
+        }
```
SamFerracin, line 195: Would this not work?
```suggestion
        # Serialize options (assuming passthrough is correctly configured)
        quantum_program.passthrough_data["post_processor"]["options"] = asdict(self.options)
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -75,6 +75,9 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
     # Extract circuit metadata if present
     circuits_metadata = post_processor_data.get("circuits_metadata", None)
 
+    # Extract options if present
+    options_metadata = post_processor_data.get("options", None)
```
```suggestion
    options_metadata = post_processor_data.get("options", {})
```
and then you don't need the `or` later

SamFerracin (commented):
Thanks! Some minor comments

qiskit_ibm_runtime/executor_estimator/estimator.py
TsafrirA: Legacy Estiamtor doesn't return all options, but rather only a portion of them. We can choose to change that, but I think at least initially it will be good to be mindful and explicit about what we include there.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: Done.

SamFerracin (dismissed):

SamFerracin (approved):

Issue #2841, opened 2026-05-26
Write transpiler passes to convert to `MidCircuitReset` and `MidCircuitMeasureReset`


yaelbh:
**What is the expected feature?**

Blocked by #2840.

We have to write the equivalent of `ConvertMidCircuitMeasure` for the new instructions.

**Acceptance criteria**

- [ ] Write `ConvertMidCircuitReset` and `ConvertMidCircuitMeasureReset`
- [ ] For optimization, consider integrating the passes into one common pass, so the dag will be scanned only once for all the conversion types

jyu00:
For `ConvertMidCircuitMeasureReset`, I guess the idea is to replace something like

```python
circuit.measure(q0, c0)
with circuit.if_test((c0, 1)):
    circuit.x(q0)
```

with 

```python
circuit.append(MidCircuitMeasureReset(), q0, c0)
```

If we want a common pass, I'd suggest making this conversion optional, since it changes more than just a simple `measure/reset` -> `measure_2/reset_2`.

yaelbh:
1. For the `measure-reset` transpilation pass one can be assisted by https://github.com/Qiskit/qiskit/blob/main/qiskit/transpiler/passes/optimization/reset_after_measure_simplification.py#L25
2. The passes can be made more efficient when https://github.com/Qiskit/qiskit/issues/16458 is done.

Issue #2840, opened 2026-05-26
Support `reset_2` and `measure_reset`, similar to `measure_2`


yaelbh:
**What is the expected feature?**

The `measure_2` instruction was introduced some time ago. 
- See [this guide](https://quantum.cloud.ibm.com/docs/en/guides/execute-dynamic-circuits#mid-circuit-measurements) about what is `measure_2`, and how to use the `MidCircuitMeasure` instruction directly
- See [doc](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/transpiler-passes-convert-to-mid-circuit-measure) about the `ConvertToMidCircuitMeasure` pass, which transforms regular mid-circuit measurements into the new intruction

The task is to add similar support to `reset_2` (mid-circuit single qubit reset) and `measure_reset` (I think it's a mid-circuit measurement followed immediately by a reset on the same qubit).

This specific issue covers only the introduction of the new instructions, excluding the corresponding transpiler passes, to be covered in a separate issue that I'll open shortly.

**Acceptance criteria**

- [ ] Write new classes `MidCircuitReset` and `MidCircuitMeasureReset`. This is probably all that's required in order to mark the issue complete.

jyu00:
> The task is to add similar support to `reset_2` (mid-circuit single qubit reset) and `measure_reset` (I think it's a mid-circuit measurement followed immediately by a reset on the same qubit).

To add more context, `reset_2` is a conditional reset that uses `measure_2` to do the measurement and flip the qubit state based on that outcome. It's useful for mid-circuit reset for the same reason `measure_2` is (shorter circuit). 

`measure_reset` is the same as `reset_2`, except it also returns the measurement outcome. This is a faster operation than doing an explicit measurement + conditional reset.

francabrera:
There will also be a `measure_reset2`.

PR #2903, opened 2026-06-10
Add reset_2

yaelbh:
### Summary

Added a class `MidCircuitReset` and updated the pass manager `ConvertToMidCircuitMeasure` to convert reset instructions to `reset_2`.

### Details and comments

Part of #2840 and #2841.

I made a decision which I now think was wrong, but I want to nevertheless push the PR as it is, in order to hear more opinions before I revert the (probably wrong) decision.

The decision was to replace the writing of a separate pass manager `ConvertToMidCircuitReset` with an update to `ConvertToMidCircuitMeasure` that covers also reset instructions.

Arguments supporting the decision:
1. Users will normally want to convert both, so it will be convenient for them to call a single pass.
2. The single pass is probably more performant than two executions of two passes.

Arguments against the decision:
1. With the default value of `mcr_name` being `reset_2` and not `reset`, it is a breaking change for anyone already running `ConvertToMidCircuitReset`.
2. May be less performant if only the measure instructions (or only the reset instructions) are to be converted.
3. More diffs in the PR, making it more difficult to review.

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Bob 1.0.3, Claude Opus 4.6


jyu00:
> 1. With the default value of `mcr_name` being `reset_2` and not `reset`, it is a breaking change for anyone already running `ConvertToMidCircuitReset`.

Especially since we are going to release `measure_2` and `reset_2` separately, so many backends will have one but not the other. 

My suggestion would be to use a single pass, but 
- issue a warning instead of an error if the target doesn't contain MCM measure/reset. This is what we did at QDC, and I thought the pass did the same way. 
- add an option to replace only measure or only reset or both. 

Also note that the `reset_2` release is delayed 😅

yaelbh:
> add an option to replace only measure or only reset or both.

This is already done, by allowing `mcm_name` to be `measure` (before it had to start with `measure_`), and similarly for `mcr_name`. The default is to replace them both with `measure_2` and `reset_2`, and this choice of defaults is the breaking change. If we want to be consistent with the current behavior, then the defaults should be `measure_2` and `reset`, but this is not I'd say esthetic.

jyu00:
> The default is to replace them both with `measure_2` and `reset_2`, and this choice of defaults is the breaking change. 

I think we can either

1. Release a new pass (`ConvertToMidCircuitMeasureReset` ?) that does both `measure_2` and `reset_2`, update the existing `ConvertToMidCircuitMeasure` to just call this new pass (with `mcr_name="reset"`), and deprecate the existing pass. After the 3 months deprecation period we'd have just the new pass. 
2. Update `ConvertToMidCircuitMeasure` to include mcr, but default to `mcr_name="reset"` with a warning saying the default will change after the deprecation period. 

Option 1 is the more proper way of making a change, but it requires users to update their code. We could also do option 1 but at the end of the deprecation period, redirect `ConvertToMidCircuitMeasure` to `ConvertToMidCircuitMeasureReset` instead of dropping it. 

By the way, I don't really like the name `ConvertToMidCircuitMeasureReset`, since we may release a `measure_reset` instruction.

qiskit_ibm_runtime/transpiler/passes/basis/convert_mid_circ_meas.py
```diff
@@ -33,46 +33,71 @@ class ConvertToMidCircuitMeasure(TransformationPass):
...
+        self.mcr_name = mcr_name
 
     def run(self, dag: DAGCircuit) -> DAGCircuit:
         """Run the pass on a dag."""
-        final_measure_nodes = calc_final_ops(dag, {"measure"})
-        for node in dag.op_nodes(Measure):
-            if node not in final_measure_nodes:
+        final_measure_nodes = calc_final_ops(dag, {"measure", "reset"})
+        for node in dag.op_nodes():
+            if isinstance(node.op, (Measure, Reset)) and node not in final_measure_nodes:
```
yaelbh: Need to fix here: it's worse performance in the case that we want only `measure_2`, i.e., if `mcr_name == "reset"`: things that were previously done efficiently in Rust are moving to Python (see e.g. the call to `isinstance`)

yaelbh:
Ready for a new review, with some comments:
- It's done in a way that doesn't require deprecation.
- There are many ways to implement the pass, with different trade-offs regarding performance. Need to consider run time of Python's `DAGCircuit.calc_final_ops` (with different parameters) and `isinstance`, and of Rust's `DAGCircuit.op_nodes`. The bottom line is that the current implementation, which at the moment I think is optimal, in practice runs two separate passes, one for measure and one for reset, without mixing the two. Perhaps there will be room for improvement if https://github.com/Qiskit/qiskit/issues/16458 is solved.
- There were already errors if the instructions are not supported by the target. I added warnings if the combination of instruction + qubits is not supported.

qiskit_ibm_runtime/transpiler/passes/basis/convert_mid_circ_meas.py
```diff
@@ -25,54 +29,130 @@
     from qiskit.transpiler import Target
 
 
-class ConvertToMidCircuitMeasure(TransformationPass):
+class ConvertToMidCircuitInstructions(TransformationPass):
     """Transpiler pass replacing mid-circuit terminal measure instructions.
```
jyu00:
```suggestion
    """Transpiler pass replacing mid-circuit measure and reset instructions.
```

qiskit_ibm_runtime/transpiler/passes/basis/convert_mid_circ_meas.py
```diff
@@ -25,54 +29,130 @@
...
+    and won't replace existing mid-circuit measurement instructions
+    (e.g., ``"measure_2" -> "measure_3"``) or convert any ``MidCircuitMeasure`` instance
+    into a ``Measure``.
 
     Args:
         target: Backend's target instance that contains one or more ``measure_`` instructions.
-        mcm_name: Name of the ``measure_`` instruction that terminal measure instructions in
+        mcm_name: Name of the ``measure`` instruction that terminal measure instructions in
             non-terminal locations will be replaced with. This instruction must be contained in
             the target. Defaults to ``measure_2``.
```
jyu00: This is a bit hard to read, how about

> mcm_name: Name of the instruction used to replace non-terminal ``Measure`` instructions. The name must start with "measure". The default name is ``measure_2``.

qiskit_ibm_runtime/transpiler/passes/basis/convert_mid_circ_meas.py
```diff
@@ -25,54 +29,130 @@
...
+        mcr_name: Name of the ``reset`` instruction that terminal reset instructions in
+            non-terminal locations will be replaced with. This instruction must be contained in
+            the target. Defaults to ``reset_2``.
 
     Raises:
-        ValueError: If the specifcied ``mcm_name`` does not conform to the ``measure_`` pattern
+        ValueError: If the specified ``mcm_name`` does not conform to the ``measure_`` pattern
             or is not contained in the provided target.
```
jyu00: This should say 

> ValueError: If the specified ``mcm_name`` does not start with "measure", or the specified ``mcr_name`` does not start with "reset", or the specified instructions are not contained in the provided target.

jyu00 (commented):
Some comments on the docstring wording but this looks good.

qiskit_ibm_runtime/transpiler/passes/basis/convert_mid_circ_meas.py
yaelbh: I copied it almost exactly in 7e9ed47b.

qiskit_ibm_runtime/transpiler/passes/basis/convert_mid_circ_meas.py
yaelbh: 56166ce2

diego-plan9:
Thanks Yael: does this need a release note?

yaelbh:
@diego-plan9 I added a release note.

diego-plan9 (dismissed):

qiskit_ibm_runtime/transpiler/passes/basis/convert_mid_circ_meas.py
```diff
@@ -25,54 +29,131 @@
     from qiskit.transpiler import Target
 
 
-class ConvertToMidCircuitMeasure(TransformationPass):
-    """Transpiler pass replacing mid-circuit terminal measure instructions.
+class ConvertToMidCircuitInstructions(TransformationPass):
```
SamFerracin: I find the name confusing, because every instruction is a mid-circuit instruction XD Can we rename it like this?
```suggestion
class ConvertToMidCircuitMeasureAndReset(TransformationPass):
```

yaelbh: We plan to add a new instruction `measure_reset`, with its class `MeasureReset` and probably also a transpiler pass `ConvertToMeasureReset`. Therefore I prefer not to adopt the proposed name `ConvertToMidCircuitMeasureAndReset`.

yaelbh: I'll do `ConvertToMidCircuitResetAndMeasure`

SamFerracin (approved):
Thanks, @yaelbh !

diego-plan9 (approved):

PR #2914, opened 2026-06-12
ExecutorSampler should be able to accept circuits that contain tuples in the metadata

### LINKED ISSUE #2913: ExecutorSampler should be able to accept circuits that contain tuples in the metadata

The existing Sampler can submit circuits with metadata objects of many different types. In contrast, the new executor-based Sampler only supports metadata that can be represented as a `DataTree`, defined as:
```python
DataTree: TypeAlias = (
    list["DataTree"] | dict[str, "DataTree"] | NDArray[float] | str | float | int | bool | None
)
"""Arbitrary nesting of lists and dicts with typed leaves."""
```
Because many Qiskit APIs commonly return tuples (for example, qubit indices, coupling maps, and similar structures), users of the existing Sampler often include tuples in metadata. The executor-based limitation therefore creates a migration burden, as seen in efforts such as porting qiskit-experiment to the new Sampler.

**Acceptance criteria**

- [ ] Users of executor sampler can submit jobs that contain circuits with tuples in the metadata
- [ ] Tuples are automatically converted to lists for internal handling and returned in list form in results

SamFerracin:
Closes #2913

### DIFF
--- a/qiskit_ibm_runtime/quantum_program/datatree.py
+++ b/qiskit_ibm_runtime/quantum_program/datatree.py
@@ -14,10 +14,14 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Any, TypeAlias
 
 import numpy as np
 from numpy.typing import NDArray
+
+logger = logging.getLogger(__name__)
+
 
 DataTree: TypeAlias = (
     list["DataTree"] | dict[str, "DataTree"] | NDArray[float] | str | float | int | bool | None
@@ -41,14 +45,18 @@ def is_datatree_compatible(data: Any) -> bool:
         return True
 
     if isinstance(data, list):
-        # Recursively check list elements
+        return all(is_datatree_compatible(item) for item in data)
+
+    if isinstance(data, tuple):
+        logger.warning(
+            "Tuples are turned into list when the payload is serialized, and attached to the "
+            "returned result list format."
+        )
         return all(is_datatree_compatible(item) for item in data)
 
     if isinstance(data, dict):
-        # Check dict keys are strings and recursively check values
         if not all(isinstance(key, str) for key in data.keys()):
             return False
         return all(is_datatree_compatible(value) for value in data.values())
 
-    # Any other type is not compatible
     return False
--- a/qiskit_ibm_runtime/quantum_program/datatree.py
+++ b/qiskit_ibm_runtime/quantum_program/datatree.py
@@ -14,14 +14,10 @@
 
 from __future__ import annotations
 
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/quantum_program/datatree.py
```diff
@@ -41,14 +45,18 @@ def is_datatree_compatible(data: Any) -> bool:
         return True
 
     if isinstance(data, list):
-        # Recursively check list elements
+        return all(is_datatree_compatible(item) for item in data)
+
+    if isinstance(data, tuple):
+        logger.warning(
+            "Tuples are turned into list when the payload is serialized, and attached to the "
```
SamFerracin: This conversion is operated automatically by the schema models when we call `dump`. In the unlikely event that a future model has a different behaviour, the integration test will fail, and we can think of more drastic solutions (such as adding support for tuples in the model itself)

qiskit_ibm_runtime/quantum_program/datatree.py
```diff
@@ -41,14 +45,18 @@ def is_datatree_compatible(data: Any) -> bool:
         return True
 
     if isinstance(data, list):
-        # Recursively check list elements
+        return all(is_datatree_compatible(item) for item in data)
+
+    if isinstance(data, tuple):
+        logger.warning(
```
diego-plan9: If  `data` contains multiple `tuples`, this would emit one warning per tuple, right? If it is a relatively common case, perhaps it would make sense to tweak the implementation so only one warning is emitted, for user-friendliness. 

An option might be to tweak the signature so the "I have found a tuple, or I have already emitted a warning" is carried over during recursiveness and emitted at the right moment, or track the existence of tuples and return it so the callers of `is_datatree_compatible` emit the warning instead.

qiskit_ibm_runtime/quantum_program/datatree.py
leonapivato: Not sure if this matters at all in practice or for the warning, but doesn't the conversion happen at model validation/creation, not serialization?

qiskit_ibm_runtime/quantum_program/datatree.py
SamFerracin: As discussed during standup, i have removed this warning

> this would emit one warning per tuple, right?

You are right, actually. I thought logger was smarter, but I hadn't tried

leonapivato (approved):

TsafrirA (approved):
LGTM.

PR #2917, opened 2026-06-12
Fix sampler shot inference with measurement twirling

### LINKED ISSUE #2915: Wrapper Sampler Measurement Twirling Post-Processing Failure
## Description

The executor-backed Wrapper Sampler can fail during result post-processing when measurement twirling is enabled.

The failure appears as:

```text
ValueError: Unable to uniquely identify the shots per PUB.
```

This has been observed for wrapper sampler cases with `twirling.enable_measure=True`, including measurement-only twirling and gate-plus-measurement twirling.

## Legacy Behavior

Legacy Runtime `SamplerV2` returns a valid `PrimitiveResult` for the same workload.

For the observed parameter-sweep case, the legacy result contains a user-facing `BitArray` with:

```text
shape=(8,)
num_shots=128
```

## Wrapper Behavior

The wrapper sampler raises during post-processing before producing a `PrimitiveResult`:

```text
ValueError("Unable to uniquely identify the shots per PUB.")
```

The failure occurs while inferring shots from raw executor result arrays.

## Steps To Reproduce

Run the following with an unfixed executor-backed wrapper sampler install:

```python
from qiskit_ibm_runtime.executor_sampler import SamplerV2 as WrapperSampler

service = QiskitRuntimeService(name="default-ibm-cloud")
backend = service.backend("test_heron_pok_1")

circuit = QuantumCircuit(2, 2)
circuit.h(0)
circuit.cx(0, 1)
circuit.measure([0, 1], [0, 1])

isa_circuit = pm.run(circuit)

sampler = WrapperSampler(mode=backend)
sampler.options.twirling.enable_gates = False
sampler.options.twirling.enable_measure = True
sampler.options.twirling.num_randomizations = 4
sampler.options.twirling.shots_per_randomization = 128
sampler.options.twirling.strategy = "active"

job = sampler.run([isa_circuit], shots=512)
result = job.result()
```

Expected result:

The call to `job.result()` should return a valid `PrimitiveResult`.

Actual result:

```text
ValueError: Unable to uniquely identify the shots per PUB.
```

## Suspected Cause

When measurement twirling is enabled, the raw executor result can contain both the user-facing measured register data and temporary measurement-twirling bookkeeping arrays:

```text
c
measurement_flips.c
```

The measured register can have a shape like:

```text
c.shape == (4, 8, 128, 3)
```

where the axes are:

```text
num_randomizations, pub_shape, shots_per_randomization, num_bits
```

The measurement flip bookkeeping array is constructed by `samplomatic` with a singleton shot axis:

```text
measurement_flips.c.shape == (4, 8, 1, 3)
```

This is expected because the flip mask is sampled per randomized circuit instance and broadcast across all shots for that instance.

The wrapper post-processor currently infers shots by inspecting the second-to-last axis of all raw arrays:

```python
{array.shape[-2] for array in result[0].values()}
```

With measurement twirling enabled, this can produce:

```text
{128, 1}
```

and therefore raises:

```text
ValueError("Unable to uniquely identify the shots per PUB.")
```

## Expected Behavior

The wrapper sampler should ignore `measurement_flips.*` helper arrays when inferring the final sampler output shot count.

Those arrays are temporary bookkeeping used to undo measurement twirling and should not be treated as user-facing output registers.

The post-processor should still validate the helper arrays separately, for example by checking that each `measurement_flips.<register>` array:

- References an existing target register
- Is broadcast-compatible with the target register data
- Has a compatible bit axis

## Suggested Fix

When inferring shots, only inspect user-facing measurement register arrays:

```python
output_arrays = [
    array
    for key, array in result[0].items()
    if not key.startswith(TWIRLING_PREFIX)
]

if not output_arrays:
    raise ValueError("Unable to identify measurement data for shots per PUB.")

if len(set_shots := {array.shape[-2] for array in output_arrays}) != 1:
    raise ValueError("Unable to uniquely identify the shots per PUB.")

shots = next(iter(set_shots))
```

leonapivato:
### Summary

Closes #2915 

### DIFF
--- a/qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
+++ b/qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
@@ -19,7 +19,12 @@ from typing import TYPE_CHECKING, cast
 from qiskit.primitives import PrimitiveResult
 
 from .converters import quantum_program_item_result_to_sampler_pub_result
-from .utils import executor_metadata_to_sampler_metadata, flatten_twirling_axes, undo_twirling
+from .utils import (
+    TWIRLING_PREFIX,
+    executor_metadata_to_sampler_metadata,
+    flatten_twirling_axes,
+    undo_twirling,
+)
 
 if TYPE_CHECKING:
     from ...results.quantum_program import QuantumProgramResult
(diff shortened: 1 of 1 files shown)


TsafrirA (approved):
Nice catch!

PR #2918, opened 2026-06-14
Pea mit prepare

### LINKED ISSUE #2896: Implement PEA prepare function
**Acceptance criteria**
Create `prepare_pea` function to create a QuantumProgram that support PEA mitigation.
Should allow `noise_factors` as part of `ZneOptions`.

dekelmeirom:
### Summary
Add support for the preparing quantum program for pea mitigation method.
Create a new prepare_pea function - force gate twirling and add noise injection annotations to the circuit. Bind the given noise scaling factors to the samplex and pass the scale factors using the passthourgh_data.

### Details and comments

Fixes #2896


dekelmeirom:
I assume in this PR that a user can't give different `noise_factors` to different pubs

TsafrirA:
> I assume in this PR that a user can't give different noise_factors to different pubs

That assumption matches the contract of legacy `EstimatorV2`.

qiskit_ibm_runtime/executor_estimator/pea_utils.py
```diff
@@ -0,0 +1,180 @@
...
+
+    passthrough_data = {
+        "post_processor": {
+            "version": "v0.1",
+            "circuits_metadata": circuits_metadata,
+            "observables": observables_list,
+            "param_basis_pairs": param_basis_pairs_list,
+            "param_shapes": param_shapes_list,
+            "measure_mitigation": "False",
+            "pea_noise_factors": noise_factors,
```
TsafrirA, line 167: So, the only hint here that we are running PEA is the field `pea_noise_factors`?
I wonder if we should add a more explicit flag to indicate what type of EM we're applying.

qiskit_ibm_runtime/executor_estimator/pea_utils.py
```diff
@@ -0,0 +1,180 @@
...
+    circuits_metadata = validate_and_extract_metadata(pubs)
+
+    passthrough_data = {
+        "post_processor": {
+            "version": "v0.1",
+            "circuits_metadata": circuits_metadata,
+            "observables": observables_list,
+            "param_basis_pairs": param_basis_pairs_list,
+            "param_shapes": param_shapes_list,
+            "measure_mitigation": "False",
```
TsafrirA:
```suggestion
            "measure_mitigation": measure_noise_learning is not None,
```

qiskit_ibm_runtime/options_models/zne_options.py
```diff
@@ -0,0 +1,154 @@
...
+    "polynomial_degree_7",
+    "fallback",
+]
+
+FloatLargerThanOne = Annotated[float, Field(ge=1)]
+NonNegativeFloat = Annotated[float, Field(ge=0)]
+
+
+@dataclass(config=PRIMITIVES_CONFIG)
+class ZneOptions:
```
TsafrirA, line 47: This is not used anywhere.
It (or at least a subset of it) should also appear in the passthrough data to be available during post processing?

TsafrirA (commented):
All in all, this looks good!

I left some comments about, mostly about the passthrough data and the options.
As it stands, the option class is not used anywhere. Where do we envision those options being added to the passthrough data? If it's in the prepare function, then that's missing. If it's outside the prepare function (and that would match recent addition to Estimator.run) maybe we can hold off on adding the option class.

qiskit_ibm_runtime/executor_estimator/pea_utils.py
dekelmeirom: As discussed elsewhere - we envision the options being added to the `passthorugh_data` outside the prepare function (and there should probably be a general flag indicating which EM was used)

qiskit_ibm_runtime/options_models/zne_options.py
dekelmeirom: These options will be added outside the prepare function. I will remove the options class from this PR, and it should be added in the relevant PR that will use it.

qiskit_ibm_runtime/executor_estimator/pea_utils.py
```diff
@@ -0,0 +1,180 @@
...
+from .trex_utils import create_trex_calibration_circuit
+
+logger = logging.getLogger(__name__)
+
+
+def prepare_pea(
+    pubs: Sequence[EstimatorPub],
+    twirling_options: TwirlingOptions,
+    shots: int,
+    noise_factors: Sequence[float],
```
SamFerracin: Slight preference for passing ZNE options, and then you can just use the `noise_factors` in those options and ignore everything else:
```suggestion
    zne_options: ZneOptions,
```
Reasons: 1) Consistency with what we do with PEC, and 2) so it's clear where the values used in the prepare funcs (in this case, `noise_factors`) need to be fished from

This is a minor, non-blocking comment though. Also, I realize this may somewhat go against something we talked about last week ("should I pass the extrapolator to the prepare?")

qiskit_ibm_runtime/executor_estimator/pea_utils.py
```diff
@@ -0,0 +1,180 @@
...
+        twirling_options.shots_per_randomization,
+    )
+
+    # validate noise_model_mapping length
+    if noise_model_mapping is None or (
+        noise_model_mapping is not None and len(noise_model_mapping) != len(pubs)
+    ):
+        raise IBMInputValueError(
+            "If PEA mitigation is used, the input must contain noise_model_mapping for each pub"
+        )
```
SamFerracin: I believe there's an error here, and likely in PEC too (sorry, I missed it in the PEC PRs). We actually want `noise_model_mapping` to be a dict, not a sequence of dict (that is, one dict for the whole set of pubs). If two pubs contain the same layer, they'll translate into boxes that contain the same ref, hence they will require the same noise map


```suggestion
    if noise_model_mapping is None:
        raise IBMInputValueError(
            "If PEA mitigation is used, the input must contain ```noise_model_mapping``."
        )
```

lmk if I'm missing something

SamFerracin: my advice here would be to avoid checking if `noise_model_mapping` contains enough maps--we can check it more efficiently later, when we create the samplex items (if the users did not specify all the necessary maps, an error will be raised)

qiskit_ibm_runtime/executor_estimator/pea_utils.py
```diff
@@ -0,0 +1,180 @@
...
+        # make parameters array broadcastable with the noise scales
+        flat_parameter_values = np.expand_dims(flat_parameter_values, 0)
+        samplex_arguments = make_samplex_arguments(
+            samplex, boxed_circuit, flat_parameter_values, change_basis
+        )
+
+        # add samplex_arguments related to noise injection
+
+        # Removing from the noise_factors 1 which is represented by the noise of the data
+        # gates layer. Also, make noise_scales broadcastable with the parameters
```
SamFerracin: I got a bit confused by the word "remove," it sounded like you wanted to remove amplification factors
```suggestion
        # Subtract 1 from the noise_factors 1, since  a value of 1 represents 
        # the noise that is present in the circuit in the absence of amplification.
        #Also, make noise_scales broadcastable with the parameters
```

test/unit/executor_estimator/test_estimator_v2_pea_utils.py
```diff
@@ -0,0 +1,336 @@
...
+        self.assertEqual(item.shape, expected_shape)
+
+        # Check that samplex_arguments contains pauli_lindblad_maps
+        self.assertIn("pauli_lindblad_maps.r2feB", item.samplex_arguments)
+        self.assertEqual(
+            item.samplex_arguments["pauli_lindblad_maps.r2feB"], noise_model_mapping[0]["r2feB"]
+        )
+
+        # Check that samplex_arguments contains noise_scales for the layer
+        self.assertIn("noise_scales.r2feB", item.samplex_arguments)
```
SamFerracin: is there a way that we can avoid placing explicit refs in tests? in this case, I'm talking about the value `"r2feB"`. I don't know what guarantees we have that these values will remain constant with future iterations of samplomatic

qiskit_ibm_runtime/executor_estimator/pea_utils.py
dekelmeirom: I removed this check completely - not even checking that it is not `None`, as the input type hint do not allow `None` as an input

qiskit_ibm_runtime/options_models/zne_options.py
```diff
@@ -0,0 +1,150 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Zero noise extrapolation mitigation options.."""
+
+from __future__ import annotations
+
+from collections.abc import Sequence  # noqa: TC003
```
SamFerracin, line 17: This shouldn't be needed. We import sequence from collections in other files
```suggestion
from collections.abc import Sequence
```

qiskit_ibm_runtime/options_models/zne_options.py
```diff
@@ -0,0 +1,150 @@
...
+            this option is selected, gate twirling will always be used whether or not it has
+            been enabled in the options. In this technique, the twirled noise model of each each
+            unique layer of entangling gates in your ISA circuits is learned beforehand, see
+            :class:`~.LayerNoiseLearningOptions` for relevant learning options. Once complete,
+            your circuits are executed at each noise factor, where every entangling layer of
+            your circuits is amplified by probabilistically injecting single-qubit noise
+            proportional to the corresponding learned noise model.
+    """
+
+    noise_factors: Sequence[FloatLargerThanOne] | Literal["auto"] = "auto"
```
SamFerracin: Consider this, since you only use `FloatLargerThanOne` once (same for the other annotation)
```suggestion
    noise_factors: Sequence[Annotated[float, Field(ge=1)]] | Literal["auto"] = "auto"
```

qiskit_ibm_runtime/options_models/zne_options.py
```diff
@@ -0,0 +1,150 @@
...
+        Any V2 estimator is guaranteed to return data fields called ``evs`` and ``stds`` that
+        report the desired expectation value estimates and errors, respectively.
+        When ZNE options are enabled in the runtime estimator, additional data is returned.
+
+        In particular, suppose an input pub has observable array shape ``obs_shape`` and parameter
+        values shape ``par_shape``, with corresponding pub shape
+        ``shape=np.broadcast_shapes(obs_shape, par_shape)``. Then the corresponding pub result will
+        additionally contain:
+
+        1. `pub_result.data.evs_extrapolated` and `pub_result.data.stds_extrapolated`,
```
SamFerracin:
```suggestion
        1. ``pub_result.data.evs_extrapolated`` and ``pub_result.data.stds_extrapolated``,
```

qiskit_ibm_runtime/options_models/zne_options.py
```diff
@@ -0,0 +1,150 @@
...
+            your circuits are executed at each noise factor, where every entangling layer of
+            your circuits is amplified by probabilistically injecting single-qubit noise
+            proportional to the corresponding learned noise model.
+    """
+
+    noise_factors: Sequence[FloatLargerThanOne] | Literal["auto"] = "auto"
+    """ noise_factors: Noise factors to use for noise amplification.
+
+    The default depends on the amplifier method - the default for pea is (1, 1.5, 2, 2.5, 3)
+    and the default for the other methods is (1, 3, 5).
```
SamFerracin:
```suggestion
    The default depends on the amplifier method - the default for pea is ``(1, 1.5, 2, 2.5, 3)``
    and the default for the other methods is ``(1, 3, 5)``.
```

qiskit_ibm_runtime/executor_estimator/pea_utils.py
```diff
@@ -0,0 +1,194 @@
...
+            reserved classical register name ``_meas``.
+        IBMInputValueError: If noise_model_mapping is missing a noise map for at least one of
+            the pubs layers.
+
+    """
+    if zne_options.amplifier != "pea":
+        raise IBMInputValueError("PEA mitigation must be used with ``pea`` as noise amplification.")
+
+    if zne_options.noise_factors == "auto":
+        noise_factors = (1.0, 1.5, 2.0, 2.5, 3.0)
```
SamFerracin:
```suggestion
        noise_factors = np.array([1.0, 1.5, 2.0, 2.5, 3.0])
```

qiskit_ibm_runtime/executor_estimator/pea_utils.py
```diff
@@ -0,0 +1,194 @@
...
+        pub_noise_model = {}
+        for spec in specs:
+            ref = spec.name.split(".")[-1]
+            if ref not in noise_model_mapping.keys():
+                raise IBMInputValueError(
+                    f"noise_model_mapping is missing noise map for layer reference {ref}"
+                )
+            pub_noise_model[ref] = noise_model_mapping[ref]
+
+        for ref in pub_noise_model:
```
SamFerracin:
```suggestion
```

qiskit_ibm_runtime/options_models/zne_options.py
SamFerracin: Actually, to make sure that these defaults remain in sync, you could:

- Define them on top of the file:
```python
PEA_DEFAULT_NOISE_FACTORS = (1, 1.5, 2, 2.5, 3)
"""The values of ``noise_factors`` used by default when PEA is selected."""
```
- Point to this in the documentation: `the default for pea is :class:`.~PEA_DEFAULT_NOISE_FACTORS`
- Use these defaults in `prepare_pea`

So that in the future we need to change these defaults, we only have to worry about one place

qiskit_ibm_runtime/options_models/zne_options.py
dekelmeirom: The tests fail without it, because it is imported only for TYPE_CHECKING in this flow

SamFerracin: The way we normally import it is as:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from collections.abc import Sequence
```

SamFerracin (approved):
Thanks!

PR #2920, opened 2026-06-15
Move json module from utils to top-level

diego-plan9:
### Summary

Move `qiskit_ibm_runtime.utils.json` to `qiskit_ibm_runtime.json`. `RuntimeEncoder` and `RuntimeDecoder` are a bit central - and having them in `utils` forced every other item using `utils` for other purposes go through it. Moving the package up reflects better his significativeness, and allows to keep `utils` lighter.

Top-level import remains unchanged.

### Details and comments

Related to #2749 (and a bit to #2894, continuing it)

As a small byproduct, we no longer need to exclude `utils/__init__.py` from `isort` rules.

### AI/LLM disclosure


SamFerracin (approved):
Thanks! Looks good!

PR #2921, opened 2026-06-15
Update ruff config for more terse imports

diego-plan9:
### Summary

Disable the [`split-on-trailing-comma`](https://docs.astral.sh/ruff/settings/#lint_isort_split-on-trailing-comma) `isort` option, in order to automatically enforce keeping imports in one line (if they fit). 

This restricts a bit the flexibility (ie. developers cannot keep imports in multiple lines if they fit in one line), but it allows for more consistency and an easier time grepping over the codebase. Plus, arguably we don't exercise that flexibility (there are a good number of imports of just one item that are using multiple lines, likely due to copy-pasting).

### Details and comments

Fixes N/A

There is also https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma, which applies not only to `import` but to the other statement. It felt a bit aggressive to apply (dry-running it, it did caught some instances where being more compact made sense, but there seemed to be enough examples for extra lines included for providing more readability).

### AI/LLM disclosure

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,6 +76,9 @@ source = ["qiskit_ibm_runtime"]
 line-length = 100
 target-version = "py310"
 
+[tool.ruff.lint.isort]
+split-on-trailing-comma = false
+
 [tool.ruff.lint]
 future-annotations = true
 select = [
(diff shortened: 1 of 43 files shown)


TsafrirA (approved):
LGTM

PR #2922, opened 2026-06-15
zne prepare function

### LINKED ISSUE #2849: Wire ZNE gate folding into the executor estimator preprocess flow
- Either update the [prepare](https://github.com/Qiskit/qiskit-ibm-runtime/blob/97ac0fc869250d5eb1cd268e040be3d00dc7ee06/qiskit_ibm_runtime/executor_estimator/prepare.py#L49) function or create a ``prepare_zne`` function, which trasnforms input estimator PUBs into a ``QuantumProgram`` implementing the gate-folded experiments.
- Ensure metadata/passthrough data are correctly accounting for the ZNE experiments and which experiments go with which noise factors

dekelmeirom:
### Summary
Add support for the preparing quantum program for zne with gate folding mitigation method.
Create a new prepare_zne function - add folded gates to the circuit. Add new item to the quantum program for each noise scale factor for each pub, and pass the scale factors using the passthourgh_data.

### Details and comments

Fixes #2849

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/zne/zne_utils.py
+++ b/qiskit_ibm_runtime/executor_estimator/zne/zne_utils.py
@@ -1,0 +1,180 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2026.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Prepare function for Executor-based EstimatorV2 primitive."""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any, Literal
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from qiskit import QuantumCircuit
+    from qiskit.primitives.containers.estimator_pub import EstimatorPub
+
+    from ...options_models.measure_noise_learning_options import MeasureNoiseLearningOptions
+    from ...options_models.twirling_options import TwirlingOptions
+
+from qiskit.transpiler import PassManager
+from samplomatic import build
+
+from ...exceptions import IBMInputValueError
+from ...executor.calculate_twirling_shots import calculate_twirling_shots
+from ...executor.passthrough_utils import validate_and_extract_metadata
+from ...quantum_program import QuantumProgram
+from ...quantum_program.quantum_program import SamplexItem
+from ..prepare import box_circuit, compute_samplex_arguments, make_samplex_arguments
+from ..trex_utils import create_trex_calibration_circuit
+from .gate_folding import GateFolding
+
+logger = logging.getLogger(__name__)
+
+
+def fold_gates(
+    circuit: QuantumCircuit,
+    noise_factor: float,
+    method: Literal["random", "front", "back"] = "random",
+    **kwargs: Any,
+) -> QuantumCircuit:
+    """Test helper: run ``GateFolding`` via a one-pass PassManager."""
+    return PassManager([GateFolding(noise_factor, method, **kwargs)]).run(circuit)
+
+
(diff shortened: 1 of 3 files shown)


dekelmeirom:
Assume that other options details needed for the post processing part, like the chosen extrapolation method will be added outside the `prepare_zne` function, like in pea.

qiskit_ibm_runtime/executor_estimator/zne/zne_utils.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Args:
+        pubs: List of estimator pubs to convert.
+        twirling_options: The twirling options.
+        shots: The number of shots to use. Will be overridden by
+            ``num_randomizations * shots_per_randomization`` when both are specified explicitly
+            and twirling is on.
+        noise_factors: The noise amplification factors.
+        folding_method: The method for folding the gates for noise amplification. Can be one of
+        `front`, `back` or `random`.
```
SamFerracin: To make sure sphinx renders them as proper code:
```suggestion
        ``front``, ``back`` or ``random``.
```

qiskit_ibm_runtime/executor_estimator/zne/zne_utils.py
```diff
@@ -0,0 +1,180 @@
...
+    for i, pub in enumerate(pubs):
+        logger.info("Processing pub %d/%d", i + 1, len(pubs))
+
+        # Prepare samplex_arguments that are common to all noise factors
+        flat_parameter_values, change_basis, param_basis_pairs = compute_samplex_arguments(pub)
+
+        for j, noise_factor in enumerate(noise_factors):
+            logger.info("Processing noise factor %d/%d", j + 1, len(noise_factors))
+
+            folded_circuit = fold_gates(pub.circuit, noise_factor, folding_method)
```
SamFerracin: I would consider simply going with:
```suggestion
            folded_circuit = PassManager([GateFolding(noise_factor, folding_method)]).run(pub.circuit)
```

Alternatively, if you want to keep the function, could you improve the docstring?

qiskit_ibm_runtime/executor_estimator/zne/zne_utils.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # Store data for passthrough
+        observables_list.append(pub.observables.tolist())
+        param_basis_pairs_list.append(param_basis_pairs)
+        param_shapes_list.append(pub.parameter_values.shape)
+
+    # Collect and validate circuit metadata from each pub
+    circuits_metadata = validate_and_extract_metadata(pubs)
+
+    passthrough_data = {
```
SamFerracin, line 163: This function is different from PEC and PEA prepare: while those ones create 1 item per PUB, `prepare_zne` creates multiple items per PUB. Thus, it is important to keep in mind that item0 is pub0 noise_factor0, item1 is pub0 noise_factor1, etc.

I wonder if it we want to be explicit here, create a map from `(pub_idx, noise_factor_idx)` to `item_idx` or vice-versa and add it to the passthrough data? It may make our lives easier in the post-processing function, and also if/when we need to debug

dekelmeirom: each pub will have the same amount of items with repeated noise factors for each pub. but I agree that it can help in post processing, especially if there will be future different features in the future - so I added such mapping from item index to (pub_index, noise_factor)

qiskit_ibm_runtime/executor_estimator/zne/zne_utils.py
```diff
@@ -0,0 +1,197 @@
...
+        "post_processor": {
+            "version": "v0.1",
+            "circuits_metadata": circuits_metadata,
+            "observables": observables_list,
+            "param_basis_pairs": param_basis_pairs_list,
+            "param_shapes": param_shapes_list,
+            "measure_mitigation": measure_noise_learning is not None,
+            "zne_noise_factors": noise_factors,
+            "item_to_pub_and_noise_factor_map": item_to_pub_and_noise_factor_map,
```
SamFerracin: Maybe a shorter `item_id` for item identifier?
```suggestion
            "item_id": item_id,
```

SamFerracin (approved):
Thanks!

PR #2925, opened 2026-06-16
Bump `Samplomatic` to 0.17.0 and `Qiskit` to 2.3.0

### LINKED ISSUE #2924: Revise compatibility with minimal versions of dependencies
**Describe the bug**

In the last integration tests run, the stage that runs the unit tests with minimal versions of the dependencies has failed - likely to an issue with the `samplomatic` version needed by recent commits.

**Steps to reproduce**

Mimic https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/.github/workflows/integration-tests.yml#L90 locally.

**Expected behavior**

All test pass.

**Suggested solutions**

Investigate the dependency that is causing the tests to fail, and check whether the codebase can be changed so it does not require minimal version dependencies to be bumped - or bump the versions required if there is no alternative.

**Additional Information**

- **qiskit-ibm-runtime version**: `main` 
- **Python version**: 3.10+
- **Operating system**: all

TsafrirA:
### Summary
Closes #2924 

### Details and comments
The tests added in #2897 hard code the references of `Samplex`'s `InjectNoise` nodes. Prior to `Samplomatic` version 0.17.0, these refs were not stable across python processes (`Samplomatic` ~~issue 337~~ PR [323](https://github.com/Qiskit/samplomatic/pull/323)). We can fix the tests to build the `Samplex` and query the refs, and it should be identical to the ones created in the `prepare` function (as the two builds run in the same process). However, this will mean that users will be unable to box in a different process than the one they use to run the wrapper Estimator. In the context of noise learning, such restriction is not acceptable. Therefore, we are forced to bump `Samplomatic` to 0.17.0, and `Qiskit` to 2.3.0 because `Samplomatic` depends on it.

### AI/LLM disclosure

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -164,9 +164,9 @@ dependencies = [
     "ibm-platform-services>=0.61.1",
     "ibm-quantum-schemas>=0.8.20260605,<1",
     "pydantic>=2.11.0",
-    "qiskit>=2.2.0",
+    "qiskit>=2.3.0",
     "pybase64>=1.4.0",
-    "samplomatic>=0.13.0"
+    "samplomatic>=0.17.0"
 ]
 
 [project.entry-points."qiskit.transpiler.translation"]


dekelmeirom:
Fix for the tests #2926 to use a flow similar to the one expected by the user (use of `get_layers` to extract the layers refs). In the tests all is run in the same process, so even though the tests are passing with Samplomatic version older than 0.17, there might still be an issue for the user, as mentioned above

diego-plan9:
Can you add a release note (a good inspiration for the note might be #2689) for the qiskit version bump?

SamFerracin (approved):

PR #2926, opened 2026-06-16
fix noise model mapping in pec

dekelmeirom:
### Summary
Change `noise_model_mapping` in `prepare_pec` from a list of dicts to a single dict containing the mapping for all pubs.

In addition, remove the hardcoded noise model refs from the tests, and instead use `get_layers` and take the refs from the layers. 

### Details and comments

related to comments raised in #2918

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/pec_utils.py
+++ b/qiskit_ibm_runtime/executor_estimator/pec_utils.py
@@ -147,7 +147,7 @@ def prepare_pec(
         IBMInputValueError: If pubs have mismatched precision,
             if a circuit contains mid-circuit measurements, or if a circuit already uses the
             reserved classical register name ``_meas``.
-        IBMInputValueError: If noise_model_mapping is missing a noise map for at least one of
+        IBMInputValueError: If ``noise_model_mapping`` is missing a noise map for at least one of
             the pubs layers.
     """
     num_randomizations, shots_per_randomization = calculate_pec_twirling_shots(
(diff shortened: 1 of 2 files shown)


TsafrirA:
Could you check to see if the tests pass with samplomatic 0.16.0? I think they should.

qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -147,21 +147,15 @@ def prepare_pec(
         IBMInputValueError: If pubs have mismatched precision,
             if a circuit contains mid-circuit measurements, or if a circuit already uses the
             reserved classical register name ``_meas``.
-        IBMInputValueError: If the length of noise_model_mapping and the length of the pubs
-            mismatched.
+        IBMInputValueError: If noise_model_mapping is missing a noise map for at least one of
```
SamFerracin: This is to avoid italic:
```suggestion
        IBMInputValueError: If ``noise_model_mapping`` is missing a noise map for at least one of
```

qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -213,10 +207,22 @@ def prepare_pec(
...
+        # Create a noise model map containing only the layers relevant for the current pub
+        specs = samplex.inputs().get_specs("pauli_lindblad_maps")
+        pub_noise_model = {}
+        for spec in specs:
+            ref = spec.name.split(".")[-1]
+            if ref not in noise_model_mapping.keys():
+                raise IBMInputValueError(
+                    f"noise_model_mapping is missing noise map for layer reference {ref}"
+                )
+            pub_noise_model[ref] = noise_model_mapping[ref]
```
SamFerracin:
```suggestion
            try:
                pub_noise_model[ref] = noise_model_mapping[ref]
            except KeyError:
                raise IBMInputValueError(
                    f"noise_model_mapping is missing noise map for layer reference {ref}"
                )
```

qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -213,10 +207,22 @@ def prepare_pec(
...
+
+        for ref in pub_noise_model:
```
SamFerracin: Can this extra for loop be avoided and compounded into the previous one?
```suggestion
```

test/unit/executor_estimator/test_estimator_v2_pec_utils.py
```diff
@@ -211,7 +213,14 @@ def test_prepare_pec_basic(self):
         noise_model = PauliLindbladMap.from_sparse_list(
             [("XX", [0, 1], 0.1), ("ZZ", [0, 1], 0.05)], num_qubits=2
         )
-        noise_model_mapping = [{"r2feB": noise_model}]
```
SamFerracin, line 214: thanks!

dekelmeirom:
yes, all tests pass with samplomatic 0.16.
but I think that some users might have problems running it with samplomatic 0.16 if they use different python processes for noise learning and execution

SamFerracin (commented):
Thanks! Some minor comments

SamFerracin (approved):
Thanks!

PR #2771, opened 2026-04-29
Add extra documentation on active instance

diego-plan9:
### Summary

Currently, there are some code paths where `QiskitIBMRuntime._active_api_client` is replaced, and they can be non-obvious to users. This seems to be part of the need for `QiskitIBMRuntime` to transparently handle multiple instances under the hood - and can ripple into unexpected side-effects.

This PR aims to improve the documentation on using multiple instances, allowing users to be more aware of this behavior.

### Details and comments

Fixes #724

PR #2931, opened 2026-06-18
Use backend's api client during QiskitRuntimeService._run

### LINKED ISSUE #2892: Sampler.run not using the right instance
**Describe the bug**

If I do 
```python
backend = service.backend("ibm_kingston")
sampler = SamplerV2(backend)

mock = service.backend("mock_kingston")
mock_sampler = SamplerV2(mock)

mock_job = mock_sampler.run([isa_circuit])
job = sampler.run([isa_circuit])
```

The second job submission fails with 
```
IBMRuntimeError: 'Failed to run program: \'403 Client Error: Forbidden for url: https://quantum.cloud.ibm.com/api/v1/jobs. {"errors":[{"code":1215,"message":"You are not authorized to run program on backend ibm_kingston or backend does not exist on this channel."
```

Even though `sampler.backend()._instance` shows it's using the right CRN. 

**Steps to reproduce**

See above.

**Expected behavior**

The job submission should use the instance picked for the original backend initialization. 

**Suggested solutions**

**Additional Information**

- **qiskit-ibm-runtime version**: 0.47.0
- **Python version**: 3.11
- **Operating system**:

diego-plan9:
### Summary

Revise `QiskitRuntimeService._run` so, instead of using the "active" API client (which can be modified by other calls - see #2771), it uses the API client that corresponds to the backend's instance.

### Details and comments

Fixes #2892

### AI/LLM disclosure


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1013,6 +1013,18 @@ def _run(
         if isinstance(backend, str):
             backend = self.backend(name=qrt_options.get_backend_name())
 
+        # Set the active client to match the backend.
+        try:
+            self._active_api_client = self._api_clients[backend._instance]
+        except KeyError:
```
diego-plan9, line 1034: An alternative is being lenient, and call `_get_or_create_cloud_client()`. The logic in the service is a bit convoluted, but I think it is fair to assume that a user would be using backends created from the same service that is being used for running.

I'm going with an exception as otherwise it would allow for backends obtained from a different `QiskitIBMRuntime` to be used (and their api clients to be added dynamically), which seems a questionable use case.

test/unit/test_qiskit_runtime_service.py
```diff
@@ -0,0 +1,86 @@
...
+from ..utils import transpile_pubs
+from .mock.fake_api_backend import FakeApiBackendSpecs
+from .mock.fake_runtime_client import BaseFakeRuntimeClient
+
+
+class TestQiskitRuntimeService(IBMTestCase):
+    """Class for testing the `QiskitRuntimeService` class."""
+
+    @run_cloud_fake
+    def test_run_active_client(self, service):
```
diego-plan9, line 33: The test goes quite low-level, mimicking how `service._api_clients` are handled (and manipulating it and the api clients directly). Part of it could be avoided by revising `FakeRuntimeService`, as at some point in time it allowed multiple `crns` to be defined - however, refactoring `FakeRuntimeService` is quite time consuming (and it seems part of its features are broken / outdated).

diego-plan9:
@jyu00  can you check your scenario against this branch?

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1000,7 +1000,7 @@ def _run(
             RuntimeProgramNotFound: If the program cannot be found.
             IBMRuntimeError: An error occurred running the program.
         """
-        self._check_instance_usage()
+        # Get the backend to use.
```
SamFerracin: Remove this? Since this is not the block where you get the backend
```suggestion
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1013,6 +1013,18 @@ def _run(
...
+        try:
+            self._active_api_client = self._api_clients[backend._instance]
+        except KeyError:
+            raise IBMRuntimeError(
+                f"The backend crn ({backend._instance}) is not among the instances supported by "
+                "this QiskitRuntimeService object. Please ensure the backend object was retrieved "
+                "from this object."
+            )
+
+        self._check_instance_usage()
```
SamFerracin, line 1041: Let me get this straight: You are moving `self._check_instance_usage()` down here because it queries `self._active_api_client`, which you set in your try/except. Is that right?

test/unit/test_qiskit_runtime_service.py
SamFerracin: yeah it's quite ugly, but i wouldn't know how to write it better..

SamFerracin (commented):
It looks good to me! But I'll let Jessie take a look too

qiskit_ibm_runtime/qiskit_runtime_service.py
diego-plan9: Correct (and for setting `self._active_client`, we need to have a backend so it can be inspected for its crn and client, hence the moving).

SamFerracin (approved):

PR #2934, opened 2026-06-19
Reduce unit test time by adjusting some tests 

diego-plan9:
### Summary

Reduce the total unit test time (by around ~30s in CI), focusing on the [currently slowest tests ](https://github.com/Qiskit/qiskit-ibm-runtime/actions/runs/27789798458/job/82236016042#step:5:266), by:

* minimizing usage of instantiating `NoiseModel`, as it is expensive
* minimize usage of `min_num_qubits` filter for local backends, as it is expensive
* avoid some seemingly unneeded `sleep`s
* moving some slow, rarely updated tests to the integration tests suite

### Details and comments

Fixes N/A

### AI/LLM disclosure


test/decorators.py
```diff
@@ -49,14 +49,12 @@ def run_cloud_fake(func):
...
-        cloud_service = FakeRuntimeService(
+        kwargs["service"] = FakeRuntimeService(
             channel="ibm_cloud",
             token="my_token",
             instance="crn:v1:bluemix:public:quantum-computing:my-region:a/...:...::",
         )
-        with self.subTest(service=cloud_service.channel):
-            kwargs["service"] = cloud_service
-            func(self, *args, **kwargs)
+        func(self, *args, **kwargs)
```
diego-plan9, line 57: Changes in this file are do not affect the total time, but simplify a bit.

test/integration/test_fake_backends.py
```diff
@@ -80,6 +92,63 @@ def test_circuit_on_fake_backend_v2_with_optimization_level(self, optimization_l
...
+        job = sampler.run([isa_circuit])
+        result = job.result()
+
+        self.assertTrue(job.done())
+        self.assertIsNotNone(result)
+
+    @combine(
+        opt_cls=[EstimatorOptions, SamplerOptions], fake_backend=[FakeManilaV2(), FakeNairobiV2()]
+    )
+    def test_simulator_set_backend(self, opt_cls, fake_backend):
```
diego-plan9, line 128: Stretching the definition of `test_fake_backends.py` a bit. These tests account for ~10s, and arguably they involve fake backends.

test/integration/test_fake_backends.py
```diff
@@ -80,6 +92,63 @@ def test_circuit_on_fake_backend_v2_with_optimization_level(self, optimization_l
...
+
+        # Assert backend property shapes are correct
+        self.assertEqual(len(backend.properties().qubits), backend.num_qubits)
+
+        # Initialize quantum circuit
+        qc = QuantumCircuit(2, 2)
+        qc.h(0)
+        qc.cx(0, 1)
+        qc.measure(0, 0)
+        qc.measure(1, 1)
```
SamFerracin: Pro tip:
```suggestion
        qc.measure_all()
```

test/integration/test_fake_backends.py
SamFerracin: How many qubits do these two backends have?

Things to consider to reduce test time further:
- do we need two fake backends in this test?
- if not, can we stick to the one with the smallest number of qubits?
- can we in fact use FakeVigo, which has 5 qubits?

test/unit/test_jobs.py
```diff
@@ -139,7 +137,6 @@ def test_run_program_failed_ran_too_long(self, service):
     def test_cancel_job(self, service):
         """Test canceling a job."""
         job = run_program(service, job_classes=CancelableRuntimeJob)
-        time.sleep(1)
```
SamFerracin, line 142: lol, good catch!

SamFerracin (commented):
Looks good! Some minor suggestions

test/integration/test_fake_backends.py
diego-plan9: As the test is now part of the integration tests (where the time is dominated by the actual integration test, instead of the unit tests that are run in a parallel stage), maybe there are questions for the future? They are 5 and 7 qubit devices anyway, and the cost is mostly from `NoiseModel.from_backend()`.

SamFerracin (approved):

PR #2938, opened 2026-06-19
Bugfix avg kerneled indexing

### LINKED ISSUE #2928: `avg_kerneled` crashes the executor Sampler post-processor
## Situation

The executor-wrapper `SamplerV2` raises `IndexError: tuple index out of range`
during client-side result decoding whenever `execution.meas_type="avg_kerneled"`.
The legacy `SamplerV2` handles the same request and returns a `PrimitiveResult`,
so this is a **wrapper-only parity break**, not a backend/service issue.

## What actually happened

`avg_kerneled` averages over shots **server-side**, so each register array comes
back with shape `(*pub_shape, num_bits)` — there is **no `num_shots` axis**. The
other two modes keep it:

| `meas_type`    | returned array shape                    |
| -------------- | --------------------------------------- |
| `classified`   | `(*pub_shape, num_shots, num_bits)`     |
| `kerneled`     | `(*pub_shape, num_shots, num_bits)`     |
| `avg_kerneled` | `(*pub_shape, num_bits)`  ← one axis fewer |

The v0.1 post-processor hard-codes the two-trailing-axes layout. The hard crash is
the integer index at
`qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py:83`:

```python
# Compute the shots from the second-to-last axis of the result arrays
if len(set_shots := {array.shape[-2] for array in meas_arrays}) != 1:
```

For a non-parametric pub (e.g. a Bell circuit, `pub_shape == ()`), the
`avg_kerneled` array is 1-D `(num_bits,)`, so `shape[-2]` indexes past the end →
`IndexError: tuple index out of range`.

The same `(num_shots, num_bits)` assumption also corrupts two downstream
computations even if the crash were avoided:

- `post_processor_v0_1.py:88` — `pub_shapes = [...shape[... : -2] ...]`
- `decoders/executor_sampler/converters.py:62` — `pub_shape = first_meas_data.shape[:-2]`
  (this one feeds the `DataBin` shape).

Note the converter even has an `avg_kerneled` branch
(`converters.py:70-71`), but execution dies at the shots inference (line 83)
before it is ever reached — so `avg_kerneled` is effectively unsupported
end-to-end.

`meas_type` is otherwise plumbed correctly: it is set on the `QuantumProgram` and
in the passthrough at `executor_sampler/sampler.py:300,310`.

## Minimal reproducible example

```python
qc = QuantumCircuit(2, 2)
qc.h(0)
qc.cx(0, 1)
qc.measure([0, 1], [0, 1])
isa = generate_preset_pass_manager(backend=backend, optimization_level=1).run(qc)

sampler = WrapperSampler(mode=backend)
sampler.options.execution.meas_type = "avg_kerneled"

job = sampler.run([isa], shots=512)
print(job.result())   # -> IndexError: tuple index out of range
```

**Expected:** a `PrimitiveResult` whose `DataBin` holds the averaged complex IQ
array of shape `(creg_size,)`, matching legacy `SamplerV2` for the same options.
**Actual:** `IndexError: tuple index out of range` from `post_processor_v0_1.py:83`.

leonapivato:
### Summary
Fixes #2928  Wrapper Sampler crash with `meas_type="avg_kerneled"`. The problem was that the `avg_kerneled` data comes back as `(*pub_shape, num_bits)` - with no shots axis, unlike the other types which are `(*pub_shape, num_shots, num_bits)`. This was not accounted for during the post-processing so the Wrapper Sampler raised `IndexError: tuple index out of range` when `avg_kerneled` was used.

### Details and comments
Since `num_shtos` is no longer available when `meas_type="avg_kerneled"` during post processing, but is necessary for generating spans in the metadata, shots have been added to the passthrough data. Kerneled measuring types are incompatible with twirling so it is assumed the user specified shots is consistent with the actual shots executed. The `num_shots` in the passthrough data is only necessary when `meas_type="avg_kerneled"`.

### AI/LLM disclosure


qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
```diff
@@ -63,29 +63,41 @@ def sampler_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveRes
...
+        shots = next(iter(set_shots))
+        shape_end = -2
+    else:
+        shape_end = -1
 
-    # Compute the shape of the input PUBs
-    pub_shapes = [next(iter(item.values())).shape[1 if twirling else 0 : -2] for item in result]
+    # PUB shape = array shape minus the leading randomization axis and the trailing intrinsic axes.
+    pub_shapes: list[tuple[int, ...]] = [
+        next(iter(item.values())).shape[shape_start:shape_end] for item in result
```
SamFerracin, line 76: does this work when a sampler pub has `shape=()`--no parameters?
it should, just checking

test/unit/decoders/executor_sampler/test_post_processor.py
```diff
@@ -51,13 +51,18 @@ def test_single_pub_multiple_registers(self):
             0, 2, size=(num_rands, num_shots_per_rand, 3), dtype=np.uint8
         )
 
+        # Treat the leading axis as a (num_rands,)-long PUB sweep; (num_shots, num_bits) trail it.
+        pub_shape = (num_rands,)
```
SamFerracin: i'm confused by this. `pub_shape` has nothing to do with `num_rands`, why is it equal here?

leonapivato: I believe it has to be consistent with 

```
meas_data_c1 = np.random.randint(
    0, 2, size=(num_rands, num_shots_per_rand, 2), dtype=np.uint8
)
meas_data_c2 = np.random.randint(
    0, 2, size=(num_rands, num_shots_per_rand, 3), dtype=np.uint8
)
```

not sure why it was put like this previously though. Happy to adjust it if needed.

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
leonapivato: I believe so. If it was working before I don't think this would change anything.

leonapivato: Also slices can't produce out of bounds errors.

qiskit_ibm_runtime/decoders/executor_sampler/converters.py
```diff
@@ -49,18 +50,14 @@ def quantum_program_item_result_to_sampler_pub_result(
...
         meas_type: The measurement type.
         circuit_metadata: The metadata attached to the circuit in the input PUB.
 
     Returns:
         A sampler pub result.
     """
-    # Infer pub_shape from the first classical register's data
-    # meas_data shape: (...pub_shape..., num_shots, num_bits)
-    first_meas_data = next(iter(item.values()))
-    pub_shape = first_meas_data.shape[:-2]
```
TsafrirA, line 62: There is an inconsistency here - This definition will return not just the pub shape, but also the leading randomization axis.

The new definition returns only the original pub shape.

TsafrirA (commented):
There seems to be an inconsistency with how the `pub_shape` parameter is defined.

Edit - I also wonder if we should add the `pub_shape` to the passthrough data, seeing the issues we're having with it.

TsafrirA (changes requested):
Post processing fails if you set either gate twirling or measurement twirling along with avg kerneled.

This combination of options is currently not supported with the legacy Sampler. With the wrapper sampler, we can't run these jobs and then fail on post processing (as we do now). Measurement twirling doesn't make sense with averaging (you average before correcting the bit flips), but I think gate twirling is not a-logical (though I don't know if it has a real use case).

(With gate twirling on the flattening fails as the shapes are assumed to have the shots axis).

qiskit_ibm_runtime/decoders/executor_sampler/utils.py
```diff
@@ -188,6 +192,9 @@ def flatten_twirling_axes(item: QuantumProgramItemResult, pub_shape: tuple[int,
             ``(3,)`` for a 1-D parameter sweep.
     """
     for creg_name, data in list(item.items()):
+        if data.ndim == len(pub_shape) + 2:
```
TsafrirA, line 196: I would add a comment explaining what case this if block covers.

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
```diff
@@ -63,29 +63,41 @@ def sampler_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveRes
...
+        num_randomizations = 0
+        shape_start = 0
+
+    # Classified/kerneled arrays end in (num_shots, num_bits): strip 2 and read shots from the
+    # second-to-last axis (PUB shots untwirled, shots_per_randomization twirled). avg_kerneled is
+    # averaged over shots, ending in (num_bits,): strip 1; its shots came from the request above.
+    if meas_type != "avg_kerneled":
+        if len(set_shots := {array.shape[-2] for array in meas_arrays}) != 1:
+            raise ValueError("Unable to uniquely identify the shots per PUB.")
+        shots = next(iter(set_shots))
```
TsafrirA: Don't you get shots from pass through data?

TsafrirA (commented):
Thanks for the quick fix. This should resolve the issue.
I think there is no test covering the error raising for the measurement type and twirling combination. There's also some inconsistency about the shots calculation in the post processing.

I am leaving it to @SamFerracin to approve, as he had some reservations about the code. I think a lot of the complexity could be avoided by just passing the pub shape in the passthrough data.

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
leonapivato: yes I do. Which is nice because now I can remove the array filtering also!

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
```diff
@@ -72,20 +69,12 @@ def sampler_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveRes
...
 
-    # Filter out measurement-twirling bookkeeping arrays; they may use a
-    # singleton shot axis for broadcasting and are not user-facing result data.
-    meas_arrays = [array for key, array in result[0].items() if not key.startswith(TWIRLING_PREFIX)]
-    if not meas_arrays:
-        raise ValueError("Unable to identify measurement data for shots per PUB.")
-
-    # Compute the shots from the second-to-last axis of the result arrays; this corresponds to
-    # PUB shots if twirling is OFF, and to ``shots_per_randomization`` if twirling is ON.
-    if len(set_shots := {array.shape[-2] for array in meas_arrays}) != 1:
```
leonapivato, line 83: shots are taken from passthrough instead. Filtered `meas_arrays` are no longer needed then.

leonapivato: I previously added `meas_arrays` for this [bug](https://github.com/Qiskit/qiskit-ibm-runtime/issues/2915)

qiskit_ibm_runtime/executor_sampler/utils.py
```diff
@@ -45,6 +45,27 @@ def validate_no_boxes(circuit: QuantumCircuit) -> None:
             )
 
 
+def validate_meas_type_twirling(meas_type: str | None, enable_measure: bool | None) -> None:
```
leonapivato, line 48: This is to fail measurement twirling and kerneled/avg_kerneled measurement type before job submission rather than letting it fail during post processing. 

This configuration is a also not supported with the legacy implementation, which has the server fail the job after submission.

qiskit_ibm_runtime/decoders/executor_sampler/utils.py
```diff
@@ -188,6 +192,10 @@ def flatten_twirling_axes(item: QuantumProgramItemResult, pub_shape: tuple[int,
             ``(3,)`` for a 1-D parameter sweep.
     """
     for creg_name, data in list(item.items()):
+        # (ndim == len(pub_shape) + 2) is true only for avg_kerneled measurements
```
leonapivato, line 195: This change is to allow for gate twirling with `avg_kerneled` measurement types. It can no longer rely on shots in the result to flatten the axis and instead must average it out. 

This is consistent with the legacy implementation.

leonapivato:
Splitting into 3 different PRs for the 3 bugs

PR #2940, opened 2026-06-22
Add local mode to wrapper sampler

### LINKED ISSUE #2683: Add wrapper sampler support for local mode

TsafrirA:
### Summary
Closes #2683

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_sampler/sampler.py
+++ b/qiskit_ibm_runtime/executor_sampler/sampler.py
@@ -19,7 +19,6 @@ import logging
 from dataclasses import asdict
 from typing import TYPE_CHECKING
 
-from qiskit.primitives import BackendSamplerV2
 from qiskit.primitives.base import BaseSamplerV2
 from qiskit.primitives.containers.sampler_pub import SamplerPub
 from samplomatic import build
@@ -29,9 +28,7 @@ from ..base_primitive import get_mode_service_backend
 from ..executor import Executor
 from ..executor.calculate_twirling_shots import calculate_twirling_shots
 from ..executor.dynamical_decoupling import apply_dynamical_decoupling
-from ..fake_provider.local_runtime_job import LocalRuntimeJob
 from ..fake_provider.local_service import QiskitRuntimeLocalService
-from ..ibm_backend import IBMBackend
 from ..options_models.sampler_options import SamplerOptions
 from ..quantum_program import QuantumProgram
 from ..quantum_program.quantum_program import CircuitItem, SamplexItem
@@ -45,10 +42,12 @@ if TYPE_CHECKING:
     from qiskit.providers import BackendV2
 
     from ..batch import Batch
+    from ..fake_provider.local_runtime_job import LocalRuntimeJob
     from ..options_models.executor_options import ExecutorOptions
     from ..quantum_program import QuantumProgramItem
     from ..runtime_job_v2 import RuntimeJobV2
     from ..session import Session
+
 
 logger = logging.getLogger(__name__)
 
@@ -349,62 +348,26 @@ class SamplerV2(BaseSamplerV2):
         Raises:
             ValueError: If using IBMBackend in local mode.
         """
-        if isinstance(self._backend, IBMBackend):
-            raise ValueError(
-                "Local testing mode is not supported when a cloud-based backend is used."
-            )
+        # Prepare options dict - this goes in the inputs["options"] field
+        options_dict = copy.deepcopy(asdict(self.options))  # type: ignore[call-overload]
 
-        # Prepare options for BackendSamplerV2
-        options_copy = copy.deepcopy(asdict(self.options))  # type: ignore[call-overload]
-
-        prim_options = {}
-        sim_options = options_copy.get("simulator", {})
-
-        # Extract seed_simulator if present
-        if seed_simulator := sim_options.pop("seed_simulator", None):
-            prim_options["seed_simulator"] = seed_simulator
-
-        # Check if BackendSamplerV2 supports run_options
-        dummy_prim = BackendSamplerV2(backend=self._backend)
-        use_run_options = hasattr(dummy_prim.options, "run_options")
-
-        run_options = {}
-        if use_run_options:
-            # Add noise model if present
-            if "noise_model" in sim_options:
-                run_options["noise_model"] = sim_options.pop("noise_model")
-
-            # Map meas_type to meas_level
-            if meas_type := options_copy.get("execution", {}).pop("meas_type", None):
-                if meas_type == "classified":
-                    run_options["meas_level"] = 2
-                elif meas_type == "kerneled":
-                    run_options["meas_level"] = 1
-                    run_options["meas_return"] = "single"
-                elif meas_type == "avg_kerneled":
-                    run_options["meas_level"] = 1
-                    run_options["meas_return"] = "avg"
-
-        # Set default_shots
+        # Add default_shots to options if provided
         if default_shots is not None:
-            prim_options["default_shots"] = default_shots
+            options_dict["default_shots"] = default_shots
 
-        if run_options:
-            prim_options["run_options"] = run_options
+        # Prepare inputs dict with pubs and options
+        inputs = {
+            "pubs": pubs,
+            "options": options_dict,
+        }
 
-        # Create BackendSamplerV2 instance
-        primitive_inst = BackendSamplerV2(backend=self._backend, options=prim_options)
+        # Prepare runtime options with backend
+        runtime_options = {"backend": self._backend}
 
-        # Run the primitive
-        primitive_job = primitive_inst.run(pubs)
-
-        # Wrap in LocalRuntimeJob
-        local_runtime_job = LocalRuntimeJob(
-            function=primitive_job._function,
-            future=primitive_job._future,
-            backend=self._backend,
-            primitive="sampler",
-            inputs={"pubs": pubs},
+        # Call the centralized _run method from local_service
+        return self._service._run(
+            program_id="sampler",
+            inputs=inputs,
+            options=runtime_options,
+            calibration_id=None,
         )
-
-        return local_runtime_job
--- a/qiskit_ibm_runtime/executor_sampler/sampler.py
+++ b/qiskit_ibm_runtime/executor_sampler/sampler.py
@@ -14,7 +14,6 @@
 
 from __future__ import annotations
 
-import copy
 import logging
 from dataclasses import asdict
 from typing import TYPE_CHECKING
@@ -353,7 +352,7 @@ class SamplerV2(BaseSamplerV2):
             A LocalRuntimeJob.
         """
         # Prepare options dict - this goes in the inputs["options"] field
-        options_dict = copy.deepcopy(asdict(self.options))  # type: ignore[call-overload]
+        options_dict = asdict(self.options)  # type: ignore[call-overload]
         options_dict["default_shots"] = shots
 
         # Prepare inputs dict with pubs and options
--- a/qiskit_ibm_runtime/executor_sampler/sampler.py
+++ b/qiskit_ibm_runtime/executor_sampler/sampler.py
@@ -14,17 +14,24 @@
 
 from __future__ import annotations
 
+import copy
 import logging
+from dataclasses import asdict
 from typing import TYPE_CHECKING
 
+from qiskit.primitives import BackendSamplerV2
 from qiskit.primitives.base import BaseSamplerV2
 from qiskit.primitives.containers.sampler_pub import SamplerPub
 from samplomatic import build
 from samplomatic.transpiler import generate_boxing_pass_manager
 
+from ..base_primitive import get_mode_service_backend
 from ..executor import Executor
 from ..executor.calculate_twirling_shots import calculate_twirling_shots
 from ..executor.dynamical_decoupling import apply_dynamical_decoupling
+from ..fake_provider.local_runtime_job import LocalRuntimeJob
+from ..fake_provider.local_service import QiskitRuntimeLocalService
+from ..ibm_backend import IBMBackend
 from ..options_models.sampler_options import SamplerOptions
 from ..quantum_program import QuantumProgram
 from ..quantum_program.quantum_program import CircuitItem, SamplexItem
@@ -107,7 +114,14 @@ class SamplerV2(BaseSamplerV2):
     ):
         super().__init__()
 
-        self._executor = Executor(mode=mode)
+        # Store mode, service, and backend for simulator detection
+        self._mode, self._service, self._backend = get_mode_service_backend(mode)
+
+        # Only create executor for non-local backends
+        # For local simulators (QiskitRuntimeLocalService), we'll use BackendSamplerV2 directly
+        self._executor = None
+        if not isinstance(self._service, QiskitRuntimeLocalService):
+            self._executor = Executor(mode=mode)
 
         # Coerced to `SamplerOptions` via `__setattr__()`.
         self.options = options if options is not None else SamplerOptions()  # type: ignore[assignment]
@@ -147,12 +161,6 @@ class SamplerV2(BaseSamplerV2):
 
         Returns:
             The submitted job.
-
-        Raises:
-            ValueError: If backend is not provided.
-            IBMInputValueError: If circuits contain :class:`~qiskit.circuit.BoxOp` instructions or
-                if shots are not properly specified.
-            NotImplementedError: If unsupported options are enabled.
         """
         # Coerce pubs to SamplerPub objects
         coerced_pubs = [SamplerPub.coerce(pub, shots) for pub in pubs]
@@ -160,6 +168,12 @@ class SamplerV2(BaseSamplerV2):
         # Determine default shots: run parameter takes precedence over options.default_shots
         default_shots = shots if shots is not None else self.options.default_shots
 
+        # Check if we're in local simulator mode
+        if self._executor is None:
+            logger.info("Running in local simulator mode")
+            return self._run_simulator(coerced_pubs, default_shots)
+
+        # Non-simulator path: use executor
         # Convert pubs to QuantumProgram and map options using the prepare method
         logger.info("Starting pre-processing")
         quantum_program, executor_options = self.prepare(coerced_pubs, default_shots)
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -107,7 +114,14 @@ def __init__(
     ):
         super().__init__()
 
-        self._executor = Executor(mode=mode)
+        # Store mode, service, and backend for simulator detection
+        self._mode, self._service, self._backend = get_mode_service_backend(mode)
```
TsafrirA, line 116: We call `get_mode_service_backend` twice. Once here and once in the `Executor`. We can remove this call once we simulate via `AerExecutor` and run through `prepare()`.

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,130 @@
...
+    pass
+
+
+# Custom config for SimulatorOptions to allow CouplingMap type
+_SIMULATOR_CONFIG = ConfigDict(
+    validate_assignment=True, extra="forbid", arbitrary_types_allowed=True
+)
+
+
+@dataclass(config=_SIMULATOR_CONFIG)
```
TsafrirA: @diego-plan9 
I have some issues with the Qiskit objects here. Had to make changes to the config. What do you think?

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,130 @@
...
+
+    basis_gates: list[str] | None = None
+    """List of basis gate names to unroll to.
+
+    For example, ``['u1', 'u2', 'u3', 'cx']``. Unrolling is not done if not set.
+
+    Default: all basis gates supported by the simulator.
+    """
+
+    @field_validator("coupling_map", mode="before")
```
TsafrirA: @diego-plan9 
What do you think of the validators?

test/unit/executor_sampler/test_simulator_options.py
```diff
@@ -0,0 +1,81 @@
+# This code is part of Qiskit.
```
TsafrirA, line 1: We currently only use these options in the sampler, and hence the file is located where it is. However, eventually it might need to move.

qiskit_ibm_runtime/options_models/sampler_options.py
```diff
@@ -66,6 +73,8 @@ class SamplerOptions:
     def to_executor_options(self) -> ExecutorOptions:
         """Map sampler options to executor options, ignoring all irrelevant fields.
 
+        Note: Simulator options are ignored as executor does not support local mode.
```
diego-plan9: Small nit: consider using `.. note:` for consistency with how we signal comments such as this.

qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -319,3 +333,78 @@ def prepare(
         executor_options = options.to_executor_options()
 
         return quantum_program, executor_options
+
+    def _run_simulator(
+        self, pubs: Sequence[SamplerPub], default_shots: int | None
+    ) -> LocalRuntimeJob:
+        """Run sampler in local simulator mode using BackendSamplerV2.
```
diego-plan9: Did you explore directly invoking `QiskitRuntimeLocalService._run_backend_primitive_v2()` (or `._run()` maybe) - and if so, can you comment on going with a dedicated method?

Mostly for consistency and responsibilities. Primitives and program end up delegating in `SomeService._run()`, and it would be good to keep it that way as it separates responsibilities better and is more inline with the rest of the programs).

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,130 @@
...
+
+            from qiskit_aer.noise import NoiseModel as AerNoiseModel
+
+            if not isinstance(model, AerNoiseModel):
+                raise ValueError(
+                    "'noise_model' can only be a dictionary or qiskit_aer.noise.NoiseModel."
+                )
+        return model
+
+    def set_backend(self, backend: BackendV2) -> None:
```
diego-plan9, line 108: Is this method needed?

qiskit_ibm_runtime/options_models/simulator_options.py
diego-plan9: Can you elaborate on the issues? I suspect `NoiseModel` and `CouplingMap` (at least) causing some serialization headaches?

Ideally we would like all new-style options to be fully and easily serializable (and `arbitrary_types_allowed=True` hinders that, even if it is in a sub-model, as it affects the parent model). The approach with the validators and the dummy class could be worth exploring further, depending on the issues encountered (as in, including an artificial conversion to guarantee keeping the options serializable).

qiskit_ibm_runtime/executor_sampler/sampler.py
TsafrirA: That's a good point. I had tunnel-vision trying to work directly with `BackendSamplerV2`, because in my mind it was similar to the pattern we will have with `AerExecutor`.

qiskit_ibm_runtime/options_models/simulator_options.py
TsafrirA: Pydantic doesn't seem to allow the Qiskit types without `arbitrary_types_allowed=True`. The simulator options are interesting because we don't need to serialize them - they don't make it to the server.

qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -147,19 +160,19 @@ def run(self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None) -> Ru
 
         Returns:
             The submitted job.
-
-        Raises:
-            ValueError: If backend is not provided.
```
SamFerracin, line 152: What happens now if a backend is not provided?

SamFerracin: Oh, looking below, I guess `self._run_simulator`

qiskit_ibm_runtime/options_models/simulator_options.py
diego-plan9: I think something along the lines of (please take it as pseudo-code in all senses) might do the trick:

```python
def deserialize_coupling_map(value):
    ...

class SimulatorOptions:
    ...
    coupling_map: (
        Annotated[
            CouplingMap,
            PlainValidator(deserialize_coupling_map),
        ]
        | None
    ) = None
```

This should achieve roughly the same as your existing validator - and avoiding the `PydanticSchemaGenerationError` and the need to use `arbitrary_types_allowed=True`. It is equivalent in some ways, but at least keeps the arbitrarity under control (so we are explicit about which non-base types are allowed).

Devil will likely be in the details:
* this only takes care of validating (and serializing)
* depending on what the consumer of the options expect, it might make sense to let both lists _and_ actual `CouplingMap` instances go through. It goes a bit against pydantic expectations, but there is some serialization cost (also depends on the level of compatibility needed with legacy options)
* if deserializing is needed (maybe for #2942?), an `PlainSerializer` can be added to the annotations ... but will not have an effect on `dataclass` (some extra machinery will be needed)

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,130 @@
...
+class SimulatorOptions:
+    """Simulator options.
+
+    Used to control local mode simulation.
+    """
+
+    noise_model: dict | NoiseModel | None = None
+    """Noise model for the simulator.
+
+    Default: ``None``.
```
diego-plan9: We can rely on the field defaults for informing users of the default value, can you do a round of removing them? If the default value is not self-explanatory, it should be fine to convert it into a proper sentence (ie. `If ``None``, it will be ...`.

qiskit_ibm_runtime/executor_sampler/sampler.py
diego-plan9: Thanks! Might make sense to go one step further, and avoid the extra function altogether? It might amount to just ~5 lines of code (excluding the call to `._run`), and we are already using a pattern of "check the kind of service and invoke" inside a number of functions [such as this one](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/qiskit_ibm_runtime/base_primitive.py#L154-L192) already, so it might end up being more readable / simpler. Non blocking.

qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -319,3 +332,37 @@ def prepare(
...
+            shots: The number of shots to run.
+
+        Returns:
+            A LocalRuntimeJob wrapping the BackendSamplerV2 job.
+
+        Raises:
+            ValueError: If using IBMBackend in local mode.
+        """
+        # Prepare options dict - this goes in the inputs["options"] field
+        options_dict = copy.deepcopy(asdict(self.options))  # type: ignore[call-overload]
```
diego-plan9: Is the `deepcopy` needed?

TsafrirA: Just to be on the safe side, because we are passing it to the simulator. I think it's not strictly needed.

qiskit_ibm_runtime/executor_sampler/sampler.py
TsafrirA: This should be caught when you call `get_mode_service_backend`, so I think it's redundant here.

qiskit_ibm_runtime/options_models/simulator_options.py
TsafrirA: It matches legacy behavior.

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,127 @@
...
+
+    For example, ``['u1', 'u2', 'u3', 'cx']``. Unrolling is not done if not set.
+    """
+
+    @field_validator("coupling_map", mode="plain")
+    @classmethod
+    def _validate_coupling_map(
+        cls, coupling_map: list[list[int]] | CouplingMap | None
+    ) -> CouplingMap:
+        """Validate and coerce a coupling map into a :class:`.CouplingMap`."""
```
TsafrirA, line 71: we don't have to coerce, but it makes validation easier.

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,127 @@
...
+from qiskit.exceptions import MissingOptionalLibraryError
+from qiskit.providers import BackendV2
+from qiskit.transpiler import CouplingMap
+from qiskit.utils import optionals
+
+from .utils import PRIMITIVES_CONFIG
+
+
+class NoiseModel:
+    """Fake noise model class for pydantic.
```
SamFerracin: I am thinking about a user reading this and becoming very confused. Could we improve the docstring?
```suggestion
    """And alias for :class:`...` for pydantic.
```

diego-plan9: It is actually not an alias - it is a bit of a funky mechanism, even if an inherited one.

I wonder if something like:
```python
# Some decent comment ...
if optionals.HAS_AER:
    from qiskit_aer.noise import NoiseModel as AerNoiseModel
    noise_model_type: TypeAlias = dict | AerNoiseModel | None
else:
    noise_model_type: TypeAlias = dict | None

...
class SimulatorOptions:
    noise_model: noise_model_type = None
```

might work/help in removing the "fake" `NoiseModel`. It can be tackled in a separate PR (on account for it being inherited), but I think we should tackle this one as it is quite misleading.

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+
+    @field_validator("coupling_map", mode="plain")
+    @classmethod
+    def _validate_coupling_map(
+        cls, coupling_map: list[list[int]] | CouplingMap | None
+    ) -> CouplingMap:
+        """Validate and coerce a coupling map into a :class:`.CouplingMap`."""
+        if coupling_map is None:
+            return None
```
diego-plan9, line 73: This is returning `None`, but the return type hint does not declare it - can you reconcile?

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,127 @@
...
+        elif isinstance(coupling_map, CouplingMap):
+            return coupling_map
+        elif isinstance(coupling_map, list):
+            try:
+                return CouplingMap(coupling_map)
+            except Exception as exc:
+                raise ValueError(f"Cannot create a valid CouplingMap from the provided list. {exc}")
+        else:
+            raise ValueError(
+                f"Expected a CouplingMap or list[list[int]] instead got {coupling_map}"
```
diego-plan9: Not sure if you meant to emit "instead got SomeOtherType" (ie. it is printing the value currently)?

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,127 @@
...
+    @classmethod
+    def _validate_noise_model(cls, model: dict | NoiseModel | None) -> dict | NoiseModel | None:
+        """Validate noise model."""
+        if model is None:
+            return model
+        if not isinstance(model, dict):
+            if not optionals.HAS_AER:
+                raise MissingOptionalLibraryError(
+                    "qiskit-aer", "Aer provider", "pip install qiskit-aer"
+                )
```
diego-plan9: Can you emit a `ValueError` instead? https://pydantic.dev/docs/validation/latest/errors/errors
> Validation code should not raise the ValidationError itself, but rather raise a ValueError or a AssertionError (or subclass thereof) which will be caught and used to populate the final ValidationError.

Also, combined with the suggestion at https://github.com/Qiskit/qiskit-ibm-runtime/pull/2940/changes#r3500509377, you _might_ be able to take advantage of not having to worry about `AerNoiseModel` when aer is not installed (might make the logic simpler, and the value error messages more precise).

qiskit_ibm_runtime/executor_sampler/sampler.py
diego-plan9: Un-resolving this to ensure it was parsed, and the trade-off evaluated.

qiskit_ibm_runtime/executor_sampler/sampler.py
diego-plan9: Can you please check or investigate further (specially as `asdict` should already be doing a copy itself)?

qiskit_ibm_runtime/options_models/simulator_options.py
TsafrirA: I think the current error is more appropriate. It's not that validation failed - your input might have been perfectly valid, it's just that you're missing the packages required.

I will test the other fixes, and we'll see if the discussion is rendered moot.

diego-plan9: Even if semantically appropriate from some perspective, from a different one, this is "validation code" - and Pydantic [explicitly expects](https://pydantic.dev/docs/validation/2.5/concepts/validators/#handling-errors-in-validators) a `ValueError` or `AssertionError`, for keeping the same behavior for all validators:

>  Any other errors (including TypeError) are bubbled up and not wrapped in a ValidationError.

Ultimately, from the end user perspective, validation does fail if this `raise` is emitted: if the user puts something in `.noise_model` that causes validation to reach that line, `noise_model` will not be able to be populated. Arguably it is slightly more natural in the [legacy class](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/qiskit_ibm_runtime/options/simulator_options.py#L74), with perhaps a tweak to the message (if aer is not present, the only value allowed is `dict`).

TsafrirA: [84c6afb](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2940/commits/84c6afb66d7925a0ef8b3ea8facf74867899ac78)

qiskit_ibm_runtime/options_models/simulator_options.py
TsafrirA: [84c6afb](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2940/commits/84c6afb66d7925a0ef8b3ea8facf74867899ac78)

qiskit_ibm_runtime/options_models/simulator_options.py
TsafrirA: Seems to be working [84c6afb](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2940/commits/84c6afb66d7925a0ef8b3ea8facf74867899ac78).

We should verify that the documentation correctly picks it up.

diego-plan9: Glad it worked!

It will be rendered in the documentation as long as `aer` is installed (and the tox environment [has it](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/tox.ini#L23)). I verified locally as well:

Screenshot 2026-07-01 at 14 06 12

qiskit_ibm_runtime/executor_sampler/sampler.py
TsafrirA: Thanks, this was redundant. As you said, `asdict` does deep copy on its own.

qiskit_ibm_runtime/options_models/simulator_options.py
```diff
@@ -0,0 +1,129 @@
...
+from pydantic import field_validator
+from pydantic.dataclasses import dataclass
+from qiskit.exceptions import MissingOptionalLibraryError
+from qiskit.providers import BackendV2
+from qiskit.transpiler import CouplingMap
+from qiskit.utils import optionals
+
+from .utils import PRIMITIVES_CONFIG
+
+# We avoid a forced dependency on qiskit-aer
```
diego-plan9:
```suggestion
# Dynamically define the `noise_model` field type at runtime, as `NoiseModel`
# is only a valid alternative if `qiskit_aer` is installed.
```

A bit nitty, but for being more explicit when it comes to surprising lines in the codebase.

diego-plan9 (approved):
Thanks! 

There is one conversation still open: https://github.com/Qiskit/qiskit-ibm-runtime/pull/2940#discussion_r3451887917, will continue it as a follow-up.

PR #2941, opened 2026-06-22
Homogenize on unittest-style asserts

diego-plan9:
### Summary

During #2934, noticed that we are mixing bare `asserts` with unitttest-style asserts (`self.assertFoo`) in the tests. While they are both acceptable and roughly equivalent in most scenarios, this PR homogenizes them in direction of unittest, with the main argument being consistency, and adds a `ruff` rule to enforce not using asserts.

### Details and comments

Other minor arguments for going with unittest are continuing to strive for a runner-agnostic suite based on `unittest`, which can provide more informative details when the asserts are not met. Minor argument for bare `assert` is readability and conciseness. 

Fixes N/A

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: claude opus 4.8 for easier replacing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,7 +90,7 @@ select = [
     "G",      # flake8-logging-format
     "I",      # isort
     "RUF013", # implicit-optional
-    "S101",   # flake8-bandit assert (enforced in test/ only)
+    "S101",   # flake8-bandit assert
     "T10",    # flake8-debugger
     "TC",     # flake8-type-checking
     "UP",     # pyupgrade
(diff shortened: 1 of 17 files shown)


pyproject.toml
```diff
@@ -90,6 +90,7 @@ select = [
     "G",      # flake8-logging-format
     "I",      # isort
     "RUF013", # implicit-optional
+    "S101",   # flake8-bandit assert (enforced in test/ only)
```
SamFerracin: Why enforcing it in tests only? Is there a reason not to ban it in source code too, and prefer raising a proper error instead?

I wonder if the problem is that we have it in docstrings

diego-plan9: Mostly as I was just thinking about the unit tests! It should be fine to enforce everywhere - dbaab23d

SamFerracin (approved):

PR #2945, opened 2026-06-22
Users should know how to verify whether a backend is active or inactive

### LINKED ISSUE #2923: Users should know how to verify whether a backend is active or inactive
Currently, backend availability (i.e., whether a backend is active or inactive) is only verified when a circuit is submitted. As a result, users may spend a significant amount of time running their scripts—for example, preprocessing circuits and transpiling them for a target backend—before discovering that the selected backend is unavailable.

There is in fact a way for users to understand whether a backend is active or inactive, namely via:
```python
backend = service.backend(name)
assert backend.status().operational and status.status_msg == "active"
```
or via:
```python
active_backends = service.backends(
    filters=lambda backend: (status := backend.status()).operational and status.status_msg == "active"
)
```
However, this should be better documented.

**Acceptance criteria**
- [ ] Add a snippet to ``service.backend`` to explain how to check if a backend is active
- [ ] Modify the snippet in ``service.backends`` so that the filter is as in the second snippet (which is arguably more useful than the snippets currently in place)

apchytr:
### Summary

Updating the runtime `.backend()` and `.backends()` docstrings to mention how users can check backend availability.

This PR closes #2923 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -561,11 +561,13 @@ class QiskitRuntimeService:
                 For example::
 
                     QiskitRuntimeService.backends(
-                        filters=lambda b: b.max_shots > 50000
+                        filters=lambda backend: (
+                            (status := backend.status()).operational
+                            and status.status_msg == "active"
+                        )
                     )
-                    QiskitRuntimeService.backends(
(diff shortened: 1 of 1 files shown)


diego-plan9 (commented):
Thanks! Some linting tweaks needed, but looking good changes-wise.

diego-plan9 (approved):

PR #2948, opened 2026-06-23
Add `process_expectation_values_pea`

dekelmeirom:
### Summary
Add a post-process function to calculate the extrapolated expectation values when using PEA EM method.

### Details and comments

Fixes #2947 

### AI/LLM disclosure

- [X] I used the following tool to generate or modify code: bob

PR #2953, opened 2026-06-24
Implement ZNE postprocess function

### LINKED ISSUE #2850: Wire ZNE into the executor estimator postprocess flow
- Either update the [postprocess](https://github.com/Qiskit/qiskit-ibm-runtime/blob/97ac0fc869250d5eb1cd268e040be3d00dc7ee06/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py) function or create a ZNE postprocessing function, which transforms an `QuantumProgramResult` from a ZNE experiment into an `PrimitiveResult[PubResult]`.
- Ensure the `PubResult` for each input circuit contains the noise factors, measurements at each noise factor, the EVs for each noise-amplified point, and the EV for the extrapolated point. Standard deviations should also be provided for all noise-amplified points and extrapolated point(s)

dekelmeirom:
### Summary
Add a post-process function to calculate the extrapolated expectation values when using ZNE with gate folding EM method.

### Details and comments

Fixes #2850 

### AI/LLM disclosure

- [X] I used the following tool to generate or modify code: bob


dekelmeirom:
Note - this PR changes the I/O of the extrapolation methods in the same way #2948 does (so after #2948 will be merged, this PR will have changes only to the zne related functions)

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -78,6 +80,9 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
     # Extract options if present
     options_metadata = post_processor_data.get("options", {})
 
+    # Extract item_id if present
+    item_id = post_processor_data.get("item_id", None)
```
TsafrirA: This goes back to the prepare function, but the name `item_id` is not very descriptive here, as we deal with a mapping.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -96,6 +101,29 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
+            pub_number = item_id[idx][0]
+            pub_items_results = [item_result]
+            # add additional items to the list until reaching item associated with another pub
+            for item_idx, item in enumerate(item_id[idx + 1 :], start=idx + 1):
+                if item[0] != pub_number:
+                    break
+                pub_items_results.append(result[item_idx])
+                skip_counter += 1
+            combined_results.append(pub_items_results)
+
```
TsafrirA, line 155: If it stays as is, I would add a TODO clarifying why `combined_results` is never used.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -96,6 +101,29 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
             passthrough_data=result.passthrough_data,
         )
 
+    # In case each pub is associated with several items - create a list in which each element
+    # is a list containing all relevant items for that pub
+    if item_id is not None:
```
TsafrirA: I think the logic here is overly general, and could be simplified because we know how the items are ordered.
Because the items of each pub are grouped together, we can convert the main loop to work in appropriate jumps and when doing so, pass all items as a group to the required post processing function.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -41,20 +42,23 @@
...
     For each entry, the requested model(s) are fit to the expectation values measured at
     the entry's noise factors and evaluated at the target noise factor(s) (``0`` for zero
     noise). Models are tried in priority order: each point takes the result of the first model
     with a valid extrapolation. An extrapolation is valid when its value and standard
     error are finite, the standard error is within the basis threshold, and
-    :math:`value \pm stderr` lies within the basis's ideal range widened by that threshold. The
-    range comes from the entry's ``ev_basis`` metadata: ``[0, 1]`` for projector-only bases,
-    ``[-1, 1]`` for bases containing Paulis, and unbounded when ``ev_basis`` is absent or
+    :math:`value \pm stderr` lies within the observable's ideal range widened by that threshold.
+    The range vary based on the observable: ``[0, 1]`` for projector-only observables,
```
TsafrirA:
```suggestion
    The range varies based on the observable: ``[0, 1]`` for projector-only observables,
```

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -41,20 +42,23 @@
...
-    extrapolated_noise_factors: float | ArrayLike = 0,
-) -> EstimatorResult:
+    exp_vals: npt.NDArray[float],
+    standard_errors: npt.NDArray[float],
+    observables: ObservablesArray,
+    zne_noise_factors: Sequence[float],
+    extrapolators: str | Sequence[str],
+    extrapolated_noise_factors: float | npt.ArrayLike = 0,
+) -> tuple[npt.NDArray[float], npt.NDArray[float], npt.NDArray[str]]:
     r"""Apply zero-noise extrapolation (ZNE) to an estimator result.
```
TsafrirA: No longer takes an estimator result as input.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -66,12 +70,14 @@ def process_extrapolated_expectation_values(
...
-            one per noise factor, aligned with the ``standard_error`` and
-            ``resilience["zne_noise_factors"]`` arrays in its metadata. An optional ``ev_basis``
-            string sets the ideal-value range used to judge extrapolation validity; an optional
-            ``ensemble_standard_error`` array is carried through to the output.
-        extrapolator: A builtin model name, or a sequence of names tried in priority order.
+        exp_vals: Raw expectation value result. Each element of measured observable and parameter
+            is a 1D array of expectation values, one per noise factor.
+        standard_errors: Raw standard deviations of the results. Have the same shape as
+            ``exp_vals``.
+        observables: The observables to calculate expectation values for. Determine the ideal-value
```
TsafrirA:
```suggestion
        observables: The observables to calculate expectation values for. The observables determine the ideal-value
```

TsafrirA (commented):
Some initial comments, mainly on the integration into the main post processor function.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: `combined_results` is never used because it will be an input to the `post_process_zne` function, and the logic to call the required prepare/post-process function based on the options was left to a PR that will manage the calls to all mitigation methods together (neither of the mitigations post-processing functions are not used). I can add a relevant TODO for this if it will help.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: If we agree that the noise factor list is global to all the pubs in the estimator and will not allowed to be different for each pub, we can indeed simplify this loop to just collect items with given jumps. We will also be able to remove the `item_id` mapping altogether from the passthrough_data, as the mapping will always be as the noise_factor list for all pubs.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: Yes, it's just not clear by looking at the function.

We could also consider not making partial changes to this function, and only implement them once everything comes together.

dekelmeirom:
The extrapolation I/O changes were moved to a separate PR which will need to be merged first - #3000 (the tests will fail until this PR will be merged, but they pass combined with this PR code)

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: This will fail for ZNE, as the number of items is not the same as the number of pubs.

Also highlights that we are missing a test for `estimator_v2_post_processor_v0_1` with a ZNE result.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -460,6 +491,132 @@ def create_pub_result_pec(
...
+            - ``"linear"``: ``a + b*x``
+            - ``"polynomial_degree_k"`` (1 <= k <= 7): a degree-k polynomial
+            - ``"exponential"``: ``a*exp(b*x)``
+            - ``"double_exponential"``: ``a*exp(b*x) + c*exp(d*x)`` (rates constrained to decay)
+            - ``"fallback"``: no fit; the measured value at the lowest noise factor
+
+    Returns:
+        An :class:`~qiskit_ibm_runtime.results.EstimatorPubResult` with an empty metadata dict.
+    """
+    # The last returned value is the selected extrapolators, that should be saved in the
```
TsafrirA:
```suggestion
    # TODO: The last returned value is the selected extrapolators, that should be saved in the
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -460,6 +491,132 @@ def create_pub_result_pec(
...
+    item_results: list[QuantumProgramItemResult],
+    observables: ObservablesArray,
+    param_shape: tuple[int, ...],
+    param_basis_pairs: list[tuple[tuple[int, ...], str]],
+    noise_factors: list[float],
+    extrapolated_noise_factors: float | int | list[float],
+    extrapolator: list[ExtrapolatorType],
+    measure_noise_data: PauliLindbladMap | np.ndarray | None,
+) -> tuple[npt.NDArray[float], npt.NDArray[float], npt.NDArray[float], list[list[str]]]:
+    """Process expectation values for a single item result.
```
TsafrirA:
```suggestion
    """Process expectation values for a single pub.
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -460,6 +491,132 @@ def create_pub_result_pec(
...
+    # Combine the data from each noise factor
+    noise_amplified_data = []
+    for i, item_result in enumerate(item_results):
+        try:
+            data = item_result["_meas"]
+        except KeyError:
+            raise ValueError("Dedicated creg ``'_meas'`` is missing from one of the results.")
+
+        if data.ndim != 4:
+            # Shape: (num_randomizations, num_noise_scales, num_configs, shots, num_bits)
```
TsafrirA:
```suggestion
            # Shape: (num_randomizations, num_configs, shots, num_bits)
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -460,6 +491,132 @@ def create_pub_result_pec(
...
+        ValueError: If ``item_result['_meas']`` has a number of axis not equal to ``4``.
+        ValueError: If ``param_shape`` and ``observables.shape`` cannot be broadcasted against
+            each other.
+    """
+    if isinstance(extrapolated_noise_factors, (float, int)):
+        extrapolated_noise_factors = [extrapolated_noise_factors]
+
+    # Combine the data from each noise factor
+    noise_amplified_data = []
+    for i, item_result in enumerate(item_results):
```
TsafrirA:
```suggestion
    for item_result in item_results:
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: What about floats here?

qiskit_ibm_runtime/executor_estimator/zne/prepare_zne.py
```diff
@@ -183,6 +183,8 @@ def prepare_zne(
             "measure_mitigation": measure_noise_learning is not None,
             "mitigation": "zne",
             "zne_noise_factors": noise_factors,
+            "extrapolated_noise_factors": zne_options.extrapolated_noise_factors,
+            "extrapolator": zne_options.extrapolator,
             "item_id": item_id,
```
TsafrirA: I think we no longer use this mapping, because we assume that items are ordered sequentially.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -102,6 +107,14 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
             passthrough_data=result.passthrough_data,
         )
 
+    # In case each pub is associated with several items - create a list in which each element
+    # is a list containing all relevant items for that pub
+    res_step = 1
+    if zne_noise_factors is not None:
```
TsafrirA: Should we use the `mitigation` flag to make it more robust?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -119,9 +132,16 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
     # Build EstimatorPubResult for each pub
     pub_results = []
-    for idx, (item_result, observables_label, param_basis_pairs, param_shape) in enumerate(
-        zip(result, observables_lists, param_basis_pairs_lists, param_shapes_list)
+    for pub_idx, (item_result, observables_label, param_basis_pairs, param_shape) in enumerate(
+        zip(result[::res_step], observables_lists, param_basis_pairs_lists, param_shapes_list)
     ):
+        # In case each pub is associated with several items - create a list in which each element
+        # is a list containing all relevant items for that pub
+        combined_results = []
```
TsafrirA: The calculation of `combined_results` only applies to the ZNE branch, so I think we should move it into that branch.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -460,6 +491,132 @@ def create_pub_result_pec(
...
+    param_basis_pairs: list[tuple[tuple[int, ...], str]],
+    measure_noise_data: PauliLindbladMap | np.ndarray | None,
+    noise_factors: list[float],
+    extrapolated_noise_factors: float | int | list[float],
+    extrapolator: list[ExtrapolatorType],
+) -> EstimatorPubResult:
+    """Calculate expectation values and errors with ZNE, and return pub result.
+
+    Args:
+        item_results: The item result.
```
TsafrirA:
```suggestion
        item_results: The item results.
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -134,7 +154,18 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
+            )
+        elif mitigation == "zne":
+            pub_result = create_pub_result_zne(
+                combined_results,
+                observables,
+                param_shape,
+                param_basis_pairs,
+                readout_noise_data,
+                noise_factors=zne_noise_factors,
+                extrapolated_noise_factors=extrapolated_noise_factors,
```
TsafrirA, line 203: The options have an `auto` value for `extrapolated_noise_factors`. I don't see where it's resolved. It's passed as is during prepare, and I don't see it being resolved here.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -460,6 +491,132 @@ def create_pub_result_pec(
...
+        if data.ndim != 4:
+            # Shape: (num_randomizations, num_noise_scales, num_configs, shots, num_bits)
+            # where num_configs is the total number of (param_index, basis) pairs
+            raise ValueError(
+                f"one of the ``item_result['_meas']`` has ``{data.ndim}`` axes, expected ``4``."
+            )
+
+        # Apply measurement flips if present
+        if "measurement_flips._meas" in item_result:
+            data ^= item_result["measurement_flips._meas"]
```
TsafrirA: We can pop the measurement flips here. Also makes sure you can't apply it twice somehow.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -134,7 +154,18 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
                 param_shape,
                 param_basis_pairs,
                 readout_noise_data,
-                pec_gamma=pec_gammas[idx],
+                pec_gamma=pec_gammas[pub_idx],
+            )
+        elif mitigation == "zne":
+            pub_result = create_pub_result_zne(
```
TsafrirA, line 196: Just like in the PEA PR, some of the values could be None, but are passed as here as actual values. Need to validate. (Again, I don't understand how this passes static checks).

TsafrirA (changes requested):
Some nitpicks, and a couple of things that need a fix -

- Length validation will fail.
- 'auto' value will fail for `extrapolated_noise_factors`.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: I think it should be resolved in `finalize_options`.

qiskit_ibm_runtime/executor_estimator/zne/prepare_zne.py
dekelmeirom: you are correct. I removed it. We can add it in the future if we will have somehow option to have different number of items for each pub

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: I resolved it in `prepare_zne`. We can move it if you think it will be better to be in `finalize_options`.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: added check for that

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -139,7 +148,25 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
                 param_basis_pairs,
                 readout_noise_data,
-                pec_gamma=pec_gammas[idx],
+                pec_gamma=pec_gammas[pub_idx],
+            )
+        elif mitigation == "zne":
+            # In case each pub is associated with several items - create a list in which each
+            # element is a list containing all relevant items for that pub
+            combined_results = []
+            if res_step != 1:
```
TsafrirA: There is no guard against a ZNE payload with only 1 noise factor. Under this edge case we'll skip the `if` and end up with no `pub_result`. I think we should add a guard in the prepare function, and assume res_step>1 here?

TsafrirA: (Is there even a guard against an empty list of noise factors?)

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: The error message is inconsistent with the check.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -712,7 +872,7 @@ def calculate_extrapolated_expectation_values(
                 )
                 noise_scaled_exp_vals.append(term_exp_val)
                 noise_scaled_ensemble_std.append(np.sqrt(term_ensemble_variance))
-                if noise_factors[noise_factor_index] == 1:
+                if noise_factors[noise_factor_index] in [1, 1.0]:
```
TsafrirA:
```suggestion
                if noise_factors[noise_factor_index] == 1.0:
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -712,7 +872,7 @@ def calculate_extrapolated_expectation_values(
...
                     non_amplified_twirl_var = term_twirl_variance
```
TsafrirA: What happens if `1` is missing from the noise factors?
Do we force it to be there?

TsafrirA (commented):
A couple of edge cases for your consideration.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: there is a guard now to check there are at least 2 points

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: removed, as we now calculate the twirled data for each of the noise_factor points

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: the if statement was removed, as we now calculate the twirled data for each of the noise_factor points

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: should be `len(result[::res_step])`.

qiskit_ibm_runtime/executor_estimator/zne/prepare_zne.py
```diff
@@ -96,6 +97,11 @@ def prepare_zne(
     else:
         noise_factors = np.array(zne_options.noise_factors)
 
+    if len(noise_factors) < 2:
```
TsafrirA: I think we can move this to the options class, and that way it will handle both PEA and ZNE together.

TsafrirA (commented):
Thanks @dekelmeirom!

There is one small issue with the circuits metadata that needs fixing. Other than the code seems solid. I am just a little worried about how this stacks up against the legacy estimator - could you explain which attributes are currently matching and which aren't, and what is your plan for aligning them? I am not against leaving some stuff to follow up PRs, but I am not sure we have clarity on exactly what's missing.

dekelmeirom:
> I am just a little worried about how this stacks up against the legacy estimator - could you explain which attributes are currently matching and which aren't, and what is your plan for aligning them? I am not against leaving some stuff to follow up PRs, but I am not sure we have clarity on exactly what's missing.

Currently, `_process_expectation_values_zne` returns the evs and the enseble_std of the extrapolated points, and evs, enemble_std and twirl_stds of the noise_factor points. in addition, it returns also the valid extrapolators that were chosen to do the extrapolation for each observable term (so the dimension of this parameter is not as the broadcasting dimension).
As far as I understand, the legacy estimator returned the same data, maybe with different names. I am not aware of data that is missing, and if there is such, I will be happy to discuss it and understand how we can calculate it.

TsafrirA:
Thanks @dekelmeirom. If it's just a question of renaming the parameters, can we get them right in this PR?

These are the fields I am seeing in the legacy data bin - 

- `evs`
- `stds`
- `evs_noise_factors`
- `stds_noise_factors`
- `ensemble_stds_noise_factors`
- `evs_extrapolated`
- `stds_extrapolated`

dekelmeirom:
I added the parameters to the data bin. As far as I understand (but I might be wrong) - `evs` and `evs_extrapolated` have the same data (and also `stds` and `stds_extrapolated`). There is still the selected extrapolators which I think that should be added somewhere in the metadata

TsafrirA:
> I added the parameters to the data bin. As far as I understand (but I might be wrong) - evs and evs_extrapolated have the same data (and also stds and stds_extrapolated). 

Thanks. As far as I can tell `evs` is a single value per parameter\observable combination, while `evs_extrapolated` is one per requested extrapolated noise level. So, the two are not the same. I can look into the exact definition if you want.

> There is still the selected extrapolators which I think that should be added somewhere in the metadata

That can be left to future efforts as part of #3046.

qiskit_ibm_runtime/executor_estimator/zne/prepare_zne.py
TsafrirA:
```suggestion
        noise_factors = np.array(ZNE_DEFAULT_NOISE_FACTORS, dtype=float)
```

qiskit_ibm_runtime/executor_estimator/zne/prepare_zne.py
```diff
@@ -96,6 +96,10 @@ def prepare_zne(
     else:
         noise_factors = np.array(zne_options.noise_factors)
```
TsafrirA:
```suggestion
        noise_factors = np.array(zne_options.noise_factors, dtype=float)
```

qiskit_ibm_runtime/executor_estimator/zne/prepare_zne.py
```diff
@@ -96,6 +96,10 @@ def prepare_zne(
...
 
+    extrapolated_noise_factors = zne_options.extrapolated_noise_factors
+    if extrapolated_noise_factors == "auto":
+        extrapolated_noise_factors = np.insert(noise_factors, 0, 0.0)
+
```
TsafrirA, line 104:
```suggestion
else:
        extrapolated_noise_factors = np.array(extrapolated_noise_factors, dtype=float)

```

qiskit_ibm_runtime/executor_estimator/zne/prepare_zne.py
TsafrirA, line 93: Because `noise factors` and `extrapolated_noise_factors` end up in the passthrough data we need to make sure they don't end up as an int, because it fails the conversion.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -90,16 +93,22 @@ def process_extrapolated_expectation_values(
...
     """
     if isinstance(extrapolators, str):
         extrapolators = [extrapolators]
 
     if isinstance(extrapolated_noise_factors, (float, int)):
         extrapolated_noise_factors = [extrapolated_noise_factors]
 
+    # Always evaluate at zero-noise point, copy list to not modify original
+    extrapolated_noise_factors = copy.deepcopy(extrapolated_noise_factors)
+    extrapolated_noise_factors.insert(0, 0)
```
TsafrirA: This fails when `extrapolated_noise_factors` is a numpy array which currently happens in the 'auto' case.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -176,7 +186,24 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
                 param_shape,
                 param_basis_pairs,
                 readout_noise_data,
-                pec_gamma=pec_gammas[idx],
+                pec_gamma=pec_gammas[pub_idx],
+            )
+        elif mitigation == "zne":
+            # In case each pub is associated with several items - create a list in which each
+            # element is a list containing all relevant items for that pub
+            combined_results = []
```
TsafrirA:
```suggestion
            combined_results = result[pub_idx * res_step:(pub_idx + 1) * res_step]
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -553,7 +580,16 @@ def create_pub_result_pea(
...
     # metadata
-    exp_vals, stds, ensemble_stds, _ = _process_expectation_values_pea(
+    (
+        zero_noise_exp_vals,
+        zero_noise_stds,
+        noise_factors_exp_vals,
+        noise_factors_ensemble_stds,
+        noise_factors_stds,
+        extrapolated_exp_vals,
+        extrapolated_ensemble_stds,
```
TsafrirA:
```suggestion
        extrapolated_stds,
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -600,10 +652,19 @@ def _process_expectation_values_pea(
             PauliLindbladMap of a noise model learned upfront, or a result of a calibration circuit.
 
     Returns:
-        A tuple ``(exp_vals, stds, ensemble_stds, selected_extrapolators)``, where ``exp_vals``
-        are expectation values, ``stds`` are standard deviations, and ``ensemble_stds`` are
-        ensemble standard errors. ``selected_extrapolators`` is a list of the valid extrapolators
-        used to extrapolate the data.
+        A tuple ``(zero_extrapolated_exp_vals, zero_extrapolated_stds, noise_factors_exp_vals,
```
TsafrirA: Maybe we can format this is a list?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -600,10 +652,19 @@ def _process_expectation_values_pea(
...
+        noise_factors_ensemble_stds, noise_factors_stds, extrapolated_exp_vals,
+        extrapolated_ensemble_stds, selected_extrapolators)``, where
+        ``zero_extrapolated_exp_vals`` are expectation values evaluated at zero noise point,
+        ``zero_extrapolated_stds`` are the standard deviations of the extrapolated expectation
+        values evaluated at zero noise point, ``noise_factors_exp_vals`` are expectation values
+        calculated at the noise_factors points, ``noise_factors_ensemble_stds`` are ensemble
```
TsafrirA: I think we should explain what that means, or refer to somewhere else where it is explained.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -600,10 +652,19 @@ def _process_expectation_values_pea(
...
+        standard errors calculated at the noise_factors points, ``noise_factors_stds`` are
+        standard errors calculated at the noise_factors points, ``extrapolated_exp_vals``
+        are expectation values evaluated at the extrapolated_noise_factors points,
+        ``extrapolated_ensemble_stds`` are ensemble standard errors evaluated at the
```
TsafrirA:
```suggestion
        ``extrapolated_stds`` are ensemble standard errors evaluated at the
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -645,6 +706,173 @@ def _process_expectation_values_pea(
...
+        noise_factors: The noise factors used to amplify the noise.
+        extrapolated_noise_factors: Noise factors to evaluate the fits at.
+        extrapolator: The extrapolator model or models to use. Models will be tried in priority
+            order. Supported models (each fits the named function of the noise factor ``x``):
+            - ``"linear"``: ``a + b*x``
+            - ``"polynomial_degree_k"`` (1 <= k <= 7): a degree-k polynomial
+            - ``"exponential"``: ``a*exp(b*x)``
+            - ``"double_exponential"``: ``a*exp(b*x) + c*exp(d*x)`` (rates constrained to decay)
+            - ``"fallback"``: no fit; the measured value at the lowest noise factor
+
```
TsafrirA, line 735: I wonder if we should explain the contents of the `data` field here.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -645,6 +706,173 @@ def _process_expectation_values_pea(
...
+    # TODO: The last returned value is the selected extrapolators, that should be saved in the
+    # metadata
+    (
+        zero_noise_exp_vals,
+        zero_noise_stds,
+        noise_factors_exp_vals,
+        noise_factors_ensemble_stds,
+        noise_factors_stds,
+        extrapolated_exp_vals,
+        extrapolated_ensemble_stds,
```
TsafrirA:
```suggestion
        extrapolated_stds,
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -645,6 +706,173 @@ def _process_expectation_values_pea(
...
+        noise_factors_ensemble_stds, noise_factors_stds, extrapolated_exp_vals,
+        extrapolated_ensemble_stds, selected_extrapolators)``, where
+        ``zero_extrapolated_exp_vals`` are expectation values evaluated at zero noise point,
+        ``zero_extrapolated_stds`` are the standard deviations of the extrapolated expectation
+        values evaluated at zero noise point, ``noise_factors_exp_vals`` are expectation values
+        calculated at the noise_factors points, ``noise_factors_ensemble_stds`` are ensemble
+        standard errors calculated at the noise_factors points, ``noise_factors_stds`` are
+        standard errors calculated at the noise_factors points, ``extrapolated_exp_vals``
+        are expectation values evaluated at the extrapolated_noise_factors points,
+        ``extrapolated_ensemble_stds`` are ensemble standard errors evaluated at the
```
TsafrirA:
```suggestion
        ``extrapolated_stds`` are ensemble standard errors evaluated at the
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -645,6 +706,173 @@ def _process_expectation_values_pea(
...
+            order. Supported models (each fits the named function of the noise factor ``x``):
+            - ``"linear"``: ``a + b*x``
+            - ``"polynomial_degree_k"`` (1 <= k <= 7): a degree-k polynomial
+            - ``"exponential"``: ``a*exp(b*x)``
+            - ``"double_exponential"``: ``a*exp(b*x) + c*exp(d*x)`` (rates constrained to decay)
+            - ``"fallback"``: no fit; the measured value at the lowest noise factor
+        measure_noise_data: Measurement noise calibration data for TREX mitigation. Can be either a
+            PauliLindbladMap of a noise model learned upfront, or a result of a calibration circuit.
+
+    Returns:
```
TsafrirA, line 817: Same questions as above.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -645,6 +706,173 @@ def _process_expectation_values_pea(
...
+        extrapolated_noise_factors points and ``selected_extrapolators`` are the valid
+        extrapolators used to extrapolate the data for each observable term.
+
+    Raises:
+        ValueError: If ``item_result`` has no ``'_meas'`` key.
+        ValueError: If ``item_result['_meas']`` has a number of axis not equal to ``4``.
+        ValueError: If ``param_shape`` and ``observables.shape`` cannot be broadcasted against
+            each other.
+    """
+    if isinstance(extrapolated_noise_factors, (float, int)):
```
TsafrirA, line 839: If we follow my suggestion above, this check will not be needed.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -676,8 +913,19 @@ def calculate_extrapolated_expectation_values(
             PauliLindbladMap of a noise model learned upfront, or a result of a calibration circuit.
 
     Returns:
-        A tuple ``(exp_vals, stds, ensemble_stds)``, where ``exp_vals`` are expectation values,
-        ``stds`` are standard deviations, and ``ensemble_stds`` are ensemble standard errors.
+        A tuple ``(zero_extrapolated_exp_vals, zero_extrapolated_stds, noise_factors_exp_vals,
```
TsafrirA: Same questions as above.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -701,9 +949,34 @@ def calculate_extrapolated_expectation_values(
         )
 
     # Compute expectation values for all observables
-    exp_vals = np.empty(shape=(len(extrapolated_noise_factors),) + output_shape, dtype=float)
-    stds = np.empty(shape=(len(extrapolated_noise_factors),) + output_shape, dtype=float)
-    ensemble_stds = np.empty(shape=(len(extrapolated_noise_factors),) + output_shape, dtype=float)
+    zero_extrapolated_exp_vals = np.zeros(shape=output_shape, dtype=float)
+    zero_extrapolated_ensemble_vars = np.zeros(shape=output_shape, dtype=float)
```
TsafrirA:
```suggestion
    zero_extrapolated_vars = np.zeros(shape=output_shape, dtype=float)
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -701,9 +949,34 @@ def calculate_extrapolated_expectation_values(
...
+    )
+    # Save also the data for all of the noise amplified points
+    noise_factors_exp_vals = np.zeros(shape=output_shape + (len(noise_factors),), dtype=float)
+    noise_factors_ensemble_variance = np.zeros(
+        shape=output_shape + (len(noise_factors),), dtype=float
+    )
+    noise_factors_twirl_variance = np.zeros(shape=output_shape + (len(noise_factors),), dtype=float)
+    noise_factors_ensemble_stds = np.empty(shape=output_shape + (len(noise_factors),), dtype=float)
+    noise_factors_twirl_stds = np.empty(shape=output_shape + (len(noise_factors),), dtype=float)
+    # save for each extrapolated observable term the selected extrapolator
```
TsafrirA: Is it per term?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -764,36 +1055,51 @@ def calculate_extrapolated_expectation_values(
...
-            # Calculate scale factor in case TREX mitigation is used
-            term_scale_factor = (
-                calculate_trex_factor(measure_noise_data, observable_term)
-                if measure_noise_data is not None
-                else 1
+            zero_extrapolated_exp_vals[bcast_index] += (
+                coeff * selected_exp_vals[0] * term_scale_factor
             )
-            for extrap_index, (extrap_exp_val, extrap_std) in enumerate(
+            if not np.isnan(selected_stds[0]):
```
TsafrirA: Does this mean that the extrapolation failed? If so, how is the exp_vals summed? Also, does the sum have any meaning if some of the elements were nan and skipped?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -764,36 +1055,51 @@ def calculate_extrapolated_expectation_values(
...
-        exp_vals[(slice(None), *bcast_index)] = exp_vals_extrapolated
-        ensemble_stds[(slice(None), *bcast_index)] = np.sqrt(ensemble_var_extrapolated)
-        # When twirling is off (num_randomizations=1), stds equals ensemble_standard_error
-        if num_randomizations == 1:
-            stds[(slice(None), *bcast_index)] = ensemble_stds[(slice(None), *bcast_index)]
-        else:
-            stds[(slice(None), *bcast_index)] = np.sqrt(
-                twirl_variance_extrapolated / num_randomizations
+                    # failed extrapolation might return NaN as std; accumulate variance
+                    if not np.isnan(extrap_std):
```
TsafrirA: Same question as above.

TsafrirA (commented):
We are getting there!

There are a couple of places which fail end-to-end (depending on some options) and require minor fixes. This indicates that we don't have a good enough test coverage here, but we might be able to merge and address that as part of the ongoing testing revamping.

Besides that
-  I think the doc strings need some work, but we could do that in a follow up PR.
- The names `extrapolated_ensemble_*` are misleading, because this is not an ensemble std like in the vanilla case.
- I am not clear on the selected_extrapolators - is that supposed to be the selected just for 0 noise, or for all extrapolation values?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: I think it is considered better practice to return a tuple in the general case, is there a reason to return a list instead?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: As far as I understand, it is not a public facing function. Should we explain it better for developers? or will it be enough to explain with more details all of the returned values in the public facing function?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: I meant to turn the doc string into a list to make it more readable :)

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: it is not a public facing function. A developer can "click" of the data type and see that it is a`Literal`. Should we still explain it further here?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: it is true if we look at this function inly as part of this full workflow. If we look at the function also outside of this flow context, we can remove this only if we change the input types to not include `int` or `float`. I am not sure that I want to do this if taking into account that this function will be moved to another repo in the future. I will think about it, but I think I prefer to leave it as is for now.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: yes, it is per term, per observable, per param confg. I edited a bit the comment

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: one can argue what should be the behavior in such a case. For simplicity, I change each `NaN` into `inf` in the extrapolation result (in the extrapolation method) so all the result elements are valid numpy elements that can be summed (so if the extrapolation of one of the terms fails, the total result will be `inf` in such case).
We can discuss what is the best behavior in such case (but maybe in another PR in case the current one is not the wanted one)

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: Same answer as above

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: Ok. That makes sense. I think I would have preferred to have both as `NaN`, but that's something we can discuss later.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: That's a fair point. I think the documentation is lacking at the moment, but it's not entirely clear where the gap should be filled.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -106,8 +105,8 @@ def process_extrapolated_expectation_values(
         extrapolated_noise_factors = [extrapolated_noise_factors]
 
     # Always evaluate at zero-noise point, copy list to not modify original
```
TsafrirA, line 107: The comment is outdated.

TsafrirA (approved):
I will open follow up issues to support some cleanup and decision making, but I think this should be merged to unblock testing.

Thanks @dekelmeirom for all the effort you put into this!

PR #2960, opened 2026-06-25
Always contruct targets with bounds on rzz gates

### LINKED ISSUE #2441: Use new `angle_bounds` and `WrapAngles` features of Qiskit 2.2 with `rzz` gate

Qiskit 2.2 [added](https://github.com/Qiskit/qiskit/pull/14406) a new `angle_bounds` parameter to `Target.add_instruction()` which can specify bounds for a the added gate's angle parameter. Additionally, it added a new `WrapAngles` transpiler pass that can act on angles that are out of bounds and bring them back within bounds. qiskit-ibm-runtime currently exposes the `rzz` gate which is limited to [0,pi/2] on IBM systems. It provides its own transpiler pass `FoldRzzAngle` for wrapping the angles. As much as possible the `rzz` code here should make use of the support in Qiskit (at least the support for wrapping angles in parameterized angles in PUBs would need to stay).

Expected benefits:

* Less code to maintain in qiskit-ibm-runtime. The base functionality of registering a transpiler pass and checking for gates to wrap is handled by Qiskit. Only the function wrapping the gate angle is needed.
* Performance. The base functionality of the wrap angles pass is written in Rust and might be faster than `FoldRzzAngle`.
* Broader ecosystem integration. If some other code wants to make use of the angle bounds, they will be recorded in a standard place in the `Target` now.

**Acceptance criteria**

1. Decide on which features to support before dropping support for Qiskit <2.2. The code could opportunistically use the new method for Qiskit >=2.2 now and fall back to the old method otherwise. Or we could wait until dropping support for Qiskit <2.2 before supporting the new method.
2. Start setting the `angle_bounds` when adding `rzz` to the backend `Target`.
3. Register a wrap angles call back for `rzz` that behaves like [this block](https://github.com/Qiskit/qiskit-ibm-runtime/blob/a4fb0efa98acc043ace14516e0ee6ad8a2d45e05/qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py#L100-L116) from `FoldRzzAngle` and stop using the `ibm_fractional` translation plugin.
4. Decide and act on a deprecation plan for `FoldRzzAngle`

The code could only add the angle bounds, register the call back, and skip using the `ibm_fractional` plugin when Qiskit >=2.2, but I don't know that it would be the best user experience for the behavior to differ across Qiskit versions like that.

yaelbh:
### Summary

- All backend targets have bounds `(0, pi/2)` for `rzz` instructions.
- The mechanism of the `FoldRzzAngle` pass is registered in the wrap angles registry.
- The fractional-dynamic plugin is not used anymore.

### Details and comments

Fixes #2441

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Bob 2.0.0


qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -315,5 +320,36 @@ def _get_value(prop_dict: dict, prop_name: str) -> Any:
...
                 name=inst_name,
+                angle_bounds=[(0, math.pi / 2)] if inst_name == "rzz" else None,
             )
+            # Register the FoldRzzAngle callback for the WrapAngles pass so that when the
+            # transpiler encounters an out-of-bounds Rzz angle it knows how to fold it.
+            if (
+                inst_name == "rzz"
+                and (wrap_angles_cls := getattr(transpiler_passes, "WrapAngles", None)) is not None
+            ):
+                from qiskit_ibm_runtime.transpiler.passes.basis.fold_rzz_angle import FoldRzzAngle
```
diego-plan9: Is placing this `import` (and the other one in the `if windings % 2:` block) here intentional?

yaelbh: It was intentional because of circular imports.

Placing the `FoldRzzAngle` import at the top of the file:
```
  File "/Users/yaelbh/miniforge3/envs/py311/lib/python3.11/site-packages/quantum_tests/plugin.py", line 16, in <module>
    from qiskit_ibm_runtime import RuntimeJobV2, SamplerV2, EstimatorV2
  File "/Users/yaelbh/work/runtime/takashi/qiskit_ibm_runtime/__init__.py", line 203, in <module>
    from .qiskit_runtime_service import QiskitRuntimeService
  File "/Users/yaelbh/work/runtime/takashi/qiskit_ibm_runtime/qiskit_runtime_service.py", line 25, in <module>
    from qiskit_ibm_runtime import ibm_backend
  File "/Users/yaelbh/work/runtime/takashi/qiskit_ibm_runtime/ibm_backend.py", line 29, in <module>
    from .utils.backend_converter import convert_to_target
  File "/Users/yaelbh/work/runtime/takashi/qiskit_ibm_runtime/utils/backend_converter.py", line 41, in <module>
    from qiskit_ibm_runtime.transpiler.passes.basis.fold_rzz_angle import FoldRzzAngle
  File "/Users/yaelbh/work/runtime/takashi/qiskit_ibm_runtime/transpiler/passes/__init__.py", line 36, in <module>
    from .basis import ConvertIdToDelay, ConvertToMidCircuitMeasure, FoldRzzAngle
  File "/Users/yaelbh/work/runtime/takashi/qiskit_ibm_runtime/transpiler/passes/basis/__init__.py", line 17, in <module>
    from .fold_rzz_angle import FoldRzzAngle
  File "/Users/yaelbh/work/runtime/takashi/qiskit_ibm_runtime/transpiler/passes/basis/fold_rzz_angle.py", line 37, in <module>
    from qiskit_ibm_runtime import EstimatorV2, SamplerV2
ImportError: cannot import name 'EstimatorV2' from partially initialized module 'qiskit_ibm_runtime' (most likely due to a circular import) (/Users/yaelbh/work/runtime/takashi/qiskit_ibm_runtime/__init__.py)
```

Placing the `GlobalPhaseGate` import at the top of the file:
Success!

So I'll move the `GlobalPhaseGate` import.

yaelbh: done in f6c5288d5d600561a551427d38d9a8890fb8aabc (for one import out of the two that are mentioned in the comment)

qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -315,5 +320,34 @@ def _get_value(prop_dict: dict, prop_name: str) -> Any:
...
+                    elif -math.pi <= wrap_angle <= -math.pi / 2:
+                        dag = _fold_rzz._quad3(wrap_angle, fresh_qubits)
+                    else:
+                        dag = _fold_rzz._quad4(wrap_angle, fresh_qubits)
+                    windings = round((angle - wrap_angle) / (2 * math.pi))
+                    if windings % 2:
+                        dag.apply_operation_back(GlobalPhaseGate(math.pi))
+                    return dag
+
+                wrap_angles_cls.DEFAULT_REGISTRY.add_wrapper("rzz", _rzz_wrapper)
```
SamFerracin: There seems to be a standard way of modifying the registry, which is slightly different than the one you have found (the `getattr`), which is:
```python
from qiskit.transpiler.passes.utils.wrap_angles import WrapAngles

registry = WrapAngles.DEFAULT_REGISTRY
registry.add_wrapper(gate_name, my_fn)
```
`DEFAULT_REGISTRY` is a class attribute of `WrapAngles`, and it will do the correct modification for all the future uses of `WrapAngles`. The docstrings in qiskit indicate to use it this way, where it says ([here](https://github.com/Qiskit/qiskit/blob/main/qiskit/transpiler/passes/utils/wrap_angles.py#L112)) that "The preferred way to access the object is as `WrapAngles.DEFAULT_REGISTRY` and we should deprecate the old access path for Qiskit 2.4."

qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -27,15 +29,18 @@
...
 from qiskit.circuit.gate import Gate
-from qiskit.circuit.library.standard_gates import get_standard_gate_name_mapping
+from qiskit.circuit.library.standard_gates import GlobalPhaseGate, get_standard_gate_name_mapping
 from qiskit.circuit.parameter import Parameter
 from qiskit.providers.backend import QubitProperties
+from qiskit.transpiler import passes as transpiler_passes
 from qiskit.transpiler.target import InstructionProperties, Target
 
 from qiskit_ibm_runtime.models.exceptions import BackendPropertyError
 from qiskit_ibm_runtime.utils.utils import is_fractional_gate
```
SamFerracin, line 39: As discussed in private, I'd move this up here:
```suggestion
from qiskit_ibm_runtime.utils.utils import is_fractional_gate

def rzz_wrapper(angles: list[float], qubits: list[Qubit]) -> DAGCircuit:
    """A wrapper that does this and that"""
    from qiskit_ibm_runtime.transpiler.passes.basis.fold_rzz_angle import FoldRzzAngle

    fold_rzz = FoldRzzAngle()
    # ... all your code ...
    return dag

registry = WrapAngles.DEFAULT_REGISTRY
"""A registry that does this and that"""

registry.add_wrapper("rzz", rzz_wrapper)
```

SamFerracin (commented):
Thanks, Yael. Definitely a tricky PR to put together, but it makes sense!

qiskit_ibm_runtime/utils/backend_converter.py
diego-plan9: Thanks for elaborating!

SamFerracin: Actually, one more comment from me, let me unresolve. It looks like you are importing `FoldRzzAngle` just so that you can call these `_quad` functions, which are private btw. We should clean this up (and thus get rid of the import) by:

- Writing `quad1`, `quad2`, etc as stand-alone functions, hopefully with better names, and place them in a util file where they can be accessed here and in `FoldRzzAngle` without circular imports
- Going even a step further, writing all this logic (which you copy/pasted from `FoldRzzAngle._run_inner`) into a stand-alone function that can be used here and in `FoldRzzAngle._run_inner` without circular imports

I am okay with turning this comment into an issue and solving it in a follow-up PR, if this bug needs to be fixed urgently

yaelbh: I don't think the bug fix is that urgent, and the changes that you're suggesting are not big. But it's convenient to stick with small PRs, so I opened an issue https://github.com/Qiskit/qiskit-ibm-runtime/issues/2968.

yaelbh:
@SamFerracin and @diego-plan9, thanks a lot for your comments, I fixed and it's not ready for a new review.

@mtreinish, @t-imamichi, and @wshanks: I'd appreciate it if you could also take a look.

t-imamichi:
Thank you. LGTM.

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
```diff
@@ -118,18 +118,18 @@ def test_controlflow(self):
     def test_fractional_plugin(self):
         """Verify that a pass manager for a fractional backend applies the rzz folding pass."""
         circ = QuantumCircuit(2)
-        circ.rzz(7, 0, 1)
+        circ.rzz(-1, 0, 1)
```
t-imamichi: Could you test both positive case (7) and negative case (-1) just in case?

yaelbh: Please check 56fa8810

diego-plan9:
Can you add a release note?

yaelbh:
I added a release note. It was added in the wrong commit - 0bda28a instead of 4917996 - sorry about that...

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
```diff
@@ -115,21 +115,22 @@ def test_controlflow(self):
...
+    @data(-1, -1 + 4 * np.pi)
+    def test_fractional_plugin(self, rzz_angle):
         """Verify that a pass manager for a fractional backend applies the rzz folding pass."""
         circ = QuantumCircuit(2)
-        circ.rzz(7, 0, 1)
+        circ.rzz(rzz_angle, 0, 1)
 
         pm = generate_preset_pass_manager(
-            optimization_level=0,
+            optimization_level=2,
```
wshanks, line 125: I think the translation method should be updated here?

SamFerracin (dismissed):
Thanks, Yael!

qiskit_ibm_runtime/utils/backend_converter.py
```diff
@@ -49,6 +54,35 @@
...
+    if windings % 2:
+        dag.apply_operation_back(GlobalPhaseGate(math.pi))
+    return dag
+
+
+wrap_angles_registry = WrapAngles.DEFAULT_REGISTRY
+"""A registry of wrappers, to instruct the transpiler what to do when it encounters
+out-of-bounds angles."""
+
+wrap_angles_registry.add_wrapper("rzz", rzz_wrapper)
```
wshanks, line 83: I wonder where the best place to call this is. This location seems fine. The import chain is backend_converter->ibm_backend->qiskit_ibm_runtime/__init__.py, so this will always get run when you use qiskit-ibm-runtime. Likely this is the only usage of `WrapAngles`. I could imagine some other library inserting a different `rzz` wrapper and the doing the `add_wrapper` call later (like at backend creation time or the beginning of transpilation if possible) would be more robust.

release-notes/unreleased/2960.bug.rst
```diff
@@ -0,0 +1,6 @@
+Fixed the folding of out-of-bounds ``rzz`` angles during transpilation. The backend's
+target now declares ``angle_bounds`` of ``(0, pi/2)`` for the ``rzz`` gate, and
+registers an ``rzz`` wrapper with the transpiler's ``WrapAngles`` pass, so that
+``rzz`` gates with angles outside ``[0, pi/2]`` are folded into the supported range.
+This mechanism replaces the usage of the ``IBMDynamicFractionalTranslationPlugin``
+plugin and the :class:`.FoldRzzAngle` transpiler pass.
```
wshanks, line 8: I wonder if this is actually a fix? It is an upgrade that lets qiskit-ibm-runtime defer some responsibility to Qiskit. Maybe there are some edge cases where the qiskit-ibm-runtime fractional plugin was not getting run but the builtin wrapping will get run?

Should this PR start the deprecation process of any of the old path?

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
yaelbh: We test here, quite indirectly, that Takashi's bug was solved. Increasing the optimization level invokes a pass that takes the angles outside of the permitted range, and we verify that the angles get corrected.

qiskit_ibm_runtime/utils/backend_converter.py
yaelbh: We can return it to where it was before (inside `convert_to_target`, for backends with `rzz`). We can check if a wrapper for rzz (ours or someone else's - it doesn't matter) was already registered, and register only if the answer is negative. @SamFerracin What do you think?

release-notes/unreleased/2960.bug.rst
yaelbh: - It's a fix in my opinion, because there was a real bug that has been solved.
- We may deprecate `FoldRzzAngle` and the plugin. It's not mandatory but I don't see a reason to keep them around. The deprecation is not in the scope of this PR.

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
wshanks: I commented here because it was the closest line I could comment on, but I am referring to the two lines below where `translation_method="ibm_dynamic_and_fractional"` is set. With this PR, that translation method is obsolete so I think it would be better to specify `ibm_dynamic_circuits` anywhere that specifies `ibm_dynamic_and_fractional` in this file now? Or maybe translation method doesn't need to be specified now in those places?

release-notes/unreleased/2960.bug.rst
wshanks: > It's a fix in my opinion, because there was a real bug that has been solved.

From reading the release note, I can't tell what bug was fixed. Transpiling with backends loaded with `use_fractional_gates=True` already folded out-of-bounds `rzz` angles. The fix is that cases where you transpile without a backend will wrap now as well? Or there was an error with the old method?

> We may deprecate FoldRzzAngle and the plugin. It's not mandatory but I don't see a reason to keep them around. The deprecation is not in the scope of this PR.

That's fair. I wonder if anyone will think to do it later but the old path likely won't hurt anything staying in the code either.

qiskit_ibm_runtime/utils/backend_converter.py
SamFerracin: I am not too sure about the implications that this would have for other libraries. I suggest seeing if @jakelishman or someone else in the Qiskit team has an opinion. In case of doubt, I'd say we be conservative: We place it where it was before (inside `convert_to_target`, for backends with rzz), and we can always move it out later

test/unit/transpiler/passes/basis/test_fold_rzz_angle.py
yaelbh: It turns out to be the only place. I also hesitate if to specify the translation method at all, I decided quite arbitrarily to keep specifying it, but fixed it of course to `ibm_dynamic_circuits`.

release-notes/unreleased/2960.bug.rst
yaelbh: I changed a bit the wording in the release note.

> I wonder if anyone will think to do it later

I'll open an issue after this PR is merged, just in case.

qiskit_ibm_runtime/utils/backend_converter.py
jakelishman: With the caveat that I don't remember the discussions Matt had with you guys at the time, so I'm inferring from the shape of the Qiskit-side code:

Since it's a global owned by `qiskit`, I'd say the best thing to do for right now is try and make it as reliable as possible, so it happens as soon as you `import qiskit_ibm_runtime`.  Having it be a side-effect of constructing the first `Target` just means there's more time where a user could be confused by the behaviour of the transpiler changing.

This system of using a single global variable seems _not ideal_ to me looking back at the code now. The alternative is that you define your own complete translation stage always, and use an instantiation of the `WrapAngles` transpiler pass with your own owned `WrapAnglesRegistry` instead.

wshanks (approved):

qiskit_ibm_runtime/utils/backend_converter.py
wshanks: I think this is the best practical option for now. If we put it inside of a function call, it might not carry over if someone used Python multiprocessing for transpilation (like the Qiskit transpiler can do or could...maybe that has moved into Rust any way at this point). My concern is just some other package also registering a WrapAngles callback at import time -- then it is a race, but I feel like it is not too likely at the moment that another package will add an rzz angle wrapping callback.

diego-plan9: > Since it's a global owned by `qiskit`, I'd say the best thing to do for right now is try and make it as reliable as possible, so it happens as soon as you `import qiskit_ibm_runtime`.

Thanks - for what is worth, this is what the current behavior is in this PR (as Will mentioned,  top-level import brings `ibm_backend.py` in, which brings `utils/backend_converter.py` in).

SamFerracin (approved):

PR #2963, opened 2026-06-25
Add precision validation for estimator pubs to prevent non-positive v…

leonapivato:
### Summary

Specifying 0 as you precision in the pub level input on the legacy EstimatorV2 will cause the following server error: `RuntimeJobFailureError: "Unable to retrieve job result. Error code 1500;`

A guard clause should be added before server submission to throw `IBMInputValueError: 'The precision value must be strictly greater than 0.'` instead.

This is consistent to specifying 0 as your precision in the run kwarg level or default options level. 

```
qc = QuantumCircuit(2)
qc.h(0)
qc.cx(0, 1)

pm = generate_preset_pass_manager(optimization_level=1, backend=backend)
isa_circuit = pm.run(qc)
isa_obs = SparsePauliOp("ZZ").apply_layout(isa_circuit.layout)

estimator = EstimatorV2(mode=backend)

# precision = 0 on the pub -> accepted by the client, then ZeroDivisionError server-side
job = estimator.run([(isa_circuit, isa_obs, None, 0.0)])
print(job.result())  # job fails: (error 1500)
```

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/utils/validations.py
+++ b/qiskit_ibm_runtime/utils/validations.py
@@ -82,6 +82,7 @@ def validate_estimator_pubs(pubs: list[EstimatorPub]) -> None:
 
     Raises:
         IBMInputValueError: If any observable array is of size 0
+        IBMInputValueError: If any precision value is not greater than 0
     """
     for pub in pubs:
         if pub.precision is not None and pub.precision <= 0:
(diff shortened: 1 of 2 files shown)


SamFerracin (approved):
Thanks, @leonapivato !
Any other precisions (for example, the one that you can set at the level of estimator) will error if set to 0, right?

leonapivato:
> Thanks, @leonapivato ! Any other precisions (for example, the one that you can set at the level of estimator) will error if set to 0, right?

Yes. Seems like pub level was the only missing validation.

PR #2977, opened 2026-06-29
Add optional `persist` argument to `FakeBackendV2.refresh()` method

### LINKED ISSUE #2976: `refresh()` does not work on system-wide installations
**Describe the bug**
Currently, the `refresh()` method on fake backends tries to modify the config and props file in the site-packages directory. This is already not a good practice (to modify package files at runtime), but it also does not work when the user does not have permissions to do those changes.

**Steps to reproduce**
1. Install Qiskit IBM Runtime in a system path, not a local venv.
2. Import a fake backend with a matching real QPU
3. Use the `refresh()` method

Example:

```python
from qiskit_ibm_runtime.fake_provider import FakeKingston

backend = FakeKingston()
backend.refresh(service)
```

This fails with:

```console
fake_backend.refresh:WARNING:2026-06-29 22:03:42,436: The refreshing of fake_kingston has failed: [Errno 13] Permission denied: '/usr/lib/python3.14/site-packages/qiskit_ibm_runtime/fake_provider/backends/kingston/conf_kingston.json'
```

**Expected behavior**
It should work

**Suggested solutions**
Qiskit IBM Runtime already uses `~/.qiskit` directory to store admin data. Let's use this directory to cache data from backends.

**Additional Information**

- **qiskit-ibm-runtime version**: 0.47.0
- **Python version**: 3.14.6

iyanmv:
### Summary

`FakeBackendV2.refresh()` writes the updated `conf_*.json` / `props_*.json` back into the package's own directory. When the package is installed system-wide that directory is read-only (`PermissionError`), and even when writable it's bad practice to mutate files inside site-packages.

~Instead, let's keep the .json files in the `~/.qiskit` directory which is already used for storing the IBM Quantum platform credentials.~

As suggested [here](https://github.com/Qiskit/qiskit-ibm-runtime/issues/2976#issuecomment-4852615378), the new proposal leaves the default behavior unchanged, i.e., `refresh()` tries to modify the backend json files in site-packages. With ~`use_temp_dir=True`~ `persist=False`, the updated json files are stored in a temp directory. This allows to use this method in installations where the package's files are read-only or the user does not have permissions to modify those files.

### Details and comments

- I tested this change locally and in the Arch Linux package and I didn't have any issues.
- I used `os.path` instead of `pathlib.Path` because this was used elsewhere. If you want I can refactor this using `pathlib`

Fixes #2976

- [x] I used the following tool to generate or modify code: Claude Opus 4.8 for writing the unit tests

### DIFF
--- a/qiskit_ibm_runtime/fake_provider/fake_backend.py
+++ b/qiskit_ibm_runtime/fake_provider/fake_backend.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import tempfile
 import warnings
 from typing import TYPE_CHECKING, Any
 
@@ -43,11 +44,6 @@ if TYPE_CHECKING:
     from ..qiskit_runtime_service import QiskitRuntimeService
 
 logger = logging.getLogger(__name__)
-
-# User-writable directory where refreshed fake backend data is cached. Updates are
-# stored here instead of inside the package installation (e.g. ``site-packages``),
-# which is often read-only and should not be modified at runtime.
-_LOCAL_DATA_DIR = os.path.join(os.path.expanduser("~"), ".qiskit", "fake_provider")
 
 
 class FakeBackendV2(BackendV2):
@@ -121,23 +117,8 @@ class FakeBackendV2(BackendV2):
         supported_features = self._conf_dict.get("supported_features") or []
         return "qasm3" in supported_features
 
-    def _local_data_dir(self) -> str:
-        """Return the user-writable directory where refreshed data for this backend is cached."""
-        return os.path.join(_LOCAL_DATA_DIR, self.backend_name)  # type: ignore[arg-type]
-
-    def _resolve_data_path(self, filename: str) -> str:
-        """Return the path to a data file.
-
-        A locally cached copy (written by :meth:`refresh`) in the user's home directory takes
-        precedence over the copy bundled with the package.
-        """
-        local_path = os.path.join(self._local_data_dir(), filename)
-        if os.path.exists(local_path):
-            return local_path
-        return os.path.join(self.dirname, filename)  # type: ignore[arg-type]
-
     def _load_json(self, filename: str) -> dict:
-        with open(self._resolve_data_path(filename)) as f_json:
+        with open(os.path.join(self.dirname, filename)) as f_json:  # type: ignore[arg-type]
             the_json = json.load(f_json)
         return the_json
 
@@ -389,25 +370,40 @@ class FakeBackendV2(BackendV2):
 
         return noise_model
 
-    def refresh(self, service: QiskitRuntimeService, use_fractional_gates: bool = False) -> None:
+    def refresh(
+        self,
+        service: QiskitRuntimeService,
+        use_fractional_gates: bool = False,
+        use_temp_dir: bool = False,
+    ) -> None:
         """Update the data files from its real counterpart.
 
-        This method pulls the latest backend data from its real counterpart and writes it to a
-        user-writable cache directory:
+        This method pulls the latest backend data files from their real counterpart and,
+        by default (``use_temp_dir=False``), overwrites the corresponding files in the local
+        installation:
 
-        *  ``~/.qiskit/fake_provider/{backend_name}/conf_{backend_name}.json``
-        *  ``~/.qiskit/fake_provider/{backend_name}/props_{backend_name}.json``
+        *  ``../fake_provider/backends/{backend_name}/conf_{backend_name}.json``
+        *  ``../fake_provider/backends/{backend_name}/props_{backend_name}.json``
 
-        The cached files take precedence over the copies bundled with the package and persist
-        through sessions, so the backend stays updated until the cache is manually removed.
-        Writing to ``~/.qiskit`` rather than the package installation means refreshing works even
-        when ``qiskit-ibm-runtime`` is installed in a read-only location (e.g. ``site-packages``)
-        and avoids modifying the installed package itself.
+        The new data files will persist through sessions so the files will stay updated unless they
+        are manually reverted locally or when ``qiskit-ibm-runtime`` is upgraded or reinstalled.
+
+        Overwriting the installed data files fails, however, when the package is installed in a
+        read-only location (e.g. a system-wide ``site-packages`` directory). In that case, or
+        whenever modifying the installed package is undesirable, pass ``use_temp_dir=True`` to write
+        the refreshed data to a temporary directory instead. The backend is still updated for the
+        current session, but the changes are not persisted to disk and are lost once the backend is
+        garbage collected or the process exits.
 
         Args:
             service: A :class:`QiskitRuntimeService` instance
             use_fractional_gates: Set True to allow for the backends to include
                 fractional gates.
+            use_temp_dir: If ``False`` (default), the refreshed data files overwrite the copies
+                bundled with the installed package so the update persists across sessions. If
+                ``True``, the data is written to a temporary directory instead, which allows
+                :meth:`refresh` to succeed even when the package is installed in a read-only
+                location, at the cost of the update not persisting between sessions.
 
         Raises:
             ValueError: if the provided service is a non-QiskitRuntimeService instance.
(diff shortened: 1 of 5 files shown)


diego-plan9 (changes requested):
Thanks for catching this, and for the preliminary work: please check #2976 for discussion on an alternative solution.

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -415,13 +433,26 @@ def refresh(self, service: QiskitRuntimeService, use_fractional_gates: bool = Fa
...
+                target_dir = self.dirname
+            else:
+                # Write to a temporary directory so that refresh() succeeds even when the
+                # package is installed in a read-only location. Keep a reference to the
+                # directory on the instance so it lives as long as the backend and is cleaned
+                # up automatically when the backend is garbage collected. Point ``dirname`` at
+                # it so subsequent data loads use the refreshed files.
+                self._tmp_data_dir = tempfile.TemporaryDirectory(  # pylint: disable=attribute-defined-outside-init
+                    prefix="qiskit_fake_backend_"
+                )
```
diego-plan9: Can you:
* define the attribute either as part of `__init__` or as part of the class attributes? It helps being able to inspect at a glance what attributes an object has without peeking at internals. Slight preference from me to using a type hint at class-level without initializing, but anything works as long as the variable is easier to find.
* trim down the comments a bit, as their current contents seems quite well covered by the docstring updates (and comments are also something that needs to be maintained too). The bit about keeping the reference is a good piece of information - perhaps as part of a comment in the place where it ends up being declared?

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -415,13 +433,26 @@ def refresh(self, service: QiskitRuntimeService, use_fractional_gates: bool = Fa
...
+                target_dir = self.dirname = self._tmp_data_dir.name
```
diego-plan9: I like updating `self.dirname` as it helps keeping consistency (ie. the functions that have a `refresh` arg use the new data files). On the other hand, by replacing it, a second call to `refresh()` without `use_temp_dir` would not have the intended effect - a bit of an edge case, but would be nice to preserve that functionality while we are modifying this.

What about:
* a new attribute defined and typed at the class or init level (so the `pylint` ignore is not needed - we want to be able to discover attributes easily), storing the temporary directory?
* a tweak to `_load_json()` so, if that attribute has a value, the name of the dir is used instead of `self.dirname`?

An alternative is avoiding to update `self.dirname` and making it explicit in the docstrings which attributes are updated as the result of the call, and that the functions that accept `refresh` will use the non-refreshed files (which could be argued to be reasonable behavior, but will likely leave `_target` inconsistent).

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -415,13 +433,26 @@ def refresh(self, service: QiskitRuntimeService, use_fractional_gates: bool = Fa
             updated_config = real_config.to_dict()
             updated_config["backend_name"] = self.backend_name
 
+            if not use_temp_dir:
+                target_dir = self.dirname
```
diego-plan9: Small nit: can you avoid `target` in this name, to minimize confusion with the backend target? Perhaps `snapshots_dir`, `data_dir`?

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -369,23 +370,40 @@ def _get_noise_model_from_backend_v2(  # type: ignore
 
         return noise_model
 
-    def refresh(self, service: QiskitRuntimeService, use_fractional_gates: bool = False) -> None:
+    def refresh(
+        self,
+        service: QiskitRuntimeService,
+        use_fractional_gates: bool = False,
+        use_temp_dir: bool = False,
```
diego-plan9: Can you tweak so it conveys functionality rather than implementation? Perhaps `persist: bool = True`?

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -117,7 +118,7 @@ def _supports_dynamic_circuits(self) -> bool:
         return "qasm3" in supported_features
 
     def _load_json(self, filename: str) -> dict:
-        with open(os.path.join(self.dirname, filename)) as f_json:
+        with open(os.path.join(self.dirname, filename)) as f_json:  # type: ignore[arg-type]
```
diego-plan9: This (and the other `# type: ignore[arg-type]` introduced) are not needed with the current `mypy` configuration, right? Even if they are actually emitted by other tools (rightfully, due to the `None`).

We are aiming to reduce the amount of selective ignores (via a long-running issue https://github.com/Qiskit/qiskit-ibm-runtime/issues/2514). Can you either:
* remove them, so we don't increase the amount (and they will surface at some point in time when we tune the `mypy` config)?
* preferably, tweak `dirname` so it becomes a type hint with just `str` rather than initialized to `None`? A bit of a trade-off, but it should be safe to do so (as in practice, all subclasses initialize it to a `str`).

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -369,23 +370,40 @@ def _get_noise_model_from_backend_v2(  # type: ignore
...
         """Update the data files from its real counterpart.
 
-        This method pulls the latest backend data files from their real counterpart and
-        overwrites the corresponding files in the local installation:
+        This method pulls the latest backend data files from their real counterpart and,
+        by default (``use_temp_dir=False``), overwrites the corresponding files in the local
+        installation:
 
         *  ``../fake_provider/backends/{backend_name}/conf_{backend_name}.json``
-        *  ``../fake_provider/backends/{backend_name}/defs_{backend_name}.json``
```
diego-plan9, line 379: Good catch!

qiskit_ibm_runtime/fake_provider/fake_backend.py
iyanmv: Done in 1588ef98

qiskit_ibm_runtime/fake_provider/fake_backend.py
iyanmv: Done in d5b2cd99

qiskit_ibm_runtime/fake_provider/fake_backend.py
iyanmv: Done in 7623de26

qiskit_ibm_runtime/fake_provider/fake_backend.py
iyanmv: And done in 3b537069

qiskit_ibm_runtime/fake_provider/fake_backend.py
iyanmv: Yeah, this was just fresh from https://github.com/Qiskit/qiskit-ibm-runtime/pull/2711 ;) (I was confused about what defs was)

iyanmv:
I think this is ready for a second review. Thanks for all the feedback!

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -49,10 +50,15 @@ class FakeBackendV2(BackendV2):
...
-    backend_name: str | None = None
+    dirname: str
+    conf_filename: str
+    props_filename: str
+    backend_name: str
+
+    # When ``refresh(persist=False)`` is called, the refreshed snapshots are written to this
+    # temporary directory instead of ``dirname``. While it is set, snapshot data is read from here,
+    # so ``dirname`` is left untouched and keeps pointing at the files bundled with the package.
+    _tmp_data_dir: tempfile.TemporaryDirectory | None = None
```
diego-plan9, line 64:
```suggestion
    dirname: str
    """Directory for the real backend snapshots."""
    
    conf_filename: str
    """Filename for the backend configuration."""

    props_filename: str
    """Filename for the backend properties."""

    backend_name: str
    """Name of the backend."""

    _tmp_data_dir: tempfile.TemporaryDirectory | None = None
    """Temporary directory for the real backend snapshots, if using ``refresh(persist=False)``.
    
    If set, snapshot data is read from this temporary directory instead of ``dirname``. This
    attribute stores a reference to ensure the temporary directory is bound to the lifetime of
    the instance (cleaned up automatically when the backend is garbage collected).
    """
```

Please also drop the `# directory and file names for real backend snapshots.` comment above (GitHub interface would not allow me to make a suggestion including that line!).

This is mostly for:
* taking advantage of you modifying it, so we end up with nice docstring for all these attributes.
* explicitly describing the intent of the attribute (and the need for it to contain a reference so we resist the temptation to inadvertently turn it into a `str` in the future)

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -415,13 +439,33 @@ def refresh(self, service: QiskitRuntimeService, use_fractional_gates: bool = Fa
             updated_config = real_config.to_dict()
             updated_config["backend_name"] = self.backend_name
 
+            # Explicitly clean up and drop any temporary directory left over from a previous
+            # ``persist=False`` call so it is removed now rather than being left to garbage
+            # collection, and ``_load_json`` reads the files written by this call.
```
diego-plan9:
```suggestion
            # Clean up and drop any temporary directory left over from a previous call.
```

qiskit_ibm_runtime/fake_provider/fake_backend.py
```diff
@@ -415,13 +439,33 @@ def refresh(self, service: QiskitRuntimeService, use_fractional_gates: bool = Fa
...
+            if persist:
+                # Persist to (and read back from) the bundled ``dirname``.
+                snapshots_dir = self.dirname
+            else:
+                # Write to a temporary directory so that refresh() succeeds even when the
+                # package is installed in a read-only location. Keep a reference to the
+                # directory on the instance so it lives as long as the backend and is cleaned
+                # up automatically when the backend is garbage collected. ``_load_json`` reads
+                # from ``_tmp_data_dir`` while it is set, so ``dirname`` is left unchanged and a
+                # later ``refresh()`` with ``persist=True`` still targets the bundled files.
```
diego-plan9:
```suggestion
                # Use a temporary directory instead of persisting to ``dirname``.
```

diego-plan9 (commented):
Thanks for the updates! Just some notes related to comments for achieving a better balance: ultimately, comments need to maintained as well, and as docstring changes and the code is quite explanatory already, by trimming them a bit we reduce the risk of them becoming stale.

qiskit_ibm_runtime/fake_provider/fake_backend.py
iyanmv: Done in a34975fb

qiskit_ibm_runtime/fake_provider/fake_backend.py
iyanmv: Done in cb11f087

qiskit_ibm_runtime/fake_provider/fake_backend.py
iyanmv: Done also in cb11f087

diego-plan9 (approved):
Thanks!

PR #2982, opened 2026-07-01
Account for measure mitigation in `num_randomization` and `shots_per_randomization` calculation

### LINKED ISSUE #2981: When measure mitigation is on, twirling is forced, but twirling properties calculation ignores it
**Describe the bug**
When measure mitigation is on, measurements are always twirled ([here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/005f9450790217174f581bcbb664138fe275def4/qiskit_ibm_runtime/executor_estimator/utils.py#L247-L249)). This matches legacy behavior. However, when we calculate `num_randomizations` and `shots_per_randomization` [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/005f9450790217174f581bcbb664138fe275def4/qiskit_ibm_runtime/executor_estimator/prepare.py#L72) we don't account for measure mitigation, and could mistakenly take the non-twirled branch.

TsafrirA:
### Summary
Closes #2981 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/prepare.py
+++ b/qiskit_ibm_runtime/executor_estimator/prepare.py
@@ -69,7 +69,11 @@ def prepare(
             if a circuit contains mid-circuit measurements, or if a circuit already uses the
             reserved classical register name ``_meas``.
     """
-    if twirling_options.enable_gates or twirling_options.enable_measure:
+    if (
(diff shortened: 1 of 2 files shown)


TsafrirA:
Issue was resolved via #3014 with an alternative solution.

PR #2983, opened 2026-07-01
Keep .utils.json (as deprecated, with no imports from .utils)

diego-plan9:
### Summary

Even if `RuntimeDecoder` and `RuntimeEncoder` are the only public items from the `json` module (via the top-level imports, which remained unchanged in #2894), reintroduce the `qiskit_ibm_runtime.utils.json` module with all their private members as a convenience for users that might be consuming private members, easing the transition.

### Details and comments

Related to #2749, #2920

This does not provide full compatibility, as the

```python
from .json import RuntimeEncoder, RuntimeDecoder, to_base64_string
```

from `qiskit_ibm_runtime/utils/__init__.py` is not included (on purpose, as it would defeat the goal of reducing import cycles).

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/utils/json.py
+++ b/qiskit_ibm_runtime/utils/json.py
@@ -1,0 +1,39 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2024-2026.
+#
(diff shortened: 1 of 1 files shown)


SamFerracin (approved):
Thank you!

PR #2985, opened 2026-07-01
Enhance DataBin construction in estimator_v2_post_processor to includ…

### LINKED ISSUE #2984: Wrapper EstimatorV2: `DataBin.shape` is `()` for broadcast pubs
The wrapper built each result `DataBin` without a `shape=`, so `DataBin.shape` stayed `()`
even when the pub broadcast to something bigger (e.g. an observable list). The `evs`/`stds`
arrays were fine — only the bin's declared shape was wrong, so it's metadata-only.

Legacy sets `shape` to the broadcast shape; the wrapper's sampler decoder already did too, so
this was just missed in the estimator post-processor
(`decoders/executor_estimator/post_processor_v0_1.py`).

Fix — pass the shape through (`exp_vals` already has it):

```python
data_bin = DataBin(
    evs=exp_vals, stds=stds, ensemble_standard_error=ensemble_stds, shape=exp_vals.shape
)
```

leonapivato:
Fixes #2984 

### AI/LLM disclosure


SamFerracin (approved):
Thank you!

PR #2991, opened 2026-07-02
Improve handling of options in executor-based Estimator and Sampler

### LINKED ISSUE #2989: Improve handling of options in executor-based Estimator
Upon initialization, the executor-based estimator accepts options of type `None | dict | EstimatorOptions`. Behind-the-scenes coercion to `EstimatorOptions` is operated via `__setattr__`. This approach is more complex than necessary, causes issues with mypy, and unintentionally allows users to assign to `estimator.options`, which is not supported by the legacy estimator.

**Acceptance criteria**
- [ ] Coerce in the init via if/elif
- [ ] Expose getter (`@property`) but not setter

### LINKED ISSUE #2990: Improve handling of options in executor-based Sampler
Upon initialization, the executor-based estimator accepts options of type `None | dict | EstimatorOptions`. Behind-the-scenes coercion to `EstimatorOptions` is operated via `__setattr__`. This approach is more complex than necessary, causes issues with mypy, and unintentionally allows users to assign to `estimator.options`, which is not supported by the legacy estimator.

**Acceptance criteria**
- [ ] Coerce in the init via if/elif
- [ ] Expose getter (`@property`) but not setter

SamFerracin:
### Summary
Upon initialization, the executor-based estimator and sampler accept options of type `None | dict | EstimatorOptions`. Behind-the-scenes coercion to `EstimatorOptions`/`SamplerOptions` is operated via `__setattr__`. This approach is more complex than necessary, causes issues with mypy, and unintentionally allows users to assign to `estimator.options`/`sampler.options`, which is not supported by the legacy estimator.

Closes #2989
Closes #2990

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import asdict
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 import numpy as np
 from qiskit.primitives.base import BaseEstimatorV2
@@ -91,8 +91,7 @@ class EstimatorV2(BaseEstimatorV2):
             for all available options.
     """
 
-    options: EstimatorOptions
-    """The options of this Estimator."""
+    _options_class: EstimatorOptions
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


diego-plan9 (commented):
Thanks - this would resurface #2719 (the combination of `@property` and the usage of dataclasses seems to cause `ipython` and `jupyter`, as it uses under the hood, to not be able to auto-complete second-level fields that are dataclasses themselves): 

Image

Image

More details at the issue and at #2739.

> causes issues with `mypy`

Can you elaborate on these?

Side-note: we do have some typing issues in _legacy_ options (see [test/integration/test_estimator_v2.py](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/test/integration/test_estimator_v2.py#L74-L97) in your IDE, with the `PyLance` warnings. They don't surface in `mypy` and seem not strictly related to `@property`, but instead due to the complexity when it comes to options (subclasses, decorators, etc). Still, I see them as an argument for minimizing complexity when it comes to exposing options to users (ie. the closer to being a boring regular attribute, the better), and streamlining those classes and their usage.

> This approach is more complex than necessary, [...],  and unintentionally allows users to assign to estimator.options/sampler.options, which is not supported by the legacy estimator.

From a different perspective, functionality-wise it brings the wrappers closer to `Executor`, which does allows assigning the settings (since #2616). It is not backward incompatible but an additional feature - and from my perspective a good one (we were allowing users to mutate them, and I cannot think of a good reason for preventing them to set/replace them, as we don't depend on the specific pointer to the object or treat them differently than a regular attribute).

Complexity-wise - I actually see using `@property` as an extra indirection level that does not buy us much in this case (except for the prevention of replacing the object: which I see as an attempt of making it read-only, but quite diluted as they can be modified), and brings complexity (which seems to be hitting us already for the auto-completion bug). In `main`, the complexity is isolated into `__setattr__`, and `options` remains a regular well typed attribute that is easier to use (by users and devs). It's a trade-off with a bit of magic, but hopefully

diego-plan9:
An alternative for the future is also considering just having the user always pass a `SamplerOptions` instance at creation time. Longer discussion as this would be a breaking change and has usability trade-offs, but since `Sampler` and `Executor` always require their corresponding options, it would be a natural pattern and trim some complexity.

SamFerracin:
Closing after private conversation with @diego-plan9

PR #2992, opened 2026-07-02
Deprecate specifying different precision and shots across pubs in Est…

### LINKED ISSUE #2987: Deprecate pub-level shots / precision on the legacy primitives
Per-pub `shots` (sampler) and `precision` (estimator) should be deprecated on the
legacy primitives, in favor of the run-level / options value.

## Why

The wrapper primitives turns a batch of pubs into a single `QuantumProgram`, and `QuantumProgram.shots`
is **one global value** — there's no per-item shots in the class or the wire schema. So the
wrapper can't honor different shots/precision per pub. Today it papers over this by requiring
every pub to agree and raising if they don't:

- Sampler — `executor_sampler/utils.py::extract_shots_from_pubs`:
  > "All pubs must have the same number of shots."
- Estimator — `executor_estimator/utils.py::resolve_precision`:
  > "All pubs must have the same precision."

So per-pub shots/precision looks supported (it's part of the pub tuple) but is really only
accepted when it collapses to a single value. That's a footgun: users think they can vary it
per pub and instead get a hard error. Legacy primitives *do* allow per-pub values, so this is
also a wrapper-vs-legacy divergence.

To prepare for this change it is best to deprecate this behavior for the legacy primitives. 

## Deprecation

- Keep accepting a single, uniform value (back-compat) but emit a `DeprecationWarning`
  pointing at `run(..., shots=)` / `run(..., precision=)`.
- Document that the wrapper uses one global shot budget per submission, so per-pub values
  aren't meaningful.
- Keep the "must all agree" error for now; the warning just makes the intent explicit ahead
  of removal.

leonapivato:
Resolves #2987 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -153,7 +153,7 @@ class EstimatorV2(BasePrimitiveV2[EstimatorOptions], Estimator, BaseEstimatorV2)
             issue_deprecation_msg(
                 msg="Specifying different 'precision' across pubs is deprecated",
                 version="0.48.0",
-                remedy="Submit one job for each desired shot count instead. "
+                remedy="Submit one job for each desired precision instead. "
                 "To reduce overhead, Consider submitting these jobs inside a Batch execution mode.",
                 stacklevel=2,
             )
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -20,6 +20,8 @@ from typing import TYPE_CHECKING
 from qiskit.primitives.base import BaseEstimatorV2
 from qiskit.primitives.containers.estimator_pub import EstimatorPub
 
+from qiskit_ibm_runtime.utils.deprecation import issue_deprecation_msg
+
 from .base_primitive import BasePrimitiveV2
 from .options.estimator_options import EstimatorOptions
 from .utils import validate_estimator_pubs
@@ -146,6 +148,17 @@ class EstimatorV2(BasePrimitiveV2[EstimatorOptions], Estimator, BaseEstimatorV2)
             if precision <= 0:
                 raise ValueError("The precision value must be strictly greater than 0.")
         coerced_pubs = [EstimatorPub.coerce(pub, precision) for pub in pubs]
+
+        if len({pub.precision for pub in coerced_pubs}) > 1:
+            issue_deprecation_msg(
+                msg="Specifying different 'precision' across pubs is deprecated",
+                version="0.48.0",
+                remedy="Use a single run-level value instead, e.g. estimator.run(pubs, precision=...). "  # noqa: E501
+                "The executor-backed primitives use a single global shot budget per submission, "
+                "so precision that varies between pubs is not supported.",
+                stacklevel=2,
+            )
+
         validate_estimator_pubs(coerced_pubs)
         return self._run(coerced_pubs)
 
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/estimator.py
```diff
@@ -146,6 +148,17 @@ def run(
             if precision <= 0:
                 raise ValueError("The precision value must be strictly greater than 0.")
         coerced_pubs = [EstimatorPub.coerce(pub, precision) for pub in pubs]
+
+        if len({pub.precision for pub in coerced_pubs}) > 1:
+            issue_deprecation_msg(
+                msg="Specifying different 'precision' across pubs is deprecated",
+                version="0.48.0",
+                remedy="Use a single run-level value instead, e.g. estimator.run(pubs, precision=...). "  # noqa: E501
```
SamFerracin:
```suggestion
                remedy="Use a single run-level value instead, e.g. ``estimator.run(pubs, precision=...)``. "  # noqa: E501
```

SamFerracin: what's noqa doing here?

SamFerracin:
Thanks! Looks good. Could you add a few tests to make sure that we're only raising warnings in cases of interest?

qiskit_ibm_runtime/estimator.py
leonapivato: line is couple characters too long. I can break it up.

SamFerracin: yes please

jyu00: Instead of telling people to use a single run-level value, we should tell them to submit different jobs, which is the only way to have different precisions - e.g.

`
Consider submit one job for each desired precision inside a Batch execution mode
`

leonapivato:
Added tests for cases of interest. Let me know @SamFerracin if you see any adjustments that should be made for them.

And for the deprecation message, @jyu00 if you could confirm the message is good.

test/unit/test_estimator.py
```diff
@@ -300,3 +302,89 @@ def test_gate_not_in_target(self):
...
+        """Conflicting pub-level precision emits one DeprecationWarning."""
+        backend = get_mocked_backend()
+        circ = QuantumCircuit(1)
+        t_circ = transpile(circ, backend=backend)
+        observable = remap_observables("Z", t_circ)
+        inst = EstimatorV2(mode=backend)
+
+        warning_msg = "Specifying different 'precision' across pubs is deprecated"
+
+        def matching_warnings(call):
```
leonapivato: Might be worth abstracting this helper into `self.assertDeprecationAppears(msg, num_times)` into the `IBMTestCase` class. I'll likely need to reuse it for the other deprecation PR

leonapivato: @SamFerracin what do you think?

qiskit_ibm_runtime/estimator.py
```diff
@@ -146,6 +148,16 @@ def run(
             if precision <= 0:
                 raise ValueError("The precision value must be strictly greater than 0.")
         coerced_pubs = [EstimatorPub.coerce(pub, precision) for pub in pubs]
+
+        if len({pub.precision for pub in coerced_pubs}) > 1:
+            issue_deprecation_msg(
+                msg="Specifying different 'precision' across pubs is deprecated",
+                version="0.48.0",
+                remedy="Submit one job for each desired shot count instead. "
```
jyu00:
```suggestion
                remedy="Submit one job for each desired precision instead. "
```

jyu00 (commented):
> And for the deprecation message, @jyu00 if you could confirm the message is good.


The message looks good (besides the 1 comment)

SamFerracin (approved):
Thanks!

PR #2994, opened 2026-07-02
Add convenience function to executor-based estimator to identify unique layers to learn

SamFerracin:
### Summary

This PR adds a function to the executor-based estimator to identify unique layers to learn, that can be passed straight to NLV3. It also flattens the output of the convenience function, to tailor it to the input required by NLV3.

Fixes #2993

### AI/LLM disclosure

- [x] I didn't use LLM tooling, or only used it privately.
- [ ] I used the following tool to help write this PR description:
- [ ] I used the following tool to generate or modify code:

PR #2939, opened 2026-06-19
[WIP] Exploring using tags to enable aer simulations

SamFerracin:
### Summary

TBD

### AI/LLM disclosure

- [x] I didn't use LLM tooling, or only used it privately.
- [ ] I used the following tool to help write this PR description:
- [ ] I used the following tool to generate or modify code:

PR #2998, opened 2026-07-06
Allow adding `add_tags` argument to the boxing pass manager used by the executor-based estimator

dekelmeirom:
### Summary
Allow adding `add_tags` argument to the boxing pass manager to allow additional debugging functionalities.

### Details and comments

updated version to the PR #2939.
This PR relies on #2994 on the changes to the `get_layers` function (the tests will pass after merging this PR)

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/prepare.py
+++ b/qiskit_ibm_runtime/executor_estimator/prepare.py
@@ -47,6 +47,7 @@ def prepare(
     twirling_options: TwirlingOptions,
     shots: int,
     measure_noise_learning: MeasureNoiseLearningOptions | None = None,
+    add_tags: bool = False,
 ) -> QuantumProgram:
     """Convert estimator PUBs to a quantum program.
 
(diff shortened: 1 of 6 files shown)


qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -47,6 +47,7 @@ def prepare(
     twirling_options: TwirlingOptions,
     shots: int,
     measure_noise_learning: MeasureNoiseLearningOptions | None = None,
+    add_tags: bool = False,
```
SamFerracin, line 50: Do you want to allow this instead?
```suggestion
    add_tags: Literal[<what is currently supported by Samplomatic>] ="none",
```
I'm thinking about users that will want to use these functions directly, as opposed to using the wrapper. They may want to specify different values

Alternatively, I'd add an explanation in the docstring to highlight what True and False mean (we set it to "none" or "inject noise")

SamFerracin (commented):
Thanks! 

Could you also add a test to the suite for `box_circuit`?

qiskit_ibm_runtime/executor_estimator/prepare.py
dekelmeirom: Initially I wanted to import from Samplomatic the options that are supported, but it seems that there is no global variable to import, meaning that if the options will change, we will need to manually change them. Might not be too bad, as I guess it will not change frequently, but something to consider.
I will add that Literal and some explanation in the docstring, and if we will think that the manual changing to the Literal in case of a change in Samplomatic is problematic, I will remove this.

TsafrirA (commented):
Looks good, but I'll leave it for @SamFerracin to approve because he had some questions.

SamFerracin (approved):
LGTM!

Issue #2863, opened 2026-06-01
Make balanced sampling the default for TREX


dekelmeirom:
**Acceptance criteria**
`Samplomatic` allows controlling the `twirling_group`. The default currently used is `pauli`.
Using `pauli_balanced` showed better results, at least when using TREX.

- [ ] We should check whether it should be the global default - if it does, just it to be the default when Twirling, if not, change it to the default only when using TREX ([balanced Pauli sampling](https://github.com/Qiskit/samplomatic/blob/b278e7e2fdd2485e793cd095223421d176f067e2/samplomatic/distributions/balanced_uniform_pauli.py#L23-L42)).
- [ ] Consider adding an additional option allow the user to control it
- [ ] Consider adding deterministic sampling, as suggested in Bravyi et al, Phys. Rev. A 103, 042605 (2021)

TsafrirA:
Closed in favor of qiskit-mitigation [24](https://github.com/Qiskit/qiskit-mitigation/issues/24).

PR #2999, opened 2026-07-06
change twirling group default

dekelmeirom:
### Summary

Change the twirling group to be "balanced_pauli" instead of "pauli" for better sampling of the twirling.

### Details and comments

Fixes the first bullet in #2863.

### AI/LLM disclosure


TsafrirA (commented):
This looks good, but I wonder if the default should be defined directly in the util function instead of the kwargs generating function? Would make it more explicit.

dekelmeirom:
> This looks good, but I wonder if the default should be defined directly in the util function instead of the kwargs generating function? Would make it more explicit.

I think it is better to be in the function that we use to define the variables dict (otherwise, I think that `options_to_boxing_pm_kwargs` should not allow setting this option at all).
But of course if more agree with your point, I can easily change that :)

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -190,6 +191,7 @@ def box_circuit(
             argument of
             :meth:`~samplomatic.transpiler.generate_boxing_pass_manager`. See the Samplomatic
             API docs for a full list of supported values.
+        twirling_group: The group to use for the twirling boxes.
```
SamFerracin, line 198:
```suggestion
        twirling_group: The group to use for the twirling boxes.
            Check the :meth:`~.samplomatic.transpiler.generate_boxing_pass_manager` documentation for supported values.
```

SamFerracin (approved):
Thanks!

PR #3000, opened 2026-07-06
Change the ZNE extrapolation I/O to align with the flow of the EM methods

dekelmeirom:
### Summary

Change the extrapolation input and output to align with the flow of the EM methods.

### Details and comments

Related to #2948 and #2953 

### AI/LLM disclosure

- [X] I used the following tool to generate or modify code: bob


qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -66,12 +70,14 @@ def process_extrapolated_expectation_values(
...
 
     Args:
-        result: Estimator result. Each entry's ``values`` is a 1D array of expectation values,
-            one per noise factor, aligned with the ``standard_error`` and
-            ``resilience["zne_noise_factors"]`` arrays in its metadata. An optional ``ev_basis``
-            string sets the ideal-value range used to judge extrapolation validity; an optional
-            ``ensemble_standard_error`` array is carried through to the output.
-        extrapolator: A builtin model name, or a sequence of names tried in priority order.
+        exp_vals: Raw expectation value result. Each element of measured observable and parameter
+            is a 1D array of expectation values, one per noise factor.
```
SamFerracin: To make sure I understand correctly, "a 1D array of expectation values" = "a 1D array of np.arrays of arbitrary shape," where the arbitrary shape is what one would expect from `pub.shape`. Is this correct?

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -208,82 +224,80 @@ def clamp_degenerate_stds(y_std: np.ndarray) -> np.ndarray | None:
 
 
 def select_zne_extrapolated_result(
-    zne_values: np.ndarray, zne_metadata: dict[str, Any]
-) -> tuple[np.ndarray, dict[str, Any]]:
-    """Choose the best extrapolated values and stack them in the top row of the values/metadata.
+    zne_values: np.ndarray,
+    zne_std_errors: np.ndarray,
```
SamFerracin: Should these also be type-hinted as below?
```suggestion
    zne_values: npt.NDArray[float],
    zne_std_errors: npt.NDArray[float],
```

SamFerracin:
Looks good! Just a couple of small questions, but looks good to go in to me!

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
dekelmeirom: yes, you are correct. I tried to explain this in the docstring by saying that "Each element of measured observable and parameter" (these elements form a np.arrays of `pub.shape`). I changed the docs to be more clear about it.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -41,20 +42,23 @@
...
-    extrapolator: str | Sequence[str],
-    extrapolated_noise_factors: float | ArrayLike = 0,
-) -> EstimatorResult:
-    r"""Apply zero-noise extrapolation (ZNE) to an estimator result.
+    exp_vals: npt.NDArray[float],
+    standard_errors: npt.NDArray[float],
+    observables: ObservablesArray,
+    zne_noise_factors: Sequence[float],
+    extrapolators: str | Sequence[str],
+    extrapolated_noise_factors: float | npt.ArrayLike = 0,
```
TsafrirA:
```suggestion
    extrapolated_noise_factors: float | npt.ArrayLike = 0.0,
```

will fail to convert to a list otherwise.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -208,82 +226,80 @@ def clamp_degenerate_stds(y_std: np.ndarray) -> np.ndarray | None:
...
-    if re.search(_pattern_ylim_01, basis):
-        val_min, val_max = (0, 1)
-    elif re.search(_pattern_ylim_pm1, basis):
-        val_min, val_max = (-1, 1)
-    else:
-        val_min, val_max = (-np.inf, np.inf)
+    # Pauli in the basis term we assume ideal <B> in [-1, 1], for only projectors [0, 1].
+    # For missing or non-standard basis don't constrain values
+    val_min, val_max = (-np.inf, np.inf)
+    for observable_term in observable.keys():
```
TsafrirA: Shouldn't we account for weights here? And the number of terms?
For example, IX+YI ranges from [-2,2] and not [-1,1]
And 0.1*IX +10YI ranges from [-10.1, 10.1]

What am I missing here?

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -208,82 +226,80 @@ def clamp_degenerate_stds(y_std: np.ndarray) -> np.ndarray | None:
...
+        if re.search(_pattern_ylim_01, observable_term):
+            val_min, val_max = (0, 1)
+        elif re.search(_pattern_ylim_pm1, observable_term):
+            val_min, val_max = (-1, 1)
+            break
```
TsafrirA: Not crucial, but it's more efficient to have the breaking branch first.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -473,24 +453,6 @@ def poly_degree(name: str) -> int | None:
...
-                0
-            ]
-        res = fit_metadata["resilience"]
-        for field in ["zne_noise_factors", "zne_extrapolator"]:
-            if field in res:
-                res[field] = res[field].flat[0]
-    return fit_values, fit_metadata
-
-
 def copy_metadata(metadata: dict[str, Any]) -> dict[str, Any]:
```
TsafrirA: I think this function is no longer called anywhere.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -83,69 +91,89 @@ def process_extrapolated_expectation_values(
...
+            extrapolated_stderr,
+            observable,
+            extrapolators,
         )
-        fit_values, fit_metadata = format_extrapolated(fit_values, fit_metadata)
 
-        result_values.append(fit_values)
-        result_metadata.append(fit_metadata)
+        # Reshape size 1 results to floats to avoid returning shaped results in this case
+        if selected_values.size == 1:
```
TsafrirA: I think you can remove this condition and still have the `(1, )` shaped value correctly broadcast into the result array, which will also have a leading dimension of `(1,)`.
Was it your intention to drop the leading dimension altogether?

TsafrirA (commented):
A few comments and questions.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
dekelmeirom: I added support for ints

dekelmeirom:
The code was changed to calculate extrapolation for each observable term separately

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -456,3 +458,157 @@ def create_pub_result_pec(
...
+        param_basis_pairs: The map between params ndindexes to basis.
+        noise_factors: The noise factors used to amplify the noise.
+        extrapolated_noise_factors: Noise factors to evaluate the fits at.
+        extrapolator: The extrapolator model or models to use. Models will be tried in priority
+            order. Supported models (each fits the named function of the noise factor ``x``):
+            - ``linear``: ``a + b*x``
+            - ``polynomial_degree_k`` (1 <= k <= 7): a degree-k polynomial
+            - ``exponential``: ``a*exp(b*x)``
+            - ``double_exponential``: ``a*exp(b*x) + c*exp(d*x)`` (rates constrained to decay)
+            - ``fallback``: no fit; the measured value at the lowest noise factor
```
```suggestion
            - ``"linear"``: ``a + b*x``
            - ``"polynomial_degree_k"`` (1 <= k <= 7): a degree-k polynomial
            - ``"exponential"``: ``a*exp(b*x)``
            - ``"double_exponential"``: ``a*exp(b*x) + c*exp(d*x)`` (rates constrained to decay)
            - ``"fallback"``: no fit; the measured value at the lowest noise factor
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -456,3 +458,157 @@ def create_pub_result_pec(
...
+    noise_factors: list[float],
+    extrapolated_noise_factors: list[float],
+    extrapolator: list[ExtrapolatorType],
+    measure_noise_data: PauliLindbladMap | np.ndarray | None,
+) -> tuple[npt.NDArray[float], npt.NDArray[float], npt.NDArray[float], list[list[str]]]:
+    """Calculate expectation values for given data, observables and params.
+
+    Args:
+        noise_amplified_data: The measured data result. Its shape should be:
+            (noise_factors, num_randomizations, num_configs, shots, num_bits)
```
SamFerracin: For sphynx, in case this becomes a public func, so that we can avoid it turn it into a bunch italics
```suggestion
            ``(noise_factors, num_randomizations, num_configs, shots, num_bits)``
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -456,3 +458,157 @@ def create_pub_result_pec(
...
+        except KeyError:
+            raise ValueError(
+                f"No measurement basis configurations found for parameter index {param_index}"
+            )
+
+        # each item should contain results for each point in extrapolated_noise_factors
+        exp_vals_extrapolated = np.zeros(len(extrapolated_noise_factors), dtype=float)
+        ensemble_std_extrapolated = np.zeros(len(extrapolated_noise_factors), dtype=float)
+        twirl_variance_extrapolated = np.zeros(len(extrapolated_noise_factors), dtype=float)
+        selected_extrapolators_per_term = []
```
SamFerracin, line 545: I assume you can't do this because later you use `+=`, right?
```suggestion
        exp_vals_extrapolated = np.empty(len(extrapolated_noise_factors), dtype=float)
        ensemble_std_extrapolated = np.empty(len(extrapolated_noise_factors), dtype=float)
        twirl_variance_extrapolated = np.empty(len(extrapolated_noise_factors), dtype=float)
        selected_extrapolators_per_term = []
```

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -41,20 +40,23 @@
...
     the entry's noise factors and evaluated at the target noise factor(s) (``0`` for zero
     noise). Models are tried in priority order: each point takes the result of the first model
     with a valid extrapolation. An extrapolation is valid when its value and standard
     error are finite, the standard error is within the basis threshold, and
-    :math:`value \pm stderr` lies within the basis's ideal range widened by that threshold. The
-    range comes from the entry's ``ev_basis`` metadata: ``[0, 1]`` for projector-only bases,
-    ``[-1, 1]`` for bases containing Paulis, and unbounded when ``ev_basis`` is absent or
+    :math:`value \pm stderr` lies within the observable's ideal range widened by that threshold.
+    The range varies based on the observable: ``[0, 1]`` for projector-only observables,
+    ``[-1, 1]`` for observables containing Paulis, and unbounded when the observable is absent or
```
SamFerracin: I don't fully understand what is meant here:
``[-1, 1] for observables containing Paulis``
by "for observables containing Paulis." The observable `10*Z` contains a Pauli, but the expectation value will be between -10 and 10

My be worth adding a note that these are expectation values for observables that contain projectors only, or Paulis, not linear combinations (as these observables are combined elsewhere). That is, if I understand correctly

Additionally, is "project-only" used correctly here? What is the interval for an observable that contains a mixture of Paulis and projectors, such as `+X`? If I'm not mistaken it should be [0, 1]. Therefore, the distinction is between Pauli-only and mixtures of Paulis and projectors

SamFerracin (commented):
Thanks! Some comments and questions, but looks good

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
SamFerracin: I think there's another function that has the same issue, may be worth switching to a literal in a followup PR and centralize docs and valid values

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
SamFerracin: Sorry, re-reading the docstring again and the type-hint, I realized that you are hereby referring to observable terms, not observables. So that addresses my first point (where maybe, it'd clearer if you referred to terms instead of to observables in the docstring)

I wonder if there is a type in `qiskit.quantum_info` that we could use to type-hint more precisely than a `str`?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
dekelmeirom: Yes, you are correct - I must initialize those to zeros and not empty arrays

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
dekelmeirom: The function indeed calculates the extrapolation for a single observable term, without its coefficient, and not the entire observable together. So for a `10*Z` observable, the function will only get `Z` as the `observable_term`.
I changed a bit the docs, hopefully clarifying it a bit better.
I am not familiar with a type in `qiskit.quantum_info` that will allow both Paulis and projectors observable terms (but maybe there is one that I am not familiar with...)

For an observable contains a mixture of Paulis and projectors, such as `+X`, the interval should be [-1,1]. 
The eigenvalues of a tensor product are the products of the eigenvalues of its components. Therefore, the result of each shot is one of {-1*0, -1*1, 1*0, 1*1} which results in one of {-1,0,1}. The sum of multiple such shots is in the interval [-1,1].

SamFerracin (approved):

PR #3002, opened 2026-07-06
Improve QPY version mismatch error when deserializing jobs

### LINKED ISSUE #2426: Cryptic message for QPY version warning

When retrieving a job that was generated using a newer version of QPY than the version installed client-side, a warning is triggered to alert of potential problems.

```python
> job = service.job("<job_id>")
> ~/python3.x/site-packages/qiskit/qpy/interface.py:316: UserWarning: The qiskit version used to generate the provided QPY file, 2.2.0, is newer than the current qiskit version 2.1.2. This may result in an error if the QPY file uses instructions not present in this current qiskit version
```

From the standpoint of a non-expert user who may be unfamiliar with QPY, the current message is not intuitive (e.g it talks about a _provided QPY file_). The warning message could be refined to provide clearer guidance and a suggestion about what to do (i.e. install the same qpy version client-side).

rosspeili:
Peek the QPY header before `qpy.load` and raise a clear `QiskitError` when the embedded Qiskit version is newer than the installed one. Fixes #2426.

When job data includes circuits serialized with a newer Qiskit than the client has installed, users used to get a vague warning from `qiskit.qpy.load` about a "provided QPY file." This PR checks the QPY header in `RuntimeDecoder` first and fails with a short message that names both versions and suggests `pip install -U qiskit`.

###Changes

- Add `_validate_qpy_qiskit_version` / `_load_qpy` in `json.py`; use them for `QuantumCircuit` and `Instruction` decoding.
- Release note: `release-notes/unreleased/2426.bug.rst`
- Tests: `TestQpyQiskitVersionValidation` (newer version → clear error; matching version still decodes)
- Did not change when `job.inputs` loads data, `service.job()` already skips params, so this covers the paths that actually call `qpy.load`.

Fixes #2426

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/json.py
+++ b/qiskit_ibm_runtime/json.py
@@ -20,8 +20,6 @@ import importlib
 import inspect
 import io
 import json
-import re
-import struct
 import sys
 import warnings
 import zlib
@@ -30,8 +28,6 @@ from typing import TYPE_CHECKING, Any, BinaryIO, get_args
 
 import dateutil.parser
 import numpy as np
-from qiskit import __version__ as QISKIT_VERSION
-from qiskit.exceptions import QiskitError
 
 from qiskit_ibm_runtime.quantum_program.params_converters import QUANTUM_PROGRAM_PARAMS_CONVERTERS
 
@@ -56,7 +52,7 @@ from qiskit.primitives.containers.estimator_pub import EstimatorPub
 from qiskit.primitives.containers.observables_array import ObservablesArray
 from qiskit.primitives.containers.sampler_pub import SamplerPub
 from qiskit.qpy import QPY_VERSION as QISKIT_QPY_VERSION
-from qiskit.qpy import dump, load
+from qiskit.qpy import dump, get_qpy_version, load
 from qiskit.qpy.binary_io.value import _read_parameter, _write_parameter
 from qiskit.result import Result
 from qiskit.transpiler import CouplingMap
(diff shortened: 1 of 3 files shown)


TsafrirA (commented):
Thanks @rosspeili for looking into this!
I have a few concerns about this proposed solution.
First of all, this will now error instead of warn. I think this is overly aggressive. There is a good chance that your QPY will be deserialized just fine.

I am also a little worried about the maintainability of `_read_qpy_qiskit_version`. It will be sensitive to changes in the QPY format. I wonder if we can just use the QPY version instead of the qiskit version. Client side, we can use `qiskit.qpy.QPY_VERSION` and `qiskit.qpy.QPY_COMPATIBILITY_VERSION`, and for the payload, we only need to look at a single byte (the one right after the 6 bytes of `QISKIT`). (You could also read the qiskit version parts directly in a similar fashion, avoiding the header reading code) This part of the format is very unlikely to change, so it should be a bit safer.

What do you think?

rosspeili:
@TsafrirA Thanks for the review and direction, I see your point and tried to push the suggested approach, also rebased to avoid confilct, and tests/lint pass locally. Let me know how it looks.

Newer embedded Qiskit version → warn with a clear message and still attempt load() (no longer errors)
Unsupported QPY format version → still raises a clear QiskitError
Header peek simplified to fixed byte offsets after "QISKIT" (no qpy_formats dependency)
Suppresses the cryptic "provided QPY file" warning when we emit our own

qiskit_ibm_runtime/json.py
```diff
@@ -127,6 +131,76 @@ def _serialize_and_encode(
...
+def _load_qpy(file_obj: BinaryIO) -> list:
+    """Load circuits from QPY with clearer version mismatch warnings."""
+    warned_qiskit = _warn_on_qpy_version_mismatch(file_obj)
+    with warnings.catch_warnings():
+        if warned_qiskit:
+            warnings.filterwarnings(
+                "ignore",
+                message=_QPY_QISKIT_VERSION_WARNING,
+                category=UserWarning,
+            )
```
diego-plan9: Instead of suppressing the original Qiskit warning, what about just keeping it and making no attempt to silence it?.

This makes the code less dependent on the specific wording of the warning (it seems quite stable at least since 2.3.0, which is the minimal version supported by `qiskit-ibm-runtime` - but it might not reasonable to expect it to be kept as-is in future versions), at the expense of a bit more of verbosity on the user side. But technically the new warning that we control should be emitted, plus users have the standard `warnings` facilities for further silencing with fine-grained control if desired.

qiskit_ibm_runtime/json.py
```diff
@@ -127,6 +131,76 @@ def _serialize_and_encode(
...
+
+
+def _read_qpy_header(file_obj: BinaryIO) -> tuple[int, tuple[int, int, int]]:
+    """Return the QPY format and Qiskit versions from the file header prefix."""
+    offset = file_obj.tell()
+    file_obj.seek(0)
+    try:
+        preface, qpy_version, major, minor, patch = struct.unpack("!6sBBBB", file_obj.read(10))
+        if preface != b"QISKIT":
+            raise QiskitError("Input is not a valid QPY payload.")
```
diego-plan9: As the functionality in `json.py` is quite far away from the public interface (ie. it involves results and decoding of inputs, which should come from the API under normal circumstances) and not interfaced with directly, we should have enough assurances that the input to `qpy.load()` is a valid QPY payload. 

By not checking here (the intent is much appreciated, though!) we reduce a bit of complexity, and trust that the upper layers have done their job correctly,

qiskit_ibm_runtime/json.py
```diff
@@ -127,6 +131,76 @@ def _serialize_and_encode(
...
+        warnings.warn(
+            "This job was serialized with Qiskit "
+            f"{version_str}, which is newer than your installed version "
+            f"({QISKIT_VERSION}). Loading may still succeed, but if it fails, "
+            "upgrade with 'pip install -U qiskit'.",
+            UserWarning,
+            stacklevel=3,
+        )
+        return True
+    return False
```
diego-plan9: Can you revise this method, aiming at simplicity? This is in the spirit of https://github.com/Qiskit/qiskit-ibm-runtime/pull/3002#pullrequestreview-4683149585.

By using [get_qpy_version()](https://github.com/Qiskit/qiskit/blob/2.3.0/qiskit/qpy/interface.py#L441) and comparing to [`qiskit.qpy.QPY_VERSION`](https://quantum.cloud.ibm.com/docs/en/api/qiskit/qpy#qiskitqpyqpy_version), hopefully there is enough information for detecting if the input comes "from the future".

If that is the case, and for the purposes of avoiding complexity in the implementation, perhaps the warning can be reworded (still pointing to upgrading the version of `qiskit`) and still be informative enough? We would sacrifice the mention of the specific Qiskit versions, but would result in a very straight forward and maintainable implementation (that also avoids peeking at internals and relies on the standard interfaces provided by Qiskit).

diego-plan9 (commented):
Thanks @rosspeili - took us a while to get back to the review, but continuing the review from Tsafrir, please find a few suggestions aimed at simplifying the implementation while still tackling the original issue.

It's a trade-off that is more pointy in this particular file. Since `json.py` is quite central (and not user-facing), we have an explicit desire of keeping it as simple and maintainable as possible (at the expenses of some "features").

qiskit_ibm_runtime/json.py
```diff
@@ -127,6 +131,76 @@ def _serialize_and_encode(
...
+
+
+def _load_qpy(file_obj: BinaryIO) -> list:
+    """Load circuits from QPY with clearer version mismatch warnings."""
```
diego-plan9: If the previous review comment sound good, you might end up with this single function. If so, please place extra emphasis on the docstrings and type hints, for ensuring maintainability.

rosspeili:
Hey @diego-plan9 thanks for the follow-up, and nw at all about timing. Just pushed a simplified version, lmk how it looks.

test/unit/test_data_serialization.py
```diff
@@ -895,3 +899,39 @@ def test_decoding_incorrect_noise_learner_v3_params_warns(self, schema_version):
...
+        ]
+        self.assertEqual(len(clear_warnings), 1)
+
+    def test_matching_qpy_qiskit_version_still_decodes(self):
+        """Circuits encoded with the installed Qiskit version still decode."""
+        circuit = QuantumCircuit(1)
+        encoded = json.dumps(circuit, cls=RuntimeEncoder)
+        decoded = json.loads(encoded, cls=RuntimeDecoder)
+        self.assertIsInstance(decoded, QuantumCircuit)
+        self.assertEqual(decoded.num_qubits, 1)
```
diego-plan9: Thanks for the attention to preserving functionality - however, this is arguably already covered by `TestDataSerialization.test_coder_qc` (and indirectly, by other tests). Can you remove this test?

diego-plan9 (commented):
Thanks again! One small tweak, and should be good to go.

rosspeili:
@diego-plan9 removed redundant test, indeed test_coder_qc already covers the matching-version path. <3 thanks for this.

diego-plan9 (approved):
Thanks to you for the effort and the PR! This provides a more friendly experience to users, and will guide users towards the right solution gently 👍 

If you are interested in further contributions, I'd be happy to point to specific ones as needed - or feel free to browse the list of issues and ping!

rosspeili:
Hey @diego-plan9, @TsafrirA, glad this was merged, and yes I am open to direct assignments. Feel free to ping me on high priority issues that need to get out of the way, but that are small/medium in size and can be trusted to external contributors. Thanks in advance. V

PR #3004, opened 2026-07-06
Deprecate measure noise learnings shots per randomization as an int

### LINKED ISSUE #2988: Deprecate `measure_noise_learning.shots_per_randomization` as an int (auto only)
`measure_noise_learning.shots_per_randomization` should only support `"auto"` on the
legacy EstimatorV2. The explicit-int path should be deprecated.

## Why

The wrapper runs its TREX calibration as a bundled `SamplexItem` inside the same
`QuantumProgram` as the estimation items, and `QuantumProgram.shots` is **one global value** —
there's no per-item shots on the item or the wire schema. So the calibration can't be given a
shot count of its own; the only lever it has is `num_randomizations` (the item shape).

An int `shots_per_randomization` pretends to give the calibration its own per-randomization
budget, but there's nowhere to put it. Today `prepare.py` just checks that the int matches the
estimation's per-randomization shots and raises otherwise
(`executor_estimator/prepare.py::prepare`):

> "shots_per_randomization must be the same for twirling and measure_noise_learning."

So an explicit int is only accepted when it equals a value the user didn't choose.

To prepare for this change it is best to deprecate the configuration for the legacy behavior.

## Deprecation

- Keep `"auto"` as the only meaningful value; emit a `DeprecationWarning` when an int is passed
  (and keep the existing "must match" error for now).
- Document that the wrapper sizes the calibration budget itself (auto), so a per-randomization
  int isn't meaningful under one global `QuantumProgram.shots`.

leonapivato:
Resolves #2988 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -20,6 +20,8 @@ from typing import TYPE_CHECKING
 from qiskit.primitives.base import BaseEstimatorV2
 from qiskit.primitives.containers.estimator_pub import EstimatorPub
 
+from qiskit_ibm_runtime.utils.deprecation import issue_deprecation_msg
+
 from .base_primitive import BasePrimitiveV2
 from .options.estimator_options import EstimatorOptions
 from .utils import validate_estimator_pubs
@@ -146,6 +148,16 @@ class EstimatorV2(BasePrimitiveV2[EstimatorOptions], Estimator, BaseEstimatorV2)
             if precision <= 0:
                 raise ValueError("The precision value must be strictly greater than 0.")
         coerced_pubs = [EstimatorPub.coerce(pub, precision) for pub in pubs]
+
+        if len({pub.precision for pub in coerced_pubs}) > 1:
+            issue_deprecation_msg(
+                msg="Specifying different 'precision' across pubs is deprecated",
+                version="0.48.0",
+                remedy="Submit one job for each desired precision instead. "
+                "To reduce overhead, Consider submitting these jobs inside a Batch execution mode.",
+                stacklevel=2,
+            )
+
         validate_estimator_pubs(coerced_pubs)
         return self._run(coerced_pubs)
 
--- a/qiskit_ibm_runtime/estimator.py
+++ b/qiskit_ibm_runtime/estimator.py
@@ -23,6 +23,7 @@ from qiskit.primitives.containers.estimator_pub import EstimatorPub
 from .base_primitive import BasePrimitiveV2
 from .options.estimator_options import EstimatorOptions
 from .utils import validate_estimator_pubs
+from .utils.deprecation import issue_deprecation_msg
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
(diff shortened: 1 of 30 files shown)


qiskit_ibm_runtime/estimator.py
```diff
@@ -167,6 +168,22 @@ def _validate_options(self, options: dict) -> None:
                 "a coupling map is required."
             )
 
+        shots_per_randomization = (
```
SamFerracin: Thanks, Leo. I think it'd be better to issue a warning in the option itself. If at any point in time (initialization or after), a user sets `shots_per_randomization`, we issue the deprecation warnings. There should be built-in tools to do that, I suspect, just as they exist for validating and issuing errors

leonapivato: Looks like the only tooling is for validating ints, and whether it is within some bounds. Nothing I can use for this case unless I'm missing something. I made a small helper that uses the same dataclass mechanism though.

test/unit/test_estimator_options.py
```diff
@@ -110,7 +158,7 @@ def test_program_inputs(self):
             "measure_mitigation": True,
             "measure_noise_learning": {
                 "num_randomizations": 1,
-                "shots_per_randomization": 20,
+                "shots_per_randomization": "auto",
```
leonapivato, line 191: this change shouldn't affect the existing test in any way. Better to avoid turning the broad serialization test into one that has to handle deprecation warnings.

leonapivato:
~waiting on #2992 so that the testing helper function can be reused.~

test/unit/test_estimator_options.py
```diff
@@ -93,6 +97,50 @@ def test_bad_inputs(self, val):
         with self.assertRaisesRegex(ValidationError, error_msg):
             EstimatorOptions(**bad_input)
 
+    def test_deprecate_measure_noise_learning_shots_per_randomization_int_on_init(self):
```
leonapivato: Might be cleaner to do subtests here.

qiskit_ibm_runtime/options/measure_noise_learning_options.py
```diff
@@ -46,3 +49,21 @@ class MeasureNoiseLearningOptions:
...
+
+    @field_validator("shots_per_randomization")
+    @classmethod
+    def _warn_shots_per_randomization_int(
+        cls, value: UnsetType | int | Literal["auto"]
+    ) -> UnsetType | int | Literal["auto"]:
+        """Warn when shots per randomization is set to a deprecated integer value."""
+        if isinstance(value, int):
+            issue_deprecation_msg(
+                msg="Specifying 'measure_noise_learning.shots_per_randomization' as an integer "
```
leonapivato, line 61: @jyu00, could you confirm whether the deprecation message looks good.

qiskit_ibm_runtime/options/measure_noise_learning_options.py
```diff
@@ -46,3 +49,21 @@ class MeasureNoiseLearningOptions:
...
+                "is deprecated",
+                version="0.48.0",
+                remedy='Use "auto" instead. The executor-backed primitives size the measurement '
```
SamFerracin: This isn't an executor-backed primitive, and users do not know what that even is (they are not documented yet). I would stick with `Use "auto" instead`

SamFerracin (approved):
Thanks, @leonapivato !

I have verified that this works as expected on a notebook

Issue #2764, opened 2026-04-28
Avoid / minimize the number of DeprecationWarnings from qiskit


diego-plan9:
**What is the expected enhancement?**

Currently, the codebase uses a number of features deprecated by Qiskit. This is undesirable by users (who get a `DeprecationWarning` that they cannot act upon, as it is emitted from our usage) and developers (the tests output becomes quite clogged with such warnings, and we risk being caught in an actual feature removal during the upgrade to a Qiskit release).

It would be great to avoid or at least minimize the number of `DeprecationWarnings` emitted during regular usage of the package, by using non-deprecated mechanisms instead. Related issues:
* https://github.com/Qiskit/qiskit-ibm-runtime/issues/2350 - tackling the likely biggest source of deprecation warnings (there is an open PR https://github.com/Qiskit/qiskit-ibm-runtime/pull/2351, with relevant context)
* https://github.com/Qiskit/qiskit-ibm-runtime/issues/2034
* https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.46.1/test/ibm_test_case.py#L65 - related to the tests, ~I recall there was an in-progress PR touching that line, but I cannot find it~ see https://github.com/Qiskit/qiskit-ibm-runtime/pull/2352. The regular expression `r"^qiskit$`" is currently too restrictive (only top-level `qiskit` considered) and relaxing it might help uncovering more warnings or enforcing action on them.
* CI and  `pytest -v -s` is also a good guide for checking emitted warnings

As an extra item to consider, when replacing usages with non-deprecated alternatives, we should make sure that they are supported by the minimal version of Qiskit supported by the package (`2.2` as of today). This will be easier to check one https://github.com/Qiskit/qiskit-ibm-runtime/issues/2575 is tackled.

**Acceptance criteria**

The number of deprecation warnings emitted by the package is 0 or greatly reduced.

diego-plan9:
When tackling, it would be good to reintroduce https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.48.0/test/ibm_test_case.py#L72-L73:

with a tweak in the regex so tests _do_ fail when we have a deprecation warning:
```python
        # fail test on deprecation warnings from qiskit
        warnings.filterwarnings("error", category=DeprecationWarning, module=r"^qiskit\.*")
```

This would force us to be more diligent when it comes to using deprecated features (being able to easily catch them when testing `rc`s).

yaelbh:
[...]

PR #3007, opened 2026-07-07
Getting rid of two warnings in tests

yaelbh:
### Summary

A lot of warnings get fired during test execution. Some of them - the more important ones - require fixes in the code body (not in tests). They may manifest for users when they run their code, and we opened Issue #2764 to eliminate them. Others require fixes only in the tests, and it's always nice to fix those too.

### Details and comments

Part of #2764.

The following warning is completely eliminated:
```
UserWarning: The <some number>-th circuit has no output classical registers so the result will be empty. Did you mean to add measurement instructions?
```

In addition one of two instances of the following warning:
```
DeprecationWarning: Treating CircuitInstruction as an iterable is deprecated legacy behavior since Qiskit 1.2, and will be removed in Qiskit 3.0. Instead, use the `operation`, `qubits` and `clbits` named attributes.
```

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Bob or Claude Code

### DIFF
--- a/qiskit_ibm_runtime/json.py
+++ b/qiskit_ibm_runtime/json.py
@@ -513,7 +513,7 @@ class RuntimeDecoder(json.JSONDecoder):
                 # circuit to deserialize load qpy circuit and return first instruction object in
                 # that circuit.
                 circuit = _decode_and_deserialize(obj_val, load)[0]
-                return circuit.data[0][0]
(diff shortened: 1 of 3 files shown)


diego-plan9 (approved):
Thanks!

PR #3008, opened 2026-07-07
Use appropriate post processing function for PEC

### LINKED ISSUE #2919: Allow users to select PEC mitigation in the wrapper estimator
#2844 added PEC support designed for the wrapper estimator. However, the `EstimatorV2` class doesn't use the functionality yet. We need to add the option classes, and dispatch from `EstimatorV2.run()` to the correct prepare function.

- [x] The correct prepare function is selected.
- [ ] The correct post processing function is selected.

**Acceptance criteria**
The user can run PEC in the wrapper estimator.

TsafrirA:
### Summary
Closes #2919
This PR modifies the post processor of the wrapper estimator to use the appropriate post processing functions when dealing with PEC error mitigation.

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
+++ b/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
@@ -78,7 +78,8 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
     # Extract options if present
     options_metadata = post_processor_data.get("options", {})
 
-    # Check if PEC mitigation was used
+    # Extract mitigation data
+    mitigation = post_processor_data.get("mitigation", None)
     pec_gammas = post_processor_data.get("pec_gammas", None)
 
     # Check if measure_mitigation was used
@@ -123,9 +124,9 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
         observables = ObservablesArray(observables_label)
         param_shape = tuple(param_shape)
 
-        # Calculate exp vals and place them in a databin
-        if pec_gammas is not None:
-            exp_vals, stds, ensemble_stds = process_expectation_values_pec(
+        # Calculate exp vals and build an EstimatorPubResult
+        if mitigation == "pec":
+            pub_result = create_pub_result_pec(
                 item_result,
                 observables,
                 param_shape,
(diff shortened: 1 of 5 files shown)


qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -121,9 +124,19 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
         param_shape = tuple(param_shape)
 
         # Calculate exp vals and place them in a databin
-        exp_vals, stds, ensemble_stds = process_expectation_values(
-            item_result, observables, param_shape, param_basis_pairs, readout_noise_data
-        )
+        if pec_gammas is not None:
```
TsafrirA: I was considering a more explicit PEC flag, but opted for this at the end.

dekelmeirom (dismissed):
LGTM

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
SamFerracin: Would you be open to consider the opposite, instead?
Namely:
- Adding a `mitigation` field to the passthrough data that can be none, pec, zne, ...
- if mitigation is pec, `pec_gammas` must be present

This will allow us to support more EM protocols more naturally, where we do not need to worry about cases where both `pec_gammas` and `pea_noise_factors` are present

SamFerracin (commented):
Thanks!

A couple of observations:
- The item metadata will need to be different for pec VS no pec (do we have an issue open for that)?
- Also the elements that go into the data bin will need to be different depending on the EM strategy (for PEA for example, you'll need to add `evs_extrapolated` etc).

So we'll need to iterate on this code when we add PEA. I would like to propose a change that will make us iteration faster and support ZNE and PEA more cleanly:

- Write a util function called `compute_pub_results` that:
  - Loops through `result, observables_lists, ..`
  - Returns the full list of `EstimatorPubResult`s
- Write a util function called `compute_pub_results_pec` that does the same for pec
- Now these functions take care of computing metadata and populating the databins correctly, and you can replace your code with:
```python
if mitigation == "none":
    pub_results = compute_pub_results(result, observables_lists, param_basis_pairs_lists, param_shapes_list)
elif mitigation == "pec":
    pub_results = compute_pub_results_pec(result, observables_lists, param_basis_pairs_lists, param_shapes_list)
elif mitigation == "zne":
    pub_results = compute_pub_results_pec(result, observables_lists, param_basis_pairs_lists, param_shapes_list)
...
```

TsafrirA:
@SamFerracin 
I added `create_pub_result` functions. They're a bit useless right now, but once we fill in the metadata (#2825) it will make more sense. I converted the tests to target these functions and underscored the original functions (as they might, for example, be moved outside of runtime).

qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
```diff
@@ -189,6 +189,7 @@ def prepare_pec(
             "param_basis_pairs": param_basis_pairs_list,
             "param_shapes": param_shapes_list,
             "measure_mitigation": "False",
+            "mitigation": "pec",
```
TsafrirA, line 192: It could be argued that this should be set in the `run` method, and I think that would make sense, but seeing that all post processors attributes are set here, I left it here. We can change later and move, for example, the version and the mitigation info to be set in `run`.

SamFerracin: > and I think that would make sense

I agree, but I also agree that we can change later!

SamFerracin (approved):
Thanks 🙏 Looks good!

Issue #2908, opened 2026-06-12
Revise wrapper unit tests


diego-plan9:
**What is the expected enhancement?**

The wrapper estimator / wrapper sampler tests have grown a bit organically, and focus has not been placed in maintainability. 

It would be good to revise them with the information gained during its development, with the goal of bringing them more in line with the rest of the test suite and ensure they can be maintained and evolved. Some ideas and pointers:
* reusing existing testing utils
* increase parametrization
* reduce the amount of patching

**Acceptance criteria**

The wrapper estimator and wrapper sampler tests are more maintainable (fuzzy metrics - using the team's criteria).

diego-plan9:
From https://github.com/Qiskit/qiskit-ibm-runtime/pull/3009/changes#r3551543264:

> Can you explore if https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/test/utils.py#L284 can be reused instead of using custom function with seemingly similar intent? I have not verified in this PR, but it seems it can be replace in `test_sampler.py`with the tests passing.

> Small throwback to https://github.com/Qiskit/qiskit-ibm-runtime/issues/2908 - tiny bit (and can be tackled during that one if needed), and along the lines of re-using existing testing facilities.

PR #3009, opened 2026-07-07
Add local mode to wrapper estimator

### LINKED ISSUE #2965: Support local mode for executor-based estimator
The executor-based estimator needs to support local mode, in the same way as the executor-based sampler does.

TsafrirA:
### Summary
Closes #2965 


SamFerracin (commented):
Thanks, Tsafrir! This looks good

I noticed that when I run the tests, I get warnings. The same happens for the tests of wrapper-sampler that check local mode. Example below. Could you please fix that, for example by popping all the entries of options that won't be needed?

```
  /Users/samferracin/Git/qiskit-ibm-runtime/qiskit_ibm_runtime/fake_provider/local_service.py:274: UserWarning: Options {'execution': {'init_qubits': True, 'rep_delay': None, 'scheduler_timing': False, 'stretch_values': False}, 'twirling': {'enable_gates': False, 'enable_measure': False, 'num_randomizations': 'auto', 'shots_per_randomization': 'auto', 'strategy': 'active-accum'}, 'dynamical_decoupling': {'enable': False, 'sequence_type': 'XX', 'extra_slack_distribution': 'middle', 'scheduling_method': 'alap', 'skip_reset_qubits': False}, 'simulator': {'noise_model': None, 'coupling_map': None, 'basis_gates': None}, 'experimental': {}, 'max_execution_time': None, 'environment': {'log_level': 'WARNING', 'job_tags': [], 'private': False, 'max_execution_time': None, 'image': None}, 'resilience': {'measure_mitigation': True, 'measure_noise_learning': {'num_randomizations': 32, 'shots_per_randomization': 'auto'}, 'pec_mitigation': False, 'pec': {'max_overhead': 100, 'noise_gain': 'auto'}, 'zne_mitigation': False, 'zne': {'amplifier': 'gate_folding', 'noise_factors': 'auto', 'extrapolator': ('exponential', 'linear'), 'extrapolated_noise_factors': 'auto'}, 'noise_model_mapping': None}} have no effect in local testing mode.
    warnings.warn(f"Options {options_copy} have no effect in local testing mode.")
```

TsafrirA:
@SamFerracin 
I noticed that two, but I think that warning is worthwhile - the user needs to know that these options are ignored and don't have any effect on the simulation. I was hoping that once @diego-plan9 carries out his planned transition to BaseModel, we can emit the warning only when something is explicitly set by the user. In the meantime I guess we can just suppress the warnings in the tests?

SamFerracin:
@TsafrirA what is the behaviour in the legacy sampler/estimator?
If those don't warn, I would not warn either

TsafrirA:
@SamFerracin 
> what is the behaviour in the legacy sampler/estimator?

Legacy estimator warns as well, but it does so only for options which were explicitly set by the user. As I said, once we switch to pydantic models, we can be more selective and only warn when something was explicitly set.

test/unit/executor_estimator/test_estimator_v2.py
```diff
@@ -443,3 +455,94 @@ def test_run_dispatches_to_prepare_zne_when_zne_enabled(self, mock_prepare_zne):
...
+        """Test level setup."""
+        import warnings
+
+        self._warning_filter = warnings.catch_warnings()
+        self._warning_filter.__enter__()
+        warnings.filterwarnings("ignore", message=".*have no effect in local testing mode.*")
+
+    def tearDown(self):
+        """Test level tear down."""
+        self._warning_filter.__exit__(None, None, None)
```
diego-plan9: I'm not sure of the intent here, but this seems a bit off - entering a context manager manually during test setup is very surprising, and I am not sure what we gain by silencing our warnings during tests (except for some extra lines in test output if ran in verbose mode - but this obscure actual behavior, as the code _does_ emit the warnings).

If you need to assert on a warning being emitted, there are several options:
* `assert_warning_appears` - https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/ibm_test_case.py#L190 (unsure how this method resisted the pass of time)
* stdlib directly - https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarnsRegex

If you need to silence them for a particular line, you can use https://docs.python.org/3/library/warnings.html#warnings.catch_warnings , and apply them to a specific block of code. This is more explicit and granular, and conveys intent better.

TsafrirA: I didn't assert that the warnings are emitted because I don't consider it something that we want to test.
Why do you prefer to catch the warnings at the test level and not the class level?

In any case, my goal is to have these warnings removed once we switch to pydantic models.

diego-plan9: For explicitness and clarity, and avoiding surprising behavior. Specially coupled with the second concern: what do we gain by silencing the warnings in the tests?

I feel we actually _lose_ instead of gain (as in, the output of these particular tests artificially doesn't behave 100% as the output we would get by running the same code outside of the tests). The only benefit I see is a bit less of cruft when running those particular tests - but our default runner (`pytest`) has options for hiding/showing/tuning warnings as needed (`-v`, `-s`, `--disable-warnings`, `-W`, etc), and we just assume warnings are emitted for the rest of the tests already.

I think the goal is good (and to be evolved as code evolves), but I don't see the benefit or how it contributes to the goal.

test/unit/executor_estimator/test_estimator_v2.py
```diff
@@ -20,30 +20,42 @@
...
 
 
+def create_mock_backend():
+    """Create a mock IBMBackend for testing (non-local, executor path)."""
+    backend = MagicMock(spec=IBMBackend)
+    backend.name = "fake_backend"
+    backend._instance = "ibm-q/open/main"
+    backend.service = MagicMock()
+    backend.target = FakeManilaV2().target
+    return backend
```
diego-plan9: Can you explore if https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/test/utils.py#L284 can be reused instead of using custom function with seemingly similar intent? I have not verified in this PR, but it seems it can be replace in `test_sampler.py`with the tests passing.

Small throwback to https://github.com/Qiskit/qiskit-ibm-runtime/issues/2908 - tiny bit (and can be tackled during that one if needed), and along the lines of re-using existing testing facilities.

test/unit/executor_estimator/test_estimator_v2.py
TsafrirA: Removed the error suppression. And opened #3021 to reduce the warnings when the models are available. (fyi @SamFerracin)

test/unit/executor_estimator/test_estimator_v2.py
TsafrirA: The proposed `get_mocked_backend` fails one of the tests, it seems to be missing some instruction duration properties. I can look into it if you want.

diego-plan9: Ok, moved it to #2908

test/unit/executor_estimator/test_estimator_v2.py
diego-plan9: Thanks.

SamFerracin (dismissed):
LG! Thanks!

diego-plan9 (approved):

PR #3011, opened 2026-07-07
NLV3 - updating `find_learning_protocol` to avoid undressed boxes for Clifford validation

apchytr:
### Summary

Updating the find_learning_protocol validation method in the noise learner v3 to avoid using the undressed box for Clifford validation.

### Details and comments

An example like
```
with circuit.box([Twirl()]):
    circuit.rz(1.2, 0)
    circuit.cz(0, 1)
```
will now raise.

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/noise_learner_v3/find_learning_protocol.py
+++ b/qiskit_ibm_runtime/noise_learner_v3/find_learning_protocol.py
@@ -42,10 +42,11 @@ def find_learning_protocol(instruction: BoxOp) -> LearningProtocol | None:
     Raises:
         IBMInputValueError: If ``instruction`` does not contain a box.
     """
-    if (name := instruction.operation.name) != "box":
+    box = instruction.operation
+    if (name := box.name) != "box":
         raise IBMInputValueError(f"Expected a 'box' but found '{name}'.")
 
-    undressed_box = undress_box(instruction.operation)
+    undressed_box = undress_box(box)
 
     if len(undressed_box.body) == 0:
         return LearningProtocol.PAULI_LINDBLAD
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/noise_learner_v3/find_learning_protocol.py
```diff
@@ -56,23 +57,23 @@ def find_learning_protocol(instruction: BoxOp) -> LearningProtocol | None:
     ]
     is_layer = len(active_qubits) == len(set(active_qubits))
 
-    # Check if the undressed box only contains two-qubit Clifford gates
-    has_only_2q_clifford_gates = all(
-        (op.is_standard_gate() and op.operation.num_qubits == 2) or op.name == "barrier"
-        for op in undressed_box.body
+    # Check if the dressed box is Clifford
+    is_clifford = all(
+        (op.is_standard_gate() and op.operation.num_qubits <= 2) or op.name == "barrier"
```
TsafrirA, line 62: What about delay?

qiskit_ibm_runtime/noise_learner_v3/find_learning_protocol.py
```diff
@@ -56,23 +57,23 @@ def find_learning_protocol(instruction: BoxOp) -> LearningProtocol | None:
...
+            Clifford(box.body)
         except QiskitError:
-            has_only_2q_clifford_gates = False
+            is_clifford = False
+
+    if is_layer and is_clifford:
+        return LearningProtocol.PAULI_LINDBLAD
 
     # Check if the undressed box only contains measurements
     has_only_meas = all(op.name in ["measure", "barrier"] for op in undressed_box.body)
```
TsafrirA, line 75: What about delay?

TsafrirA (commented):
Seems good. I think delays should be considered?

apchytr:
> Seems good. I think delays should be considered?

Yes, that's a good point. With this PR `delay` is handled the same as is currently in `main`. As in, `find_learning_protocol` will return `None` because it fails all of the checks. I'm not sure whether we intend to support `delay` or not but if we do I think we need to create a new issue and investigate

SamFerracin:
Thanks @TsafrirA for pointing out delays. Given that they are not currently supported, I'd say:
- @apchytr please open an issue to investigate. I am not even sure pec-runtime can handle this
- We merge this PR as is, as it is fixing the bug without reducing functionality

apchytr:
Issue for delay support here: https://github.com/Qiskit/qiskit-ibm-runtime/issues/3013

SamFerracin (approved):

PR #3014, opened 2026-07-08
Add a method in executor-based estimator to finalize the options based on the resilience level

### LINKED ISSUE #2978: Add resilience level to wrapper estimator
Legacy estimator has a resilience level which allows the user to select pre-defined sets of error mitigation options. We need to introduce a similar field to the estimator options, and when set, apply the appropriate options. This should be done with care, as without the `Unset` option, you can't differentiate defaults from user set values.

**Acceptance criteria**
Users can set resilience level in the wrapper estimator, and get a similar behavior to that of the legacy estimator. Discrepancies are documented.

Depends on ~~#2919, #2979,~~ #2980

SamFerracin:
### Summary
<Coming when taking out of WIP>

### Details and comments

Fixes #2978, #2981

### AI/LLM disclosure


qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -140,13 +140,79 @@ def find_unique_layers(self, pubs: Iterable[EstimatorPubLike]) -> list[CircuitIn
...
+        user_options_dict = asdict(self.options)  # type: ignore[call-overload]
+        if self.options.twirling.enable_gates is None:
+            user_options_dict["twirling"].pop("enable_gates")
+        if self.options.twirling.enable_measure is None:
+            user_options_dict["twirling"].pop("enable_measure")
+        if self.options.resilience.measure_mitigation is None:
+            user_options_dict["resilience"].pop("measure_mitigation")
+        if self.options.resilience.zne_mitigation is None:
+            user_options_dict["resilience"].pop("zne_mitigation")
+        options_dict.update(user_options_dict)
```
leonapivato: Might be a small bug here if .update() isn't recursive, which I don't think it is.

When you call .update() it only replaces the top level keys. So `options_dict["twirling"]`, for example, will be updated to be the `user_options_dict["twiriling"]`, which just popped several important keys.

SamFerracin: Thanks! I fixed as we discussed!

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -140,13 +140,37 @@ def find_unique_layers(self, pubs: Iterable[EstimatorPubLike]) -> list[CircuitIn
...
+            inherit the resilience-level defaults.
+        3. Enforce required option dependencies. Specifically:
+            * Enabling measurement mitigation automatically enables measurement twirling.
+            * Enabling gate-based mitigation techniques (such as PEA-based ZNE or PEC) automatically
+              enables both gate and measurement twirling.
+
+        Returns:
+            The finalized :class:`~.EstimatorOptions` object.
+        """
+        return finalize_options(self.options)
```
diego-plan9: What is the purpose of keeping `finalize_options` as a separately defined and internal function? If it is not to be consumed other than its current usage, the current implementation adds one level of indirection that we could avoid.

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -178,18 +202,21 @@ def run(
         # Coerce pubs to EstimatorPub objects
         coerced_pubs = [EstimatorPub.coerce(pub, precision) for pub in pubs]
 
+        # finalize options
+        options = self.finalize_options()
```
diego-plan9: Can you expand the comment slightly, to keep track of why the finalization is taking place at runtime (currently)?

Following up on an internal discussion, I'd really like us to explore finalizing options at option creation or modification time, rather than at runtime. As far as I can see:
* all finalizing (and the introduced checks in the `prepare_*` functions) only need knowledge about the current options
* if they are not modified, we are finalizing every time `run()` is invoked, producing the same set of options
* we cannot do `self.options = self.finalize_options()` just because of wanting to do it at runtime

Something for later, but please document the motivation in the comment.

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -140,13 +140,37 @@ def find_unique_layers(self, pubs: Iterable[EstimatorPubLike]) -> list[CircuitIn
             The unique boxed layers found across the given PUBs.
         """
         coerced_pubs = [EstimatorPub.coerce(pub, None) for pub in pubs]
+        options = self.finalize_options()
         return find_unique_layers(
             pubs=coerced_pubs,
```
diego-plan9, line 175: Not part of this PR, but related to the question on `self.finalize_options -> finalize_options`.

qiskit_ibm_runtime/executor_estimator/estimator.py
SamFerracin: I placed this here to facilitate using it in isolation, which is useful to @leonapivato in his repo as well as in the tests.
If I hadn't created the utils method, one would need to instantiate an instance of estimator, which in some situations (for example unit tests) adds some complications, such as using mock backends. But we face these complications in other places, so I am not against removing the indirection. Whatever people prefer 😄

SamFerracin: Removed as per slack discussions: [3991f33](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3014/commits/3991f3399f0152085de8a739cc85686a9bc30375)

qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
```diff
@@ -87,6 +87,16 @@ def prepare_pec(
         IBMInputValueError: If ``noise_model_mapping`` is missing a noise map for at least one of
             the pubs layers.
     """
+    if twirling_options.enable_gates is None:
+        raise ValueError(
+            "Expected twirling options field ``enable_measure`` set to ``True`` or ``False``, "
+            "found ``None``."
+        )
+    if twirling_options.enable_gates is not True:
```
apchytr:
```suggestion
    if not twirling_options.enable_gates:
```

qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
```diff
@@ -87,6 +87,16 @@ def prepare_pec(
...
+        raise ValueError("PEC requires enabling twirling for gates.")
+    if measure_noise_learning is not None and twirling_options.enable_measure is not True:
```
apchytr:
```suggestion
    if measure_noise_learning is not None and not twirling_options.enable_measure:
```

qiskit_ibm_runtime/executor_estimator/estimator.py
SamFerracin: Essentially, we can see the difference in terms of static finalization as opposed to dynamic finalization. 

To see the difference, look at this two examples:
```
options.resilience_level = 2
options.resilience_level.zne_mitigation = False

options.resilience_level.zne_mitigation = False
options.resilience_level = 2
```
With today's "static" mechanism, in both legacy and wrapper estimator, the finalized options (those sent to the server by the legacy, and those sent to the `prepare` function by the wrapper) will end up with ZNE being `False`.
But with the "dynamical" mechanism, they'd end up being `False` and `True` respectively, or to raise an error if we decide that setting resilience level =2 and ZNE False is disallowed. So we may have some backward compatibility issues

I also need to admit that I haven't thought too deeply about this, for lack of time. I definitely see it as an interesting direction to explore in the coming weeks, to have more reasoned answers

qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
SamFerracin: Solved a bit better: [836b0fb](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3014/commits/836b0fb5fccafdd3f8e99ca9f67d242d0cc39754)

qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
SamFerracin: [3bd8219](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3014/commits/3bd8219dc3a09c1d6959d52a06ce6fbc0c698b06)

qiskit_ibm_runtime/executor_estimator/estimator.py
SamFerracin: [d1d77a2](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3014/commits/d1d77a2463592d2806b2bcb71fdb8526eb7d5f4d)

diego-plan9 (approved):

PR #3017, opened 2026-07-08
Move prepare out of executor-based Sampler

### LINKED ISSUE #3016: Move prepare out of executor-based Sampler
Reasons:
- Consistency with executor-based Estimator
- `prepare` can be invoked without instantiating a backend and/or a Sampler, which is more handy for debugging and testing
- Allows introducing `finalize_options` consistently across Sampler and Estimator

SamFerracin:
### Summary
This PR moves `prepare` out of `SamplerV2` and into a dedicated file.

Fixes #3016 

### AI/LLM disclosure


SamFerracin:
I have not removed or changed tests, but simply split them and adjusted them

qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -174,7 +167,16 @@ def run(self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None) -> Ru
         # Non-simulator path: use executor
         # Convert pubs to QuantumProgram and map options using the prepare method
         logger.info("Starting pre-processing")
-        quantum_program, executor_options = self.prepare(coerced_pubs, default_shots)
+        quantum_program, executor_options = prepare(coerced_pubs, self.options, default_shots)
+
+        # Apply dynamical decoupling if enabled
```
leonapivato, line 172: Would it be better if this was still in the prepare function, and the backend got passed in as a parameter. Seems like `apply_dynamical_decoupling` is apart of the overall prepare process, no?

leonapivato (approved):

qiskit_ibm_runtime/executor_sampler/sampler.py
SamFerracin: As discussed privately, this is not possible due to the fact that DD is not vendor agnostic (needs a backend)

PR #3019, opened 2026-07-08
Port zero precision guard clause

### LINKED ISSUE #3015: Wrapper Estimator ZeroDivisionError
## Summary
Specifying 0 as your precision in the run level input on the wrapper estimator will cause a `ZeroDivisionError`. This is because it calculates the shots from the precision on the client side and never ported the guard clause.

leonapivato:
Fixes #3015 

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/utils.py
+++ b/qiskit_ibm_runtime/executor_estimator/utils.py
@@ -150,7 +150,7 @@ def resolve_precision(
         IBMInputValueError: If pubs have different precision values.
     """
     # Extract precision from pubs
-    pub_precisions = {pub.precision or run_precision for pub in pubs}
+    pub_precisions = {pub.precision if pub.precision is not None else run_precision for pub in pubs}
 
     if len(pub_precisions) != 1:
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -150,15 +150,18 @@ def resolve_precision(
...
+    pub_precisions = {pub.precision if pub.precision is not None else run_precision for pub in pubs}
 
     if len(pub_precisions) != 1:
         raise IBMInputValueError(
             f"All pubs must have the same precision. Found: {pub_precisions}"
             "(possibly via the run provided precision parameter)"
         )
 
-    return next(iter(pub_precisions))
+    if (precision := next(iter(pub_precisions))) == 0:
```
TsafrirA: Let's check that it's greater than zero then.

TsafrirA (commented):
Nice!
Your error message says that it must be larger than zero, but you check only equality. Let's check for negative values as well.

TsafrirA (approved):
Thank you!

PR #3020, opened 2026-07-09
Correctly retrieve jobs from retired backends

### LINKED ISSUE #3012: Allow querying jobs from retired backend

This code 

```python
jobs = service.jobs(limit=1, backend_name="ibm_torino")
```

fails with 

```
Unable to create configuration for ibm_torino. '404 Client Error: Not Found for url: https://quantum.cloud.ibm.com/api/v1/backends/ibm_torino/configuration. 
```

Because `ibm_torino` is retired. But I'm only trying to retrieve the job information, _not_ backend configuration, and the Qiskit Runtime API allows me to retrieve jobs from retired backends. 

If the reason is that a Job must have a backend instance, then perhaps it can use [IBMRetiredBackend](https://github.com/Qiskit/qiskit-ibm-runtime/blob/be561d0b0228bd945852a0ff3e1e2ff4bbf2be4c/qiskit_ibm_runtime/ibm_backend.py#L493). 

**Acceptance criteria**

Allow users to query for jobs from retired backends.

diego-plan9:
### Summary

Revise the handling of jobs from retired backends (by having an inner function raise when their configuration could not be accessed), so the `job()` and `jobs()` methods produce an `IBMRetiredBackend` in that case.

### Details and comments

Fixes #3012

### AI/LLM disclosure


diego-plan9:
Issued this one against `main` for extra confidence via integration tests.

diego-plan9:
Moving to draft to further inspect some potential issues with `._active_api_client`.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -738,6 +741,9 @@ def _create_backend_obj(
 
         Returns:
             A backend object.
+
+        Raises:
+            ``QiskitBackendNotFoundError`` if the backend cannot be created.
```
diego-plan9, line 746: This is the main source of the rest of the changes. This function will now raise if the configuration for a backend could not be retrieved, instead of returning `None`. Two functions depended on it: 
* `_decode_job` was kind of already prepared for it (and the responsible for creating `IBMRetiredBackend`
* `backends` was not - I added equivalent exception handling for safety.

test/unit/test_job_retrieval.py
```diff
@@ -91,9 +88,9 @@ def test_backend_instance_warnings(self, service):
         with self.assertNoLogs("qiskit_ibm_runtime", level="WARNING"):
             service.job(job.job_id())
 
-    def test_jobs_skip_limit(self):
+    @run_cloud_fake
+    def test_jobs_skip_limit(self, service):
```
diego-plan9, line 92: These changes in decorators and signature are for consistency and proper usage of `run_cloud_fake`. Eventually most of this file can be rewritten using the registry approach - but it requires filtering to be handled correctly, so leaving it for the future.

diego-plan9:
> Moving to draft to further inspect some potential issues with `._active_api_client`.

This was unrelated to this change, but related to `jobs()`. Updated the docstring to reflect it.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -781,7 +787,9 @@ def _create_backend_obj(
                 self._backend_configs[backend_name] = config
         except Exception as ex:
             logger.warning("Unable to create configuration for %s. %s ", backend_name, ex)
-            return None
+            raise QiskitBackendNotFoundError(
+                f"Unable to create configuration for {backend_name}, assuming backend is retired"
```
rainerschoe: The assumption that the backend is retired is not happening here. It is possibly done by the caller when the exception is mapped to an IBMRetiredBackend.

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -802,7 +810,10 @@ def _create_backend_obj(
                 calibration_id=calibration_id,
                 physical_qubits=physical_qubits,
             )
-        return None
+
+        raise QiskitBackendNotFoundError(
+            f"Unable to create configuration for {backend_name}, assuming backend is retired"
```
rainerschoe: maybe something like this:
```suggestion
            f"Unable to create configuration for {backend_name}. This might happen for example when a backend is retired."
```

qiskit_ibm_runtime/qiskit_runtime_service.py
rainerschoe: Maybe like this?
```suggestion
                f"Unable to create configuration for {backend_name}. This might happen for example when a backend is retired."
```

rainerschoe (dismissed):
LGTM, some minor comment about the text in the exception. It confused me initially.

qiskit_ibm_runtime/qiskit_runtime_service.py
diego-plan9: Thanks - tweaked both at 84def66 (not applying as the line length would cause another ruff commit).

rainerschoe (approved):
LGTM!

Issue #2979, opened 2026-06-30
Allow users to select PEA error mitigation in wrapper estimator


TsafrirA:
**What is the expected enhancement?**

**Acceptance criteria**
The user can set PEA options, and enable it.

- [x] The correct prepare method is used.
- [x] The correct post processor is applied.

rosspeili:
Hey @TsafrirA can I give this a shot?

TsafrirA:
Thank you @rosspeili for taking interest in our code base! We highly value external contributions. Unfortunately, this particular issue is part of a new effort which is undergoing changes on quite literally a daily basis. It is therefore not well suited for external contributors. If you want, we can try and find an issue which is better suited for you!

rosspeili:
Hey @TsafrirA nw at all, and appreciate it. Yes would love to work on something around qiskit. Thanks in advance. I already have the CLA and a few minor prs merged if it's the same across repos.

TsafrirA:
Closed via #2948 and #3029.

PR #3029, opened 2026-07-13
Dispatch wrapper estimator with PEA to appropriate prepare function

TsafrirA:
### Summary
Part of #2979

### Details and comments
This PR dispatches to the correct prepare function when PEA is enabled. Also verifies that ZNE and PEC are not set simultaneously. 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -186,7 +186,8 @@ class EstimatorV2(BaseEstimatorV2):
             pubs=coerced_pubs,
             twirling_options=options.twirling,
             measure_noise_learning=options.resilience.measure_noise_learning,
-            inject_noise=options.resilience.pec_mitigation,  # TODO: Add PEA once available
+            inject_noise=options.resilience.pec_mitigation
+            or (options.resilience.zne_mitigation and options.resilience.zne.amplifier == "pea"),
         )
 
     def finalize_options(self) -> EstimatorOptions:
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -302,15 +308,32 @@ def run(
...
-                measure_noise_learning=options.resilience.measure_noise_learning
-                if options.resilience.measure_mitigation
-                else None,
-            )
+            if options.resilience.zne.amplifier == "pea":
+                if options.resilience.noise_model_mapping is None:
+                    raise IBMInputValueError(
+                        "When PEA mitigation is enabled, you must provide a noise model "
+                        "via options.resilience.noise_model_mapping"
+                    )
```
SamFerracin: I would just go with this:
```suggestion
                noise_model_mapping = options.resilience.noise_model_mapping or {}
```
`prepare_pea` expects a dict, but it doesn't specify whether or not it needs to have items in it. So I'd leave the validation up to `prepare_pea`--which is already happening when it raises an error saying that `f"noise_model_mapping is missing noise map for layer reference {ref}"`

TsafrirA: [2be01b6](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3029/commits/2be01b658ba3b443f638ba801b95677e0806b493)

SamFerracin (approved):
Thanks!

diego-plan9 (approved):

PR #3031, opened 2026-07-13
Pass finalized options in WrapperEstimator

diego-plan9:
### Summary

As part of landing #3014 and #3009, I did not spot during the final review of #3009 that `_run_simulator` is using the non-finalized options. This PR fixes it so both wrapper primitives behave the same way.

### Details and comments

Related to #2965

### AI/LLM disclosure


TsafrirA (approved):
Thanks for spotting this!

PR #3025, opened 2026-07-10
Add missing release notes

SamFerracin:
### Summary
This PR adds missing release notes for PRs #3004 , #2992 , #2973 

### AI/LLM disclosure

- [x] I didn't use LLM tooling, or only used it privately.
- [ ] I used the following tool to help write this PR description:
- [ ] I used the following tool to generate or modify code:

PR #3034, opened 2026-07-14
Prepare 0.48.0 release

diego-plan9:
### Summary

Prepare the changelog for upcoming `0.48.0` release.

### Details and comments

Fixes N/A

### AI/LLM disclosure


docs/conf.py
```diff
@@ -31,7 +31,7 @@
 # The short X.Y version
 version = ""
 # The full version, including alpha/beta/rc tags
-release = "0.46.1"
+release = "0.48.0"
```
SamFerracin, line 34: Oops, looks like I forgot to change this when I released 0.47

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
...
+  * ``EstimatorPubResult`` (previously part of ``qiskit_ibm_runtime.utils.estimator_pub_result``)
+
+  Importing from the previous locations is deprecated, and the references will be removed in a future
+  release. (`2835 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2835>`__)
+
+
+New Features
+------------
+
+- The :meth:`qiskit-ibm-runtime.RuntimeJobV2.usage` method now supports an optional ``partial``
```
SamFerracin: All of these `:meth:` and `:class:` stuff renders ugly right now, see rich diff. I suspect if we add `~`, it should render properly
```suggestion
- The :meth:`~qiskit-ibm-runtime.RuntimeJobV2.usage` method now supports an optional ``partial``
```

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
...
+- Added a user warning when a user sets ``pub.shots`` in SamplerV2, and ``pub.precision`` in EstimatorV2. (`2992 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2992>`__)
+- Added a user warning when a user sets the ``shots_per_randomization`` field of ``MeasureNoiseLearningOptions```
+  to an ``int``. (`3004 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/3004>`__)
+
+
+Deprecation Notes
+-----------------
+
+- All classes that represent results are now available via the ``qiskit_ibm_runtime.results`` package,
+  resulting in the following classes being moved to the results package:
```
abbycross:
```suggestion
- All classes that represent results are now available via the ``qiskit_ibm_runtime.results`` package;
 consequently, the following classes moved to the results package:
```

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
...
+  stretches for individual circuits when running Executor jobs. (`2790 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2790>`__) (`2789 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2789>`__)
+- Added support for ``instance="auto"`` in :class:`.QiskitRuntimeService` and
+  :meth:`.QiskitRuntimeService.save_account`. Passing ``instance="auto"`` explicitly
+  opts in to automatic instance selection without triggering the "instance was not set"
+  warning or the "Loading/Using instance" warnings during backend retrieval. The value
+  can be saved to the account file so it takes effect on every subsequent instantiation. (`2854 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2854>`__)
+- Added a ``use_fractional_gates`` parameter to :meth:`.QiskitRuntimeService.least_busy`,
+  matching the behavior of the same parameter in :meth:`.QiskitRuntimeService.backend`.
+  When set to ``True``, only backends that include fractional gates in their basis gates are
+  considered; backends without it are skipped in favor of the next least busy qualifying
```
abbycross:
```suggestion
  considered; backends without it are skipped in favor of the next least-busy qualifying
```

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
...
+  backend. (`2855 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2855>`__)
+- Support for passing multiple result decoders to :class:`qiskit_ibm_runtime.RuntimeJobV2` and
+  :meth:`qiskit_ibm_runtime.RuntimeJobV2.result` has been reintroduced. (`2899 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2899>`__)
+- Added support ``QuantumProgram.meas_level = "both"``, which can be used to request that
```
abbycross:
```suggestion
- Added support for ``QuantumProgram.meas_level = "both"``, which can be used to request that
```

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
...
+Bug Fixes
+---------
+
+- Fixed a bug in the :class:`.FoldRzzAngle` transpiler pass where ``Rzz`` gates whose angle
+  wraps exactly onto the ``+-pi`` boundary (for example ``3*pi``, ``-3*pi``, ``7*pi``) were
+  folded into a circuit that differed from the original by a global phase of ``pi``. The
+  pass now derives the dropped ``2*pi`` winding parity directly from the wrapped angle, so
+  the folded circuit preserves the full unitary (including global phase) for all angles. (`2856 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2856>`__)
+- When using backends from different instances, the service used by :meth:`.SamplerV2.run`,
+  :meth:`.EstimatorV2.run` and other primitives' ``.run()`` methods now ensures that the API client
```
abbycross:
```suggestion
  :meth:`.EstimatorV2.run`, and other primitives' ``.run()`` methods now ensures that the API client
```

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
...
+  target to ``(0, pi/2)`` for the ``rzz`` gate, and
+  registering an ``rzz`` wrapper with the transpiler's ``WrapAngles`` pass.
+  This mechanism replaces the usage of the ``IBMDynamicFractionalTranslationPlugin``
+  plugin and the :class:`.FoldRzzAngle` transpiler pass. (`2960 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2960>`__)
+
+
+Other Notes
+-----------
+
+- The library now uses the version ``2026-04-15`` of the Quantum Runtime REST API. (`2845 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2845>`__)
```
abbycross: Should this be Qiskit Runtime?

diego-plan9: Yes - 0148513

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
+0.48.0 (2026-07-14)
+===================
+
+Upgrade Notes
+-------------
+
+- The minimal version of Qiskit required is now ``2.3.0``, and that of Samplomatic is ``0.18.0``. (`2925 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2925>`__)
+- Added a user warning when a user sets ``pub.shots`` in SamplerV2, and ``pub.precision`` in EstimatorV2. (`2992 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2992>`__)
```
jyu00: This is only if shots/precision contain different values, right? E.g. I can still specify `pub.shots` as long as I use the same value for all pubs?

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
...
+- Added a user warning when a user sets the ``shots_per_randomization`` field of ``MeasureNoiseLearningOptions```
+  to an ``int``. (`3004 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/3004>`__)
```
jyu00: Why aren't these in the deprecation section, given the warnings are deprecation warning?

release-notes/0.48.0.rst
diego-plan9: Thanks, fixed at c4ef04d. This was likely due to including `qiskit-ibm-runtime` (dashes instead of underscores). I went without `~` as [that would keep the text](https://www.sphinx-doc.org/en/master/usage/domains/python.html#target-specification) just showing `usage()` in this case, and in other cases where it is more reasonable to show the class name too.

release-notes/0.48.0.rst
diego-plan9: Good question!

@SamFerracin, taking advantage of you working in #3025, can you provide a suggestion for converting them more into a deprecation note?
> - Setting MeasureNoiseLearningOptions.shots_per_randomization to an int is deprecated [...]

(also for the first item, please check https://github.com/Qiskit/qiskit-ibm-runtime/pull/3034#discussion_r3579932045 too).

SamFerracin: I would suggest something similar to what you have proposed:
```
Setting `MeasureNoiseLearningOptions.shots_per_randomization` to an `int` is deprecated and will be removed in future releases. The value `"auto"` remains supported (or something along those lines, but basically telling users on "int" to switch to "auto").
```

release-notes/0.48.0.rst
SamFerracin: Here, I would suggest something along the lines of:
```
Running a Sampler job with different `pub.shots` values across PUBs is deprecated. Similarly, running an Estimator job with different `pub.precision` values across PUBs is deprecated.
```

@leonapivato comments?

release-notes/0.48.0.rst
diego-plan9: Thanks - tweaked and moved in 79ca8de.

release-notes/0.48.0.rst
diego-plan9: Thanks - tweaked and moved in 79ca8de.

leonapivato: > This is only if shots/precision contain different values, right? E.g. I can still specify `pub.shots` as long as I use the same value for all pubs?

This is correct, but there is also the case when pubs resolve to different pub-level shots.

 Example: I have two pubs, pub 1 has shots = 100, pub 2 has shots = None. In this case, if run shots != 100, pub 2 will resolve a different shot count than pub 1, so the message will appear.  But if run shots = 100, then it will resolve to the same shot count, so no message will appear.

Not sure if that changes the comment at all. 

> ```
> Running a Sampler job with different `pub.shots` values across PUBs is deprecated. Similarly, running an Estimator job with different `pub.precision` values across PUBs is deprecated.
> ```

Looks good to me.

release-notes/0.48.0.rst
```diff
@@ -0,0 +1,82 @@
+0.48.0 (2026-07-14)
+===================
+
+Upgrade Notes
+-------------
+
+- The minimal version of Qiskit required is now ``2.3.0``, and that of Samplomatic is ``0.18.0``. (`2925 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2925>`__)
```
SamFerracin:
```suggestion
- The minimal version of Qiskit required is now ``2.3.0``, and that of Samplomatic is ``0.18.0``. (`2925 <https://github.com/Qiskit/qiskit-ibm-runtime/pull/2925>`__) Users of `Executor` can now twirl their circuits with the `"local_c1"` twirling group, which is supported by Samplomatic version ``0.17.0`` and above.
```

diego-plan9: Thanks - 89393de3 for some quotes handling and leaving the link at the end.

SamFerracin (approved):
LGTM!

PR #3035, opened 2026-07-14
Size TREX calibration randomizations from twirling

### LINKED ISSUE #3023: Executor Estimator: TREX calibration shot budget is mis-sized
## Summary

The executor-based `EstimatorV2` sizes the TREX (measure-noise-learning)
calibration item incorrectly. The calibration still returns `evs`, so the
divergence is silent, but the shot cost is wrong.

## Where

- `qiskit_ibm_runtime/executor_estimator/trex_utils.py` (`create_trex_calibration_circuit`)
- `qiskit_ibm_runtime/executor_estimator/prepare.py` (and `prepare_pea` / `prepare_pec` / `prepare_zne`)
- `qiskit_ibm_runtime/options_models/measure_noise_learning_options.py`
  (`shots_per_randomization` default `"auto"`, `num_randomizations` default `32`)

## Problem

A `QuantumProgram` has a single global per-randomization shot budget, shared by
the estimation items and the TREX calibration item, so the TREX item cannot carry
its own `shots_per_randomization`. Today the TREX item just runs
`num_randomizations` (default 32) at that global per-randomization budget. Because
the budget grows ~`1/precision²`, TREX cost grows without bound as precision
tightens (e.g. `program_shots ≈ 10k` → ~320k TREX shots).

## Proposed fix

Keep `shots_per_randomization` at `"auto"` and never resolve it to a concrete
per-randomization value (an explicit int should not be supported — it is being
deprecated). Since the TREX item always runs at the fixed global per-randomization
budget, size the calibration purely by **scaling `num_randomizations`** so the
total TREX shots land at an appropriate, bounded budget instead of
`32 × program_shots`.

leonapivato:
### Summary
Measure_noise_learning.num_randomizations now defaults at "auto", which will re-use the twirling num_randomizations. If it is explicitly set to an int, then it will use that int.

Fixes #3023 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
+++ b/qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
@@ -97,6 +97,10 @@ def prepare_pec(
         twirling_options.num_randomizations,
         twirling_options.shots_per_randomization,
     )
+    # Preserve the twirling randomization count: ``num_randomizations`` is scaled
+    # per-pub by the PEC sampling overhead below, but the TREX calibration should
+    # follow the (unscaled) twirling value.
+    twirling_num_randomizations = num_randomizations
 
     # set max_overhead
     max_overhead = pec_options.max_overhead
@@ -222,7 +226,9 @@ def prepare_pec(
             raise IBMInputValueError(
                 "shots_per_randomization must be the same for twirling and measure_noise_learning"
             )
-        trex_item = create_trex_calibration_circuit(pubs, measure_noise_learning)
+        trex_item = create_trex_calibration_circuit(
+            pubs, measure_noise_learning, twirling_num_randomizations
+        )
         quantum_program.items.append(trex_item)
         passthrough_data["post_processor"]["measure_mitigation"] = "True"
 
--- a/qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
+++ b/qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
@@ -34,7 +34,7 @@ from samplomatic import build
 from ...exceptions import IBMInputValueError
 from ...quantum_program import QuantumProgram
 from ...quantum_program.quantum_program import SamplexItem
-from ..trex_utils import create_trex_calibration_circuit
+from ..trex_utils import create_trex_calibration_circuit, resolve_trex_num_randomizations
 from ..utils import (
     box_circuit,
     compute_samplex_arguments,
@@ -226,9 +226,10 @@ def prepare_pec(
             raise IBMInputValueError(
                 "shots_per_randomization must be the same for twirling and measure_noise_learning"
             )
-        trex_item = create_trex_calibration_circuit(
-            pubs, measure_noise_learning, twirling_num_randomizations
+        trex_num_randomizations = resolve_trex_num_randomizations(
+            measure_noise_learning, twirling_num_randomizations
         )
+        trex_item = create_trex_calibration_circuit(pubs, trex_num_randomizations)
         quantum_program.items.append(trex_item)
         passthrough_data["post_processor"]["measure_mitigation"] = "True"
 
--- a/qiskit_ibm_runtime/executor_estimator/prepare.py
+++ b/qiskit_ibm_runtime/executor_estimator/prepare.py
@@ -164,7 +164,9 @@ def prepare(
             raise IBMInputValueError(
                 "shots_per_randomization must be the same for twirling and measure_noise_learning"
             )
-        trex_item = create_trex_calibration_circuit(pubs, measure_noise_learning)
+        trex_item = create_trex_calibration_circuit(
+            pubs, measure_noise_learning, num_randomizations
+        )
         quantum_program.items.append(trex_item)
         passthrough_data["post_processor"]["measure_mitigation"] = "True"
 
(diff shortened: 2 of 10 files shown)


qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
```diff
@@ -97,6 +97,10 @@ def prepare_pec(
         twirling_options.num_randomizations,
         twirling_options.shots_per_randomization,
     )
+    # Preserve the twirling randomization count: ``num_randomizations`` is scaled
+    # per-pub by the PEC sampling overhead below, but the TREX calibration should
+    # follow the (unscaled) twirling value.
+    twirling_num_randomizations = num_randomizations
```
leonapivato, line 103: Just wanted to confirm that TREX should use the num_randomizations prior to the scaling done below for pec. @SamFerracin @TsafrirA

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -164,7 +164,9 @@ def prepare(
             raise IBMInputValueError(
                 "shots_per_randomization must be the same for twirling and measure_noise_learning"
             )
-        trex_item = create_trex_calibration_circuit(pubs, measure_noise_learning)
+        trex_item = create_trex_calibration_circuit(
+            pubs, measure_noise_learning, num_randomizations
+        )
```
SamFerracin: I'd suggest we keep it simple. The function to resolve is a single line, so we can keep the logic in the prepare functions instead. And we pass a single value (named simply `num_randomizations`, instead of `twirling_num_randomizations`) to `create_trex_calibration_circuit` rather than two
```suggestion
        num_randomizations = num_randomizations if measure_noise_learning.num_randomizations is "auto" else measure_noise_learning.num_randomizations
        trex_item = create_trex_calibration_circuit(
            pubs, num_randomizations
        )
```

qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
```diff
@@ -30,8 +30,12 @@ class MeasureNoiseLearningOptions:
         technique that requires measurement noise learning.
     """
 
-    num_randomizations: int = 32
-    """The number of random circuits to draw for the measurement learning experiment."""
+    num_randomizations: int | Literal["auto"] = "auto"
+    """The number of random circuits to draw for the measurement learning experiment.
+
+    If "auto", the calibration uses the same number of randomizations
+    as the estimation twirling.
```
SamFerracin:
```suggestion
    as specified in the estimator's twirling options.
```

qiskit_ibm_runtime/executor_estimator/pec/prepare_pec.py
SamFerracin: yes, this seems correct to me

qiskit_ibm_runtime/executor_estimator/prepare.py
leonapivato: If we plan to incorporate the "num_term" literal as an option for num_randomziations, it might be easier to have a utils function `resolve_trex_num_randomizations`. I agree though that it is better to do this call/calculation in the prepare function itself.

What do you think, are we still planning to include the num_terms literal? Should the utils function be moved into prepare vs the raw calculations?

SamFerracin: I am not opposed to having a util function instead of the single line. Feel free to centralize that line to a util used in every prepare if you prefer 😄 And the if and when we need to change it in the future, we have one place to change it

I would move it to prepare though, and just pass `num_randomizations` to the trex util. That's the part that I wanted to point out more strongly

qiskit_ibm_runtime/executor_estimator/trex_utils.py
```diff
@@ -30,16 +30,43 @@
...
 ) -> SamplexItem:
     """Creates a TREX calibration circuit.
 
     The calibration circuit is based on all circuit terminal measurement layers in all pubs.
 
     Args:
         pubs: List of estimator pubs to extract relevant qubits from.
-        measure_noise_learning: Measure noise learning options.
+        num_randomizations: The number of TREX calibration randomizations (already
+            resolved, see :func:`resolve_trex_num_randomizations`).
```
SamFerracin: This is the sphynx rendering issue that I was mentioning in the team meeting
```suggestion
            resolved, see :func:`~.resolve_trex_num_randomizations`).
```

qiskit_ibm_runtime/executor_estimator/trex_utils.py
```diff
@@ -30,16 +30,43 @@
...
 
 
+def resolve_trex_num_randomizations(
+    measure_noise_learning: MeasureNoiseLearningOptions,
+    twirling_num_randomizations: int,
+) -> int:
+    """Resolve the number of TREX calibration randomizations.
+
+    ``measure_noise_learning.num_randomizations`` may be an explicit ``int`` or
+    ``"auto"`` (the default). When ``"auto"``, the TREX calibration uses the same
```
SamFerracin: This may change in the future, and there will is no automated way for us to remember taht we need to change this docstring
```suggestion
    ``measure_noise_learning.num_randomizations`` may be an explicit ``int`` or
    ``"auto"``. When ``"auto"``, the TREX calibration uses the same
```

SamFerracin (approved):

PR #3036, opened 2026-07-14
Fix parameter binding for wrapper sampler and estimator

### LINKED ISSUE #3024: Executor Sampler: dict-keyed parameter values bind in the wrong order
## Summary

The executor-based `SamplerV2` flattens a pub's parameter values to a positional
array **without** passing the circuit's parameter order. For a pub whose values are
given as a dict / `BindingsArray` whose key order differs from the circuit's
canonical `circuit.parameters` order, the values bind to the **wrong parameters** —
silently, with no error and no warning.

## Where

- `qiskit_ibm_runtime/executor_sampler/prepare.py:103` (no-twirl path) and `:137`
  (twirl path) — both call `pub.parameter_values.as_array()`.

## Problem

`BindingsArray.as_array()` called with no argument *"preserves the order of both the
dict and the parameters in the keys"* (qiskit `bindings_array.py:189-194`) — i.e. the
user's dict insertion order. `SamplerPub.coerce` does not re-canonicalize a dict
(`sampler_pub.py:135` → `BindingsArray.coerce(values)`), so the keys stay in user
order.

The resulting `CircuitItem.circuit_arguments` is a bare positional float array with
no parameter names (`quantum_program/quantum_program.py:82-114`), so the executor can
only bind it positionally against the circuit's canonical `circuit.parameters`
(sorted). When the dict order ≠ canonical order, parameters are swapped.

Reproduced offline: circuit with params canonical `[a, b]`, values `{b: 0.0, a: π/2}`
→ wrapper array `[0.0, 1.57]` (dict order b, a) vs correct `[1.57, 0.0]` (canonical
a, b). `a` and `b` are silently swapped.

- **Trigger:** dict / `BindingsArray`-keyed multi-parameter pub with non-canonical key order.
- **Safe:** plain positional-array pubs — `SamplerPub.coerce` wraps them as
  `{tuple(circuit.parameters): values}`, so they're already canonical.
- **Impact:** silent wrong results (no exception). Legacy binds by name → unaffected.

## Fix

Pass the circuit's parameters to `as_array` so the flattened array is ordered
canonically:

```python
# executor_sampler/prepare.py  (both the no-twirl and twirl paths)
param_values = pub.parameter_values.as_array(pub.circuit.parameters)
```

`as_array(parameters)` reorders to the given parameter order (and validates the set
matches), which is exactly the canonical order the positional `circuit_arguments`
axis is bound against.

## Also affects the estimator

The executor **`EstimatorV2`** has the same flaw: `compute_samplex_arguments`
calls `parameter_values.as_array()` with no parameter list
(`executor_estimator/utils.py:390`), feeding all four prepare paths (`prepare.py`,
`prepare_pea.py`, `pec/prepare_pec.py`, `zne/prepare_zne.py`). Same fix — pass the
pub's circuit parameters into `as_array` there too.

leonapivato:
Fixes #3024

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/utils.py
+++ b/qiskit_ibm_runtime/executor_estimator/utils.py
@@ -395,7 +395,9 @@ def compute_samplex_arguments(
     for ndindex, basis in param_basis_map.items():
         for bases in basis:
             change_basis[basis_idx] = pauli_to_ints(bases)
-            flat_parameter_values[basis_idx] = parameter_values.as_array()[ndindex]
+            flat_parameter_values[basis_idx] = parameter_values.as_array(pub.circuit.parameters)[
+                ndindex
+            ]
             basis_idx += 1
 
     # Step 4. Log info.
--- a/qiskit_ibm_runtime/executor_sampler/prepare.py
+++ b/qiskit_ibm_runtime/executor_sampler/prepare.py
@@ -100,9 +100,10 @@ def prepare(
             logger.info("Processing pub %d/%d", i + 1, len(pubs))
             validate_no_boxes(pub.circuit)
 
-            # Convert parameter values to numpy array
+            # Convert parameter values to numpy array. Pass the circuit's
+            # parameters so the columns are ordered to match ``circuit.parameters``.
             if pub.parameter_values.num_parameters > 0:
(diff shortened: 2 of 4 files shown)


SamFerracin (approved):
Thanks!

PR #3038, opened 2026-07-15
Correctly sum contributions to the std in the extrapolation module

### LINKED ISSUE #3037: `calculate_extrapolated_expectation_values` mis-calculates the standard deviations
**Describe the bug**
The `calculate_extrapolated_expectation_values` function [sums ](https://github.com/Qiskit/qiskit-ibm-runtime/blob/3d36c55fa8f4c2ddac8880c4099f21f0cadd413d/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py#L596-L598) the contributions to the std linearly. This is incorrect, as only variances can be summed like that (as is done in the next line). We need to compute the variance, and take the sqrt as a last step.

TsafrirA:
### Summary
Fixes #3037.

### Details and comments
We can do a more basic change and change the return value of `process_extrapolated_expectation_values` to be the variance instead of the standard deviation.

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
+++ b/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
@@ -540,7 +540,7 @@ def calculate_extrapolated_expectation_values(
 
         # each item should contain results for each point in extrapolated_noise_factors
         exp_vals_extrapolated = np.zeros(len(extrapolated_noise_factors), dtype=float)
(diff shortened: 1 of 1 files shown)


dekelmeirom (approved):
LGTM

SamFerracin (commented):
Good catch! This looks correct to me, thanks for fixing it!

SamFerracin (approved):

PR #3039, opened 2026-07-15
Small improvements to prepare functions

SamFerracin:
### Summary
This PR adds two small improvements to the existing functions:
- Firstly, it modifies `prepare_pec` so that when `gamma` is `1`, the calculation of `noise_gain` does not raise a "dividing-by-zero" warning, which currently shows up (eg in tests) when users try `prepare_pec` passing trivial maps.
- Secondly, it adds a graceful error message to `calculate_gamma` for the case where a ref is missing from the noise models map.
- Finally, it rewrites all the error messages for the case of missing refs consistently.

### AI/LLM disclosure


dekelmeirom (approved):
LGTM

PR #3040, opened 2026-07-15
Homogenize imports towards relative imports

diego-plan9:
### Summary

We have a mixture of absolute and relative imports in the codebase. This PR:
* converts to relative imports in `qiskit_ibm_runtime/`
* converts to relative imports in `test/` (these should stay relative, as `test` is not a proper package and can cause conflicts with other packages that might be named `test`)

### Details and comments

I could not find a way to enforce it using `ruff`. Personally, I am more in favor of absolute imports - but changes would be more intrusive and for historical reasons this PR aims for relative imports.

Fixes N/A

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/api/clients/runtime.py
+++ b/qiskit_ibm_runtime/api/clients/runtime.py
@@ -17,8 +17,7 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING, Any
 
-from qiskit_ibm_runtime.api.session import RetrySession
-
+from ...api.session import RetrySession
 from ..rest.runtime import Runtime
 from .backend import BaseBackendClient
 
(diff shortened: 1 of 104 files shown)


SamFerracin (approved):
Thanks!

When you say that you couldn't find a ruff rule, I guess that means that we rely on reviewers to check this?

diego-plan9:
> When you say that you couldn't find a ruff rule, I guess that means that we rely on reviewers to check this?

Yes, unfortunately.

TsafrirA:
Could you shed some light on the reasons behind this effort?

diego-plan9:
Consistency, which is a overall theme we pursue. In this case, having mixed styles of imports hinders readability and grep-style searches and refactors.

PR #3042, opened 2026-07-15
Use BaseModel for options_models

### LINKED ISSUE #3041: Use pydantic models for new style options

Historically, we have been using `pydantic.dataclass` as the way to defined Options, with a number of constructs on top of it (with `@primitive_dataclass` + `OptionsV2` as the main driver: a hierarchy of classes, with `arbitrary_types_allowed`, and logic for converting, extracting information, coping with older versions of the options and visualization). For new programs, we have been using `pydantic.dataclass` with a leaner configuration (no hierarchy, explicitly avoiding `Unset` and arbitrary types), which increases explicitness and simplifies implementation.

Arguably, Options are involved in:
* interface design (a way for users to configure the behavior of a program in an ergonomic discoverable way)
* validation (ensure correctness and explicitly declare the allowed values and constrains)
* serialization (used for constructing part of the payload that is sent to the API)

Keeping Options thin and using Pydantic BaseModels seems to naturally provide solutions to those needs, allowing to take advantage of pydantic features fully and remove the need for custom functionality, resulting in more maintainable code and providing more features "for free". This story is about exploring such approach.

**Acceptance criteria**

Using `BaseModels` for `option_models` is explored, and evaluated by the team.

diego-plan9:
### Summary

Proof of concept for using `BaseModel` for new program options at `options_models`:
* Add `BaseOptionsModel` as the pure pydantic base class for options
   * replicates `update()` and `asdict()` for backwards compatibility
   * trims the `dir()` return for auto-completion conciseness
* Revise existing `option_models`
   * inherit from `BaseOptionsModel` instead of using `pydantic.dataclass`
   * avoid `Field(default_factory=dict)` pattern, for readability
   * use `Annotated` pattern for validators (replacing validators for non-serializable classes with `InstanceOf`)
   * revise docstrings (length, rely on types for users discovering options, other misc tweaks)
   * focus on keeping them as containers and lean
      * conversion responsibilities moved to `converters.py`
      * revise class-level constants  
* Adjust the rest of the codebase
   * replace `asdict()` with `.model_dump()`
   * other misc adjustments
* Tests
   * add tests for the base model
   * adjust existing tests

### Details and comments

Fixes #3041

### AI/LLM disclosure


diego-plan9:
Marking as draft, as this is intended to be a vehicle for discussion (and can be re-issued as smaller PRs as needed).

qiskit_ibm_runtime/options_models/base.py
```diff
@@ -0,0 +1,89 @@
...
+
+        This provides support for applying `dataclasses.asdict()` by simulating that the model has
+        dataclass fields (populating the variable inspected by `asdict()`). It is a brittle
+        approach and only intended to provide some level of supports while users transition to
+        using `model_dump()` directly.
+        """
+        warn(
+            "Using `dataclasses.asdict()` on option models is deprecated as of qiskit_ibm_runtime "
+            "v0.49.0 and will be removed in a future release. Please use Pydantic features for "
+            "converting to dict (`options.model_dump()`) instead.",
```
SamFerracin, line 68: Nice! I like this idea of adding tech debt for backward compatibility, but with a deprecation warning from the get-go. It's the best that we can do, imo

diego-plan9: It's also a bit tied to the rollout strategy (and likely a follow-up prior to next release). It might make sense to emit the deprecation warnings for the older options as well (or just for them, depending on how to evolve). In the meantime, the warnings are mostly a signal of direction.

SamFerracin:
This looks good to me @diego-plan9. It is a change that comes at no cost for the user (it's a drop-in-replacement change), but makes our lives easier as we can use pydantic's built-in methods

qiskit_ibm_runtime/options_models/simulator.py
```diff
@@ -48,13 +47,15 @@ class SimulatorOptions:
     seed_simulator: int | None = None
     """Random seed to control sampling."""
 
-    coupling_map: list[list[int]] | CouplingMap | None = None
+    coupling_map: (
+        list[list[Annotated[int, Field(ge=0)]]] | Annotated[CouplingMap, InstanceOf] | None
+    ) = None
```
diego-plan9, line 51: This change involves reverting the coercing into `CouplingMap` that was taking place at the validator. The main motivation is adhering to how `coupling_map` is declared (in this PR, and in `main`, we are signalling "this field can contain lists, `CouplingMap` or `None`", but in practice in `main` it could not contain lists due the coercion).

I have adjusted the tests accordingly.

qiskit_ibm_runtime/options_models/base.py
diego-plan9: In the end, this is the only `DeprecationWarning` kept in this PR. While it is possible to use `asdict()`, we are not aspiring to full dataclass compatibility, and since there is already a clean alternative in place, I believe we are better off encouraging users to move already.

qiskit_ibm_runtime/options_models/base.py
```diff
@@ -0,0 +1,80 @@
...
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
diego-plan9, line 71: Minor note that this warning will be issued more than once if there are nested models. Technically we could inspect the stack and attempt to prevent it, but we don't control the original call (ie. it is `asdict()`, this function is invoked under the hood) and detecting if it is the "initial" call or a nested one seems quite dependent on `dataclass.asdict()` implementation, which might be different between Python versions. I have not found a clean way to do it - so opting explicitly for potentially multiple warnings (as the alternative is risking not emitting a warning).

SamFerracin (approved):

PR #3044, opened 2026-07-15
Remove `MeasureNoiseLearningOptions.shots_per_randomization` attribute in executor-based estimator

SamFerracin:
### Summary
In the Executor-based Estimator, every TREX circuit is executed with the same number of shots as every other item in the program. This PR removes the ability to specify `MeasureNoiseLearningOptions.shots_per_randomization`, which is currently ignored after PR #3035, and had probably been forgotten

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
+++ b/qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
@@ -33,6 +33,6 @@ class MeasureNoiseLearningOptions:
     num_randomizations: int | Literal["auto"] = "auto"
     """The number of random circuits to draw for the measurement learning experiment.
 
-    If "auto", the calibration uses the same number of randomizations
+    If ``"auto"``, the calibration uses the same number of randomizations
     as specified in the estimator's twirling options.
     """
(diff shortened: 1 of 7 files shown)


qiskit_ibm_runtime/options_models/measure_noise_learning_options.py
```diff
@@ -36,9 +36,3 @@ class MeasureNoiseLearningOptions:
     If "auto", the calibration uses the same number of randomizations
```
leonapivato: Unrelated to the changes in this PR, but is this supposed to be ``auto`` instead of "auto"?

leonapivato:
```suggestion
    If ``auto``, the calibration uses the same number of randomizations
```

if that is the case

SamFerracin: It's `"auto"` actually: [a23a1bc](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3044/commits/a23a1bca78bba8984bc65349d968a28200e00e84)

leonapivato (approved):

PR #3047, opened 2026-07-16
Convert `measure_mitigation` from str to bool

TsafrirA:
### Summary

The `measure_mitigation` field in the metadata is stored as string. This PR converts it to bool. Also consolidates the field setting a bit.


SamFerracin (commented):
This looks good in principle, but I wonder if we did this intentionally due to the fact that in the to/from json conversion, we get a string back from server rather than a bool. Could you do a quick check by running on a test heron, and see if we get a bool?

TsafrirA:
@SamFerracin 
> and see if we get a bool?

Yes, it goes through end to end.

SamFerracin (approved):

PR #3049, opened 2026-07-16
Revise IBMTestCase, cleanup test utils and extra code

### LINKED ISSUE #3045: Revise IBMTestCase and common test functionality

Over time, `IBMTestCase` and the common tests facilities have grown a bit organically, with some unused items and extra complexity.

It would be good to take a pass at `IBMTestCase`, the hierarchy, utils, aimed at reducing complexity. 

Some small relation to #2908 (both as an exercise in gaining knowledge of the test facilities, and ideally paving a way for it).

**Acceptance criteria**

`IBMTestCase` is slimmed down, and more test cases inherit from it.

diego-plan9:
### Summary

Round of revising `IBMTestCase` and its hierarchy, aimed at streamlining `IBMTestCase` and overall simplifying / cleaning:
* Revise `IBMTestCase` for lightness when it comes to setup:
   * Remove custom setup for logging (as same features can be achieved by `pytest` and handled at test-running time)
   * Remove custom hiding of warnings
   * Remove attributes and methods that are needed just by specialized tests
   * Consolidate some methods that provides assertions into this class
* Revise tests:
   * Ensure they inherit from one of our TestCases (disabling inheriting from `unittest.TestCase` via `ruff`)
   * Cleanup unused functions and files
   * Minor adjustments to tests, focused on reusing functionality

### Details and comments

Fixes #3045

Some comments inline.

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: IBM Bob for further catching unused functionality


diego-plan9:
Issuing this against the main repo for extra assurances when it comes to integration tests.

test/ibm_test_case.py
```diff
@@ -45,149 +43,78 @@
...
+                        return False
+
+            return True
+
+        if not _dict_keys_equal(a, b, exclude_keys):
+            raise AssertionError(f"Dicts don't have the same keys: {a}, {b}")
 
     @contextmanager
-    def assert_warning_appears(
+    def assertWarnsStrict(
```
diego-plan9, line 117: Camelcase is an stylistic choice, so we get closer to `unittest.TestCase` other `assertFoo` methods.

test/ibm_test_case.py
```diff
@@ -45,149 +43,78 @@
...
+                if key not in dict2:
+                    return False
+                if isinstance(val, dict):
+                    if not _dict_keys_equal(val, dict2[key]):
+                        return False
+
+            return True
+
+        if not _dict_keys_equal(a, b, exclude_keys):
+            raise AssertionError(f"Dicts don't have the same keys: {a}, {b}")
```
diego-plan9, line 114: These come from `utils.py` - with an extra layer on top for trying to be in harmony with `unittest.testCase.assertDict`.

.github/workflows/ci.yml
```diff
@@ -127,10 +125,8 @@ jobs:
       QISKIT_IBM_URL: ${{ secrets.QISKIT_IBM_URL }}
       QISKIT_IBM_INSTANCE: ${{ secrets.QISKIT_IBM_INSTANCE }}
       QISKIT_IBM_QPU: ${{ secrets.QISKIT_IBM_QPU }}
-      LOG_LEVEL: DEBUG
-      STREAM_LOG: True
       QISKIT_IN_PARALLEL: True
-      PYTEST_ADDOPTS: "-v -s --durations=0"
+      PYTEST_ADDOPTS: "-v -s --durations=0 --log-level=DEBUG"
```
SamFerracin, line 129: how many new warnings do we gain from this change?

test/ibm_test_case.py
SamFerracin: But do we have an `assertDict` available now? I only see methods to assert dict keys, not values

SamFerracin (commented):
This looks good, one question about the assertion for dicts

.github/workflows/ci.yml
diego-plan9: We should be having the same level of _log entries_  - this has the same effect as the `LOG_LEVEL: DEBUG` removed above. Basically, `IBMTestCase` has some machinery to "set the log level in the tests to the `LOG_LEVEL` variable", through https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.48.0/test/ibm_test_case.py#L60. This PR removes that machinery and just uses the pytest equivalent flag.

diego-plan9: If you refer to actual `Warnings`:

```
main:    1815 passed, 4 skipped, 2211 warnings
this PR: 1815 passed, 4 skipped, 2252 warnings
```

This is due to removing https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.48.0/test/ibm_test_case.py#L65-L70. A bit of a regression but a drop in the ocean, and we have https://github.com/Qiskit/qiskit-ibm-runtime/issues/2764 for tackling the warnings properly.

test/ibm_test_case.py
diego-plan9: My bad as I should have written `assertDictEqual` (ie. https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertDictEqual).

These moved functions aim to mimic the signature and wording of it.

.github/workflows/ci.yml
SamFerracin: okay, thanks. yes, no big deal, and even if it was a big deal, i agree that having these warnings out is better than not knowing about them due to silencing

SamFerracin (approved):
Thanks!

PR #3050, opened 2026-07-16
Replace mock backends with fake backends in some tests for executor-based primitives

SamFerracin:
### Summary
Now that local mode is supported in both sampler and estimator, we can change some tests so that they instantiate a fake backend rather than mocking one.

### AI/LLM disclosure


TsafrirA (approved):
LGTM

PR #3051, opened 2026-07-17
Move options-related tests to package, trim module names

diego-plan9:
### Summary

Reorganize unit tests, so options_models-related tests are in `test/unit/options_models`:
* move tests related to `estimator`, `resilience` and `sampler`
* drop tests that exercise assignment, as it should be covered by other tests and/or regular functionality
* some tweaks to some tests

Additionally, renames the `options_models` module names removing the `_option` suffix, for conciseness and consistency with other recently introduced modules (`results`,  `decoders`).

### Details and comments

Related to #2908, #3042

This is actually part of #3042, and one of the bits that cause more conflicts. This change is independent (ie. it improves organization by containing options-related tests in the predictable place), and also issuing this PR as a separate one helps reduce review surface.

### AI/LLM disclosure


SamFerracin (approved):
Thanks for doing this, Diego!

PR #3062, opened 2026-07-20
Raise an error when wrapper estimator is run with empty PUBs.

### LINKED ISSUE #3061: Wrapper estimator fails ungracefully with empty PUBs
**Describe the bug**
When the wrapper estimator is run with empty PUBs, the precision fails to resolve, and the error message reads `All pubs must have the same precision`. This, in contrast to the legacy estimator which will run, and return a failed job (`The job contains no PUBs. Please submit a job with valid PUBs.`).

**Expected behavior**
Empty PUBs should fail gracefully, or return a failed job (preferably without submitting to the server?).

TsafrirA:
### Summary
Fixes #3061 

I opted to raise an error instead of returning a failed job (or sending an empty payload to the server for it to fail). This deviates from legacy behavior, but I think it's for the better.

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -265,11 +265,13 @@ class EstimatorV2(BaseEstimatorV2):
 
         Raises:
             ValueError: If backend is not provided.
-            IBMInputValueError: If precision is not properly specified or if unsupported
-                options are detected.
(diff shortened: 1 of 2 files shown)


SamFerracin (approved):
thanks!

diego-plan9 (approved):
Thanks, I was checking if there was some relation to the "inadvertent iteration exhaustion" similar to with #2759, but I think we are good. As far as I can see, the iterable is consumed when creating `coerced_pubs` in `run()` already, and this does not change it.

PR #3065, opened 2026-07-20
Improve the error handling of unsupported metadata types in `Executor.run` 

### LINKED ISSUE #3063: Improve wrapper sampler `metadata` handling
For the wrapper sampler, `metadata` on the `QuantumCircuit` is converted to `passthrough_data` on the `QuantumProgram` and thus is no longer needed when running with the `Executor`. We should update things to clear the metadata so that an example like:

```
circuit.metadata['np_type'] = np.int64(4)

for q in range(circuit.num_qubits):
    circuit.x(qubit=q)
circuit.measure_all()

isa_circuit = pm.run(circuit)

sampler = SamplerV2(backend)
job = sampler.run([circuit])
job.job_id()
```
can run without encountering a `TypeError: Object of type int64 is not JSON serializable` from the QPY encoding of the circuits.

**Acceptance criteria**

- [ ]  Update the wrapper sampler to clear circuit metadata without mutating the user's original circuits
- [ ] (Stretch) Add a more informative error message

apchytr:
### Summary

Given an example such as:
```
circuit = QuantumCircuit(100)

for q in range(circuit.num_qubits):
    circuit.x(qubit=q)
circuit.measure_all()

circuit.metadata['np_type'] = np.int64(4)

isa_circuit = pm.run(circuit)

qp = QuantumProgram(shots=10, passthrough_data={"np_type": np.int64(4)})
qp.append_circuit_item(circuit=isa_circuit)

executor: Executor = Executor(mode=backend)
job = executor.run(qp)
```
the error message raised looks like `TypeError: Object of type int64 is not JSON serializable` which is not particularly informative for the user. Here we update that to `TypeError: Program metadata contains a field of unsupported type 'int64'.`

### Details and comments

Closes #3063 stretch goal

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor/executor.py
+++ b/qiskit_ibm_runtime/executor/executor.py
@@ -117,13 +117,21 @@ class Executor:
 
         Returns:
             A job.
+
+        Raises:
+            ValueError: If converters are not found for a given schema version.
+            TypeError: If the program metadata contains an unsupported type.
         """
         try:
             converter = QUANTUM_PROGRAM_PARAMS_CONVERTERS[self._SCHEMA_VERSION]
         except KeyError:
             raise ValueError(f"No converters for schema version {self._SCHEMA_VERSION}.")
 
-        params = converter.encoder(program, self.options)
+        try:
+            params = converter.encoder(program, self.options)
+        except TypeError as e:
+            type_name = str(e).split("type ")[-1].split(" is not")[0]
+            raise TypeError(f"Metadata contains a field of unsupported type '{type_name}'.") from e
         runtime_options = asdict(self.options.environment)  # type: ignore[call-overload]
         runtime_options["backend"] = self._backend.name
         runtime_options["instance"] = self._backend._instance
--- a/qiskit_ibm_runtime/executor/executor.py
+++ b/qiskit_ibm_runtime/executor/executor.py
@@ -131,7 +131,9 @@ class Executor:
             params = converter.encoder(program, self.options)
         except TypeError as e:
             type_name = str(e).split("type ")[-1].split(" is not")[0]
-            raise TypeError(f"Metadata contains a field of unsupported type '{type_name}'.") from e
+            raise TypeError(
+                f"Program metadata contains a field of unsupported type '{type_name}'."
+            ) from e
         runtime_options = asdict(self.options.environment)  # type: ignore[call-overload]
         runtime_options["backend"] = self._backend.name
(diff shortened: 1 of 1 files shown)


qiskit_ibm_runtime/executor/executor.py
```diff
@@ -117,13 +117,23 @@ def run(self, program: QuantumProgram) -> RuntimeJobV2:
...
+            TypeError: If the program metadata contains an unsupported type.
         """
         try:
             converter = QUANTUM_PROGRAM_PARAMS_CONVERTERS[self._SCHEMA_VERSION]
         except KeyError:
             raise ValueError(f"No converters for schema version {self._SCHEMA_VERSION}.")
 
-        params = converter.encoder(program, self.options)
+        try:
+            params = converter.encoder(program, self.options)
```
SamFerracin, line 131: This can fail for reasons unrelated to metadata, such as, trivially:
```python
from qiskit_ibm_runtime.quantum_program.params_converters import QUANTUM_PROGRAM_PARAMS_CONVERTERS

converter = QUANTUM_PROGRAM_PARAMS_CONVERTERS["v1.1"]
converter.encoder("ciao", "hola")
-------------
-> 
AttributeError: 'str' object has no attribute 'items'
```

apchytr: Yes, which is why we are only catching `TypeError` here (unless I'm missing the point youre trying to raise!). I guess an extra safeguard could be to use a regex pattern just to ensure other potential `TypeError`s aren't accidentally being caught here

SamFerracin: My point is: Can you point me to public documentation in QPY/ibm-quantum-schemas that says that if serialization returns a TypeError, it is most definitely due to the fact that the circuit's metadata are not serializable, and it cannot be anything else other than that?

SamFerracin: > I guess an extra safeguard could be to use a regex pattern

Similarly, can you point to public docs that says that "Object of type ndarray is not JSON serializable" is only given when the metadata fail to serialize?

apchytr: Ah, I see the issue is more about pointing users to the metadata. My bad I was under the assumption that the concern was we were catching unintentional errors. I think that's a fair point it's probably more accurate to point the user to the circuit as that is the object being encoded here.

PR #3068, opened 2026-07-21
Change `mypy` config to not ignore broken imports

TsafrirA:
### Summary
Mypy config is changed to catch broken imports within the repo. Dependencies with no stubs are still ignored in this regard.

Some broken imports were fixed.

Thanks @diego-plan9 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,6 @@ fallback_version = "0.46.1"
 python_version = "3.10"
 packages = ["qiskit_ibm_runtime", "test"]
 namespace_packages = true
-ignore_missing_imports = true
 warn_redundant_casts = true
 warn_unreachable = true
 strict_equality = true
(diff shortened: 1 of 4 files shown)


diego-plan9 (commented):
Thanks - can you extend the scope of this PR to do a round of looking into all the `if TYPE_CHECKING` blocks and fixing those? We had a few of them creeping past, as our current configuration does not catch wrong imports inside the block.

`mypy` will likely help, although it needs some deeper revision. Can you add:

```toml
[[tool.mypy.overrides]]
module = [
    "ibm_cloud_sdk_core.*",
    "ibm_platform_services.*",
    "ibm_quantum_schemas.*",
    "numpy.*",
    "plotly.*",
    "pydantic.*",
    "qiskit_aer.*",
    "qiskit.*",
    "requests_ntlm.*",
    "rustworkx",
    "samplomatic.*",
    "urllib3.*",
]
ignore_missing_imports = true
```

after the existing `[[tool.mypy.overrides]]` entry, and remove the `ignore_missing_imports=True` in the `[tool.mypy]` block?

This would hopefully surface the `qiskit-ibm-runtime` problematic imports (while not warning about the rest), and we will make it a bit cleaner once we have time revising the `mypy` configuration more thoroughly (we should not need to explicitly exclude all those modules when it comes to imports eventually). You might need to remove the `mypy` cache if iterating locally.

TsafrirA:
@diego-plan9 
I am seeing a lot of `[import-not-found]` errors, but all of them seem to be from other packages without stubs (qiskit, numpy, etc).

Should we replace `ignore_missing_imports = true` with `disable_error_code = ["import-untyped", "import-not-found"]
`?

diego-plan9 (approved):
Thanks!

PR #3069, opened 2026-07-21
Reject NaN RZZ angles during primitive validation

### LINKED ISSUE #2452: rzz angle validation could check for nan

The rzz validation code in these two spots:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/32360252e9d5c8fef2b1f569519c947057e47187/qiskit_ibm_runtime/utils/utils.py#L151

https://github.com/Qiskit/qiskit-ibm-runtime/blob/32360252e9d5c8fef2b1f569519c947057e47187/qiskit_ibm_runtime/utils/utils.py#L205

currently does not check for `nan` as the angle value but it could.

**Acceptance criteria**

`is_valid_rzz_pub` returns an error message for an rzz gate with an angle of `nan`. Example code:

```python
from qiskit.primitives.containers.sampler_pub import SamplerPub

from qiskit_ibm_runtime.utils.validations import is_valid_rzz_pub

qc = QuantumCircuit(2)
qc.rzz(float("nan"), 0, 1)
assert is_valid_rzz_pub(SamplerPub(qc)) != ""
```

sankalpsthakur:
### Summary

Reject fixed and parameterized `RZZ` angles whose evaluated value is `NaN`. This keeps invalid input inside the existing `IBMInputValueError` validation path instead of accepting a fixed NaN or leaking a raw Qiskit binding error.

### Details and comments

- evaluate parameter expressions with `bind_all` before validating the resulting angle
- preserve valid expressions where a NaN-valued parameter is algebraically eliminated
- extend fixed and parameterized RZZ regression coverage, including the eliminated-NaN case
- add the required Towncrier bug-fix fragment

Validation:

- `python -m pytest -q test/unit/test_sampler.py`: 43 passed, 3 subtests passed
- `ruff check` on touched Python files: passed
- `ruff format --check` on touched Python files: passed
- `mypy --no-incremental` on touched Python files: passed
- `git diff --check`: passed
- additional automated code review: no findings after the eliminated-parameter revision

## Why this matters

NaN gate angles should fail fast at validation time with a clear error, not propagate into backend execution or produce undefined results on real hardware.

## Related

Fixes #2452

### AI/LLM disclosure

- [x] I used OpenAI Codex to help write this PR description.
- [x] I used OpenAI Codex to generate or modify code and tests.

### DIFF
--- a/release-notes/unreleased/2452.bug.rst
+++ b/release-notes/unreleased/2452.bug.rst
@@ -1,0 +1,3 @@
+Reject fixed and parameterized ``RZZ`` angles that evaluate to ``NaN`` during
+primitive input validation, instead of accepting the value or leaking a raw
+Qiskit binding error.
--- a/qiskit_ibm_runtime/utils/utils.py
+++ b/qiskit_ibm_runtime/utils/utils.py
@@ -165,7 +165,7 @@ def _is_valid_rzz_pub_helper(circuit: QuantumCircuit) -> str | set[Parameter]:
             angle = instruction.operation.params[0]
             if isinstance(angle, ParameterExpression):
                 angle_params.add(angle)
-            elif angle < 0.0 or angle > np.pi / 2 + 1e-10:
+            elif np.isnan(angle) or angle < 0.0 or angle > np.pi / 2 + 1e-10:
                 return (
                     "The instruction rzz is supported only for angles in the "
                     f"range [0, pi/2], but an angle ({angle}) outside of this "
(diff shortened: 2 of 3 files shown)


sankalpsthakur:
Closing; no maintainer review in 14+ days.

release-notes/unreleased/2452.bug.rst
```diff
@@ -0,0 +1,3 @@
+Reject fixed and parameterized ``RZZ`` angles that evaluate to ``NaN`` during
+primitive input validation, instead of accepting the value or leaking a raw
+Qiskit binding error.
```
yaelbh, line 3: I understand what you mean by "leaking a raw Qiskit binding error", because I went over the PR, but a reader of the release notes will not understand.

yaelbh (commented):
I agree with checking for `NaN`, as suggested in #2452, but I disagree with allowing `NaN` as long as it doesn't affect the parameter expression evaluation. I'd assume that no user means to submit a `NaN` value, so it should raise anyway.

What's the situation with other gates, such as `rz` and `rx`? We should be consistent in the behavior, across gates. And, if possible, we should write code in one place for all the gates. For example, we can check parameter values, regardless of the gates, and raise if any value is `NaN`.

Issue #3046, opened 2026-07-16
Align wrapper estimator's result metadata with legacy estimator


TsafrirA:
**What is the expected enhancement?**
The wrapper estimator needs to return the same metadata with its results as does the legacy estimator.

To this end we need to:

- [ ] Map legacy metadata under various conditions (twirling on\off, trex on\off, PEC, PEA, ZNE)
- [ ] Make sure the wrapper estimator matches the legacy values
- [ ] Decide if we want to include more metadata based on the underlying Executor metadata

**Acceptance criteria**
Checklist is completed, and the wrapper estimator metadata is aligned with the legacy implementation.

PR #3072, opened 2026-07-21
Correctly set quantum program metadata in wrapper estimator

TsafrirA:
### Summary
Part of #3046 

This PR fixes the `QuantumProgram` level metadata. `shots` and `target_precision` are added. In the legacy estimator they were set at the pub level, but now they are set at the program level, so it makes sense to have those values here (as well as the pub metadata? TBD).

### AI/LLM disclosure


qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -365,12 +365,19 @@ def run(
...
+        # Prepare and set program metadata (assuming passthrough is correctly configured)
+        program_metadata = asdict(options)  # type: ignore[call-overload]
+        for flag_key, options_key in [
+            ("zne_mitigation", "zne"),
+            ("pec_mitigation", "pec"),
+            ("measure_mitigation", "measure_noise_learning"),
+        ]:
+            if not program_metadata["resilience"][flag_key]:
+                program_metadata["resilience"].pop(options_key)
+        program_metadata["target_precision"] = resolved_precision
```
TsafrirA: FYI @leonapivato these fields were not present at this level in the legacy implementation.

test/unit/executor_estimator/test_estimator_v2.py
```diff
@@ -249,17 +249,55 @@ def test_run_adds_options_to_passthrough_data(self):
...
         self.assertEqual(options_metadata["dynamical_decoupling"]["enable"], False)
         self.assertEqual(options_metadata["resilience"]["measure_mitigation"], True)
 
+    @data(
+        ("zne_mitigation", "zne"),
+        ("pec_mitigation", "pec"),
+        ("measure_mitigation", "measure_noise_learning"),
+    )
+    @unpack
+    def test_run_drops_inactive_resilience_sub_options(self, flag_key, options_key):
```
SamFerracin: This test introduces the same kind of issues that we will need to address in issue #2908: It is a big test with a mock, which just tries to test something trivial.

I suggest you:
- Write a helper function that takes `options` and returns metadata
- Rewrite this test for that function

After we write the high-level `prepare` that dispatches to the `prepare_*` functions and returns a quantum program, we can add a unit test to see that the metadata are added to the quantum program

TsafrirA: Done. https://github.com/Qiskit/qiskit-ibm-runtime/pull/3072/commits/fe75b0dfc9deb236ee5848d2ffcda6193e25208b

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -363,14 +396,9 @@ def run(
...
-        resilience_options = asdict(options.resilience)  # type: ignore[call-overload]
-        resilience_options.pop("noise_model_mapping")
-        # Serialize options (assuming passthrough is correctly configured)
-        quantum_program.passthrough_data["post_processor"]["options"] = {  # type: ignore[index, call-overload]
-            "twirling": asdict(options.twirling),  # type: ignore[call-overload]
-            "dynamical_decoupling": asdict(options.dynamical_decoupling),  # type: ignore[call-overload]
-            "resilience": resilience_options,
-        }
+        # Prepare and set program metadata (assuming passthrough is correctly configured)
+        program_metadata = build_program_metadata(options, resolved_precision, shots)
```
SamFerracin: Why is this built inside prepare? Is it because in post-process, you don't have access to `resolved_precision`?
Not opposed in principle to this, just trying to understand

TsafrirA: My initial plan was to do it on the post processing side, but that required me to add a lot of things to the pass through data (or resolve things from the data, like number of shots), so eventually I decided to just do it on the prepare side.

In the future we could add more fields on the post processing side if needed.

SamFerracin: Two thoughts:
- The passthrough data are becoming kind of confusing, as they have quite a few fields now
- In creating the metadata, we are doing a lot of gymnastic with the options. `build_program_metadata` (which is really a `build_result_metadata`) is one, and the other ones are:
```python
        # Serialize options (assuming passthrough is correctly configured)
        quantum_program.passthrough_data["post_processor"]["options"] = {  # type: ignore[index, call-overload]
            "twirling": self.options.twirling.model_dump(),
            "dynamical_decoupling": self.options.dynamical_decoupling.model_dump(),
            "resilience": self.options.resilience.model_dump(exclude={"noise_model_mapping"}),
        }
```

### Proposal:
Let's just do
```python
options_dict = asdict(options)
options_dict["resilience"].pop("noise_model_mapping")  # since this would be a lot of data
quantum_program.passthrough_data["post_processor"]["options"] = options_dict
quantum_program.passthrough_data["post_processor"]["shots"] = shots  # needed?
quantum_program.passthrough_data["post_processor"]["resolved_precision"] = resolved_precision  # needed?
```
and have the full set of options in the result. This will keep the passthrough data clean, while giving us all the info that we need in the result.

TsafrirA: [7b1e7c2](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3072/commits/7b1e7c2ca3df21dff3af1f209efcf313edb3cde0)

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -341,12 +341,10 @@ def run(
...
                 quantum_program=quantum_program,
             )
-
-        quantum_program.passthrough_data["post_processor"]["options"] = {  # type: ignore[index, call-overload]
-            "twirling": options.twirling.model_dump(),
-            "dynamical_decoupling": options.dynamical_decoupling.model_dump(),
-            "resilience": options.resilience.model_dump(exclude={"noise_model_mapping"}),
-        }
+        # Store raw options, shots and precision for post-processing side to compute metadata
+        quantum_program.passthrough_data["post_processor"]["options"] = options.model_dump()  # type: ignore[index, call-overload]
```
SamFerracin, line 345: What is your take on this?
```suggestion
        quantum_program.passthrough_data["post_processor"]["options"] = options.model_dump(exclude={"resilience": {"noise_model"}})  # type: ignore[index, call-overload]
```
Mine is that these noise models can become pretty big, and therefore we could consider removing them. But we can also evaluate later

SamFerracin (approved):
Thanks, I appreciate you making the changes! 
Approved, and I left a non-blocking comment/question

PR #3075, opened 2026-07-21
Fix docs for Estimator find_unique_layers

### LINKED ISSUE #3074: PEC noise-model recipe errors on the default (or silently wastes QPU)
**Component:** `executor_estimator` `EstimatorV2` — building `resilience.noise_model_mapping` via
`find_unique_layers` + `NoiseLearnerV3` + `NoiseLearnerV3Results.to_dict`.

## Symptom

Building a PEC noise-model mapping the documented way — pass `find_unique_layers` output to
`NoiseLearnerV3.run`, then `to_dict` — raises on the default `require_refs=True`:

```
ValueError: Found an instruction without an inject noise annotation.
            Consider setting 'require_refs' to ``False``.
```

Taking that suggestion (`require_refs=False`) makes it run, but the job then spends real QPU
TREX-characterizing a full-width measurement box whose result is discarded — it has no `InjectNoise` ref, so
it can't enter `noise_model_mapping`, and readout mitigation is done by an in-run TREX pass regardless.

## Repro

```python
est = EstimatorV2(mode, options)          # options.resilience.pec_mitigation = True
layers  = est.find_unique_layers(pubs)    # gate layer(s) + an appended measurement box
results = NoiseLearnerV3(mode).run(layers).result()
mapping = results.to_dict(layers)         # <-- raises: measurement box has no InjectNoise ref
```

`find_unique_layers` returns an **extra measurement box** on top of the gate layer(s). Inspecting its output
on Bell/ZZ (`inject_noise` on):

```python
for layer in est.find_unique_layers(pubs):
    ann = get_annotation(layer.operation, InjectNoise)
    print(layer.operation.num_qubits, "ref=", ann.ref if ann else None)
# 2    ref= 'raaaA'   <- gate layer         (InjectNoise; PEC reads this)
# 156  ref= None      <- appended measurement box (no InjectNoise; nothing reads it)
```

The `156` box is the full-width measurement layer the boxing appends for execution — never present in the
user's circuit, and the box `to_dict` raises on.

## Expected vs actual

- **Expected:** the documented recipe builds the mapping with default settings.
- **Actual:** it raises; the `require_refs=False` workaround succeeds but silently adds a wasted TREX pass to
  every PEC run.

## Likely cause

`find_unique_layers` returns every unique box, including ones with no `InjectNoise` ref: the measurement box
`box_circuit` synthesizes for the *execution* path (estimator pubs carry no measurements), and — depending on
transpilation/twirling — non-injected gate layers too. None are consumed by PEC/PEA: only
`InjectNoise`-annotated layers key into `noise_model_mapping`, `prepare_pec` re-boxes independently
(`prepare_pec.py:128`), and readout is an in-run TREX calibration. So for `inject_noise` runs the recipe
should hand back only the `InjectNoise` layers; any other box is ref-less (→ the `to_dict` raise) and, if
passed to the learner, wasted QPU.

## Impact (measured, ibm_fez, Bell/ZZ)

The wasted TREX pass pushes the wrapper's `run + learn` above legacy inline learning; dropping it brings them
to parity (±~2–4 s job overhead). Largest relative cost on shallow circuits.

leonapivato:
Fixes to #3074 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -170,9 +170,7 @@ class EstimatorV2(BaseEstimatorV2):
         """Return the unique boxed layers found across the given PUBs.
 
         The returned list contains one instance of each distinct boxed layer (represented as a
-        :class:`~.CircuitInstruction`) appearing in the input PUBs. This list can be passed
-        directly to the :meth:`~.qiskit_ibm_runtime.noise_learner_v3.NoiseLearnerV3.run` method
-        for characterization, avoiding redundant learning of identical layers.
+        :class:`~.CircuitInstruction`) appearing in the input PUBs.
 
         Args:
             pubs: The list of PUBs to return a list of unique boxes for.
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -191,7 +191,7 @@ class EstimatorV2(BaseEstimatorV2):
 
             results = NoiseLearnerV3(mode).run(layers).result()
             
-            # This can be use to set the ``noise_model`` field of to ``est.options.resilience``
+            # This can be used to set the ``noise_model`` field of to ``est.options.resilience``
             # (please reword as you see fit, and check the correct place where this should end up in the options)
             noise_model = results.to_dict(layers)
 
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -173,7 +173,7 @@ class EstimatorV2(BaseEstimatorV2):
         :class:`~.CircuitInstruction`) appearing in the input PUBs.
 
         For noise learning, keep only the boxes that carry an :class:`~samplomatic.InjectNoise`
-        annotation -- their refs key the PEC/PEA ``resilience.noise_model_mapping``:
+        annotation:
 
         .. code-block:: python
 
@@ -190,6 +190,9 @@ class EstimatorV2(BaseEstimatorV2):
             ]
 
             results = NoiseLearnerV3(mode).run(layers).result()
+            
+            # This can be use to set the ``noise_model`` field of to ``est.options.resilience``
+            # (please reword as you see fit, and check the correct place where this should end up in the options)
             noise_model = results.to_dict(layers)
 
         Args:
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -190,10 +190,10 @@ class EstimatorV2(BaseEstimatorV2):
             ]
 
             results = NoiseLearnerV3(mode).run(layers).result()
-            
-            # This can be used to set the ``noise_model`` field of to ``est.options.resilience``
-            # (please reword as you see fit, and check the correct place where this should end up in the options)
             noise_model = results.to_dict(layers)
+
+            # Assign the learned model so PEC uses it on the next run.
+            est.options.resilience.noise_model_mapping = noise_model
 
         Args:
             pubs: The list of PUBs to return a list of unique boxes for.
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -172,6 +172,26 @@ class EstimatorV2(BaseEstimatorV2):
         The returned list contains one instance of each distinct boxed layer (represented as a
         :class:`~.CircuitInstruction`) appearing in the input PUBs.
 
+        For noise learning, keep only the boxes that carry an :class:`~samplomatic.InjectNoise`
+        annotation -- their refs key the PEC/PEA ``resilience.noise_model_mapping``:
+
+        .. code-block:: python
+
+            from samplomatic import InjectNoise
+            from samplomatic.utils import get_annotation
+
+            est = EstimatorV2(mode, options)
+            est.options.resilience.pec_mitigation = True
+
+            layers = [
+                layer
+                for layer in est.find_unique_layers(pubs)
+                if get_annotation(layer.operation, InjectNoise)
+            ]
+
+            results = NoiseLearnerV3(mode).run(layers).result()
+            noise_model = results.to_dict(layers)
+
         Args:
             pubs: The list of PUBs to return a list of unique boxes for.
 


qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -301,9 +301,16 @@ def find_unique_layers(
     )
     boxed_circuits = (box_circuit(circuit=pub.circuit, **pm_kwargs) for pub in pubs)
     instructions = (box for boxed_circuit in boxed_circuits for box in boxed_circuit)
-    return find_unique_box_instructions(
+    unique_layers = find_unique_box_instructions(
         instructions=instructions, normalize_annotations=None, undress_boxes=True
     )
+    if inject_noise:
```
SamFerracin: I'm not sure that this is the best way to deal with the issue you're finding.. 🤔 

This function promises to return the unique layers, and it is currently doing exactly that. Your PR introduces undocumented behaviour that goes against that promise: with this change, in some situations (i.e., when inject noise is true), this function only returns unique layers of a certain kind, not all of them.

I would suggest keeping our public interface light, and use documentation instead to explain users what they should do, namely using functions that already exist. For example:
```python
est = EstimatorV2(mode, options)
options.resilience.pec_mitigation = True

# For the purposes of noise learning, we only care about
# learning the noise of boxes that have an inject-noise annotation
layers  = [layer for layer in est.find_unique_layers(pubs) if get_annotation(layer.operation, InjectNoise)]

results = NoiseLearnerV3(mode).run(layers).result()

noise_model = results.to_dict(layers) 
```
I would place this in a guide, but why not, you could also expand the docstring of `EstimatorV2.find_unique_layers`, since this is already mentioning NLV3.

What do you think?

leonapivato: Not sure, this is the bug @TsafrirA and I had discussed briefly, and I’m curious to hear more of his thoughts on it. 

I think documentation change is inevitable though since I believe it is says that passing find_unique_layers output to
NoiseLearnerV3.run should work.

SamFerracin: I don't think it's a bug though. Everything is working correctly, no? The function is finding the unique layers correctly, the noise learner is learning the noise of all the layers that you are requesting, `to_dict` is also working as expected. The issue is that you are combining three lines of code in a suboptimal way, where you end up having to throw away results

TsafrirA: I agree with @leonapivato we have a usability issue here that is border lining a bug (everything works as documented, but you still end up wasting shots). I also agree with @SamFerracin that this is not the place to fix it.

How about we add a flag to `Estimator.find_unique_layers()` allowing the user to select whether or not noise learning is needed for the final measurement box? Or alternatively, a new method `find_unique_layers_for_something` which will do that?

dekelmeirom: I think that the docstring should clarify that `Estimator.find_unique_layers` find the unique layers **according to the options given to the Estimator**. If, for example, no mitigation and no twirling is requested in the estimator, and the user for some unknown reason calls `Estimator.find_unique_layers()`, he will get an empty list (and I don't think it is a bug).
Once we agree on that, I think that the question becomes related to whether we want to allow the user to learn the measurement noise along the gate noise (and skip the additional TREX calibration circuit in the execution) or not.
If we won't allow it - I think that `Estimator.find_unique_layers` should not return the measurement layer in  this case. If we want to allow it in the future (and I think we should) - we should add a flag either to`find_unique_layers` function or as an option in the Estimator as part of `MeasureNoiseLearningOptions` of whether the measurement noise should be learned along the gates or as part of the execution.

SamFerracin: Message that I was going to write this morning, before entering a block of meetings.
I chatted with Leo at standup, which I think is why he has changed the PR scope.

> To summarize, I think that more than a bug, we have a lack of convenience for the users. Users do have a way to run what they want, but it requires a few lines of code and know-how
> 
> Here is a proposal, then:
> - Perhaps let's use this PR to update the docs of the function, adding a snippet that shows how to use it effectively in the context of NLV3. We'll end up with a version of main that is not misleading.
> - In a future PR, we can consider making the current method more convenient, or changing it altogether--for example, in favour of a function called something like `find_layers_that_require_nlv3` (horrible name, but you get the idea).

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -169,9 +170,27 @@ def find_unique_layers(self, pubs: Iterable[EstimatorPubLike]) -> list[CircuitIn
...
+            est.options.resilience.pec_mitigation = True
+
+            layers = [
+                layer
+                for layer in est.find_unique_layers(pubs)
+                if get_annotation(layer.operation, InjectNoise)
+            ]
+
+            results = NoiseLearnerV3(mode).run(layers).result()
+            noise_model = results.to_dict(layers)
```
SamFerracin, line 193:
```suggestion
            
            # This can be use to set the ``noise_model`` field of to ``est.options.resilience``
            # (please reword as you see fit, and check the correct place where this should end up in the options)
            noise_model = results.to_dict(layers)
```

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -169,9 +170,27 @@ def find_unique_layers(self, pubs: Iterable[EstimatorPubLike]) -> list[CircuitIn
         """Return the unique boxed layers found across the given PUBs.
 
         The returned list contains one instance of each distinct boxed layer (represented as a
-        :class:`~.CircuitInstruction`) appearing in the input PUBs. This list can be passed
-        directly to the :meth:`~.qiskit_ibm_runtime.noise_learner_v3.NoiseLearnerV3.run` method
-        for characterization, avoiding redundant learning of identical layers.
+        :class:`~.CircuitInstruction`) appearing in the input PUBs.
+
+        For noise learning, keep only the boxes that carry an :class:`~samplomatic.InjectNoise`
+        annotation -- their refs key the PEC/PEA ``resilience.noise_model_mapping``:
```
SamFerracin: If you accept my other comment, this could go away:
```suggestion
        annotation:
```

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -169,9 +170,30 @@ def find_unique_layers(self, pubs: Iterable[EstimatorPubLike]) -> list[CircuitIn
...
+
+            layers = [
+                layer
+                for layer in est.find_unique_layers(pubs)
+                if get_annotation(layer.operation, InjectNoise)
+            ]
+
+            results = NoiseLearnerV3(mode).run(layers).result()
+            
+            # This can be use to set the ``noise_model`` field of to ``est.options.resilience``
```
leonapivato:
```suggestion
            # This can be used to set the ``noise_model`` field of to ``est.options.resilience``
```

SamFerracin (approved):
Thanks!

PR #3078, opened 2026-07-22
Improve the lazy import of `aer_executor` via `HAS_AER`

apchytr:
### Summary
A small improvement to how `qiskit_ibm_runtime.aer_executor` handles the lazy import of `qiskit-aer`. Namely, we can leverage `HAS_AER.require_in_call`/`HAS_AER.require_in_instance` to only check for the import of `qiskit-aer` as needed. We also update other unit tests that make use of `qiskit-aer` to be skippable if not installed.

### AI/LLM disclosure


apchytr:
If a user does not have `qiskit-aer` installed but is trying to ~~import~~ initialize `AerExecutor`:
Screenshot 2026-07-23 at 10 03 57 AM

SamFerracin:
Thanks!

What if I try to import aer-executor as: `from qiskit_ibm_runtime.aer_executor.aer_executor import AerExecutor`? We'd still get an ungraceful error if qiskit-aer is not installed, eh?

apchytr:
> Thanks!
> 
> What if I try to import aer-executor as: `from qiskit_ibm_runtime.aer_executor.aer_executor import AerExecutor`? We'd still get an ungraceful error if qiskit-aer is not installed, eh?

We'll still get the same error message! This is because any invocation of `qiskit_ibm_runtime.aer_executor` goes through the `__init__.py` so that check always happens. The only one I might be unsure of is e.g. relative imports in another file but that should never be something the user encounters

diego-plan9 (approved):
Thanks!

PR #3079, opened 2026-07-22
Typehint cleanup and renaming to backend in `aer_executor`

apchytr:
### Summary
Porting over the type hint cleanup from #3078 and renaming `qasm_simulator` to `backend` to match `Executor` 

### AI/LLM disclosure


SamFerracin (approved):
Thanks!

Could you please keep a log of all the changes that we make to the user API? So that we can provide internal users with a transition guide when we're done.

A good place that I could think of would be the epic for Debugging Tools Phase I, in the form of a TODO list in a comment somewhere

PR #3080, opened 2026-07-23
Add "physical_qubits" property to backends

### LINKED ISSUE #3071: Adjust to new hardware metrics update

Since last week, the changes on the metrics for measuring the performance of a quantum computer is live. This story is about updating the package to better expose selected features.

**Acceptance criteria**

Physical qubits is exposed through the package.

diego-plan9:
### Summary

Add `physical_qubits` as a property of `IBMBackend` and (for non-retired ones) `FakeBackendV2`.

### Details and comments

Fixes #3071

Some forces made the implementation less straightforward than ideal:
* The new property is available at the general `/backends` endpoint
* `IBMBackend` is always instantiated through `QiskitRuntimeService`
* `IBMBackend` design is quite geared towards using the configuration/properties backend-specific information 
* `QiskitRuntimeService` has several layers of state, caching, and logic, and the point in time where `/backends` is queried is separate from the point in time where backends are instantiated. 

This caused adding a new argument to `IBMBackend` constructor (which I would like to have avoided), and re-target the `QiskitRuntimeService._backends_list` into a more generic `QiskitRuntimeService_backends_info_per_instance` (so it contains all the information from the `/backends` endpoint for all instances).

Ideally, we should invest time in refactoring `IBMBackend` and `QiskitRuntimeService` for better modelling of the concepts and maintainability.

### AI/LLM disclosure


diego-plan9:
Issuing this PR against the main repo, for integration tests.

test/unit/test_fake_backends.py
```diff
@@ -119,3 +119,22 @@ def test_backend_configuration_attributes(self):
...
+            "fake_kingston",
+            "fake_marrakesh",
+            "fake_miami",
+            "fake_pittsburgh",
+        ]
+
+        if backend.name in active_backends:
+            self.assertIsInstance(backend.physical_qubits, int)
+        else:
+            self.assertIsNone(backend.physical_qubits)
```
SamFerracin, line 143: Nit: A backend that's active today will become inactive tomorrow. Therefore, I would suggest a different solution for this, such as:

```suggestion
    def test_backend_physical_qubits(self, backend):
        """Test the `physical_qubits` property of backends.
        
        `physical_qubits` was added to all the backends that were active at the time.
        [Can be rephrased]
        """
        backends_with_physical_qubits = [
            "fake_aachen",
            "fake_berlin",
            "fake_boston",
            "fake_fez",
            "fake_kingston",
            "fake_marrakesh",
            "fake_miami",
            "fake_pittsburgh",
        ]

        if backend.name in backends_with_physical_qubits:
            self.assertIsInstance(backend.physical_qubits, int)
        else:
            self.assertIsNone(backend.physical_qubits)
```

qiskit_ibm_runtime/fake_provider/backends/aachen/fake_aachen.py
```diff
@@ -24,3 +24,4 @@ class FakeAachen(FakeBackendV2):
     conf_filename = "conf_aachen.json"
     props_filename = "props_aachen.json"
     backend_name = "fake_aachen"
+    _physical_qubits = 332
```
SamFerracin, line 27: I didn't check these numbers, but I assume that they are the correct ones taken from the non-fake backends?

SamFerracin (commented):
Looks good. A small comment for you

qiskit_ibm_runtime/fake_provider/backends/aachen/fake_aachen.py
diego-plan9: Yes - they should match https://quantum.cloud.ibm.com/computers

test/unit/test_fake_backends.py
diego-plan9: Thanks, tweaked at 98a41eb.

SamFerracin (approved):
Thanks! Looks good

PR #3083, opened 2026-07-23
`aer_executor` minimal version fix - using `get_qubit_sparse_pauli_list_copy`

apchytr:
### Summary
Minimal version tests are failing with a `'qiskit.quantum_info.PauliLindbladMap' object has no attribute 'generators'` error. This comes from the fact that `generators` is a convenience introduced in `qiskit >2.4.0` for `get_qubit_sparse_pauli_list_copy`. 

### AI/LLM disclosure


diego-plan9 (approved):
Thanks for the quick fix!

PR #3088, opened 2026-07-23
Write tests to verify parameter-basis expansion performed in executor-based Estimator

### LINKED ISSUE #3087: Write tests to verify parameter-basis expansion performed in executor-based Estimator
When mapping a `pub = (circuit, observables, parameter_values)` into a quantum program item, the various `prepare_*` functions used in executor-based Estimator undertake the same steps:
- They pre-compute `evs.shape` as `np.broadcast_shapes(observables.shape, parameter_values.shape)`.
- They create a map between elements of `parameter_values` and the observables that need to be measured for each subset.
- They replace the observables in that map with the minimal set of Pauli basis that can be used to measure all such observables.

As an example, consider the case where:
- `observables = ObservablesArray(["ZZZ", "XXX", "YYY", "IYI"]).reshape((2, 2))` has shape `(2, 2)`.
- `parameter_values` has shape `(2, 2)` as well.

In this case, `evs` has shape `(2, 2)`, and such mapping amounts to:
- `((0, 0), "ZZZ")`, meaning that the "first" element in `parameter_values`, indexed with `(0, 0)`, needs to be measured with basis `"ZZZ"`
- `((0, 1), "XXX")`
- `((1, 0), "YYY")`
- `((1, 1), "IYI")`

This mapping is currently stored inside the passthrough data of the quantum program.

### Acceptance criteria:

- [ ] Add tests that demonstrate that every `prepare_*` function computes this mapping correctly.
- [ ] Make these tests consistent across every `prepare_*` function

SamFerracin:
### Summary
When mapping a `pub = (circuit, observables, parameter_values)` into a quantum program item, the various `prepare_*` functions used in executor-based Estimator undertake the same steps:
- They pre-compute `evs.shape` as `np.broadcast_shapes(observables.shape, parameter_values.shape)`.
- They create a map between elements of `parameter_values` and the observables that need to be measured for each subset.
- They replace the observables in that map with the minimal set of Pauli basis that can be used to measure all such observables.

As an example, consider the case where:
- `observables = ObservablesArray(["ZZZ", "XXX", "YYY", "IYI"]).reshape((2, 2))` has shape `(2, 2)`.
- `parameter_values` has shape `(2, 2)` as well.

In this case, `evs` has shape `(2, 2)`, and such mapping amounts to:
- `((0, 0), "ZZZ")`, meaning that the "first" element in `parameter_values`, indexed with `(0, 0)`, needs to be measured with basis `"ZZZ"`
- `((0, 1), "XXX")`
- `((1, 0), "YYY")`
- `((1, 1), "IYI")`

This mapping is currently stored inside the passthrough data of the quantum program.

This PR: 

- Add tests that demonstrate that every `prepare_*` function computes this mapping correctly.
- Make these tests consistent across every `prepare_*` function

### Details and comments

Fixes #3087

### AI/LLM disclosure


test/unit/executor_estimator/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    observables_shape: tuple[int, ...]
+    """The shape of the observables array."""
+
+    expected_pairs: Sequence[tuple[tuple[int, ...], str]]
+    """Expected mapping from each parameter index to the observable(s) assigned after
+    broadcasting the parameter and basis dimensions."""
+
+
+@dataclass(frozen=True)
+class ParamBasisTestCases:
```
diego-plan9: Small nit: is there an alternative to naming this (and the rest of the mentions in the file) as "Test Case"? Mostly to avoid confusion with `unittest.TestCase` (the concept and for practicalities).

diego-plan9 (commented):
Thanks, looks neat!

test/unit/executor_estimator/utils.py
SamFerracin: Let's see if we like "scenario" more than "test case": [test_case -> scenarios](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3088/commits/0ef447868dad542d11a7f76dbcad5ec74c6ec9e0)

diego-plan9: Sounds good, thanks!

diego-plan9 (commented):
Thanks! I think you got caught up in the coveralls / github hiccup, but this looks good to me as soon as it gets green.

SamFerracin:
@diego-plan9 thanks! Green now

TsafrirA (commented):
We can address in a separate PR, but I think the scenarios are missing some delicate cases especially when it comes to scalar params\observables.

diego-plan9 (approved):
Thanks!

I tweaked the linked PR in the description, as it was pointing to closing the epic - please double check the right issue has been closed.

> We can address in a separate PR, but I think the scenarios are missing some delicate cases especially when it comes to scalar params\observables.

Sounds reasonable, please sync as needed for potential follow-ups. With this PR in place, we start to have a good harness for structuring and completing the testing needs when it comes to expansions - it's good to have this in.

Issue #3053, opened 2026-07-17
Phase I of "Debugging Tools": Release aer-executor


SamFerracin:
**What is the expected feature?**
This epic captures the issues related to Phase I of the "Debugging Tools" saga, which requires us to take ownership of the aer-executor prototype in qiskit-noise-learning ([link](https://github.com/Qiskit/qiskit-noise-learning/tree/main/qiskit_noise_learning/aer_executor)), turn it into production-level code, and release it.

Here is a list of steps that will need to be taken in the context of this epic, with details to be elaborated in dedicated sub-issues:

- [x] Copy/paste aer-executor files from qiskit-noise-learning into qiskit-ibm-runtime.
- [x] Rename aer-executor--named TBD, cc @jyu00 
- [ ] Identify and fix existing bugs
- [ ] Identify and implement missing features
- [ ] Add tests, or improve the existing ones
- [ ] Add docs, or improve the existing ones

**Note**
aer-executor requires qiskit-aer, which is not a core dependency of qiskit-ibm-runtime. We do not want to make qiskit-aer a core dependency.

**Acceptance criteria**

- [ ] aer-executor can be installed as part of qiskit-ibm-runtime package

apchytr:
User API changes (for internal tracking):

- [x] Rename `qasm_simulator` to `backend` to match `Executor` (#3079)
- [x] Rename `AerExecutor` and `AerRuntimeJob` to `SimExecutor` and `SimRuntimeJob`, respectively (#3100)

jyu00:
Currently the way to pass a noise map to aer-executor is something like

```python
pm = generate_boxing_passmanager(tags="tag1", inject_noise=True)

samplex_arguments = {"inject_noise_ref1": some_noise_model}
map = {"tag1": another_noise_model}

executor = AerExecutor(backend=AerSimulator(), map)
job = executor.run(program)
```

There was a discussion on why `map` here is specified in `AerExecutor` constructor, when this `map` likely has a strong coupling to the input quantum program. The motivation was to avoid having to change the `run()` method invocation when switching between a simulator and a real QPU. 

This syntax might become a bit of problem when we want to add local mode support for Executor (#3055), which would allow

```python
backend = AerSimulator()
# backend = service.backend("ibm_torino")    # uncomment for real QPU

executor = Executor(backend)
job = executor.run(program)
```

[...]

PR #3089, opened 2026-07-23
Revisit tests for `aer_executor`

apchytr:
### Summary
Revisiting the tests in `test/unit/aer_executor` setting things up to continue development

### Details and comments

Part of #3053 

Changes include:

- The tests from `test_noisy` and `test_aer_executor_clifford` are moved to `test_run_quantum_program` to better reflect the responsibilities of `run_quantum_program`
- Updated `test_samplex_item_with_parameter_sweep` to instead use samplex as it was a clone of `test_circuit_item_with_circuit_arguments` 
- Renamed `test_aer_executor_clifford` to `test_aer_executor`
- Removed `test_noisy`

### AI/LLM disclosure


test/unit/aer_executor/test_run_quantum_program.py
```diff
@@ -63,6 +160,348 @@ def test_angle_rounding_snaps_near_clifford(self):
...
+
+        ba_c = BitArray.from_bool_array(result[0]["c"])
+        cts = ba_c.get_counts()
+        if noise:
+            self.assertGreater(num_shots_tot, cts.get("0" * len(active_qubits), 0))
+        else:
+            self.assertEqual(num_shots_tot, cts.get("0" * len(active_qubits), 0))
+
+    def test_samplex_item_with_parameter_sweep(self):
+        """Run a parameterized SamplexItem by supplying parameter_values directly.
```
SamFerracin: Nit, cosmetic:
```suggestion
        """Run a parameterized Samplex item by supplying parameter values directly.
```
Not sure how consistent we are, but since you changed it, this seems more sensible

test/unit/aer_executor/test_run_quantum_program.py
```diff
@@ -12,39 +12,136 @@
...
+    if n < 1:
+        raise ValueError("n must be at least one")
+    iterator = iter(iterable)
+    while batch := tuple(islice(iterator, n)):
+        if strict and len(batch) != n:
+            raise ValueError("_batched(): incomplete batch")
+        yield batch
 
+
+def _circ_a():
```
SamFerracin: Very bad names, and no descriptions at all about what circuits A and B are supposed to represent. Could we:
- Remove the underscore
- Find more descriptive names
- Add docstrings

?

qiskit_ibm_runtime/aer_executor/run_quantum_program.py
```diff
@@ -73,8 +73,7 @@ def run_quantum_program(
     backend.set_max_qubits(10000)
     aer_sampler = AerSamplerV2.from_backend(backend)
 
-    # _seed is private but is the only way to obtain the sampler's RNG seed for reproducibility.
-    rng = np.random.default_rng(aer_sampler._seed)
+    rng = np.random.default_rng(aer_sampler.seed)
```
SamFerracin, line 76: used to be private at some point but it is public now?

SamFerracin (commented):
Thanks for splitting this out into dedicated commits. I went through it commit by commit, and it makes sense!

test/unit/aer_executor/test_run_quantum_program.py
```diff
@@ -12,39 +12,136 @@
...
 
 from ...ibm_test_case import IBMTestCase
 
 if optionals.HAS_AER:
     from qiskit_aer import AerSimulator
 
-    from qiskit_ibm_runtime.aer_executor.run_quantum_program import run_quantum_program
 
+def _batched(iterable, n, *, strict=False):
+    # _batched('ABCDEFG', 3) → ABC DEF G
```
diego-plan9: Extending Sam's comment to this one too: even if these come from a move rather than being new lines of code, can we take advantage of you going through the tests in order to increase their maintainability?

On top of Sam's suggestions, could you:
* add type hints
* revise if `*` provides value
* revise if `strict` is used or likely to be used

test/unit/aer_executor/test_run_quantum_program.py
```diff
@@ -12,39 +12,136 @@
...
+    coupling_map = fez_backend.coupling_map
+    active_qubits = list(range(fez_backend.num_qubits))
+
+    qc_boxed = QuantumCircuit(fez_backend.num_qubits, fez_backend.num_qubits)
+    with qc_boxed.box(
+        annotations=[
+            Twirl(dressing="left"),
+            Tag(ref="r0"),
+        ]
+    ):  # pyright: ignore[reportGeneralTypeIssues]
```
diego-plan9: Another "taking advantage" bit: can you do a pass for removing `pyright` excludes from the codebase? They rightfully hide warnings in some IDEs, but we don't declare `pyright` as part of our tooling (and we would need way more excludes to properly cover all its warnings - which would add to clutter).

test/unit/aer_executor/test_run_quantum_program.py
apchytr: Went through and updated all of them for consistency: 0cb526e1a8c865bcfc8fcbaaa6a0b39e36f9cf6b

test/unit/aer_executor/test_run_quantum_program.py
apchytr: Cleaned up here: 0cb526e1a8c865bcfc8fcbaaa6a0b39e36f9cf6b

qiskit_ibm_runtime/aer_executor/run_quantum_program.py
apchytr: Looking at `qiskit-aer` it's been public for at least two years so it's not something that was introduced recently. Not sure why it was private here

test/unit/aer_executor/test_run_quantum_program.py
apchytr: Applied here: 0cb526e1a8c865bcfc8fcbaaa6a0b39e36f9cf6b

test/unit/aer_executor/test_run_quantum_program.py
apchytr: Updated here: 0cb526e1a8c865bcfc8fcbaaa6a0b39e36f9cf6b

SamFerracin:
I left one extra question, but this looks good to me now. I'll let @diego-plan9 take another look, since he had comments too

diego-plan9 (approved):
Looks good, thanks!

PR #3099, opened 2026-07-24
Fix metadata serialization in Wrapper EstimatorV2 by using finalized options

leonapivato:
### Summary

#3042 introduced a regression, it passes the user configured options instead of the finalized options to the passthrough data, which has been raising flags in my parity tests for the metadata. This fixes that issue.

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code:


diego-plan9 (approved):
Thanks for catching this! Indeed, usage of `self.options` was not intentional.

Issue #3081, opened 2026-07-23
Run minimal version tests as part of PR workflows


diego-plan9:
**What is the expected enhancement?**

In #2769, the workflow was updated in order to check nightly compatibility with the minimal version of dependencies. It has proven to discover issues more often than expected - so this issue is about making it part of the regular PR checks, in order to detect and fix issues more easily.

**Acceptance criteria**

PR workflow contains a test against minimal versions.

PR #3104, opened 2026-07-26
Run lowest versions test only in main repo

diego-plan9:
### Summary

While looking at the minimal version tests in the recent issue, noticed that they are being run in the forks - which can cause a number of daily notifications for developers in their forks, as well as consume some actions time unnecessarily. This updates the job in the workflow so it is only run for the main fork, same as the others jobs in that workflow.

### Details and comments

Related to #3081

### AI/LLM disclosure


TsafrirA (approved):
LGTM

Issue #3091, opened 2026-07-23
In executor-based primitives, move DD pass inside prepare functions


SamFerracin:
In the executor based sampler and estimator, the `prepare` function takes care of mapping primitives (sampler or estimator) inputs to Executor inputs. Currently an extra prepare-like step happens afterwards, namely adding DD gates to the quantum program items. This:
- Feels a bit unnatural, since it's still part of preparation, conceptually
- Creates some difficulties with testing, since the only way of testing is by calling the run method and mocking the executor, see the test called `test_run_applies_dynamical_decoupling_after_prepare`

TODOs:

- [ ] Move the DD pass inside the prepare function
- [ ] Rewrite the tests in a way that does not requires us to use mocks

Blocked by #3090

PR #3105, opened 2026-07-27
Move DD of wrapper estimator to the prepare function

TsafrirA:
### Summary
Part of #3091
This PR moves the DD application of the wrapper estimator into the prepare function.

### AI/LLM disclosure


test/unit/executor_estimator/test_estimator_v2_prepare.py
```diff
@@ -104,3 +106,67 @@ def test_pea_path(self):
         self.assertIsInstance(program, QuantumProgram)
         self.assertIsInstance(executor_options, ExecutorOptions)
         self.assertEqual(program.passthrough_data["post_processor"]["mitigation"], "pea")
+
+    @data(True, False)
+    def test_dd_applied_when_enabled(self, twirling_enabled):
```
TsafrirA, line 111: Technically, this only tests the vanilla pathway, but I think it's sufficient.

TsafrirA:
(Fails CI because of #3119)

test/unit/executor_estimator/test_estimator_v2_prepare.py
SamFerracin: yes, I agree

SamFerracin (approved):
Looks good, thanks! I will approve again when it passes, if needed

PR #3106, opened 2026-07-27
Move DD of wrapper sampler to prepare

### LINKED ISSUE #3091: In executor-based primitives, move DD pass inside prepare functions

TsafrirA:
### Summary
Closes #3091 (with #3105)

This PR moves the DD application of the wrapper sampler into the prepare function.

### AI/LLM disclosure


SamFerracin (approved):
This looks good! I have been thinking about this `backend = None`, but I don't think there's a better way around this.. 🤔 Something that we can explain in the docs is that `prepare` is indifferent to the backend, except for the step where it applies DD, which depends on it

TsafrirA:
> Something that we can explain in the docs is that prepare is indifferent to the backend, except for the step where it applies DD, which depends on it

Will add in #3105.

Issue #3110, opened 2026-07-27
Adding `SimulatorOptions`


apchytr:
**What is the expected enhancement?**
Currently, `SimExecutor` options are the additional arguments `noise_dict`, `angle_decimals`, and `warn_absent` in the `__init__.py`. As suggested by @jyu00, we should look into instead leveraging the already existing `SimulatorOptions` class. This will bring the `SimExecutor` closer to being a "drop in" replacement for `Executor` e.g.
```
# executor = Executor(backend)   # replace next line with this for real QPU
executor = AerExecutor(backend=AerSimulator())
executor.options.simulator.noise_map = {"tag1": another_noise_model}   # ignored if using real QPU
executor.options.simulator.seed_simulator = 123   # ignored if using real QPU

job = executor.run(program)
```

**Acceptance criteria**

- [x] Create a new `SimulatorOptions` containing `nose_dict`, `angle_decimals`, and `warn_absent`
- [x] Use the new `SimulatorOptions` in `ExecutorOptions`

PR #3111, opened 2026-07-27
`SimExecutor` - new `SimulatorOptions` class

apchytr:
### Summary
Here we introduce `SimulatorOptions` for the `sim_executor` options `noise_model`, `angle_decimals`, and `warn_absent`. This brings us closer to having the `SimExecutor` be a drop in replacement for `Executor`:

```
# executor = Executor(backend)   # replace next line with this for real QPU
executor = AerExecutor(backend=AerSimulator())
executor.options.simulator.noise_map = {"tag1": another_noise_model}   # ignored if using real QPU
executor.options.simulator.seed_simulator = 123   # ignored if using real QPU

job = executor.run(program)
```
Making use of this in a follow up PR.

### Details and comments

Part of #3110 

### AI/LLM disclosure


qiskit_ibm_runtime/options_models/simulator.py
```diff
@@ -63,6 +63,18 @@ class SimulatorOptions(BaseOptionsModel):
...
+    """
+    Gate angles are rounded to the nearest multiple of π/2 at this
+    decimal precision before simulation.  This prevents floating-point drift from
+    preventing Clifford-method simulation when angles are nominally Clifford.
+    """
+
+    warn_absent: bool = True
+    """
+    Whether to emit a warning when a tagged barrier's tag is not found in ``noise_model``.
+    """
```
diego-plan9: I know it is WIP - can you update the docstring so they are conforming before taking out of WIP? We cannot enforce them automatically via `ruff` on member-docstrings, but can you revise `D200` and `D205`?

Also, not mandatory but recommended for consistency (at least in `option_models`) and for a bit of readability, feel free to use full width (100 chars).

qiskit_ibm_runtime/sim_executor/sim_executor.py
```diff
@@ -59,22 +60,22 @@ def __init__(
...
-        )
+        self._result = None
 
     def job_id(self) -> str:
         """Return the unique job ID."""
         return self._job_id
 
     def result(self, *_, **__) -> QuantumProgramResult:  # type: ignore[no-untyped-def]
         """Return the result of the program execution."""
+        if self._result is None:
```
SamFerracin, line 72: Thanks for this change! It may be worth making a dedicated issue and PR for it, for the purpose of tracking

qiskit_ibm_runtime/options_models/executor.py
```diff
@@ -30,3 +31,6 @@ class ExecutorOptions(BaseOptionsModel):
 
     experimental: dict = {}
     """Experimental options that are passed to the executor."""
+
+    simulator: SimulatorOptions = SimulatorOptions()
+    """Simulator options."""
```
SamFerracin: could you please expand the docs a little in here? to explain for example when they are skipped

SamFerracin (dismissed):

qiskit_ibm_runtime/options_models/executor.py
```diff
@@ -30,3 +31,6 @@ class ExecutorOptions(BaseOptionsModel):
 
     experimental: dict = {}
     """Experimental options that are passed to the executor."""
+
+    simulator: SimulatorOptions = SimulatorOptions()
```
SamFerracin: As discussed in private, I would hold on on adding new fields to the existing SimulatorOptions because this would create a strange situation where only some fields are relevant for SimExecutor and others (like `coupling_map`) can be set but not used

qiskit_ibm_runtime/options_models/simulator.py
```diff
@@ -34,8 +35,27 @@
     noise_model_type: TypeAlias = dict | None  # type: ignore[no-redef, misc]
 
 
+class SimulatorOptions(BaseOptionsModel):
+    """Simulator options utilized by :mod:`qiskit_ibm_runtime.sim_executor`."""
+
+    gate_angle_precision: int = 5
+    """Rounding precision for gate angles (in units of π/2)."""
+
+    layer_noise_dict: dict[str, Annotated[PauliLindbladMap, InstanceOf]] | None = None
```
SamFerracin: Does `PauliLindbladMap` create issues when you try model-dumping the options?

qiskit_ibm_runtime/options_models/simulator.py
```diff
@@ -34,8 +35,27 @@
     noise_model_type: TypeAlias = dict | None  # type: ignore[no-redef, misc]
 
 
+class SimulatorOptions(BaseOptionsModel):
+    """Simulator options utilized by :mod:`qiskit_ibm_runtime.sim_executor`."""
```
SamFerracin: I'm not sure if we want to make that module public, so I'd be vague for now.
```suggestion
    """Simulator options."""
```
When you add the options to EstimatorOptions, you can explain that they get used whenever we run a simulation via local mode, and ignored otherwise; and I think this will give enough context

What do you think?

qiskit_ibm_runtime/options_models/simulator.py
SamFerracin: Slight preference for this, since we don't have the duality anymore, as it's more concised. Which becomes a strong preference for the "dict" part of it (we know it's a dict from the type hint)
```suggestion
    noise_model: dict[str, Annotated[PauliLindbladMap, InstanceOf]] | None = None
```

qiskit_ibm_runtime/options_models/simulator.py
```diff
@@ -34,8 +35,27 @@
...
+
+    gate_angle_precision: int = 5
+    """Rounding precision for gate angles (in units of π/2)."""
+
+    layer_noise_dict: dict[str, Annotated[PauliLindbladMap, InstanceOf]] | None = None
+    """A map from tagged barrier references to Pauli-Lindblad noise maps.
+
+    Each key must match the ``ref`` of a :class:`~samplomatic.Tag` annotation.
+    """
```
SamFerracin: The barrier is an implementation detail that most users of executor won't know, and they may get confused. So, I'd not mention it
```suggestion
    """A map from ``ref`` of a :class:`~samplomatic.Tag` annotations to PauliLindblad noise maps.
    """
```

Also, the second part (Each key must match) is not true.. You can have keys that don't match

qiskit_ibm_runtime/options_models/simulator.py
```diff
@@ -34,8 +35,27 @@
     noise_model_type: TypeAlias = dict | None  # type: ignore[no-redef, misc]
 
 
+class SimulatorOptions(BaseOptionsModel):
+    """Simulator options utilized by :mod:`qiskit_ibm_runtime.sim_executor`."""
+
+    gate_angle_precision: int = 5
+    """Rounding precision for gate angles (in units of π/2)."""
```
SamFerracin: I kind of like. the name (which was concised) and above all, the description we havd in sim executor. Any reason to change it?
```suggestion
    angle_decimals:  int = 5
    """Gate angles are rounded to the nearest multiple of ``np.pi/2`` at this
            decimal precision before simulation.  This prevents floating-point drift from
            preventing Clifford-method simulation when angles are nominally Clifford.
     """
```

SamFerracin (commented):
Looks good, some cosmetic suggestions

qiskit_ibm_runtime/options_models/simulator.py
apchytr: Yes, it does only when trying `model_dump_json`. I'm guessing this creates issues if the simulator options within the executor options are not ignored?

qiskit_ibm_runtime/options_models/simulator.py
apchytr: Sgtm 👍

qiskit_ibm_runtime/options_models/simulator.py
apchytr: 296e74a709fab4c3b312e4432a567cfe3dad7f50

qiskit_ibm_runtime/options_models/simulator.py
apchytr: 296e74a709fab4c3b312e4432a567cfe3dad7f50

qiskit_ibm_runtime/options_models/simulator.py
apchytr: 296e74a709fab4c3b312e4432a567cfe3dad7f50

SamFerracin: Yeah, i'm curious to see if anything breaks when you include these options in the executor options

SamFerracin (approved):

PR #3112, opened 2026-07-27
Add decorator for mocking http responses

### LINKED ISSUE #3109: Facilitate unit testing of QiskitRuntimeService

Currently, `QiskitRuntimeService` contains quite a number of "state" variables, that are populated and modified through complex nested logic. A number of these variables also depend on the instances and account configuration and properties which are difficult to replicate - overall, this limits our ability to effectively test it locally and have good confidence about it. The natural trend is to resort to integration testing, but this is also limited when it comes to testing instance, plans and accounts combinations.

We have `FakeRuntimeService` and the family at `test/unit/mock` as a way to mitigate. These classes provide alternative implementations, but override/mock some of the methods, which causes the real logic not to be exercised (and require these counterparts to be maintained and kept on par). A number of tests depend on specific "tweaks" over the state variables, and have a risk of testing the fake counterparts rather than the actual classes. At some point in time `FakeRuntimeService` and related classes allowed for customization of instances and other features, but part of their functionality seems to be out of date.

This story is about providing testing facilities that allow for testing `QiskitRuntimeService` with more confidence.

**Acceptance criteria**

More unit tests in the codebase use `QiskitRuntimeService` directly.

diego-plan9:
### Summary

Add a `@mock_responses` decorator that allows a test to use mocked responses against the API. It is based on `responses`, providing a slightly customized `BaseRegistry` for specifying which instances and backends are to be available in the test (and generating the responses based on those).

A couple of test files have been revised in order to make use of the decorator.

### Details and comments

Fixes #3109

The intent is not to extend the decorator/registries functionality to cover all API usage, but instead to allow comfortable testing of `QiskitRuntimeService`, and specially its instance/backend functionality. If this goes through, the plan is to continue revising the unit tests to convert them to this approach, being selective (as there is the risk of increasing complexity in the callbacks).

### AI/LLM disclosure


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -274,6 +274,7 @@ def __init__(
                 )
             self._default_instance = True
             self._api_clients = {self._account.instance: RuntimeClient(self._client_params)}
+            self._active_api_client = self._api_clients[self._account.instance]
```
diego-plan9, line 277: This is intended, and arguably a small bug uncovered by using the new decorator when revising `test_qiskit_runtime_service.py`.

If `instance` is passed to the constructor, the active api client would be set to the one created at line 260 - but the stored list of api clients would point to a _different_ instance for that instance. The latter would be the one set to active during normal operations (ie. when the code attempts to get a client based on the instance, as it looks into `self._api_clients`), and the former discarded and left out of scope. This change ensures the active api client is set to the same object that is in the list of api clients.

pyproject.toml
```diff
@@ -70,7 +70,7 @@ disallow_untyped_defs = false
 
 [[tool.mypy.overrides]]
 module = [
-    "ddt",
+    "ddt.*",
```
SamFerracin, line 73: Could you give some context for this change?

test/registries.py
```diff
@@ -0,0 +1,422 @@
...
+
+if TYPE_CHECKING:
+    from requests import PreparedRequest
+
+    from qiskit_ibm_runtime.fake_provider.fake_backend import FakeBackendV2
+
+# ``responses`` callback return value: (status code, headers, body).
+CallbackResult: TypeAlias = tuple[int, dict[str, str], str]
+
+PricingType: TypeAlias = Literal["free", "trial", "paygo", "paid", "subscription", "unknown"]
```
SamFerracin, line 36: I assume that these are not already hardcoded anywhere else, or query-able?

pyproject.toml
diego-plan9: It was an unintentional slip on #3068 (my bad when providing the suggestion). We want to say "for _all_ ddt usages, ignore the missing imports, not just for top-level import" (ie. same as the rest of the members in this block). So it is a tweak taking advantage that this PR also touches the section.

test/registries.py
diego-plan9: They come from https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/qiskit_runtime_service.py#L378. The callbacks and the design is quite tuned to `QiskitRuntimeService` implementation, by design (ie. it does not aim to replicate all API features or provide full API responses, and the usability side is focused on the current testing needs).

SamFerracin (commented):
High-level comment:

This is a tricky PR to review, as it requires one to refresh one's memory about an object, QiskitRuntimeService, that hasn't been touched in a while. It also adds a lot of lines of code in the `registries.py` file, which aren't super simple to grasp. However, I don't see them as adding maintenance burden (if a bug comes up in QiskitRuntimeService, fixing it will require us to refresh our memories about it, and `registries.py` becomes clear as a consequence of that), and I see the potential the decorator has to help us standardize other tests and remove churn

I willl leave some time for other to review, but happy for this to go in

test/registries.py
SamFerracin: Would it make sense to introduce `PricingType` in `qiskit_runtime_service.py` and to import it here?

diego-plan9: Conceptually it contains the list of valid pricing types that comes _from the API_, which is not likely to change.

On the `qiskit_runtime_service.py`, it is used for an implementation detail (ie. deciding which instance to use when multiple). On the `registries.py`, it is a convenience type hint to effortlessly guide the devs towards the valid choices.

Actually, keeping `registries.py` not importing `qiskit_runtime_service.py` is arguably a goal. Registries should be kept close to the API, and while some of the design choices depend on `QiskitRuntimeService` features, the goal is not to make the registry depend directly (ie. import) from the service. In this case, `Instance` explicitly allows a pricing type, as one of the features of QiskitRuntime service is choosing the least costly instance if multiple are available - and we want to be able to replicate that (by mocking the response at the API level).

SamFerracin (approved):

diego-plan9:
> This is a tricky PR to review

It is, thanks for the review!

> as it requires one to refresh one's memory about an object, QiskitRuntimeService, that hasn't been touched in a while.

I challenge / rephrase this a bit as:
* the `registries.py` file (in particular the callbacks) do require _some_ insight into `QiskitRuntimeService`, and more insight into the API responses. For the latter, I tried to solve via documentation. For the former, I could not find a way to avoid it.
* when tests are rewritten to use the decorator, they actually use the real `QiskitRuntimeService`. This makes it easier to refresh our memory of it - and actually, allows us to exercise the real implementation properly and rebuild that knowledge backed by reproducible tests. Previously, we could only use `FakeRuntimeService` (and `FakeRuntimeClient`, etc), which override methods and mock (python-land) items, so we could not really exercise `QiskitRuntimeService` in unit tests.

> It also adds a lot of lines of code in the registries.py file, which aren't super simple to grasp. However, I don't see them as adding maintenance burden

It does, and I actually see more a risk of a maintenance burden (it does have some boilerplate, the real complexity is in the callbacks and how they relate to `Backend` and `Instance`). The goal is to be selective on adding new features to the registry, and decide as needed (and not be shy on just opting not to cover some functionality). This will be better seen as we explore moving other files to the decorator.

PR #3119, opened 2026-07-27
Fix flaky test

### LINKED ISSUE #3118: Tests relying on specs order are flaky
### Details
The specs returned by `samplex.inputs().specs` are sorted alphabetically. However, spec index is advanced numerically `1,2,3,...`, and when the number of decimal places in the index changes (`99, 100` for example), the alphabetical order does not match the insertion order. If the test relies on the order of the specs to judge the insertion order, it will be flaky and could fail depending on the exact value of the reference counter. The test will then fail or pass depending on test execution order, OS and other factors making it extremely hard to debug.

For example this test [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/6c7a627186f33b8b418d2bbd12cb61966385b153/test/unit/executor_estimator/test_estimator_v2_prepare_vanilla.py#L224).

TsafrirA:
### Summary
Closes #3118 

I haven't been able to find any other tests that rely on specs order.

### AI/LLM disclosure

### DIFF
--- a/test/unit/executor_estimator/test_estimator_v2_prepare_vanilla.py
+++ b/test/unit/executor_estimator/test_estimator_v2_prepare_vanilla.py
@@ -255,16 +255,31 @@ class TestPrepareVanilla(IBMTestCase):
         # 7 randomizations, 3 basis
         self.assertEqual(program.items[0].shape, (7 if enable_gates or enable_measure else 1, 3))
 
-        name0 = program.items[0].samplex.inputs().specs[0].name
-        self.assertTrue(name0.startswith("basis_changes"))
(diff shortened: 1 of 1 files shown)


TsafrirA:
At the moment #3105 is failing CI because of the flaky test.

SamFerracin (approved):
Thanks, approving to make CI/CD pass.

Though I'd like to ask, did you manage to figure out why this only started failing now, and  was not failing until today?

TsafrirA:
> Though I'd like to ask, did you manage to figure out why this only started failing now, and was not failing until today?

It's hard to say exactly. It depends on so many things. It should have to do with the exact makeup of the tests, so any change in the tests (running samplomatic) could potentially cause issues.

PR #3121, opened 2026-07-28
Apply DD to PEC, ZNE in wrapper estimator

TsafrirA:
### Summary
#3105 moved the DD application to the prepare function, but missed the early returns in that function causing DD to not be applied to ZNE\PEC, and making this [comment ](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3105#discussion_r3654873317) particularly tragic. This PR fixes it.

### AI/LLM disclosure


SamFerracin (approved):
Good catch, thanks!

Issue #3073, opened 2026-07-21
Executor-based Primitives: Quality Assurance


SamFerracin:
These is an epic to plan and track our efforts towards promoting executor-based sampler and estimator to production-level code.

The main objectives of the epic are:

- To identify and fix bugs.
- To improve user experience and docs.
- To improve and consolidate the testing strategy.

diego-plan9:
Marking as completed after merging the last bit, #3333

PR #3122, opened 2026-07-28
Minor cleanup of wrapper primitives

TsafrirA:
### Summary
Related to #3073 

Just some minor things I came across - 
- Doc string fix
- Removed duplicate import
- Removed unnecessary check

### AI/LLM disclosure


SamFerracin (approved):
Thanks!

Issue #3026, opened 2026-07-10
Add integration tests for wrapper-estimator


diego-plan9:
**What is the expected feature or enhancement?**

While we have some unit test coverage of the wrapper-estimator internals there is currently no integration test exercising the full functionality of running `WrapperEstimator` jobs.

Similar to #2645 and #2644, we we should have some integration tests capturing the functionality. 

**Acceptance criteria**

At least one integration test that runs a meaningful `WrapperEstimator` job against a device is added.

SamFerracin:
During a recent meeting with the addons team, it was brought up that we should monitor the RTT (round-trip time) of executor-based estimator (and sampler) jobs, to make sure that we don't take a hit on performance in the future without detecting it. 

RTT is defined as the time between pressing ENTER on `estimator.run(pubs)` to the time that I get `job.result()` back, minus the time spent in the queue and the "quantum" time spent running quantum circuits on the QPU. All these info is available somewhere, likely around the job metrics

I would be curious to know if we can augment the scope of the integration tests to also track RTT, for the reason that:
- Calculating RTT should be pretty straightforward, based on the info above
- We are already submitting jobs, may as well take full advantage of them

Monitoring (i.e., the act of comparing RTT with RTT of previous jobs) seems trickier though, and likely what needs investigation.

Just dropping this here, so that we don't forget

diego-plan9:
Thanks, and indeed we should not forget about it. One factor at play is integration test total runtime: we do want to keep them within reasonable limits.

This issue was opened with a smaller scope, though: specifically, having at least one integration tests for the estimator in the same spirit as the existing integration tests for Executor and WrapperSampler, so we can detect breakages fast (and have a quick way of testing that the estimator works when developing). By having a minimal test (or small number of them), we gain confidence without significant impact to test time.

TsafrirA:
Closing via #3189 and #3192.

Will open a separate issue for RTT.

PR #3123, opened 2026-07-28
Add Estimator through Executor integration test

rainerschoe:
### Summary
part of #3026 

Adds a very simple integration test, which runs against EstimatorV2 through implementation:

As per discussion I did not re-use existing Estimator integration tests, as we want to develop them afresh with a feature-driven testing strategy in mind.

This is the basic test with simple circuit and no options. more to follow in separate PR.

### Details and comments
Initially I planned to parametrize this against native and Executor based implementation, but as per discussions with Sam, let's keep the scope of this limited to Executor for now.

tested by running tests via

```
uv run --extra dev pytest test/integration/test_executor_estimator.py -s -v
```

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Bob (for adding imports only)


qiskit_ibm_runtime/executor_estimator/__init__.py
rainerschoe, line 15: this was missing

test/integration/test_estimator_v2.py
```diff
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2024-2026.
+# (C) Copyright IBM 2026.
```
SamFerracin: Could you rename the file as `test_executor_estimator.py`? For consistency with `test_executor_sampler.py`. 

The names of the test files are a bit messy right now (especially for unit tests), we should clean them up

test/integration/test_estimator_v2_native.py
```diff
@@ -0,0 +1,137 @@
...
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Integration tests for Estimator V2.
+
+Tests here run against the server-side Estimator V2 implementation only.
```
SamFerracin: Are these new tests, or just the old ones, rebranded?

SamFerracin (commented):
Thanks, Rainer! This looks like a good starting point!

A few points that we can think about:
- I am not too sold on running the new test twice, for legacy and new estimator. In my opinion, this integration test is about the new one, and its goal should be to inform us when we merge changes that break the e2e experience. Not about checking that the two have the same API (this is being tested independently in other tests). But we can discuss the wider scope of integration tests, perhaps in the team meeting today, and see where the team lands :) 
- The PUBs look good! I wonder if it makes sense to make them just a bit more complex, for example by using param values and observables with non-trivial shape. I don't have a definitive take here--again, probably depends on the scope that we decide to set on

rainerschoe:
@SamFerracin I removed the parametrization and un-did the re-naming. this should answer / satisfy all your quesitons above

TsafrirA (commented):
Given that we don't test the correctness here, I think we should expand coverage to include more variations on the observables side (more operators, more data types).

rainerschoe:
> Given that we don't test the correctness here, I think we should expand coverage to include more variations on the observables side (more operators, more data types).

Cool I agree. then let's merge this and work on increasing coverage in a new PR. However I need help here, as I do not understand what more operators and more data types mean :-/

SamFerracin:
@TsafrirA, do you have specific cases in mind for what you propose?

I agree with Rainer's proposal to bring this in and iterate

SamFerracin (approved):

PR #3124, opened 2026-07-28
Add mirror_circuit test util

### LINKED ISSUE #3138: Add mirror circuit utils function for testing purposes
Add a reusable test utility function, `mirror_circuit`, to `test/utils.py`. This will be helpful for integration tests regarding the wrapper primitives.

leonapivato:
### Summary
Adds a mirror circuit util to be used for testing purposes.

Fixes #3138 

### AI/LLM disclosure

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -38,7 +38,7 @@ from qiskit_ibm_runtime.runtime_job_v2 import RuntimeJobV2
 if TYPE_CHECKING:
     import logging
 
-    from qiskit.providers.backend import Backend
+    from qiskit.providers.backend import Backend, BackendV2
 
 
 def most_busy_backend(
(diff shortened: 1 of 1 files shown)


test/utils.py
```diff
@@ -291,6 +292,40 @@ def bell():
...
+        for qubit in range(num_qubits):
+            forward.rz(rng.uniform(0, 2 * np.pi), qubit)
+            forward.sx(qubit)
+            forward.rz(rng.uniform(0, 2 * np.pi), qubit)
+            forward.sx(qubit)
+            forward.rz(rng.uniform(0, 2 * np.pi), qubit)
+
+    circuit = QuantumCircuit(num_qubits, name=f"mirror_{num_qubits}q")
+    circuit.compose(forward, inplace=True)
+    circuit.barrier()  # keep U and U† from telescoping to identity under transpilation
```
SamFerracin: I have no idea what telescoping means 😅 
```suggestion
    circuit.barrier()  # keep U and U† from resolving to identity under transpilation
```

test/utils.py
```diff
@@ -291,6 +292,40 @@ def bell():
...
+    circuit.compose(forward.inverse(), inplace=True)
+    circuit.barrier()
+    for qubit in range(num_qubits):
+        circuit.rx(theta, qubit)
+    return circuit
```
SamFerracin, line 353: Otherwise, we'll have to add this in every function that uses it
```suggestion
    circuit.measure_all()
    return circuit
```

test/utils.py
```diff
@@ -291,6 +292,40 @@ def bell():
...
+    every single-qubit <Z_j>(theta) = cos(theta) -- useful as a per-qubit noise probe.
+    """
+    rng = np.random.default_rng(seed)
+    theta = Parameter("theta")
+    forward = QuantumCircuit(num_qubits, name="mirror_forward")
+    for _ in range(layers):
+        for control in range(0, num_qubits - 1, 2):
+            forward.cx(control, control + 1)
+        for control in range(1, num_qubits - 1, 2):
+            forward.cx(control, control + 1)
```
SamFerracin:
```suggestion
        forward.cx(range(0,num_qubits-1,2), range(1,num_qubits,2))
        forward.cx(range(1,num_qubits-1,2), range(2,num_qubits,2))
```

test/utils.py
```diff
@@ -291,6 +292,40 @@ def bell():
     return quantum_circuit
 
 
+def mirror_circuit(num_qubits: int = 2, layers: int = 4, *, seed: int | None = 7) -> QuantumCircuit:
```
SamFerracin: It may be good to pass a `layout: Layout | None = None` here, so that if users specify it, we can:
- Choose basis gates that exist in the layout
- Entangle qubits that are indeed neighbours 

If we go with what you have here, we wouldn't be able to, for example, use the resulting circuit for integration tests if we wanted to--the `test_backend` would raise a non-ISA error.

Do you want to give it a try?

TsafrirA (commented):
Is the point of the trailing `rx` gates to have a predictable parameter-dependent effect? (This isn't how mirror circuits are typically constructed. Typically, the extra effect is placed between the mirrors and not after them)

test/utils.py
```diff
@@ -291,6 +292,64 @@ def bell():
...
+    layer2_pairs = [
+        (control, target)
+        for control, target in odd_pairs
+        if backend.target.instruction_supported(entangler, (control, target))
+        and control < num_qubits
+        and target < num_qubits
+    ]
+    layer2 = QuantumCircuit(num_qubits)
+    for control, target in layer2_pairs:
+        layer2.getattr(entangler, (control, target))
```
SamFerracin:
```suggestion
        getattr(layer2, entangler)(control, target)
```

test/utils.py
```diff
@@ -291,6 +292,64 @@ def bell():
...
+    layer1_pairs = [
+        (control, target)
+        for control, target in even_pairs
+        if backend.target.instruction_supported(entangler, (control, target))
+        and control < num_qubits
+        and target < num_qubits
+    ]
+    layer1 = QuantumCircuit(num_qubits)
+    for control, target in layer1_pairs:
+        layer1.getattr(entangler, (control, target))
```
SamFerracin:
```suggestion
        getattr(layer1, entangler)(control, target)
```

SamFerracin: edited

test/utils.py
SamFerracin: edited

test/utils.py
```diff
@@ -291,6 +292,64 @@ def bell():
...
+    circuit = QuantumCircuit(num_qubits)
+    for _ in range(layers):
+        circuit.compose(layer1, inplace=True)
+        circuit.rx(rng.uniform(0, 2 * np.pi), range(num_qubits))
+        circuit.compose(layer2, inplace=True)
+        circuit.rx(rng.uniform(0, 2 * np.pi), range(num_qubits))
+
+    circuit.barrier()  # keep U and U† from resolving to identity under transpilation
+    circuit.compose(circuit.inverse(), inplace=True)
+    circuit.barrier()
```
SamFerracin: I'd do this, to avoid the double barrier between mirror and inverse
```suggestion
    mirror = QuantumCircuit(num_qubits)
    for _ in range(layers):
        mirror.compose(layer1, inplace=True)
        mirror.rx(rng.uniform(0, 2 * np.pi), range(num_qubits))
        mirror.compose(layer2, inplace=True)
        mirror.rx(rng.uniform(0, 2 * np.pi), range(num_qubits))

    circuit = QuantumCircuit(num_qubits)
    circuit.compose(mirror, inplace=True)
    circuit.barrier()  # keep U and U† from resolving to identity under transpilation
    circuit.compose(mirror.inverse(), inplace=True)
    
    circuit.barrier()
```

SamFerracin (approved):

PR #3126, opened 2026-07-28
Fix documentation typos in resilience and ZNE options

leonapivato:
### Summary
Used AI tool to find some doc issues with resilience options classes.

### AI/LLM disclosure


diego-plan9 (approved):
Thanks! At some point in time it would be a good idea to render the documentation (ie. add the sphinx entries to the index, list all the new classes - ensuring they don't get committed yet) temporarily, and take a look at the sphinx warnings and also inspect the rendered files (with or without AI), as I imagine it will uncover more items.

PR #3128, opened 2026-07-28
Restore numeric range validators on wrapper options

### LINKED ISSUE #3114: Wrapper options silently drop numeric validators present in legacy options
Legacy `EstimatorOptions` rejects a handful of clearly-invalid numeric inputs at construction time — `default_precision <= 0`, `default_shots < 0`, `twirling.num_randomizations/shots_per_randomization = 0`, `measure_noise_learning.num_randomizations = 0`, and `pec.max_overhead = 0` (this last one got tightened from `gt=0` to `ge=0`, so negatives are still caught, but 0 now slips through).

None of these validators exist on the executor-wrapper side. A value that used to raise immediately now gets silently accepted and only fails later (or not at all) once it hits run() or the server.

leonapivato:
Fixes #3114 

### AI/LLM disclosure


qiskit_ibm_runtime/options_models/estimator.py
```diff
@@ -28,7 +30,7 @@
 class EstimatorOptions(BaseOptionsModel):
     """Options for the executor-based EstimatorV2."""
 
-    default_precision: float = 0.015625
+    default_precision: Annotated[float, Field(gt=0)] = 0.015625
```
diego-plan9, line 33: Glad to see the annotated pattern - and I'm looking forward to see it too in the next wave of PRs!

test/unit/options_models/test_numeric_validators.py
```diff
@@ -0,0 +1,75 @@
...
+
+from qiskit_ibm_runtime.options_models.estimator import EstimatorOptions
+from qiskit_ibm_runtime.options_models.measure_noise_learning import MeasureNoiseLearningOptions
+from qiskit_ibm_runtime.options_models.pec import PecOptions
+from qiskit_ibm_runtime.options_models.twirling import TwirlingOptions
+
+from ...ibm_test_case import IBMTestCase
+
+
+class TestNumericValidators(IBMTestCase):
```
diego-plan9: Can you remove these tests? On one hand, I agree that we would have caught this earlier if similar tests were present. On the other hand, they are effectively testing standard `pydantic` features as we are just applying a well-documented validator, and unlikely to change as we move forward, so they don't provide too much value.

In this case, I'm partial to the arguments on the second side. For validators that contain custom logic or more complex conditions, that would tip the balance towards the first side.

SamFerracin:
Thanks!

I can think of a few other fields, off the top of my head. For example, what about the noise levels in ZNE? Do these have to be positive too?

leonapivato:
> I can think of a few other fields, off the top of my head. For example, what about the noise levels in ZNE? Do these have to be positive too?

Yeah, I opened a new issue for each of the 4 different entries that we discussed. One of them should cover the ZNE and the other validations.

SamFerracin (approved):

PR #3135, opened 2026-07-29
Support IAM private endpoints via IAM_URL env var

### LINKED ISSUE #3134: Enable VPE support for Qiskit Functions by adding `QISKIT_IBM_PRIVATE_ENDPOINT` environment variable

Qiskit Functions run `QiskitRuntimeService` inside a container to submit and manage quantum jobs. We have been able to access the runtime service through the public network, but as we shift towards a VPE-based network model, containers must route all IBM Cloud traffic through **private** endpoints. 

While a `private_endpoint` option exists in the runtime client, `private_endpoint=True` can only be set in code. This is different to all other connection parameters (`QISKIT_IBM_URL`, `QISKIT_IBM_TOKEN`, `QISKIT_IBM_INSTANCE`) which can be injected as environment variables at deploy time. Not having an environment value makes it impossible to enable private endpoints in Qiskit Function containers without modifying the function source code.

**Additional bugs found while investigating private endpoints**

- The private endpoint hostnames generated for Global Search and Global Catalog are incorrect and do not match the actual VPE endpoints, causing `instances()` and CRN validation to fail.
- When a fully-qualified CRN is passed as instance, `QiskitRuntimeService.__init__` still calls `instances()` to validate it, which is redundant and triggers unnecessary catalog and search traffic.

**Acceptance criteria**

Setting `QISKIT_IBM_PRIVATE_ENDPOINT=true` should be equivalent to passing private_endpoint=True, consistent with how other connection parameters work. Combined with `QISKIT_IBM_URL`, this allows a Qiskit Function container to be fully configured for a VPE environment via environment variables injected at deploy time, with no code changes required.

**Note**

I already have a local fix for the 3 points mentioned above, verified with the Qiskit Functions VPE, and will follow up with a PR shortly. I still wanted to open an issue to document the context around these changes.

ElePT:
### Summary

Adds an `IAM_URL` environment variable to override the IAM endpoint URL used for authentication, allowing IAM traffic to be routed through a private endpoint (e.g. https://private.iam.cloud.ibm.com) in VPE environments without code changes. This follows the same convention as the IBM SDK's `GLOBAL_SEARCH_URL` and `GLOBAL_CATALOG_URL` variables.

Additionally, skips the `instances()` validation call during `QiskitRuntimeService` initialization when `QISKIT_FUNCTIONS_EXPERIMENTAL` is set. This is an undocumented setting for environments where Global Search/Catalog may not be reachable via private endpoints.

### Details and comments

Fixes #3134.

I will keep the PR on draft until I have been able to re-verify that the URL changes I implemented hold in both production and staging.
### Pending tests

- [x] Test private endpoints in staging
- [x] Test private endpoints in production

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude Sonnet 4.6


diego-plan9 (approved):
Thanks! I tweaked style in order to allow moving this forward.

PR #3136, opened 2026-07-29
Change default value of `noise_model_mapping` to dict

TsafrirA:
### Summary
Currently `noise_model_mapping` in `ResilienceOptions` defaults to None. PEA overrides it to an empty dict, while PEC doesn't and fails ungracefully if the noise model is not provdied. This PR changes the default value to an empty dict, such that both pathways will fail in the same way.

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/options_models/resilience.py
+++ b/qiskit_ibm_runtime/options_models/resilience.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 from typing import Annotated
 
-from pydantic import InstanceOf
+from pydantic import Field, InstanceOf
 from qiskit.quantum_info import PauliLindbladMap
 
 from .base import BaseOptionsModel
@@ -69,10 +69,12 @@ class ResilienceOptions(BaseOptionsModel):
     zne: ZneOptions = ZneOptions()
     """Additional zero noise extrapolation mitigation options."""
 
-    noise_model_mapping: dict[str, Annotated[PauliLindbladMap, InstanceOf]] | None = None
+    noise_model_mapping: dict[str, Annotated[PauliLindbladMap, InstanceOf]] = Field(
+        default_factory=dict
+    )
     """A noise model mapping for PEC mitigation.
 
     Maps layer references (strings) to :class:`~qiskit.quantum_info.PauliLindbladMap` objects that
     describe the noise characteristics of that layer. The dict contains layers from all PUBs. This
-    is required when using PEC mitigation.
+    is required when using PEC mitigation, or ZNE with PEA amplifier.
     """
--- a/qiskit_ibm_runtime/options_models/resilience.py
+++ b/qiskit_ibm_runtime/options_models/resilience.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 from typing import Annotated
 
-from pydantic import Field, InstanceOf
+from pydantic import InstanceOf
 from qiskit.quantum_info import PauliLindbladMap
 
 from .base import BaseOptionsModel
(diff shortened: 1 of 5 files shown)


test/unit/executor_estimator/test_estimator_v2.py
```diff
@@ -387,6 +387,39 @@ def test_run_raises_error_when_pec_and_zne_both_enabled(self):
...
+        """Test that run raises error when pec_mitigation is enabled without noise model."""
+        estimator = EstimatorV2(mode=self.backend)
+        estimator.options.resilience.pec_mitigation = True
+        # noise_model_mapping is None by default
+
+        circuit = QuantumCircuit(2)
+        circuit.h(0)
+        observable = SparsePauliOp.from_list([("ZZ", 1)])
+
+        with self.assertRaisesRegex(
```
SamFerracin: This particular example doesn't error in main, because in the case, PEC/PEA does not require a noise model (there are no boxes). Which is why I have removed this error, in one of my latest PR

If you add a CZ to your example, PEA does fail gracefully because we are passing `noise_model or {}` [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/executor_estimator/prepare.py#L112), whereas in PEC we do not do the `or {}` [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/executor_estimator/prepare.py#L99). Therefore, I would fix this by simply adding the `or {}`

TsafrirA: I don't fully agree. I think running PEC without a noise model should not be allowed. What's the point of running PEC on a circuit with no layers? If we follow that logic, I think the error added here is more straightforward. If we take the opposite view, I think the correct solution is to just have an empty dict as a default value. Skip the `None` altogether.

SamFerracin: > I think running PEC without a noise model should not be allowed.

I don't agree. Setting pec to True when there are no boxes is fine, it is logically equivalent to running vanilla. It doesn't harm the user, for example by causing them to run more circuits. On the other hand, defaulting to `{}` instead of `None` would cause shared mutable state bugs, so that's also not an option

TsafrirA: What's wrong with `Field(default_factory=dict)`?

SamFerracin: You are right, I forgot this gets set with `default_factory`, and I had in mind a case like `noise_model: dict = {}`. Yes, that should work then

TsafrirA: Changed the default value.

qiskit_ibm_runtime/options_models/resilience.py
```diff
@@ -69,10 +69,12 @@ class ResilienceOptions(BaseOptionsModel):
     zne: ZneOptions = ZneOptions()
     """Additional zero noise extrapolation mitigation options."""
 
-    noise_model_mapping: dict[str, Annotated[PauliLindbladMap, InstanceOf]] | None = None
+    noise_model_mapping: dict[str, Annotated[PauliLindbladMap, InstanceOf]] = Field(
+        default_factory=dict
+    )
```
diego-plan9: Separate from the other conversation (ie. my angle comes from consistency rather than deeper discussion on the semantic implications): it is [safe to use](https://pydantic.dev/docs/validation/latest/concepts/fields/#mutable-default-values) `= {}` here and is equivalent to the `default_factory` approach, and since #3042 we have been just using the simple form (without `default_factory`), for conciseness. 

Can you tweak so we keep consistency?

TsafrirA: Thanks! Did not know that it's safe!

test/unit/executor_estimator/test_estimator_v2.py
TsafrirA: (Note the comment below indicating that a factory is not needed)

SamFerracin (approved):

PR #3140, opened 2026-07-29
`SimRuntimeJob` subclass `PrimitiveJob`

### LINKED ISSUE #3130: Switch `SimRuntimeJob` to use `LocalRuntimeJob`
We should leverage the preexisting `LocalRuntimeJob` for local jobs from `sim_executor` rather than introducing a new `SimRuntimeJob` 

**Acceptance criteria**

- [ ] Update `sim_executor` to use `LocalRuntimeJob`

apchytr:
### Summary
Making `SimRuntimeJob` subclass `PrimitiveJob` before adding local mode support in `Executor`.

### Details and comments

Closes #3130 

### AI/LLM disclosure


qiskit_ibm_runtime/sim_executor/sim_executor.py
```diff
@@ -47,27 +48,26 @@ class SimRuntimeJob:
...
         self,
-        backend: AerSimulator,
+        backend: AerSimulator | FakeBackendV2,
         program: QuantumProgram,
         noise_dict: dict[str, PauliLindbladMap] | None = None,
         angle_decimals: int = 5,
         warn_absent: bool = True,
     ):
-        self._backend = backend
+        super().__init__(function=None, future=None, backend=backend, primitive="sim", inputs={})  # type: ignore[arg-type]
```
diego-plan9: Overall, the changes do bring `SimRuntimeJob` closer to `LocalRuntimeJob` (which is based in `qiskit.PrimitiveJob`). However, from some angles it can be considered _not_ a subclass of `LocalRuntimeJob`:

* `function` and `future` are part of the mechanism that allows the work to be handled in a `ThreadPoolExecutor` (ultimately part of [Qiskit](https://github.com/Qiskit/qiskit/blob/main/qiskit/primitives/primitive_job.py#L25), with the Primitive taking care of submitting the job. Without using them, the current proposal has `result()` does the hard work without the "usual" cycle (create, submit, runs in parallel, get results).
* minor - `primitive` ideally should be passed somehow and end up containing a valid program id.
* `inputs` seems to have the same semantic meaning of `program` (ie. contain the inputs of the job)

I have not evaluated fully the rest of the existing methods in `LocalRuntimeJob` and `PrimitiveJob`, but likely might have some impact as a result of the mismatch.

My concern is not about whether some specific features are needed (for example, it might not be meaningful to do the work in parallel, or to preserve some of the arguments). It is more about leaving `SimExecutor` in a state where it is inconsistent with `LocalRuntimeJob` interface and design, that might cause more issues than solve them - and looking into solving by either getting both more aligned, or by just not considering them actual subclasses.

apchytr: Switched to `PrimitiveJob` here: 8b197696ffe3ecb25a338cc45f5048ebf4d9ef0c which results in a much more consistent interface!

qiskit_ibm_runtime/sim_executor/sim_executor.py
```diff
@@ -53,34 +49,18 @@ def __init__(
...
-        if self._result is None:
-            self._result = run_quantum_program(
-                backend=self._backend,
-                program=self._program,
-                noise_dict=self._noise_dict,
-                angle_decimals=self._angle_decimals,
-                warn_absent=self._warn_absent,
-            )
-        return self._result
+        self._submit()
```
diego-plan9, line 61: I wonder if we gain more similarity by just having the higher layers (ie. not instantiation) submit the job.

qiskit_ibm_runtime/sim_executor/sim_executor.py
```diff
@@ -26,15 +25,12 @@
     from qiskit.quantum_info import PauliLindbladMap
 
     from ..quantum_program import QuantumProgram
-    from ..results import QuantumProgramResult
 
 
-@HAS_AER.require_in_instance
-class SimRuntimeJob:
-    """Job object returned by :meth:`~.SimExecutor.run`.
+class SimRuntimeJob(PrimitiveJob):
```
diego-plan9, line 30: It would be nice to explore if the usage of generics under the hood allows us to properly type the output of `result()`.

diego-plan9 (approved):
Thanks! I left a couple of comments for potential improvements and inspecting if worth tackling, mostly as self-reminders, but they are not meant to be blockers. It will also be easier to think about them when looking at the full picture outside this PR.

qiskit_ibm_runtime/sim_executor/sim_executor.py
apchytr: Good point! I did come across this working through this PR. The issue is the return type we expect is a `QuantumProgramResult` but `BasePrimitiveJob` is generic on a type variable `ResultT` where `ResultT` is bounded to `_BasePrimitiveResultV1 | PrimitiveResult`. Since `QuantumProgramResult` doesn't inherit from anything it doesn't quite fit the type here. I'm curious if there is a good solution 🤔

Issue #3129, opened 2026-07-28
Update wrappers to switch to `SimRuntimeJob` for local mode


apchytr:
**What is the expected feature?**
With the new `SimRuntimeJob` the wrapper sampler and estimator should switch from using the `QiskitRuntimeLocalService` to the new aer based approach when running jobs in local mode

**Acceptance criteria**

- [x] Update executor to use `sim_executor` in local mode
- [x] Update wrapper executor to use `sim_executor` in local mode
- [x] Remove `SimExecutor`

PR #3142, opened 2026-07-29
Executor local mode with `sim_executor`

apchytr:
### Summary
Here we update the `Executor` to support local mode via `sim_executor` 

### Details and comments

Part of #3129 

### AI/LLM disclosure


qiskit_ibm_runtime/executor/executor.py
```diff
@@ -22,6 +22,7 @@
 from ..options_models.converters import to_runtime_options
 from ..options_models.executor import ExecutorOptions
 from ..quantum_program.params_converters import QUANTUM_PROGRAM_PARAMS_CONVERTERS
+from ..sim_executor import SimRuntimeJob
```
SamFerracin, line 25: Since we're getting rid of `SimExecutor`, perhaps we should rename `SimRuntimeJob` to something more descriptive, like `LocalRuntimeJob`--which is taken. Let's think about it, see if there are better alternatives?

SamFerracin (approved):
Looks good! Thanks!

qiskit_ibm_runtime/executor/executor.py
apchytr: `LocalRuntimeJobV2` ?

PR #3143, opened 2026-07-29
Add a simple e2e test for Executor-based Sampler that submits jobs via local-mode executor

### LINKED ISSUE #3131: Add a simple e2e test for Executor-based Sampler that submits jobs via local-mode executor
The goal is to add a simple test that we can look at, agree on, and use as a golden standards for writing more extensive tests.

At a minimum, the test should check that the results are correct--ie that when they a noiseless simulation is run, they are close to the ideal ones, in values, shapes, etc

SamFerracin:
### Summary
This PR adds a test for executor-based Sampler that verifies the correctness of the returned results.

Opening the PR prematurely for discussion, but merging will require addressing the two TODOs.

### Details and comments

Fixes #3131

### AI/LLM disclosure


test/unit/executor_sampler/simulations/test_sampler.py
```diff
@@ -0,0 +1,120 @@
...
+        """Test level setup."""
+        super().setUp()
+
+        self.backend = FakeManilaV2()
+
+        # The tolerance used when verifying simulated counts with exact probabilities via
+        # Hellinger distance.
+        self.tolerance = 0.01
+
+    def verify_pub_result(self, pub: SamplerPub, pub_result: SamplerPubResult) -> None:
```
TsafrirA, line 51: What's the value you see in comparing two simulators against one another?

SamFerracin: The two methods of computing results are entirely different and  independent of each other. The results coming from StateVector are obtained through exact calculation in qiskit and are tested on their own. They provide us with a baseline to compare the results of the simulation to, one that we can trust.

If we wanted to check the correctness of the results without using StateVector, we would have to stick to easy instances of the circuit for which we know the answer, hardcode large arrays of expected results, etc. This method gives us the flexibility to avoid that

test/unit/executor_sampler/simulations/test_sampler.py
```diff
@@ -0,0 +1,120 @@
...
+        It computes the probabilities of each outcome exactly using Qiskit's StateVector class.
+        Then, it compares them to the given ``counts`` via Hellinger distance.
+        """
+        # Avoid modifying the given circuit
+        circuit_cp = pub.circuit.copy()
+        circuit_cp.remove_final_measurements()
+
+        parameter_values = pub.parameter_values
+        if len(parameter_values.shape) == 0:
+            parameter_values = parameter_values.reshape((1,) + parameter_values.shape)
```
apchytr: minor simplification

```
parameter_values = np.atleast_1d(pub.parameter_values)
```

test/unit/executor_sampler/simulations/test_sampler.py
rainerschoe: I think even if the simulators would be the same the test still makes sense, as we are mostly interested in testing primitives responsibilities and not if the QPU / Simulator behaves correctly. We only need to make sure that the control-path we assert against is not using the code under test.

test/unit/executor_sampler/simulations/test_sampler.py
```diff
@@ -0,0 +1,109 @@
...
+    """SamplerV2 tests centered around qiskit-aer simulations.
+
+    All the tests in this class perform noiseless simulations.
+    """
+
+    def setUp(self):
+        """Test level setup."""
+        super().setUp()
+
+        self.backend = AerSimulator()
```
rainerschoe, line 45: Why AerSimulator and not a Fake backend?

test/unit/executor_sampler/simulations/test_sampler.py
```diff
@@ -0,0 +1,109 @@
...
+                fidelity := hellinger_fidelity(counts, probabilities),
+                1.0,
+                msg=f"Fidelity: {fidelity}",
+                delta=self.tolerance,
+            )
+
+    @data([True, False], [False, True], [True, True], [False, False])
+    @unpack
+    def test_twirling_configurations(self, enable_gates, enable_measure):
+        """Test sampler with different configurations of twirling."""
```
rainerschoe: This test is not testing any effect of twirling configurations, I think the name or description is misleading.
It merely tests that the sampler has a result quality close to statevector simulation. It tests this for different twirling configurations with the same threshold.

rainerschoe (dismissed):
Left two minor comments / questions, but it looks fine otherwise :-)

test/unit/executor_sampler/simulations/test_sampler.py
SamFerracin: Because simulating with a fake backend results in a noisy simulation, and then the hellinger fidelity would not be 1

rainerschoe (approved):
Looks good now :-)

PR #3147, opened 2026-07-30
Adjust `ensemble_stds_noise_factors` calculation in ZNE

TsafrirA:
### Summary
Minor cleanup post #2953.

### AI/LLM disclosure


dekelmeirom (approved):
good catch!

SamFerracin (approved):
Thanks !!

PR #3120, opened 2026-07-28
Bump mypy to latest version (2.3.0)

diego-plan9:
### Summary

Bump the `mypy` version to the latest one, and tackle a couple of new errors in the process.

### Details and comments

Related to #2514

This prepares for further `mypy` tweaks - using the latest version facilitates revising the configuration, and seems to provide some performance benefits.

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude Opus  4.8

PR #3148, opened 2026-07-30
Fix conflict around mypy when installing with `[dev]`

rainerschoe:
### Summary
#3120 introduced new `mypy` version, but we need to use consistent versions in `dev` optional dependencies to avoid conflict.

to reproduce the issue:
```
uv run --extra dev pytest test/integration/test_executor_estimator.py -s 
-v
  × No solution found when resolving dependencies for split (markers:
  │ python_full_version >= '3.14'):
  ╰─▶ Because qiskit-ibm-runtime[dev] depends on mypy==1.19.0
      and qiskit-ibm-runtime[common] depends on mypy==2.3.0,
      we can conclude that qiskit-ibm-runtime[common] and
      qiskit-ibm-runtime[dev] are incompatible.
      And because your project requires qiskit-ibm-runtime[common]
      and qiskit-ibm-runtime[dev], we can conclude that your project's
      requirements are unsatisfiable.
```

### Details and comments

Fixes #

### AI/LLM disclosure

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -254,12 +254,12 @@ performance = [
 # Developer extras.
 
 style = [
-    "mypy==1.19.0",
+    "mypy==2.3.0",
     "pre-commit>=4",
 ]
 
 style_local = [
-    "mypy==1.19.0",
+    "mypy==2.3.0",
     "ruff~=0.14.10",
 ]
 


diego-plan9 (approved):
Thanks! Indeed, those are the right sections.

PR #3149, opened 2026-07-30
Use mirror circuit for Estimator through Executor Integration test

rainerschoe:
### Summary

We decided to try to use the same example circuit "mirror circuit" for all tests if possible.
This PR changes the custom bell circuit in estimator through executor integration test to use the mirror circuit.

Also I change the import alias / naming to only `EstimatorV2` instead of `EstimatorV2ThroughExecutor`

### AI/LLM disclosure

- [x] I didn't use LLM tooling, or only used it privately.
- [ ] I used the following tool to help write this PR description:
- [ ] I used the following tool to generate or modify code:

PR #3153, opened 2026-07-31
[test] Executor Integration test: Use mirror circuit instead of custom inline circuit

rainerschoe:
### Summary
see https://github.com/Qiskit/qiskit-ibm-runtime/pull/3149#pullrequestreview-4820318873
We want to use the same circuit for as much tests as possible.

### Details and comments

Fixes #

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Bob (Entirely written by bob under supervision)


test/integration/test_executor.py
```diff
@@ -13,14 +13,13 @@
...
 import numpy as np
-from qiskit import QuantumCircuit
-from qiskit.circuit import Parameter
 from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
 from samplomatic import build
 from samplomatic.transpiler import generate_boxing_pass_manager
 
 from qiskit_ibm_runtime import Executor, QuantumProgram
 from qiskit_ibm_runtime.results import QuantumProgramItemResult, QuantumProgramResult
+from test.utils import make_mirror_circuit_with_phases
```
SamFerracin: Small cosmetic change, just for consistency
```suggestion
from .utils import make_mirror_circuit_with_phases
```

rainerschoe: Applied your suggestion, however I was explicitly prompting this import style, trying to adhere to PEP 8 import guidelines https://peps.python.org/pep-0008/ recommending to prefer absolute imports. However it states that relative imports are acceptable in some cases.
Not sure which guidelines we follow here.

rainerschoe: Oh your proposal did not work and I reverted it: `utils` is located one folder up the tree, so cannot use relative import. We would need to use `..`, which I think is not good style. It is much more readable to use absolute import then, making the path clear and makes the code more easily to be moved around.

diego-plan9: We are adhering to relative imports, mostly for historical reasons. With some extra subtleties:
* in `qiskit_ibm_runtime/`, relative imports
* in `test/*:
    * `qiskit_ibm_runtime` as absolute imports (as it is a package that is expected to be installed when running the tests, same as other dependency)
    * items from `test` as relative imports, by chaining `...` as needed - as `test` is not formally a package (so this prevents clashing with other packages that might be named `test`).

So +1 for applying Sam's change (with `..`).

rainerschoe: Thanks for explaining. Added a commit :-)

diego-plan9: Thanks - it still seems to be present?

rainerschoe: oh, sorry. only replaced one occurrence. Now fixed

SamFerracin (approved):

PR #3157, opened 2026-07-31
Add validation for measure_noise_learning against measure_mitigation at construction  

### LINKED ISSUE #3115: Wrapper doesn't validate measure_noise_learning against measure_mitigation at construction
In legacy options, setting `resilience.measure_noise_learning` fields while `measure_mitigation` is `False` trips a `model_validator` and raises right away — you find out immediately that the config doesn't make sense.

On the wrapper side there's no equivalent validator. The value just gets silently forwarded as `None` at `run()` time, so the user gets no signal that their `measure_noise_learning` config was ignored.

leonapivato:
### Summary
Adds options validator to the new options model that was present in the legacy options.

Fixes #3115 

### AI/LLM disclosure

### DIFF
--- a/test/unit/executor_estimator/test_estimator_v2.py
+++ b/test/unit/executor_estimator/test_estimator_v2.py
@@ -524,6 +524,16 @@ class TestFinalizeOptions(IBMTestCase):
         self.assertFalse(finalized_options.resilience.measure_mitigation)
         self.assertTrue(finalized_options.resilience.zne_mitigation)
 
+    def test_measure_noise_learning_customized_with_mitigation_unset(self):
+        """Test customized mnl is accepted when measure_mitigation finalizes to True."""
+        estimator = EstimatorV2(self.backend)
+        estimator.options.resilience_level = 1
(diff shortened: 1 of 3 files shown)


test/unit/executor_estimator/test_estimator_v2.py
```diff
@@ -524,6 +524,16 @@ def test_set_values_are_preserved(self, resilience_level):
         self.assertFalse(finalized_options.resilience.measure_mitigation)
         self.assertTrue(finalized_options.resilience.zne_mitigation)
 
+    def test_measure_noise_learning_customized_with_mitigation_unset(self):
```
leonapivato, line 527: This tests a slight usability bug I found in the legacy.

Basically the legacy would raise a validation error if the user had configured noise learning options when `measure_mitgation` was not set to True, even if `resilience_level` was 1 and `measure_mitigation` would be turned on anyway

The new one doesn't copy this behavior by checking only if the user has explicitly set `measure_mitigation` to false. 

Just something to be aware of.

PR #3161, opened 2026-07-31
Implement noise factors validation for ZNE preparation

### LINKED ISSUE #3116: Wrapper accepts ZNE configs with too few noise_factors for the chosen extrapolator
Legacy raises at construction if you pick an extrapolator that needs more points than you gave it — e.g. `double_exponential` needs at least 4 `noise_factors`, and the legacy `model_validator` catches an under-specified combination immediately.

The wrapper has no such check. The bad config is accepted, and the under-specified fit just fails later — either deeper in client code or server-side — instead of failing fast with a clear message.

Worth adding the same cross-field validation so users get the error where they made the mistake, not several layers downstream.

leonapivato:
Fixes #3116

For more information as to why this validation no longer happens at the options class level look at the cited issue.

### AI/LLM disclosure


TsafrirA (commented):
Thanks @leonapivato, this is an important fix.

There are some inconsistencies here which I think require some more work. In `ZneOptions`, there is already a validation for `len(noise_factors)>2`. That validation happens at construction time, while the one added here happens at run time. Which do we prefer?

Furthermore, the new validation is only applied in `prepare_zne` but is missing from `prepare_pea`. So, even if we want to keep the validation at run time, I think it should be triggered at the `prepare` function level and not at the individual `prepare_x` level.

I think I am fine with leaving the validation at run time (doing it at construction time opens us up to issues with updates and their ordering), but we need to make sure we cover all the bases.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -102,6 +104,16 @@ def prepare(
             add_tags=add_tags,
         )
     elif options.resilience.zne_mitigation:
+        if options.resilience.zne.noise_factors == "auto":
```
TsafrirA: This creates a redundancy, as we do the same thing at the `prepare_x` level (over there we also have explicit type casting which might be missing here?).

I see two options and personally I am fine with either:

- Keep the code here, and remove the `auto` handling from the sub-prepare functions (type hints, resolution code etc).
- Remove the code here, and call the validation at the sub-prepare level.

I think this mostly depends on the interface the sub-prepares want to offer (as they will be moving out of the repo). If they want to have the `auto` available to them, I think it makes sense to validate there. If they prefer to not have the `auto` then we can resolve it here.

TsafrirA (commented):
Question about code redundancy.

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: @SamFerracin @dekelmeirom 
What do you think?

dekelmeirom: I don't see a good reason to keep the `auto` option available in the sub-prepares once they move, but I guess I will have to do such validation also in the sub-prepare level once it is moved, because it might be used directly and not through the wrapper, so although the `auto` can be resolved in the `prepare` level, we might want to do the validation in the sub-prepare level.
Anyway, I think we should keep the casting to numpy arrays in the sub-prepares because we will keep a list as a valid input type to the sub-prepares.

TsafrirA: Thanks @dekelmeirom for the input. Looking at the situation from a broader perspective, I think we have to do both resolution of `auto` and validation at the sub-prepare level. Doing it at the global prepare level creates a contract mismatch, because it still passes `ZneOptions` to the sub-prepares which are then forced to consider the `auto` scenario (and potentially validate).

@leonapivato - Do you agree?

TsafrirA (approved):
Thank you!

PR #3163, opened 2026-08-02
Change executor simulator metadata

### LINKED ISSUE #3162: Populate metadata in `run_quantum_program` results

There are two types of metadata:
- Global - attached to the execution result
- Local - per input pub

No metadata is populated currently.

Internally we run Aer's `SamplerV2`, which returns both global and local metadata:
- The global metadata contains a single key: `version`
- The local metadata keys are `shots`, `circuit_metadata`, `simulator_metadata`

For the executor, I suggest to:
- Ignore the version for now, and return empty global metadata
- Return the same local metadata as the sampler

Note two problems with my suggestion:
- The real executor returns empty local metadata (possibly to be removed), since it uses the pass through data for returning this type of information
- The local metadata doesn't match the executor schema

However also for the other simulators, for example the sampler simulator, the metadata doesn't match the schema, and what's returned from the real device. It makes sense to me, because the schemas' context is the contract with the server via Rest API, which is irrelevant for the simulators.

**Acceptance criteria**

For the executor simulator, return the same local metadata as the sampler.

yaelbh:
### Summary

This pull request implements the acceptance criterion of #3162:
> For the executor simulator, return the same local metadata as the sampler.

### Details and comments

Fixes #3162.

~~I'm going to mark it as Draft, until we decide what we want to return in the metadata, following [@TsafrirA's comment](https://github.com/Qiskit/qiskit-ibm-runtime/issues/3162#issuecomment-5156605749).~~ Tsafrir and I agreed that we can go ahead with this PR.

### AI/LLM disclosure


yaelbh:
Current code doesn't populate the metadata in the sampler's local mode, this will be done separately.

yaelbh:
Readers of this PR may become confused because the terms `sampler` and `local` are overloaded:
- There are two samplers: the aer sampler and the wrapper sampler. Both are sometimes called only "sampler", and the reader is assumed to determine from the context which sampler it is. The executor is a layer between these two samplers.
- Local mode and local metadata - local mode is the wrapper sampler when it runs with a simulator; local metadata is the metadata of a pub result, in contrast to a program's result.

qiskit_ibm_runtime/executor_local_mode/run_quantum_program.py
apchytr, line 148: Does metadata here need to be updated in anyway?

apchytr (approved):
Besides a small question looks good

qiskit_ibm_runtime/decoders/executor_sampler/converters.py
```diff
@@ -75,17 +77,19 @@ def quantum_program_item_result_to_sampler_pub_result(
...
-            "circuit_duration": item.metadata.scheduler_timing.circuit_duration,
-        }
-    if item.metadata.stretch_values:
-        pub_metadata.setdefault("compilation", {})
-        pub_metadata["compilation"]["stretch_values"] = [
-            asdict(stretch_value, dict_factory=expanded_values_to_lists)
-            for stretch_value in item.metadata.stretch_values
-        ]
+
+    if isinstance(item.metadata, ItemMetadata):
```
SamFerracin, line 81: Here, despite the fact that there are a lot of red and green lines, you only added this if statement, correct?

SamFerracin (approved):

qiskit_ibm_runtime/executor_local_mode/run_quantum_program.py
yaelbh: I'm not aware of any information that we want to return to the user, at the level of the entire program.

qiskit_ibm_runtime/decoders/executor_sampler/converters.py
yaelbh: Yes, it's a clumsy way to distinguish between local and non-local modes.

PR #3165, opened 2026-08-03
Fix FakeProviderForBackendV2 missing 8 newer fake backends

### LINKED ISSUE #3164: FakeProviderForBackendV2 missing 8 newer fake backends (FakeAachen, FakeBerlin, FakeBoston, FakeBrussels, FakeKingston, FakeMiami, FakePittsburgh, FakeStrasbourg)
## Environment
qiskit-ibm-runtime: 0.48.0 (latest on PyPI)

## What happened
`FakeProviderForBackendV2` is documented as containing "all fake V2 backends available in the `qiskit_ibm_runtime.fake_provider`" module, and its docs page states it uses the list from the module's own backend page.

However, name-based lookup fails for the following backends that clearly exist and are directly importable:
- `FakeAachen`, 
- `FakeBerlin`, 
- `FakeBoston`, 
- `FakeBrussels`, 
- `FakeKingston`,
- `FakeMiami`,
- `FakePittsburgh`, and
- `FakeStrasbourg`

```python
from qiskit_ibm_runtime.fake_provider import FakeKingston
FakeKingston().num_qubits  # 156, works fine

from qiskit_ibm_runtime.fake_provider import FakeProviderForBackendV2
FakeProviderForBackendV2().backend(name="fake_kingston")
# QiskitBackendNotFoundError
```

Looking at [`FakeProviderForBackendV2.__init__`](https://github.com/Qiskit/qiskit-ibm-runtime/blob/stable/0.48/qiskit_ibm_runtime/fake_provider/fake_provider.py#L29-L116), `self._backends` is a hardcoded list literal that simply never includes `FakeKingston` (or 7 others). There's no conditional logic, feature flag, or comment explaining the omission.

## Reproduction: full list of missing classes

```python
from qiskit_ibm_runtime.fake_provider import FakeProviderForBackendV2
import qiskit_ibm_runtime.fake_provider as fp

provider = FakeProviderForBackendV2()
registered = {type(b).__name__ for b in provider.backends()}
all_classes = {n for n in dir(fp) if n.startswith("Fake") and n != "FakeProviderForBackendV2"}
print(sorted(all_classes - registered))
```
```
['FakeAachen', 'FakeBerlin', 'FakeBoston', 'FakeBrussels', 'FakeKingston',
 'FakeMiami', 'FakePittsburgh', 'FakeStrasbourg']
```

## Expected behavior
`FakeProviderForBackendV2().backend(name="fake_kingston")` (and the other 7) should resolve, matching the class's own documentation and matching user expectations given the classes exist and are directly importable.

## Suggested fix
Add the 8 missing classes to the hardcoded list, ideally via a more maintainable mechanism (e.g. auto-discovering `Fake*` subclasses of `FakeBackendV2` in the module) so this doesn't silently recur as new devices are added in future releases.

hat-im:
### Summary

`FakeProviderForBackendV2`'s hardcoded `_backends` list was missing 8 fake backend classes that
exist and are directly importable from `qiskit_ibm_runtime.fake_provider`: `FakeAachen`,
`FakeBerlin`, `FakeBoston`, `FakeBrussels`, `FakeKingston`, `FakeMiami`, `FakePittsburgh`, and
`FakeStrasbourg`. As a result, name-based lookup (e.g. `provider.backend(name="fake_kingston")`)
raised `QiskitBackendNotFoundError` for these backends despite them being valid.

### Details and comments

- Adds all 8 classes to `_backends`, alphabetically, matching the existing ordering.
- Adds a regression test, `test_all_fake_backends_registered`, asserting every `Fake*BackendV2`
  subclass in the `fake_provider` module is present in `FakeProviderForBackendV2().backends()`.
  Verified it fails against the pre-fix code (listing exactly the 8 backends from the issue) and
  passes after the fix.
- Adds a release note.

Fixes #3164

### AI/LLM disclosure


TsafrirA (commented):
Thank you @hat-im!
This looks solid to me, but I'll ping @diego-plan9 to provide a formal review.

test/unit/fake_provider/test_fake_backends.py
```diff
@@ -22,12 +22,14 @@
 from qiskit.utils import optionals
 
 from qiskit_ibm_runtime import QiskitRuntimeService, SamplerV2
+from qiskit_ibm_runtime import fake_provider as fake_provider_module
```
diego-plan9: Minor, but does `fake_provider` clash with other item? If not, I'd suggest not renaming the import - as it is arguably more clear over what we are iterating later on (ie. we can rely on the usual name).

release-notes/unreleased/3164.bug.rst
```diff
@@ -0,0 +1,7 @@
+Fix :class:`.FakeProviderForBackendV2` to include the 8 fake backends that were
+importable from :mod:`qiskit_ibm_runtime.fake_provider` but missing from the provider's
+backend list: :class:`.FakeAachen`, :class:`.FakeBerlin`, :class:`.FakeBoston`,
+:class:`.FakeBrussels`, :class:`.FakeKingston`, :class:`.FakeMiami`, :class:`.FakePittsburgh`,
+and :class:`.FakeStrasbourg`. Previously, calling
+``FakeProviderForBackendV2().backend(name="fake_kingston")`` (and the other 7) raised a
+``QiskitBackendNotFoundError`` even though the backend classes existed.
```
diego-plan9: Can you tweak so it better points to the method that has been tweaked?

```
:meth:`.FakeProviderForBackendV2.backends` now returns ...
```

This guides users better through specific changes.

diego-plan9 (commented):
Thanks hat-im, good catch! Some small suggestions but is looking good!

test/unit/fake_provider/test_fake_backends.py
hat-im: Thank you for the catch. I missed that in my final version. I've undone the rename.

release-notes/unreleased/3164.bug.rst
hat-im: I've reworded it in my latest commit. Please let me know if this is better.

diego-plan9: It does, thanks!

diego-plan9 (approved):
Thanks for the fix and for the neat PR!

Issue #3144, opened 2026-07-30
Decide the fate of `selected_extrapolators`


TsafrirA:
**What is the expected enhancement?**
Following the merger of #2953 the ZNE extrapolation pathway computes the `selected_extrapolators`, but it is not reflected in the result handed back to the user. The property is motivated by a similar property that existed in the past in the legacy estimator (but no longer is).

As it stands, the `selected_extrapolators` object contains the choice of extrapolator for every term in every observable in every configuration (=parameter set) for every extrapolated noise level (with 0 added to the values requested by the user). This choice is not very informative, as different extrapolated values are computed differently. Namely:

- `evs` - extrapolation to 0 noise which is done with a mixed strategy. Each term is extrapolated on its own, the best extrapolator is chosen, and then summed to get observable expectation value. This means that more than one extrapolation method could be used in the estimation of a single observable. This is where the selection process is relevant.
- `evs_extrapolated` - extrapolation to the user specified noise levels. Here the returned values are not using the same mixed extrapolation strategy but rather include the expectation value for each extrapolation method, where terms are using the same strategy (i.e. one result where all terms use linear extrapolation, one result where all terms use exponential extrapolation and so on, and no mixed extrapolation). In this case no selection needs to be made (though it is).

So, the `selected_extrapolators` are only relevant when the mixed extrapolation strategy is used, but that in turn only happens with the mandatory extrapolation to 0 noise. The selection for other extrapolated noise levels doesn't have any effect on the results the user gets back (and does not help with calculating the results, because the user doesn't get back the results per term).

**Acceptance criteria**
We need to:

1. Decide what, if any, do we want to return to the user when it comes to `selected_extrapolators`?
2. Compute only the needed values, and return them to the user.

TsafrirA:
#3166 will resolve many of the problems we've encountered. Furthermore, contrary to what the issue says the selected extrapolators are reported in the legacy pathway (in a way consistent with #3166) as part of the pub metadata.

PR #3166, opened 2026-08-03
keep only the zero point for selected extrap

dekelmeirom:
### Summary
Return the selected extrapolators for each observable term only for the zero noise extrapolation point.

### Details and comments

Fixes one of the issues raised in #3144 .
There is no use in the selected extrapolations for the other extrapolation points, as the we return expectation value calculation for each extrapolator type, and not a calculation of mixed extrapolator types (like we do for the zero noise point).
The data is still not returned to the user, and should be added somewhere in the metadata (if decided to return this information at all).

### AI/LLM disclosure


qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -1059,7 +1059,8 @@ def calculate_extrapolated_expectation_values(
                 )
             )
 
-            selected_extrapolators_per_term.append(sel_extrapolators)
+            # Only the selected extrapolator of the zero point should be returned
+            selected_extrapolators_per_term.append(sel_extrapolators[0])
```
TsafrirA: This definitely solves it, but aren't we doing needless work? Calculating `sel_extrapolators` for every noise factor and then throwing it way?

dekelmeirom: This is part of a broader discussion of whether the functions should support only the path we use in the wrapper-estimator, or whether some functions will become public and might be used as part of custom processing. As part of this discussion, we need to decide if `process_extrapolated_expectation_values` (in which the `sel_extrapolators` is first calculated) should be a public function with option to use it directly.
If we keep this function private - I agree that there is some redundant processing

TsafrirA: Following an in-person discussion we agreed to remove the over-calculation currently done in the extrapolation process (i.e not calculating the mixed strategy fits for all extrapolated noise factors, and consequently, not calculating the selected extrapolators for each one). @dekelmeirom's broader question could be revisited when the code moves out of `qiskit-ibm-runtime`.

(FYI - @SamFerracin)

dekelmeirom:
Selected extrapolators is calculated only for the needed (zero point extrapolation) points. In addition, a function to calculate the combined selected extrapolator per observable was added (if all of the observable's terms used the same extrapolator - that extrapolator is returned, and if different extrapolators were used for different terms of the same observable, "multiple" is returned)

TsafrirA (commented):
Thanks @dekelmeirom! This looks good. With your permission, I will take a bit more time to test this end to end.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
```diff
@@ -219,11 +218,11 @@ def clamp_degenerate_stds(y_std: np.ndarray) -> np.ndarray | None:
 
 
 def select_zne_extrapolated_result(
-    zne_values: npt.NDArray[float],
-    zne_std_errors: npt.NDArray[float],
+    zne_values: npt.Array[float],
```
TsafrirA: I don't see an `Array` in npt. Are you sure about this? Because we hide it behind the TYPE_CHECKING, it can slip through the tests.

qiskit_ibm_runtime/executor_estimator/zne/extrapolation.py
TsafrirA: The names were not updated.

test/unit/executor_estimator/test_zne_extrapolation.py
```diff
@@ -265,26 +265,23 @@ def test_selects_valid_or_fallback(
...
-        res_values, res_stderrs, res_extraps = select_zne_extrapolated_result(
+        zero_point_exp_val, zero_point_stderr, res_extrap = select_zne_extrapolated_result(
             values, stderrs, observable_term, extraps
         )
-        self.assertEqual(res_values[0], expected_value)
-        self.assertEqual(res_extraps[0], expected_extrapolator)
+        self.assertEqual(zero_point_exp_val, expected_value)
+        self.assertEqual(res_extrap, expected_extrapolator)
 
     def test_selects_per_column_independently(self):
```
TsafrirA: The test name is outdated.

TsafrirA (commented):
Thanks @dekelmeirom just a few minor nitpicks and we can merge.

test/unit/executor_estimator/test_zne_extrapolation.py
dekelmeirom: The `test_selects_valid_or_fallback` test now tests the same as this test, so I removed it

TsafrirA: Even better. Thank you.

TsafrirA (approved):
Thank you!

PR #3168, opened 2026-08-03
Add unit test for wrapper estimator Samplex arguments

### LINKED ISSUE #3167: Write tests to verify the arguments of `SamplexItem` created by wrapper estimator prepare functions
We need to unify the testing logic for the arguments of the Samplex created by the wrapper estimator prepare functions.

This is similar to #3087 which did a similar thing to the a different output of the prepare functions (the param-basis mapping).

TsafrirA:
### Summary
Fixes #3167 
This PR unifies the testing of another output of the prepare functions - the Samplex arguments. I think the validation of the Samplex shape should be done separately, because the shape (assuming we include the shot axis in it) varies differently than the arguments, and will require different test cases.

### AI/LLM disclosure


test/unit/executor_estimator/utils.py
```diff
@@ -91,3 +98,216 @@ class ParamBasisScenarios:
...
+distinct noise-injected gate layers).
+"""
+
+
+def assert_samplex_item(
+    test_case: IBMTestCase,
+    item: SamplexItem,
+    scenario: SamplexCircuitScenario,
+    inject_noise: bool,
+) -> None:
```
diego-plan9: Can you create a new `IBMTestCase` subclass in `ibm_test_case.py`, and make this function a method of it (ideally following CamelCase, which is a bit of an oddity, but arguably brings us closer to the unittest assert "flavor")?

This would make it inline with #3049, where the assert-like methods were consolidated in such Cases, as well as the hierarchy revised. It allows for more natural assertions (and in practice, prevents having to pass `TestCase` instances - which does work but can be a bit surprising).

TsafrirA: So, the ~~test cases~~ test scenarios themselves would be in utils, or directly under the new TestCase?

diego-plan9: `utils` seems good - I should have disclaimed it was not a full review, just on that bit.

The assert seems generic enough (needs a `SamplexCircuitScenario`, but does not depend on all of them or how they are defined), and the new `Scenarios` seem similar to the existing `Scenarios` to still live in utils.

TsafrirA: [31c9ecb](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3168/commits/31c9ecb2670921caf7fff2e9a686a1a42a76c81b)

test/unit/executor_estimator/utils.py
```diff
@@ -91,3 +94,107 @@ class ParamBasisScenarios:
...
+            num_basis_changes=1,
+            num_noise_maps=2,
+        ),
+    ]
+
+
+SAMPLEX_CIRCUIT_SCENARIOS: list[SamplexCircuitScenario] = _make_samplex_circuit_scenarios()
+"""Four circuit scenarios for samplex argument tests.
+
+Covers: regular, parametric, mid-circuit measurement, and multi-layer (two
```
SamFerracin: Slight preference for
```suggestion
Covers: non-parametric, parametric, mid-circuit measurement, and multi-layer (two
```
or for
```suggestion
Covers: static (i.e., non-parametric) , parametric, mid-circuit measurement, and multi-layer (two
```
since "regular" is generic. I'd also update the label accordingly

test/ibm_test_case.py
```diff
@@ -169,6 +172,117 @@ def assertWarnsStrict(
                 )
 
 
+class IBMEstimatorPrepareTestCase(IBMTestCase):
+    """TestCase with assertions for estimator prepare-function tests."""
+
+    def assertSamplexItem(
```
SamFerracin: Nit: "asserting a samplex item" isn't correct wording, I'd add a hint as to what you're actually asserting, for example: 
```suggestion
    def assertSamplexItemIsCorrect(
```

SamFerracin (commented):
This looks very neat, thanks @TsafrirA! I have left a couple of minor comments!

test/unit/executor_estimator/utils.py
TsafrirA: Done.

test/ibm_test_case.py
TsafrirA: Done.

SamFerracin (approved):

PR #3170, opened 2026-08-03
Fix dtype of noise factors in prepare_pea function for consistency

leonapivato:
### PEA integer `noise_factors` fails QPY encoding

`prepare_pea.py` builds `noise_factors = np.array(zne_options.noise_factors)` with no `dtype=float` (unlike `prepare_zne.py`, which casts explicitly). An integer `noise_factors` sequence therefore encodes as `int64`, which the QPY/tensor schema rejects: `ValueError: Unexpected NumPy dtype 'int64', one of ('f64', 'bool', 'u8', 'c128') expected.`

### AI/LLM disclosure


dekelmeirom (approved):
Good catch!

diego-plan9 (approved):

PR #3171, opened 2026-08-04
Preserve QiskitRuntimeService state when using private environment

diego-plan9:
### Summary

Revision to #3135, moving the skipping of validation of instances lower into the stack. This should allow for the state variables in `QiskitRuntimeService` to be more consistent with the regular path, in particular the `._all_instances` variable which guards further attempts to retrieve instance information from the global services - and minimize the risk of inconsistent behavior when using that feature.

### Details and comments

Related to #3134

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/accounts/account.py
+++ b/qiskit_ibm_runtime/accounts/account.py
@@ -335,6 +335,18 @@ class CloudAccount(Account):
 
     def list_instances(self) -> list[dict[str, Any]]:
         """Retrieve all crns with the IBM Cloud Global Search API."""
+        # Bypass calling Global Search and Global Catalog.
+        if os.environ.get("QISKIT_FUNCTIONS_EXPERIMENTAL"):
+            return [
+                {
+                    "crn": self.instance,
+                    "plan": "plan",
+                    "name": "name",
+                    "tags": [],
+                    "pricing_type": "unknown",
+                }
+            ]
+
         authenticator = self.get_iam_authentificator()
         client = GlobalSearchV2(authenticator=authenticator)
         catalog = GlobalCatalogV1(authenticator=authenticator)
--- a/test/unit/test_qiskit_runtime_service.py
+++ b/test/unit/test_qiskit_runtime_service.py
@@ -101,7 +101,7 @@ class TestQiskitRuntimeService(IBMTestCase):
         service = QiskitRuntimeService(token="my_token")
         # `_all_instances` contains all the instances available.
         self.assertEqual({instance["crn"] for instance in service._all_instances}, crns_in_registry)
-        # `_backend_configs` is empty.
+        # `_backend_configs` is empty (populated by `backends()`).
         self.assertEqual(service._backend_configs, {})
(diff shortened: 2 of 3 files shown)


test/unit/test_qiskit_runtime_service.py
```diff
@@ -79,3 +88,64 @@ def test_run_active_client(self, registry):
...
+            for instance in registry.instances.values()
+        }
+
+        service = QiskitRuntimeService(token="my_token")
+        # `_all_instances` contains all the instances available.
+        self.assertEqual({instance["crn"] for instance in service._all_instances}, crns_in_registry)
+        # `_backend_configs` is empty (populated by `backends()`).
+        self.assertEqual(service._backend_configs, {})
+        # `_api_clients` contains one client per instance available.
+        self.assertEqual(set(service._api_clients.keys()), crns_in_registry)
```
rainerschoe, line 107: how can client API keys be equal to CRNs? (In my understanding CRNs identify an instance, not a client key)

qiskit_ibm_runtime/accounts/account.py
```diff
@@ -335,6 +335,18 @@ def resolve_crn(self) -> None:
 
     def list_instances(self) -> list[dict[str, Any]]:
         """Retrieve all crns with the IBM Cloud Global Search API."""
+        # Bypass calling Global Search and Global Catalog.
+        if os.environ.get("QISKIT_FUNCTIONS_EXPERIMENTAL"):
+            return [
+                {
+                    "crn": self.instance,
```
rainerschoe, line 342: with this, it does not matter chich instance / crn the user passes if this flag is set?

test/unit/test_qiskit_runtime_service.py
diego-plan9: This is probably naming and knowledge of the structure: https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/qiskit_runtime_service.py#L278

`._api_clients` is a dict with CRNs as keys, and  `RuntimeClient` instances as values. So this assert checks on the keys of the dictionary (which should be CRNs).

qiskit_ibm_runtime/accounts/account.py
diego-plan9: It does matter as the instance passed at service instantiation (`QiskitRuntimeService(..., instance="foo")`) is the one that ends up in `Account.instance` (if you comb through the nested calls, you will see that the `Account` is initialized with `foo`). It is expected that users using that environment variable will always pass the argument to the service constructor (and this is a bit of undocumented behavior, but assumed by the envisaged consumers).

test/unit/test_qiskit_runtime_service.py
rainerschoe: ahhh thanks. I was somehow thinking of crypto keys, not seeing them being python dict keys. makes sense

rainerschoe (approved):
Looks good. Thanks for the explanations :-)

PR #3174, opened 2026-08-05
Test the impact of resilience levels for Executor-based Estimator (Local mode simulation)

rainerschoe:
### Summary
This uses local mode of EstimatorV2 via Executor to assert on the expected effects of resilience levels (0,1,2).

part of #3210 

### Details and comments

WIP, inspired by #3143

### AI/LLM disclosure

### DIFF
--- a/test/unit/executor_estimator/simulations/test_estimator.py
+++ b/test/unit/executor_estimator/simulations/test_estimator.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+import numpy as np
+from ddt import ddt
+from qiskit.primitives import StatevectorEstimator
+from qiskit.quantum_info import SparsePauliOp
+from qiskit_aer import AerSimulator
+
+# TODO: This import is not yet available. Maybe we can directly skip this,
+# once local mode of Estimator is supported
+from qiskit_ibm_runtime.sim_executor import SimExecutor
+
+from qiskit_ibm_runtime.executor_estimator import EstimatorV2
+from qiskit_ibm_runtime.fake_provider import FakeManilaV2
+
+from ....ibm_test_case import IBMTestCase
+from ....utils import make_mirror_circuit_with_phases
+
+
+@ddt
+class TestSampler(IBMTestCase):
+    def setUp(self):
+        super().setUp()
+        self.backend = FakeManilaV2()
+
+    def test_simple_mirror_circuit_correct_evs(self):
+        circuit = make_mirror_circuit_with_phases(self.backend)
+
+        # TODO: I think we need to explicitly define parameters in order to write assertion on EVs
+        parameters = np.random.random((2, 3) + (circuit.num_parameters,))
+
+        estimator = EstimatorV2(self.backend)
+        estimator._executor = SimExecutor(AerSimulator())
+
+        observable = SparsePauliOp("ZZ")
+        pub = (circuit, parameters, [observable])
+        result = estimator.run([pub]).result()
+
+        statevector_estimator = StatevectorEstimator()
+        statevector_result = statevector_estimator.run([pub]).result()
+
+        # TODO: make this a class assertion + check for proximity rather than equivalence
+        assert result[0].evs[0] == statevector_result[0].evs[0]
--- a/test/unit/executor_estimator/simulations/test_estimator.py
+++ b/test/unit/executor_estimator/simulations/test_estimator.py
@@ -1,14 +1,24 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2026.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Tests Executor based EstimatorV2 implementation using simulator through local mode."""
+
 from __future__ import annotations
 
 import numpy as np
 from ddt import ddt
 from qiskit.primitives import StatevectorEstimator
 from qiskit.quantum_info import SparsePauliOp
-from qiskit_aer import AerSimulator
-
-# TODO: This import is not yet available. Maybe we can directly skip this,
-# once local mode of Estimator is supported
-from qiskit_ibm_runtime.sim_executor import SimExecutor
+from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
 
 from qiskit_ibm_runtime.executor_estimator import EstimatorV2
 from qiskit_ibm_runtime.fake_provider import FakeManilaV2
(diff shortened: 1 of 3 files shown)


test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -0,0 +1,43 @@
...
+
+        observable = SparsePauliOp("ZZ")
+        pub = (circuit, parameters, [observable])
+        result = estimator.run([pub]).result()
+
+        statevector_estimator = StatevectorEstimator()
+        statevector_result = statevector_estimator.run([pub]).result()
+
+        # TODO: make this a class assertion + check for proximity rather than equivalence
+        assert result[0].evs[0] == statevector_result[0].evs[0]
```
TsafrirA: Note that the sampler returns samples and not expectation values, so you want to look in data, and then under the correct creg name (should be `meas` for this circuit). Also, when comparing samples you are essentially comparing distributions, so a different comparison strategy will be needed. It might be more complex depending on the parameters.

rainerschoe: This is Estimator :-)

TsafrirA: :)
The class name through me completely off!

rainerschoe: Haha, early draft. I based this on Sam's draft for Sampler. https://github.com/Qiskit/qiskit-ibm-runtime/pull/3143

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -0,0 +1,101 @@
...
+            max_error = np.max(np.abs(evs - statevector_evs))
+            # Assert a base-level of quality across all resilience levels:
+            self.assertLessEqual(max_error, 0.15)
+            mean_errors[resilience_level] = float(mean_error)
+
+        print(f"Mean errors per resilience level: {mean_errors}")
+
+        # Increased resilience level should translate into increased expectation value quality:
+        self.assertGreater(mean_errors[0], mean_errors[1])
+        self.assertGreater(mean_errors[1], mean_errors[2])
```
rainerschoe: This works surprisingly well, which makes me wonder how local mode simulation is configured by default. With this, I start to think we already have a useful noise model injected there.

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -0,0 +1,101 @@
...
+        parameters = np.random.uniform(
+            0,
+            2 * np.pi,
+            size=(num_randomizations, circuit.num_parameters),
+        )
+
+        # Prepare a PUB with an observable to estimate expectation values on.
+        # Using "ZZ" observable, as the mirror circuit has parametric rx gates, which should yield
+        # variations on Z projection.
+        observable = SparsePauliOp("ZZ").apply_layout(isa_circuit.layout)
```
```suggestion
        observable = ObservableArray.from_list(["ZZ", "IZ", "ZI", "01", ...])
        observable = observable.reshape(...)  # may be a different test
        observable = observable.apply_layout(isa_circuit.layout)
```

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -0,0 +1,101 @@
...
+        Compares the results against a statevector simulation.
+        """
+        # Use standard mirror circuit, but without measurements, as StatevectorEstimator does
+        # not support them.
+        circuit = make_mirror_circuit_with_phases(self.backend, add_measurement=False)
+        isa_circuit = self.preset_pass_manager.run(circuit)
+
+        # We want to run multiple random variants of the circuit to get different expectation
+        # values and to improve statistical significance:
+        num_randomizations = 5
```
```suggestion
        num_parameter_sets = 5
```

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+
+        # Prepare a PUB with an observable to estimate expectation values on.
+        # Using "ZZ" observable, as the mirror circuit has parametric rx gates, which should yield
+        # variations on Z projection.
+        observable = SparsePauliOp("ZZ").apply_layout(isa_circuit.layout)
+        pub = (isa_circuit, [observable], parameters)
+
+        # Calculate ground truth to compare the results against via a statevector simulation:
+        statevector_estimator = StatevectorEstimator()
```
SamFerracin, line 79: i wonder if state vector can handle projectors, like measuring 0 and 1 instead of Z

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -0,0 +1,101 @@
...
+            mean_error = np.mean(np.abs(evs - statevector_evs))
+            max_error = np.max(np.abs(evs - statevector_evs))
+            # Assert a base-level of quality across all resilience levels:
+            self.assertLessEqual(max_error, 0.15)
+            mean_errors[resilience_level] = float(mean_error)
+
+        print(f"Mean errors per resilience level: {mean_errors}")
+
+        # Increased resilience level should translate into increased expectation value quality:
+        self.assertGreater(mean_errors[0], mean_errors[1])
```
```suggestion
        self.assertGreater(mean_errors[0], mean_errors[1], msg=f"Mean errors per resilience level: {mean_errors}")
```

test/unit/executor_estimator/simulations/test_estimator.py
rainerschoe: Sam explained: By using fake backends instead of Aer directly, we have noise :-)

test/unit/executor_estimator/simulations/test_estimator.py
rainerschoe: I tried. It cannot :-( Sticking to Z-like for now

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -0,0 +1,111 @@
...
+            size=(1, circuit.num_parameters),
+        )
+
+        # Prepare a PUB with multiple observables to estimate expectation values on.
+        # Using "Z" observables, as the mirror circuit has parametric rx gates, which should yield
+        # variations on Z projection.
+        observables = [
+            SparsePauliOp(pauli_string).apply_layout(isa_circuit.layout)
+            for pauli_string in ["ZZ", "IZ", "ZI"]
+        ]
```
rainerschoe, line 75: Tried `ObservablesArray` but this causes CI to fail for lowest qiskit version

rainerschoe: Are those observables enough?
I tried adding exotic things like `XY`, but they seem quite unaffected and then I fail to see a difference in resilience level causing the test to fail.

SamFerracin: XY are valid, but due to the circuit you are considering, you'll see expectation values being 0, whether or not you have noise.

For evs(X) = 1, you need to start (or end) the circuit with an `h` gate. For evs(Y) = 1, you need to apply some combination of `h` with `s` or `sdg` that we can find out tomorrow

SamFerracin (approved):

PR #3175, opened 2026-08-05
Add (noiseless) executor local mode support to wrapper sampler

apchytr:
### Summary
Updating the wrapper sampler to support the new executor based local mode simulations when the experimental setting `local_mode` is set to `True`.  Note: this PR only supports the noiseless case with noise to be followed up in subsequent PRs.

### Details and comments

Part of #3129 and #3177

### AI/LLM disclosure


test/unit/executor_sampler/test_sampler.py
```diff
@@ -372,21 +372,21 @@ def test_simulator_mode_uses_backend_sampler(self):
...
         result = job.result()
 
         # Verify we got results
         self.assertEqual(len(result), 1)
-        self.assertIsNotNone(result[0].data)
+        self.assertIsNotNone(result[0])
 
         # Verify the results are valid Bell state measurements
-        counts = result[0].data.c.get_counts()
+        counts = BitArray.from_bool_array(result[0]["c"]).get_counts()
```
SamFerracin: This line should not change

qiskit_ibm_runtime/executor_local_mode/executor_local_mode.py
```diff
@@ -64,13 +66,20 @@ def __init__(
         program: QuantumProgram,
         options: ExperimentalSimulatorOptions,
     ):
+        from ..decoders.quantum_program.decoder import QuantumProgramResultDecoder
+
+        def _fn(
```
diego-plan9: This comment will hopefully be rendered moot by other comments, but in general, please strive for meaningful names for items.

qiskit_ibm_runtime/executor_local_mode/executor_local_mode.py
```diff
@@ -64,13 +66,20 @@ def __init__(
...
+            backend: BackendV2, program: QuantumProgram, options: ExperimentalSimulatorOptions
+        ) -> QuantumProgramResult | PrimitiveResult:
+            return QuantumProgramResultDecoder._apply_post_processing(
+                run_quantum_program(backend, program, options)
+            )
```
diego-plan9: In the spirit of #3140 review, `PrimitiveJob` model differentiates between:
* instantiating the job
* submitting the job
* retrieving the result of the job

And `RuntimeJobV2` also follows it. If we consider that post processing is related to the _results_ of the job, you can maintain the same pattern by (pseudo):

```python
    def result(self) -> something:
        from ..decoders.quantum_program.decoder import QuantumProgramResultDecoder
        
        return QuantumProgramResultDecoder._apply_post_processing(super().result())
```

which makes `SimRuntimeJob` closer to the rest of job-related classes, and allows to contain / signal better when using the imported items are needed.

qiskit_ibm_runtime/executor_local_mode/executor_local_mode.py
diego-plan9, line 1: Technically not part of this PR, but can you take the chance to rename this file? `sim_runtime_job.py` , for example - so it provides useful info / hints of what it contains.

qiskit_ibm_runtime/executor_local_mode/executor_local_mode.py
```diff
@@ -64,13 +66,20 @@ def __init__(
         program: QuantumProgram,
         options: ExperimentalSimulatorOptions,
     ):
+        from ..decoders.quantum_program.decoder import QuantumProgramResultDecoder
```
diego-plan9: I cannot see an easy way of avoiding the import cycle without some heavier refactoring (the fact that we need to request post processing might be a signal for revising how post processing dispatching takes place). But hopefully it can be moved further down and specific to the function that needs it.

qiskit_ibm_runtime/executor_local_mode/run_quantum_program.py
```diff
@@ -52,31 +52,32 @@ def _round_to_clifford(values: np.ndarray, decimals: int) -> np.ndarray:
...
-        warn_absent: Passed to :class:`~.InsertNoisePass`.
+        options: The simulator options to use.
 
     Returns:
         Results of simulation.
     """
+    noise_dict = options.noise_model
+    angle_decimals = options.angle_decimals
+    warn_absent = options.warn_absent
+    seed = options.seed_simulator
```
diego-plan9: Minor: if these variables are not modified (I did not check for mutations when passed to other functions), and only used (at most) twice - might be easier for readability if you just pass `options.foo` to those functions? Currently they verge on being similar to aliasing them and adding an extra (and tiny) layer of indirection.

qiskit_ibm_runtime/executor_local_mode/run_quantum_program.py
```diff
@@ -144,9 +142,11 @@ def run_quantum_program(
         else:
             raise TypeError(f"Unsupported QuantumProgramItem type: {type(prog_item)}")
 
-    return QuantumProgramResult(
+    ret = QuantumProgramResult(
         data=result_list,
         # metadata=dict(enumerate(metadata_list)),
         metadata=None,  # TODO: Figure this out
```
SamFerracin: Can we make the suggested change, and open an issue instead (if it's not open yet)?
```suggestion
metadata=None,
```

test/unit/executor_sampler/test_sampler.py
SamFerracin, line 362: @apchytr we should consider if it's worth keeping this test class (plus/minus some of the test functions in it) or if it will be redundant with the integration tests that we're writing elsewhere

qiskit_ibm_runtime/options_models/converters.py
```diff
@@ -67,6 +68,11 @@ def sampler_option_to_executor_options(options: SamplerOptions) -> ExecutorOptio
             if execution_key.get("stretch_values", False):
                 executor_options.execution.stretch_values = True
 
+        if options.experimental.get("local_mode", False):
+            executor_options.experimental["simulator_options"] = options.experimental.get(
+                "simulator_options", ExperimentalSimulatorOptions()
+            )
```
SamFerracin: What is this doing? Is it necessary?

test/unit/executor_sampler/test_sampler.py
apchytr: Makes sense 👍 imo I think a combination of unit and integration probably makes sense where we clean these up to be a bit more focused

SamFerracin (approved):
Looks "good"--that private method in `SimRuntimeJob` is really ugly, but we can probably clean it up somewhat in a followup PR

PR #3181, opened 2026-08-06
Add tests to verify the structure of the template circuit after prepare in the wrapper estimator

### LINKED ISSUE #3180: Write tests to cover the template circuit after the prepare stage of wrapper estimator
Similar to #3167, we need to verify the structure of the template circuit coming out of the various `prepare_x` stages. This includes the twirling structure (for example, the number of parameters in the template circuit).

**Acceptance criteria**
Tests are added and duplicates removed.

TsafrirA:
### Summary
~~NO MERGE - builds on top of #3167~~
Fixes #3180 

### AI/LLM disclosure


test/ibm_test_case.py
```diff
@@ -282,6 +290,31 @@ def assertSamplexItemIsCorrect(
                 msg=f"[{scenario.label}] pauli_lindblad_maps must be absent; keys={keys}",
             )
 
+        # --- template circuit checks ---
```
TsafrirA: The main thing we want to verify here is that the circuits were properly twirled. This is surprisingly tricky to do, as it depends heavily on the twirling strategy.

It will arguably be easier to look at the Samplex and count things like twirling nodes, but that's not a user facing interface. So, we end up counting parameters\rz gates, which indicate where twirling was added (again, would be sensitive to changes in the default decomposition).

I think it's not worth it to test modifications to the twirling strategy here.

test/unit/executor_estimator/utils.py
```diff
@@ -124,6 +124,33 @@ class SamplexCircuitScenario:
     Depends on the number of mid-circuit boxes in the circuit.
     """
 
+    expected_num_clbits: int
+    """Expected number of classical bits in the template circuit.
+
+    Equal to ``original_qubits + original_clbits``.
+    """
+
+    num_circuit_parameters_gates_on: int
```
TsafrirA, line 153: This and the next two values are meant to capture how the number of parameters changes in the circuit as you play with enable gates\measures and do ZNE with gate folding.

This will probably not work for every circuit out there, but I don't see a reason to formalize a bullet-proof way of describing this for every circuit.

TsafrirA:
@SamFerracin 
My biggest question was - should these assertions be enforced together with the samplex arguments ones? On the one hand, it complicates things a bit, and removes some future flexibility (say you want to extend to twiriling strategy - one set of assertions doesn't care, while the other does). On the other hand, it is logically consistent to have all validations of the samplex items together. Since I don't think we need separate test cases for each set of assertions, I opted to have them together.

Looking ahead, I believe when we add the tests to assert the shape of the samplex item, the decision would be different - there you need different test cases, and it makes less sense to have them together.

What do you think?

SamFerracin:
Thanks for this!

A few comments and suggestions:
- I get your point that "this complicates things a bit, and removes some future flexibility," and I agree that this may be coming from the fact that more than one thing (samplex arguments and template) is being tested with the same test function. I believe that it would be simpler if these were different tests. In particular, I would:
  - Make two different test functions, `test_samplex_arguments` and `test_template_circuits`--one concern per test function
  -  Rename `assertSamplexItemIsCorrect` as `assertSamplexArgumentsAreCorrect` 
  -  Add `assertTemplateIsCorrect`
- "Looking ahead, I believe when we add the tests to assert the shape of the samplex item, ..." Yes, I agree on this too. And I'd be in favour of adding a third test function for that, since this is a different concern

Separate from the above:
If the test functions are different, there's an opportunity to simplify even further. Each one of them can have their own `SCENARIO`, you don't need to use `SAMPLEX_CIRCUIT_SCENARIOS`. And I suspect you could choose circuits that allow you to simplify your assertions even further. For example, you could consider testing with this circuit, which is simpler--you either have parametric gates or not, depending on whether you are or are not applying twirling, and it's easy to test the various strategies too if you feel like you need to
```python
circuit = QuantumCircuit(3)
circuit.cx(0, 1)
circuit.cx(1, 2)
circuit.measure_all()
```

TsafrirA:
Thanks @SamFerracin 
I split into two tests and assertion methods. I used a single test case, but made it more complex than the one you suggested, because it showcases richer effects. I don't think others are needed.

I did not parameterize over the measurement noise learning flag. I think it should be removed from the samplex arguments tests as well, but I'll leave it for another PR.

SamFerracin (approved):
Thanks Tsafrir, looks great!

PR #3182, opened 2026-08-07
Warn when `finalize_options()` overrides a property explicitly set by the user

TsafrirA:
### Summary
This PR adds a warning when `finalize_options()` overrides a field which was explicitly set by the user. It uses the recent transition to pydantic models which allow the distinction between defaults and values explicitly set by the user.

### AI/LLM disclosure


TsafrirA:
@SamFerracin 
I verified that legacy estimator silently overrides without erroring or warning, at least for this example:
```python
estimator = EstimatorV2(backend)
estimator.options.resilience.pec_mitigation = True
estimator.options.twirling.enable_gates = False
```
The job runs and the metadata shows
```
 'twirling': {'enable_gates': True,
  'enable_measure': True,
  'num_randomizations': 'auto',
  'shots_per_randomization': 'auto',
  'interleave_randomizations': True,
  'strategy': 'active-accum'},
```

SamFerracin (approved):
Thanks, Tsafrir!

Issue #3151, opened 2026-07-30
Add backward-compatibility tests to check that results of old executor-based jobs can be loaded


SamFerracin:
**What is the expected feature?**
IQP retains data for several years. This means that technically, a user with a years-old `job_id` may expect the following code to work:
```python
old_job = service.job(job_id)
result = old_job.result()
```

We have no tests in place to guarantee that this is the case. Adding tests becomes even more important with the introduction of executor-based primitives, that rely on versioned post-processing functions to calculate the `result` object.

**Acceptance criteria**
- [ ] Identify a strategy for testing that results of old jobs can be loaded.
 
This will require addressing questions of the following type:
- How can we store job ids in a way that we can retrieve them in CI/CD?
- What is the cadence with which we want to store job ids?
- What sampler and estimator "paths" do we need to cover?

- [ ] Write the tests

PR #3183, opened 2026-08-07
Allow storing of jobs for backwards compatibility tests

diego-plan9:
### Summary

Extend the approach from #3112 in order to allow mocking the calls related to the job minimally - to the extent needed for easily using specific raw responses for `details` and `results` during tests.

Add a script for retrieving an existing job and storing it for consumption with the tests, as well as an initial test that loads a stored job (for `executor`) and asserts on some basic properties.

### Details and comments

Part of #3151

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude Opus 4.8


diego-plan9:
This now includes an `executor` job, as it previously used a `WrapperSampler` job from `0.49.0` that cannot be loaded against `main` (for good reasons).

test/registries.py
```diff
@@ -116,6 +116,28 @@ def from_(
...
+class Job:
+    """Registry representation of a job."""
+
+    id: str
+    """Job id."""
+
+    backend_name: str
+    """Backend name."""
+
+    program: Literal["sampler", "estimator"] = "sampler"
```
SamFerracin: executor as well?

test/registries.py
```diff
@@ -116,6 +116,28 @@ def from_(
...
+
+    status: Literal["queued", "running", "completed", "cancelled", "failed"] = "completed"
+    """Job status."""
```
SamFerracin: What's the reason to give defaults here?

SamFerracin (commented):
Thanks! Looks good! Two minor questions

test/registries.py
diego-plan9: Does not hurt! 11b9a19

test/registries.py
diego-plan9: For user convenience. The dataclasses in this file act as the "handy interface for populating the registry with items". For example:

```
        registry.add_job(
            Job(job_id, "ibm_fez", raw_details=job_details, raw_results=job_results), "a"
        )
```

By having a sensible default that is likely to be the most common (or a property that does not matter for the test at hand), no need for typing the `status="completed"` all the time.

SamFerracin (approved):

PR #3185, opened 2026-08-07
Add shape parameter to create_pub_result_pea and create_pub_result_zn…

leonapivato:
`create_pub_result_zne` and `create_pub_result_pea` in `post_processor_v0_1.py` built their `DataBin` without passing `shape=`, so it silently defaulted to `()` regardless of the pub's actual broadcast shape — while `create_pub_result`/`create_pub_result_pec` already passed it correctly. The evs/stds arrays themselves were always correctly shaped; only the declared `DataBin.shape` metadata was wrong, so any broadcasting pub (multiple observables, or a parameter sweep) under ZNE or PEA reported a scalar shape instead of the true one

### AI/LLM disclosure


SamFerracin (approved):

TsafrirA:
@leonapivato 
Does this match legacy behavior? Note that the different evs\stds don't share the same shape. They all have the same basic shape (given from the broadcasting of parameters and observables) but each one has extra dimensions that are different.

leonapivato:
It should match the legacy. Not only does it now pass my tests, but I checked, and the legacy essentially does `databin = DataBin(shape=data["evs"].shape, **data)` for every occurrence.

PR #3191, opened 2026-08-09
Fix the noise injection pass for dynamic circuits

### LINKED ISSUE #3133: `SimExecutor`'s noise injection is not recursive in control flow operations

From looking at the code (I didn't run anything to confirm this suspicion): the noise injection pass skips control flow operations, and does not inject noise inside their blocks. This should be fixed.

**Acceptance criteria**

- [ ] Noise injection works properly for dynamic circuits. Alternatively raise an error for the combination of noise injection and dynamic circuits.

yaelbh:
### Summary

Fixes #3133.

#3133 discusses whether to fix the noise injection pass for dynamic circuits, considering that this is not a use case that we envisage. This PR takes the approach of fixing the pass: I find it simpler than figuring out exactly which use cases we're expected to support, raising errors or warnings for what we don't support, and documenting all this.

### Details and comments

Still in progress. In particular it will be probably modified to align with decisions about qubit ordering that were taken in https://github.com/Qiskit/qiskit-noise-learning/pull/36.

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_local_mode/insert_noise_pass.py
+++ b/qiskit_ibm_runtime/executor_local_mode/insert_noise_pass.py
@@ -74,9 +74,7 @@ class InsertNoisePass(TransformationPass):
         physical_index = {q: dag.find_bit(q).index for q in dag.qubits}
         return self._process_dag(dag, physical_index)
 
-    def _process_dag(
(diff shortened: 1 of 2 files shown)


yaelbh:
@SamFerracin It's ready for review.

PR #3192, opened 2026-08-09
Add ZNE integration test

TsafrirA:
### Summary
This PR adds a ZNE integration test, and cleans up the wrapper estimator's integration tests.

### AI/LLM disclosure

### DIFF
--- a/test/integration/test_executor_estimator.py
+++ b/test/integration/test_executor_estimator.py
@@ -123,7 +123,7 @@ class TestEstimator(IBMIntegrationTestCase):
         # 2 Expectation values should have been calculated for 1 to 1 parameter mapping:
         self.assertEqual(results[1].data.evs.shape, (2,))
 
-    @data("auto", "pea")
+    @data("gate_folding", "pea")
     def test_zne_estimator(self, amplifier):
         """Test the ZNE path for estimator, parameterized over the noise amplifier.
 
--- a/test/integration/test_executor_estimator.py
+++ b/test/integration/test_executor_estimator.py
@@ -15,14 +15,17 @@
 from __future__ import annotations
 
 import numpy as np
-from ddt import ddt
-from qiskit.primitives.base import BaseEstimatorV2  # noqa: TC002
+from ddt import data, ddt
 from qiskit.quantum_info import PauliLindbladMap, SparsePauliOp
 from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
 from samplomatic import InjectNoise
 from samplomatic.utils import get_annotation
 
 from qiskit_ibm_runtime.executor_estimator import EstimatorV2
+from qiskit_ibm_runtime.options_models.zne import (
+    PEA_DEFAULT_NOISE_FACTORS,
+    ZNE_DEFAULT_NOISE_FACTORS,
+)
 
 from ..ibm_test_case import IBMIntegrationTestCase
 from ..utils import make_mirror_circuit_with_phases
@@ -31,8 +34,6 @@ from ..utils import make_mirror_circuit_with_phases
 @ddt
 class TestEstimator(IBMIntegrationTestCase):
     """An integration test, testing EstimatorV2 implemented through Executor."""
-
-    estimator_variant: type[BaseEstimatorV2] | None = None
 
     def setUp(self):
         """Test level setup."""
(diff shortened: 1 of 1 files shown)


TsafrirA:
(Verified locally that the test pass)

test/integration/test_executor_estimator.py
```diff
@@ -43,44 +44,53 @@ def setUp(self):
...
-                ),
-                # Map each parameter set to one observable:
-                (
-                    isa_circuit,
-                    [zz_with_offset, xx_with_offset],
-                    [[0, np.pi / 4], [np.pi, 5 * np.pi / 4]],
-                ),
-            ]
-        )
+    def _make_stub_noise_model(self, estimator):
```
TsafrirA: This one doesn't run for every test, so it's a coin flip whether to have it as a property or a method.

SamFerracin: NIT: I don't fully know what "stub" means here, I'd drop it
```suggestion
    def make_noise_model(self, estimator):
```

SamFerracin: I would leave it inside the two tests (PEC, PEA), since it's just a few lines of code. It also makes them nice-looking, for people that want to see a working example or are looking for a snippet to copy/paste

test/integration/test_executor_estimator.py
```diff
@@ -146,3 +122,58 @@ def test_pec_estimator(self):
...
+              ``(*pub_shape, num_extrapolators, num_extrapolated_noise_factors)``
+        """
+        estimator = EstimatorV2(self.backend)
+        estimator.options.resilience.zne_mitigation = True
+        estimator.options.resilience.zne.amplifier = amplifier
+
+        if amplifier == "pea":
+            estimator.options.resilience.noise_model_mapping = self._make_stub_noise_model(
+                estimator
+            )
```
SamFerracin: If my previous comment is taken into account, this becomes:
```suggestion
            layers = estimator.find_unique_layers(self._pubs)
            noise_model_mapping = {
                annotation.ref: PauliLindbladMap.from_list([("X" * layer.operation.num_qubits, 0.001)])
                for layer in layers
                if (annotation := get_annotation(layer.operation, InjectNoise))
                }
            estimator.options.resilience.noise_model_mapping = noise_model_mapping
```

test/integration/test_executor_estimator.py
```diff
@@ -146,3 +122,58 @@ def test_pec_estimator(self):
...
+            self.assertEqual(data_bin.stds_noise_factors.shape, expected_nf_shape)
+            self.assertEqual(data_bin.ensemble_stds_noise_factors.shape, expected_nf_shape)
+
+            # extrapolated arrays: (*pub_shape, num_extrapolators, num_extrapolated_noise_factors)
+            expected_extrap_shape = expected_pub_shape + (
+                expected_num_extrapolators,
+                expected_num_extrapolated,
+            )
+            self.assertEqual(data_bin.evs_extrapolated.shape, expected_extrap_shape)
+            self.assertEqual(data_bin.stds_extrapolated.shape, expected_extrap_shape)
```
SamFerracin, line 181: @rainerschoe and I had considered making very few assertions here --iirc, only the evs-- and leaving the metadata fields out of assertions until they are documented, deciding in future PRs if we wanted to add them here or test them only in the simulator tests. Would you be against doing that?

test/integration/test_executor_estimator.py
```diff
@@ -43,44 +44,53 @@ def setUp(self):
...
         zz_with_offset = SparsePauliOp.from_list([("ZZ", 1.0), ("II", 9.0)]).apply_layout(
             isa_circuit.layout
         )
-
         xx_with_offset = SparsePauliOp.from_list([("XX", 1.0), ("II", 3.0)]).apply_layout(
             isa_circuit.layout
         )
 
-        estimator = EstimatorV2(self.backend)
+        self._pubs = [
```
SamFerracin:
```suggestion
        self.pubs = [
```

SamFerracin (commented):
Thanks, Looks good! A few small comments!

test/integration/test_executor_estimator.py
TsafrirA: Once we have the simulator-based integration tests, I think we can remove it from here, to keep the tests simple (remembering that these tests don't run with every push to main). But until then I would personally prefer to have it tested.

test/integration/test_executor_estimator.py
TsafrirA: Removed the function.

test/integration/test_executor_estimator.py
TsafrirA: Done.

test/integration/test_executor_estimator.py
TsafrirA: Done.

test/integration/test_executor_estimator.py
TsafrirA: Removed the function.

rainerschoe:
Cool, I still want to look at this more closely, but an initial question: Why are we writing this as an integration test (E2E)? Why cannot we run this as a test against local mode simulation, which @apchytr introduced last Friday?

SamFerracin:
> Why are we writing this as an integration test (E2E)?

The failure that we fixed on friday happened during serialization of the payload. Estimator was adding elements to the passthrough data that cannot be serialized. Since local mode does not serialize the payload, it would not be able to catch it

test/integration/test_executor_estimator.py
```diff
@@ -43,44 +44,44 @@ def setUp(self):
...
-                    [[0, np.pi / 4], [np.pi, 5 * np.pi / 4]],
-                ),
-                # Map each parameter set to one observable:
-                (
-                    isa_circuit,
-                    [zz_with_offset, xx_with_offset],
-                    [[0, np.pi / 4], [np.pi, 5 * np.pi / 4]],
-                ),
-            ]
-        )
```
rainerschoe, line 81: I don't understand this change. The test is a noop test now, as estimator is not run. Also why assign to `self.pubs` ?

TsafrirA: It's part of the setup. We create the pubs once and reuse them for all tests.

rainerschoe: Ah thanks :P Read the diff wrong. not I see the `setUp` 👍

SamFerracin (approved):

PR #3195, opened 2026-08-10
Move `as_array` out of main loop in wrapper estimator utils

TsafrirA:
### Summary
The parameters are converted to array with each iteration of the main loop in the wrapper estimator utils. This is redundant and could be done once ahead of the loop.

### AI/LLM disclosure


SamFerracin (commented):
Good catch!

SamFerracin (approved):

PR #3196, opened 2026-08-10
Add `SamplerV2.find_unique_layers` and `add_tags` always true for executor-based primitives

### LINKED ISSUE #3177: Update wrapper sampler to use `sim_executor` in local mode
With the new `SimRuntimeJob` the wrapper sampler should switch from using the `QiskitRuntimeLocalService` to the new aer based approach when running jobs in local mode

**Acceptance criteria**

- [x] Update wrapper sampler to use `sim_executor` in local mode (noiseless)
- [x] Add in find_unique_layers for wrapper sampler (with tags)
- [x] Add noise support for local mode

apchytr:
### Summary
Adding in `find_unique_layers` for the wrapper sampler and setting `add_tags` to `True` for both wrappers. This provides users with layers that are always tagged for noise injection in the local mode simulator 

### Details and comments

Closes #3177

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_sampler/sampler.py
+++ b/qiskit_ibm_runtime/executor_sampler/sampler.py
@@ -139,13 +139,33 @@ class SamplerV2(BaseSamplerV2):
             The unique boxed layers found across the given PUBs.
         """
         coerced_pubs = [SamplerPub.coerce(pub, None) for pub in pubs]
+        options = self.finalize_options()
         return find_unique_layers(
             pubs=coerced_pubs,
-            twirling_options=self.options.twirling,
+            twirling_options=options.twirling,
             measure_noise_learning=None,
             inject_noise=False,
             add_tags=True,
         )
+
+    def finalize_options(self) -> SamplerOptions:
+        """Construct and finalize the Sampler options.
+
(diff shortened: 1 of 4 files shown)


qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -124,6 +126,27 @@ def __setattr__(self, name: str, value: Any) -> None:
...
+        The returned list contains one instance of each distinct boxed layer (represented as a
+        :class:`~.CircuitInstruction`) appearing in the input PUBs.
+
+        Args:
+            pubs: The list of PUBs to return a list of unique boxes for.
+
+        Returns:
+            The unique boxed layers found across the given PUBs.
+        """
+        coerced_pubs = [SamplerPub.coerce(pub, None) for pub in pubs]
```
SamFerracin, line 141: I would follow the same path as for estimator, where we expose `finalize_options` and apply it before finding unique layers.
```suggestion
        coerced_pubs = [SamplerPub.coerce(pub, None) for pub in pubs]
        options = self.finalize_options()
```

We make minor adjustments to the options, but still, it ensures that we are not missing something crucial

SamFerracin (approved):

PR #3204, opened 2026-08-11
[WIP] Enhange Estimator test to cover more observables

rainerschoe:
### Summary
part of #3210 

Extend the existing test to work on a 3 qubit circuit, estimating X, Y, Z projections and a combined XYZ observable.

I played around without simulator seed, but then I needed to loosen the resilience level 2 error assertion from 0.025 to 0.04 to eliminate flakiness. with 0.025 I had one failing test roughly every 8 runs. 0.03 every 12 runs, 0.04 one failure after 37 runs.
Based on this, I kept the fixed simulator seed, which ran 40 runs without error even with 0.025 threshold (I stopped then).

### AI/LLM disclosure


test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -54,21 +55,34 @@ def test_result_quality_for_different_resilience_levels(self):
...
+            3 * np.pi / 8,
+        ]
 
         # Prepare a PUB with multiple observables to estimate expectation values on.
-        # Using "Z" observables, as the mirror circuit has parametric rx gates, which should yield
-        # variations on Z projection.
         observables = [
             SparsePauliOp(pauli_string).apply_layout(isa_circuit.layout)
-            for pauli_string in ["ZZ", "IZ", "ZI"]
+            for pauli_string in ["XII", "IYI", "IIZ", "XYZ"]
```
rainerschoe, line 83: @SamFerracin The thing which was causing all the issues yesterday was that I had the order of Paulis wrong: `IIX` instead of `XII` etc. Then we estimated 0 very often which was not susceptible to noise, causing mitigation to not improve results. Took us quite a while to understand this. Tsafrir helped me debug and found the issue.

SamFerracin: Lol, classic! I always get it wrong too!

rainerschoe:
this has been covered by other PRs by now.

PR #3205, opened 2026-08-11
Improve resilience-level simulation test for Executor-based Estimator: Use static parameters, increase shots, tighten precision assert

rainerschoe:
### Summary

We want to assert a more tight boundary for mitigated estimator.
To account for statistical errors I increase number of shots and also keep the rz angles fixed to allow mitigation techniques to be more effective. 

part of #3210 

### AI/LLM disclosure

### DIFF
--- a/test/unit/executor_estimator/simulations/test_estimator.py
+++ b/test/unit/executor_estimator/simulations/test_estimator.py
@@ -59,12 +59,7 @@ class TestEstimator(IBMTestCase):
         isa_circuit = self.preset_pass_manager.run(circuit)
 
         # Select values for the rz gates:
-        np.random.seed(42)
-        parameters = np.random.uniform(
-            0,
-            2 * np.pi,
-            size=(1, circuit.num_parameters),
-        )
+        parameters = np.array([3.5 * np.pi / 4] * circuit.num_parameters)
 
         # Prepare a PUB with multiple observables to estimate expectation values on.
         # Using "Z" observables, as the mirror circuit has parametric rx gates, which should yield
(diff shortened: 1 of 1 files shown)


test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -100,12 +98,11 @@ def test_result_quality_for_different_resilience_levels(self):
             evs = result[0].data.evs
             errors[resilience_level] = np.abs(evs - statevector_evs)
 
-            # TODO: also look at stds.
-
+        print(statevector_evs)
```
```suggestion
```

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -59,12 +59,7 @@ def test_result_quality_for_different_resilience_levels(self):
         isa_circuit = self.preset_pass_manager.run(circuit)
 
         # Select values for the rz gates:
```
SamFerracin: I think these are rx gates, eh?
```suggestion
        # Select values for the rx gates:
```

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -92,6 +87,9 @@ def test_result_quality_for_different_resilience_levels(self):
                 # instead of connecting to a real backend.
                 experimental={"local_mode": True},
             )
+            options.twirling.num_randomizations = 100
+            options.twirling.shots_per_randomization = 200
+            options.default_shots = 100 * 200
```
TsafrirA, line 96: When you set all three of those fields together, default shots is just ignored.

SamFerracin (approved):
This looks good to me, thanks @rainerschoe !
I've re-run it locally a few times, printing the `evs`. They are quite stable now, with:

- Resilience level 0 amounting to an error ~0.15
- Level 1 with ~0.08 error
- Level 2 error ~0.01 error

Re-running gives pretty much the same values

test/unit/executor_estimator/simulations/test_estimator.py
rainerschoe: Interesting. Also for resilence level 0? I observed some difference in test runtime after setting default_shots, but I might be mistaken.

PR #3206, opened 2026-08-11
Import QuantumProgram and related classes from Samplomatic

SamFerracin:
### Summary
Depends on https://github.com/Qiskit/samplomatic/pull/396

### AI/LLM disclosure


SamFerracin:
A guide to the reviewer:
- Start from the `QuantumProgram` class in `qiskit_ibm_runtime/quantum_program/quantum_program.py`.
- Proceed with the result-related classes in `qiskit_ibm_runtime/results/quantum_program.py`.
- Check the various init files, for example `qiskit_ibm_runtime/quantum_program/__init__.py`, to see how issue deprecation warnings. A CTRL+F of `issue_deprecation_msg` will reveal more of these deprecations.
- Everything else should just be miscellanea that comes from the first three bullet points.
- Finally, check out the release notes.

qiskit_ibm_runtime/quantum_program/__init__.py
```diff
@@ -39,4 +39,21 @@
...
 """
 
-from .quantum_program import CircuitItem, QuantumProgram, QuantumProgramItem, SamplexItem
+from ..utils.deprecation import issue_deprecation_msg
+from .quantum_program import QuantumProgram
+
+_DEPRECATED_NAMES = frozenset({"CircuitItem", "QuantumProgramItem", "SamplexItem"})
+
+
+def __getattr__(name: str) -> object:
```
apchytr: Just raising that the LSP won't pick up the deprecated objects here. So, a user could import`CircuitItem` for example, get the deprecation warning, and not get type hints, docstrings, etc

SamFerracin: LSP?

apchytr: Language Server Protocol

qiskit_ibm_runtime/quantum_program/datatree.py
```diff
@@ -1,24 +0,0 @@
...
-
-"""DataTree."""
-
-from __future__ import annotations
-
-from typing import TypeAlias
-
-from numpy.typing import NDArray
-
-DataTree: TypeAlias = (
```
yaelbh, line 21: Is there a reason not to deprecate `DataTree`?

qiskit_ibm_runtime/visualization/draw_chunk_timings.py
```diff
@@ -12,183 +12,14 @@
...
-_HOVER_HEADER = "<br>".join(
-    [
-        "<b>{name}[{idx}]</b>",
-        "<b>&nbsp;&nbsp;&nbsp;Start:</b> {start:%Y-%m-%d %H:%M:%S.%f}",
-        "<b>&nbsp;&nbsp;&nbsp;Stop:</b> {stop:%Y-%m-%d %H:%M:%S.%f}",
-        "<b>&nbsp;&nbsp;&nbsp;Duration:</b> {duration:.4g}s",
-        "<b>&nbsp;&nbsp;&nbsp;Size:</b> {size}",
-        "<b>&nbsp;&nbsp;&nbsp;Parts ({n_parts}):</b>",
-    ]
+issue_deprecation_msg(
```
yaelbh: There are two `issue_deprecation_msg` for `draw_chunk_timings`: here and in `__init__.py`. On the other hand for other deprecations, such as `CircuitItem`, there is only one `issue_deprecation_msg`.

release-notes/unreleased/3206.feat.rst
```diff
@@ -0,0 +1,2 @@
+- `QuantumProgram`, `QuantumProgramResult`, and related classes are now imported from Samplomatic. Importing them from qiskit-ibm-runtime is deprecated and will be removed in a subsequent release.
```
yaelbh: Specifically `QuantumProgram` and `QuantumProgramResult` are still to be imported by the user from qiskit-ibm-runtime, and the import from samplomatic is done only internally.

test/integration/test_executor.py
```diff
@@ -17,7 +17,8 @@
 from samplomatic import build
 from samplomatic.transpiler import generate_boxing_pass_manager
 
-from qiskit_ibm_runtime import Executor, QuantumProgram
+from qiskit_ibm_runtime import Executor
+from qiskit_ibm_runtime.quantum_program import QuantumProgram
```
yaelbh: Why? I don't see a change in `qiskit_ibm_runtime.__init__`.

test/unit/executor/test_dynamical_decoupling.py
```diff
@@ -26,6 +26,7 @@
     TimeUnitConversion,
 )
 from samplomatic import Twirl, build
+from samplomatic.quantum_program import CircuitItem, QuantumProgram, SamplexItem
```
yaelbh: Do you want to import here samplomatic's `QuantumProgram` or our `QuantumProgram`?

test/unit/executor_sampler/test_prepare.py
```diff
@@ -19,14 +19,13 @@
 from qiskit import QuantumCircuit
 from qiskit.circuit import Parameter
 from samplomatic import Tag
+from samplomatic.quantum_program import CircuitItem, QuantumProgram, SamplexItem
```
yaelbh: Same question as before.

test/unit/quantum_program/test_circuit_item.py
```diff
@@ -14,8 +14,7 @@
 
 import numpy as np
 from qiskit.circuit import Parameter, QuantumCircuit
-
-from qiskit_ibm_runtime.quantum_program.quantum_program import CircuitItem
+from samplomatic.quantum_program import CircuitItem
```
yaelbh: Do we still want the tests for `CircuitItem`, not that it belongs to samplomatic? Probably this file can be deleted.

test/unit/quantum_program/test_params_converters.py
```diff
@@ -15,8 +15,8 @@
 from ddt import data, ddt
 from qiskit.circuit import QuantumCircuit
 
-from qiskit_ibm_runtime import QuantumProgram
 from qiskit_ibm_runtime.options_models import ExecutorOptions
+from qiskit_ibm_runtime.quantum_program import QuantumProgram
```
yaelbh: Same question as before.

test/unit/quantum_program/test_samplex_item.py
```diff
@@ -16,8 +16,7 @@
 from qiskit.circuit import Parameter, QuantumCircuit
 from qiskit.quantum_info import PauliLindbladMap
 from samplomatic import InjectNoise, Twirl, build
-
-from qiskit_ibm_runtime.quantum_program.quantum_program import SamplexItem
+from samplomatic.quantum_program import SamplexItem
```
yaelbh: Same question as for the `CircuitItem` tests.

test/unit/quantum_program/test_result.py
```diff
@@ -16,11 +16,9 @@
 
 import numpy as np
 from ddt import ddt
+from samplomatic.quantum_program import ChunkPart, ChunkSpan, ChunkTiming
 
-from qiskit_ibm_runtime.results.quantum_program import (
-    ChunkPart,
```
yaelbh, line 21: Tests of `ChunkPart` etc. can be probably removed.

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
```diff
@@ -51,7 +51,7 @@ def sampler_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveRes
             f"'{type(result.passthrough_data)}'."
         )
 
-    passthrough = cast("dict", result.passthrough_data or {})
+    passthrough = result.passthrough_data or {}
```
diego-plan9, line 54: Is this intentional?

The `cast` allows the following `if` to not complain in IDEs when using `passthrough.get` (and prevents inferring `circuits_metadata` type correctly and other items down the file). Can be kept if intentional on the grounds that we are not enforcing anything that is not caught by `mypy` at the moment, at the expense of "some more red lines in IDEs".

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
diego-plan9, line 92: Not changed in the PR currently, but likely worth chasing.

Pylance (when keeping the `cast` in the comment above) highlights this, as well as in `flatten_twirling_axes`  and `quantum_program_item_result_to_sampler_pub_result` with:

```
Argument of type "QuantumProgramItemResult" cannot be assigned to parameter "item" of type "QuantumProgramItemResult" in function "undo_twirling"
  "samplomatic.quantum_program.quantum_program_result.QuantumProgramItemResult" is not assignable to "qiskit_ibm_runtime.results.quantum_program.QuantumProgramItemResult"
```

Which is legit - `undo_twirling` does expect an `ours.QuantumProgramItemResult`, but this method receives an `our.QuantumProgramResult` which currently does _not_ override `samplomatic.QuantumProgramResult.__iter__` (and other methods) signatures, so `result[0]` is an `samplomatic.QuantumProgramItemResult`. 

I think we have two alternatives:
* make `undo_twirling` and other methods wider, accepting `samplomatic.*` instead of `ours.*` if those methods do not need to worry about out customizations. This keeps the method more general (which can be a good thing), although is likely not applicable blindly (at some point, we will need to worry about our customizations)
* revise `our.*` so the signatures are more narrow (ie. `QuantumProgramResult.__iter__` is `Iterator[our.QuantumProgramItemResult]` explicitly). I have not peeked into how safe (per Liskov's) it is or the extra complexity, but should be a safe general narrowing.

I think this can be tackled post-this PR to avoid extra complexity.

qiskit_ibm_runtime/quantum_program/quantum_program.py
```diff
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2025-2026.
+# (C) Copyright IBM 2025, 2026.
```
diego-plan9: This is likely not intentional?

qiskit_ibm_runtime/results/quantum_program.py
diego-plan9: Are these needed?

qiskit_ibm_runtime/quantum_program/__init__.py
diego-plan9: I share some concerns about the generality and robustness of this approach, mostly for being tied to import / discoverability mechanisms that we probably want to be extra-cautious about as it might have unintended side effects in certain environment or setups, as Anthony mentioned. As another small (trivial, solvable, and not really harmful) effect, for example, `from qiskit_ibm_runtime.quantum_program import *` will not pick up the items as-is.

I suspect the motivation is being able to mark an item in a module as deprecated _when_ the user imports it (but not when we internally import it) _while_ the module can contain other non-deprecated items (and thus we cannot deprecate the module fully [such as here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/utils/noise_learner_result.py)).

Perhaps we can keep them regular imports and rely on documentation and release notes? We do sacrifice the strong visibility that `DeprecationWarnings` have, but have more certainty and assurances.

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
SamFerracin: After the changes in this PR, the `cast` here has started giving me some issues. Particularly, mypy errors, saying that:
```
qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py:54: error: Redundant cast to "dict[Any, Any]"  [redundant-cast]
```
I haven't investigated why yet

qiskit_ibm_runtime/quantum_program/params_converters.py
```diff
@@ -41,7 +41,7 @@
     from ibm_quantum_schemas.common import BaseParamsModel
 
     from ..options_models import ExecutorOptions
-    from .quantum_program import QuantumProgram
+    from ..quantum_program import QuantumProgram
```
SamFerracin:
```suggestion
    from .quantum_program import QuantumProgram
```

test/unit/executor/test_dynamical_decoupling.py
```diff
@@ -34,11 +34,7 @@
 )
 from qiskit_ibm_runtime.fake_provider import FakeManilaV2
 from qiskit_ibm_runtime.options_models.dynamical_decoupling import DynamicalDecouplingOptions
-from qiskit_ibm_runtime.quantum_program.quantum_program import (
-    CircuitItem,
-    QuantumProgram,
-    SamplexItem,
-)
+from qiskit_ibm_runtime.quantum_program import CircuitItem, QuantumProgram, SamplexItem
```
SamFerracin:
```suggestion
from qiskit_ibm_runtime.quantum_program.quantum_program import (
    CircuitItem,
    QuantumProgram,
    SamplexItem,
)
```

qiskit_ibm_runtime/results/quantum_program.py
SamFerracin: no, i had forgotten them. removed: [7bb2958](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206/commits/7bb2958c70c73c919d038a590cf1b63720b6804b)

test/unit/quantum_program/test_result.py
SamFerracin: as mentioned today, let's leave all the tests for now, and remove them in a subsequent PR. they can be useful as a proof that our changes aren't breaking anything

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
SamFerracin: This is a very good point. I thought about this, and I feel like the second option (revising `our.*`) is appropriate, given that when I iterate over our result, I get objects of type `our.*`. And this should address your comment.
[cdf5445](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206/commits/cdf5445ed02497a1ad11908b36e0155201742bc2)

Happy to hear your follow-up thoughts!

SamFerracin:
Thank you for your comments!

I was really having troubles wrapping my head around this PR, and I concluded that I was trying to do too much all at once. Therefore, I decided to take a step back (at the cost of making a high number of pushes, apologies for all the spam emails 😅). But now I'm at a point where I can clearly say what this PR tries to do.

### What this PR does

[see release note]
- `QuantumProgram`, `QuantumProgramResult`, and `QuantumProgramItemResult` now inherit from the corresponding classes in Samplomatic.
- `CircuitItem`, `SamplexItem`, `DataTree`, `ChunkPart`, `ChunkSpan`, `ChunkTiming`, and `draw_chunk_timings` are now imported from Samplomatic. Importing them from qiskit-ibm-runtime is still supported.

Example to understand the second bullet: Where I had `class SamplexItem`, now I have `from samplomatic.quantum_program import SamplexItem` ([link](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206/changes#diff-cc5f3a8f9adc768ea42145e2ebb7111e9bd7eb17874d331b6cde4a76e9ea869fR24-R28)). 
This guarantees that:
- Scripts that contained `from qiskit_ibm_runtime.quantum_program import SamplexItem` keep working.
- No deprecation warnings
- Discoverability doesn't break

All in all, nothing changes. To prove it, I have reverted every other change, including:
- Import statements in other files: all reverted back to what they are in main
- Tests: No changes in the test files

### What this PR does not do

There are a few follow-up steps that we need to take. We can either take them in this PR or in subsequent PRs. I'll let the reviewers express their preference, and I'll open issues if we go with the latter.

#### 1. Update imports
- [ ] Update every single file (source, and tests), replacing imports like `from qiskit_ibm_runtime.quantum_program import SamplexItem` with imports like `from samplomatic.quantum_program import SamplexItem`.

#### 2. Deprecate importing from qiskit-ibm-runtime
- [ ] `from qiskit_ibm_runtime.quantum_program import SamplexItem` should be deprecated. Users should just import Samplomatic objects from Samplomatic via `from samplomatic.quantum_program import SamplexItem`. 

There are a few different ways of doing this. One is through a deprecation warning like I had in previous commits ([link](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206/changes/6e94c9b3505c3999cc1342aaf2ea6091d17d5396..db2e2d9c8eef8c04cdbf1a440b67dab483f120c4#diff-08660ec084d53960e4ecce73969a20eb706c83132669b58bab5113bb1edbef49R42-R55)), which is fine but breaks the LSP. Another one is through docstrings, as Diego suggests, which does not break LSP but may go unnoticed by users.

#### 3. Remove unnecessary tests
If `SamplexItem` lives in Samplomatic, then it should be tested in Samplomatic. We can remove the tests. 
- [ ] Remove unit tests that target `CircuitItem`, `SamplexItem`, `DataTree`, `ChunkPart`, `ChunkSpan`, `ChunkTiming`, and `draw_chunk_timings`.

qiskit_ibm_runtime/results/quantum_program.py
```diff
@@ -361,7 +187,7 @@ def timing(self) -> ChunkTiming:
 
         .. code-block:: python
 
-            from qiskit_ibm_runtime.visualization import draw_chunk_timings
+            from samplomatic.visualization.draw_chunk_timings import draw_chunk_timings
```
yaelbh: In this PR it should still remain the old import.

qiskit_ibm_runtime/results/quantum_program.py
```diff
@@ -374,3 +200,9 @@ def timing(self) -> ChunkTiming:
             A :class:`~.ChunkTiming` collection.
         """
         return ChunkTiming(self.metadata.chunk_timing)
+
+    def __getitem__(self, index: int | slice) -> QuantumProgramItemResult:
+        return cast("QuantumProgramItemResult", super().__getitem__(index))
+
+    def __iter__(self) -> Iterator[QuantumProgramItemResult]:
+        return cast("Iterator[QuantumProgramItemResult]", super().__iter__())
```
yaelbh: Why is it a bit different from the way it used to be in the old code?

qiskit_ibm_runtime/visualization/__init__.py
```diff
@@ -26,7 +26,6 @@
     :toctree: ../stubs/
     :nosignatures:
 
-    draw_chunk_timings
```
yaelbh, line 29: Not sure about this diff.

yaelbh:
I agree with your proposal to defer some of the modifications to a subsequent PR. The PR as it is now looks good to me, up to three comments that I left.

qiskit_ibm_runtime/results/quantum_program.py
```diff
@@ -374,3 +200,9 @@ def timing(self) -> ChunkTiming:
             A :class:`~.ChunkTiming` collection.
         """
         return ChunkTiming(self.metadata.chunk_timing)
+
+    def __getitem__(self, index: int | slice) -> QuantumProgramItemResult:
+        return cast("QuantumProgramItemResult", super().__getitem__(index))
```
TsafrirA: If you pass a slice here, the return type hint is wrong.

qiskit_ibm_runtime/results/quantum_program.py
```diff
@@ -271,30 +130,15 @@ def __init__(
...
-        del self._result[key]
-
-    def __iter__(self) -> Iterator[str]:
-        return iter(self._result)
-
-    def __len__(self) -> int:
-        return len(self._result)
-
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self._result}, metadata={self.metadata})"
```
TsafrirA: Potentially unsafe to rely on private attributes of the base class.

TsafrirA (commented):
Thanks @SamFerracin. I think it makes sense to finish this up with a follow up PR.

Two nitpicks for your consideration.

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
diego-plan9: Interesting - worth investigating as follow-up, let's keep the change in the meantime as it is far from critical.

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
diego-plan9: Sounds good - there might be some subtleties left related to typing that will be better checked post-merge, as our `mypy` configuration is a bit weak. I'll open an issue to capture this and the previous cast-related comment, this topic looks resolve this PR-wise to me.

As an aside - having `samplomatic` define the classes as generics would allow for easily saying "this MyQuantumProgramResult is the same as the base one, but contains MyQuantumProgramItemResults instead of the base ones".

diego-plan9: https://github.com/Qiskit/qiskit-ibm-runtime/issues/3279

qiskit_ibm_runtime/decoders/executor_sampler/post_processor_v0_1.py
diego-plan9: https://github.com/Qiskit/qiskit-ibm-runtime/issues/3279

qiskit_ibm_runtime/results/quantum_program.py
SamFerracin: see discussion here: https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206#discussion_r3862259517

qiskit_ibm_runtime/results/quantum_program.py
SamFerracin: Good catch: [9db6b7e](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206/commits/9db6b7eb3f30b46eda53391e529098f7ef3133a6)

qiskit_ibm_runtime/visualization/__init__.py
SamFerracin: I put it back: [7361305](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206/commits/73613054f30f4fb60eaa6d76e1b10e467deb8ac7)

qiskit_ibm_runtime/results/quantum_program.py
SamFerracin: Thanks: [bf3e522](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206/commits/bf3e522bbe5ff403671f59dfb2b7795c1e9099a7)

qiskit_ibm_runtime/results/quantum_program.py
SamFerracin: Good point, I reverted: [607a0ca](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206/commits/607a0cab65ca50d4945be86b99512fc0c55dd500)

diego-plan9 (approved):

Issue #3207, opened 2026-08-11
Initial feedback on Wrapper primitives


jyu00:
**What is the expected enhancement?**

I just started using Wrapper primitives for the first time, and here's some feedback before I forget:

- [x] It'd be nice if WrapperSampler also return all the input options in result metadata, to match that of WrapperEstimator (not high priority)
- [ ] It'd be nice if WrapperEstimator can also return ExecutionSpans (not high priority)
- [x] WrapperEstimator logging says  "Submitting 1 pub to executor with 128 shots", which is a bit misleading since it's really 128 shots _per randomization_. 
- [x] My code used TREX and I expected some logging for post-processing but there wasn't any.
- [ ] Right now it looks like `find_unique_layers()` returns both 2Q and measurement layers, and the user needs to strip out measurement layer based on (the lack of) `InjectNoise` annotation. It'd be less confusing to just have a `find_unique_2q_layers()` method. 

**Acceptance criteria**

See above.

SamFerracin:
Thanks, @jyu00 !
We'll go through these, and if you have more inputs, please add them here!

As per private discussion, perhaps let's not take action on the last one ("Right now it looks like find_unique_layers() returns both 2Q and measurement layers, ..") until we have a workflow sketched out for the local-mode, to make sure that we have the full picture in mind

TsafrirA:
Closing this as complete via #3212, #3213, #3245.
As for the last two items:
- Execution spans are added to the estimator metadata.
- `find_unique_layers` now has a `type` argument which allows for filtering.

PR #3213, opened 2026-08-12
Add the options to the result metadata in wrapper sampler

TsafrirA:
### Summary
Part of #3207 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_sampler/sampler.py
+++ b/qiskit_ibm_runtime/executor_sampler/sampler.py
@@ -219,6 +219,8 @@ class SamplerV2(BaseSamplerV2):
         quantum_program, executor_options = prepare(
             coerced_pubs, options, default_shots, backend=self._backend
         )
+        # Store raw options for post-processing side to compute metadata.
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -219,6 +219,8 @@ def run(self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None) -> Ru
         quantum_program, executor_options = prepare(
             coerced_pubs, options, default_shots, backend=self._backend
         )
+        # Store raw options for post-processing side to compute metadata.
+        quantum_program.passthrough_data["post_processor"]["options"] = options.model_dump()  # type: ignore[index, call-overload]
```
TsafrirA, line 223: I opted to put it under `options` because we already have `execution` field for the execution spans, and the execution options from the options can override it.

SamFerracin: Sounds good! It's the same as in estimator anyway!

SamFerracin (approved):

Issue #3210, opened 2026-08-12
Add simulation-based integration test for the efficacy of error mitigation techniques


TsafrirA:
**What is the expected enhancement?**
We want to add simulation-based test that will verify the efficacy of various error mitigation techniques. We want to demonstrate that given a noisy simulation, the estimate improves as the error mitigation is applied. Tests should cover different observables and parameters, and cover the following cases:

- [x] Resilience levels (vanilla, TREX, ZNE with gate folding) #3174
- [x] PEC
- [x] PEA

**Acceptance criteria**
Tests are added.

Issue #3209, opened 2026-08-12
Add simulation-based integration tests for noiseless estimates


TsafrirA:
**What is the expected enhancement?**
We want to add a simulation-based integration test to verify the correctness of estimates without noise. Without noise, one can compare directly against theoretical prediction.

With ZNE based tools it's a bit harder to predict the uncertainty of the extrapolation, so this should be limited to Vanilla and PEC.

**Acceptance criteria**
Tests are added.

PR #3214, opened 2026-08-12
[WIP] Add simulation-based integration test for the correctness of the estimator

TsafrirA:
### Summary
Part of #3209 (also draws a clear separation between the tests of #3210)

### AI/LLM disclosure

### DIFF
--- a/test/unit/executor_estimator/simulations/test_estimator.py
+++ b/test/unit/executor_estimator/simulations/test_estimator.py
@@ -18,7 +18,9 @@ from typing import TYPE_CHECKING
 
 import numpy as np
 from ddt import ddt
-from qiskit.primitives import StatevectorEstimator
+from qiskit import QuantumCircuit
+from qiskit.primitives import ObservablesArray, StatevectorEstimator
+from qiskit.providers.fake_provider import GenericBackendV2
 from qiskit.quantum_info import SparsePauliOp
 from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
 
@@ -35,8 +37,12 @@ if TYPE_CHECKING:
 
 
 @ddt
-class TestEstimator(IBMTestCase):
-    """Tests Executor based EstimatorV2 implementation using simulator through local mode."""
+class TestEstimatorErrorMitigationEfficacy(IBMTestCase):
+    """Tests the efficacy of error mitigation in the Executor based EstimatorV2 implementation.
+
+    The tests use noisy simulations to verify that expectation values get closer to the ideal
+    result as the mitigation is applied.
+    """
 
     def setUp(self):
         """Test level setup."""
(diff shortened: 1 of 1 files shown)


test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -109,3 +115,88 @@ def test_result_quality_for_different_resilience_levels(self):
...
+        ]
+        observables = ObservablesArray.coerce([obs_string for obs_string, _ in observable_pairs])
+        isa_observables = observables.apply_layout(isa_circuit.layout)
+        pub = (isa_circuit, isa_observables)
+
+        options = EstimatorOptions(
+            resilience_level=0,
+            experimental={
+                "local_mode": True,
+                # "simulator_options": ExperimentalSimulatorOptions(seed_simulator=42),
```
TsafrirA, line 170: With a random seed the test fails about half of the times. This is a bit more than I expected, I will go over the math again.

In any case, I think we should pick a passing seed, and aim for a relatively high failure rate with a random seed - it means the test is on the very edge, and would detect regressions well.

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -109,3 +115,88 @@ def test_result_quality_for_different_resilience_levels(self):
...
+        """Test level setup."""
+        super().setUp()
+        self.backend = GenericBackendV2(5, noise_info=False, seed=972)
+        # self.backend = AerSimulator()
+        self.preset_pass_manager = generate_preset_pass_manager(
+            optimization_level=1, target=self.backend.target
+        )
+        self.tolerance = 2  # In terms of stansdard deviations
+
+    def test_vanilla_correctness(self):
```
SamFerracin, line 138: I think this test is a bit too optimistic when it comes to testing precision. Regardless of what our docs say or do not say (and we can improve that), given well-established statistical arguments, all we can guarantee is a precision `p` that is `O(n_shots**-0.5)`. Any attempt at finding a constant `C` such that the errors are below `C/n_shots**0.5` is doomed to fail, either because the test becomes too strict and fails often or because it is too lenient and tests nothing [which echoes things we said today].

In the absence of noise, we can make two promises:
- Our estimates are unbiased
- If we increase `n_shots`, our `evs` get closer to the ideal values, and our `stds` shrink.

Therefore, I would split this test into two tests that verify those two concerns. For example, for vanilla you could have:
- A test where you fix a precision (default?) and try to see that the expectation values are unbiased. [For example, I have tried running your test with a single assert, namely `assert np.all(np.array(errors) < 0.02)`, without fixing the seed. I had to remove that one observable for which we expect `0`, which we know is problematic, but other than that, it always passes]
- Another test where you fix one observable, and run with 2/3 values of precision, such as 0.1, 0.01, 0.001. Then you test that what you call `errors` go down and that the stds go down, like Rainer did in his other test

If you go this way, an additional bonus is that you will not need seeds for the tests to pass

SamFerracin: In addition to this, you could consider parametrizing these tests over resilience levels 0, 1, and 2. So that we end up with 2 tests (times 3 -> 6 tests), and we only need to write two more for PEA and PEC

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -109,3 +115,88 @@ def test_result_quality_for_different_resilience_levels(self):
...
+            ("IIZ", np.cos(theta)),
+            ("IYZ", np.cos(theta)),
+            ("XIZ", np.cos(theta)),
+            ("1IZ", 0.5 * np.cos(theta)),
+            ("IYr", np.sin(np.pi / 4 - theta / 2) ** 2),
+            ("X+I", 0.5),
+            ("0IZ", 0.5 * np.cos(theta)),
+            ("IYl", np.cos(np.pi / 4 - theta / 2) ** 2),
+            ("X-I", 0.5),
+            ("ZXY", 0),
```
SamFerracin, line 160: This looks like a great test suite!

test/unit/executor_estimator/simulations/test_estimator.py
TsafrirA: > Any attempt at finding a constant C such that the errors are below C/n_shots**0.5 is doomed to fail, either because the test becomes too strict and fails often or because it is too lenient and tests nothing [which echoes things we said today].

This is true for noisy results, and also true for things like ZNE. However, when you're dealing with noiseless simulations with vanilla or PEC, you can accurately predict the distribution of results you expect based on statistical arguments.

Even if switch over to testing the `n_shots**-0.5` trend, we will still need to ask ourselves exactly the same questions - how close are we expecting our results and the trend to be? If I quadruple the shot count, is the precision going to increase by 2.1? 2.01? 2.2? It's the same issue. I believe there is no reason not to use statistics here. It could give us a lot more confidence in our tests.

TsafrirA: > In addition to this, you could consider parametrizing these tests over resilience levels 0, 1, and 2. So that we end up with 2 tests (times 3 -> 6 tests), and we only need to write two more for PEA and PEC

As I stated in the issue, I don't think we should test ZNE (resilience level 2) in the same way. It doesn't have a well defined expected distribution like vanilla and PEC do. I am not sure about TREX. With a noise less simulation what are we expecting to observe? (This is an honest question, I can't think of a clear effect to demonstrate but I am not sure there really aren't any).

TsafrirA: I had a discussion with @rainerschoe which helped me better articulate some of my views.

Given an expected quantum state and an observable, you can calculate the expected uncertainty rigorously. This calculation requires **no knowledge of qiskit ibm runtime** - it is fully dictated by the quantum state, the observable, and the number of shots.

The point of the test is then to make sure that the results we get follow the distribution we expect to get. Again, this doesn't have to do with what we promise in our contract. My test is problematic in that it mixed the two sides of the problem - it directly tied the target precision with the expected uncertainty. This is only partially true, because the target precision only controls the number of shots. I think the test should be changed to decouple these two different aspects.

TsafrirA: (

>  I am not sure about TREX. With a noise less simulation what are we expecting to observe? (This is an honest question, I can't think of a clear effect to demonstrate but I am not sure there really aren't any).

Here is one thing that could have been caught- #3225.

)

rainerschoe: well we caught this in the noisy tests now.

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -109,3 +115,88 @@ def test_result_quality_for_different_resilience_levels(self):
...
+
+    def setUp(self):
+        """Test level setup."""
+        super().setUp()
+        self.backend = GenericBackendV2(5, noise_info=False, seed=972)
+        # self.backend = AerSimulator()
+        self.preset_pass_manager = generate_preset_pass_manager(
+            optimization_level=1, target=self.backend.target
+        )
+        self.tolerance = 2  # In terms of stansdard deviations
```
yaelbh, line 136:
```suggestion
        self.tolerance = 2  # In terms of standard deviations
```

TsafrirA:
Closing in favor of #3250.

PR #3215, opened 2026-08-12
Add Wrapper Estimator noise-less simulation correctness-tests

rainerschoe:
### Summary
Part of #3209 
This PR became a bit bigger than anticipated:
- Added noise-less simulation tests for
   - all resilience levels
   - PEC
- Re-structured the test to re-use as much code from noisy sim tests as possible
- Took the example circuit and observables based on the idea of #3214
- Addressed discussion around seed and flakyness:
   - In noisy simulations removed absolute error assertion. Let's look at this in no noise correctness asserts, implemented later.
   - Do not use fixed simulator seed

### Details and comments

Fixes #

### AI/LLM disclosure


test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -104,8 +141,76 @@ def test_result_quality_for_different_resilience_levels(self):
...
         np.testing.assert_array_less(errors[1], errors[0], err_msg=debug_message)
 
         # Resilience level 2 should give very accurate expectation value:
+        # With fixed simulator seed, we can have 0.025 here.
+        # Without we should set to something like 0.04 to reduce flakiness.
         np.testing.assert_array_less(errors[2], 0.025, err_msg=debug_message)
+
+
+@ddt
+class TestEstimatorWithoutNoise(IBMTestCase):
```
SamFerracin, line 81: This PR is indeed pretty similar in spirit to @TsafrirA's #3214, at least for the part where it tests that the evs are close to the expected values. One difference is that #3214 has a larger suite of observables to test

I support the idea of creating one PR out of these two sibling ones

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -104,8 +141,76 @@ def test_result_quality_for_different_resilience_levels(self):
...
+        )
+
+    def test_correct_estimates_without_mitigation(self):
+        """Tests that vanilla EstimatorV2 produces correct results in a noise-less environment.
+
+        Compares the results against a statevector simulation.
+        """
+        pub, statevector_evs = create_estimator_test_data(self.backend, self.preset_pass_manager)
+
+        estimator = create_local_mode_estimator(self.backend)
```
SamFerracin, line 109: I tried a few instances of this where I removed the seed, and I set 
```suggestion
        estimator = create_local_mode_estimator(self.backend)
        estimator.options.default_precision = 0.01
        estimator.options.resilience_level = x
```
for `x` in `[0, 1, 2]`. They all passed. Perhaps that's the way to A parametrize over resilience levels (as I suggest [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3214/changes#r3768836310)) and B remove the need for a seed?

rainerschoe:
@TsafrirA I changed this to use something similar to your example circuit in #3214  and set of observables + got rid of StateVector simulation (It does not support projectors).

WDYT, could we use this PR as a base and treat your work on top of this to foxus on getting more precise assertions on the EV distributions?

Edit: I needed to remove ZXY, as it was always Zero and not affected by noise, causing assertions around mitigations to fail.

test/unit/executor_estimator/simulations/test_estimator.py
rainerschoe: cannot reproduce. This fails for me regularly (every ~3-4 runs) after removing the seed. I now set the threshold to `0.04` instead of `0.025`.

rainerschoe: As discussed. removing this assertion, instead have it in noise-less correctness tests only

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -51,61 +49,113 @@ def test_result_quality_for_different_resilience_levels(self):
...
-        pub = (isa_circuit, observables, parameters)
-
-        # Calculate ground truth to compare the results against via a statevector simulation:
-        statevector_estimator = StatevectorEstimator()
-        statevector_result = statevector_estimator.run([pub]).result()
-        statevector_evs = statevector_result[0].data.evs
+        # FIXME: add_projector_observables=False is only needed,
+        # due to a bug in TREX post-processing:
+        # https://github.com/Qiskit/qiskit-ibm-runtime/issues/3225
+        pub, ground_truth_evs = create_estimator_test_data(
```
SamFerracin: NIT, non-blocking comment:

In the literature of papers that try to benchmark EM techniques, these are often called "expected" or "ideal" instead of "ground-truth"
```suggestion
        pub, expected_evs = create_estimator_test_data(
```

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -51,61 +49,113 @@ def test_result_quality_for_different_resilience_levels(self):
...
+            )
+            for layer in layers
+        }
+
+        estimator.options.resilience.noise_model = noise_model
+
+        result = estimator.run([pub]).result()
+        # We get one expectation value per observable:
+        evs = result[0].data.evs
+        print(evs)
```
```suggestion
```

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -51,61 +49,113 @@ def test_result_quality_for_different_resilience_levels(self):
...
+@ddt
+class TestEstimatorWithoutNoise(IBMTestCase):
+    """Tests Executor based EstimatorV2 using noise-less simulator through local mode."""
+
+    def setUp(self):
+        """Test level setup."""
+        super().setUp()
+        self.backend = AerSimulator()
+        self.preset_pass_manager = generate_preset_pass_manager(
+            optimization_level=1, basis_gates=["cz", "rz", "sx", "x"]
```
SamFerracin, line 89: I'm pretty sure that one can set the basis gates in the AerSimulator itself, and it may be a nice conceptual cleanup (we have a backend that can only execute set gates). But we can do it in a followup

SamFerracin, line 89: I'm pretty sure that one can set the basis gates in the AerSimulator itself, and it may be a nice conceptual cleanup (we have a backend that can only execute set gates). But we can do it in a followup

test/unit/executor_estimator/simulations/utils.py
```diff
@@ -0,0 +1,94 @@
...
+def create_estimator_test_data(backend, preset_pass_manager, add_projector_observables=True):
+    """Create a pub and ground truth expectation values for it."""
+    # Use standard mirror circuit.
+    # - No measurements, as StatevectorEstimator does not support them.
+    # - No trailing Rx gates, as we want to add our own rotations
+    circuit = make_mirror_circuit_with_phases(
+        backend, num_qubits=3, add_measurement=False, add_rx=False
+    )
+
+    # Add rotations to become sensitive to X, Y, Z observables:
```
SamFerracin:
```suggestion
    # Add rotations to produce eigenstates of X, Y, and Z
```

test/unit/executor_estimator/simulations/test_estimator.py
rainerschoe: used `ideal` as I think "expected expectation values" is confusing

SamFerracin (approved):

PR #3221, opened 2026-08-12
`add_tags=True` in `prepare` when running local mode for executor-based primitives

### LINKED ISSUE #3217: Missing `add_tags=True` in executor-based primitives
**Describe the bug**
When `prepare` is called in the executor-based primitives we are missing an `add_tags=True` which means that simulated noise is not applied when running local mode simulations. 

**Steps to reproduce**

**Expected behavior**
The circuits in `run_quantum_program` must have tags when executed via the executor-based primitives

**Suggested solutions**
Update `prepare` in the executor-based primitives to include `add_tags=True` 

**Additional Information**

- **Python version**:
- **Operating system**:

apchytr:
### Summary
Because we rely on tags to specify noise for local mode simulations ,`prepare` in the executor-based primitives needs to include `add_tags=True`. This PR adds an `add_tags` argument to `prepare` for `SamplerV2` (`EstimatorV2` already has this) and sets `add_tags=True` when running local mode.

### Details and comments

Closes #3217 

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/prepare.py
+++ b/qiskit_ibm_runtime/executor_estimator/prepare.py
@@ -58,11 +58,10 @@ def prepare(
         shots: The number of shots to use. Will be overridden by
             ``num_randomizations * shots_per_randomization`` when both are specified explicitly
             and twirling is on.
-        add_tags: Whether to include tags for the boxes. Relevant mainly for debugging.
-            ``False`` will cause no tags to be added (will pass the "none" value to the relevant
-            attribute), while ``True`` will cause tags with the twirled boxes hash to be added
-            (using the "unique_box" value of the relevant attribute). These tags can help
-            injecting noise in simulators.
+        add_tags: Whether to include tags for the boxes. ``False`` will cause no tags to be added
+            (will pass the "none" value to the relevant attribute), while ``True`` will cause tags
+            with the twirled boxes hash to be added (using the "unique_box" value of the relevant
+            attribute). These tags are used to inject noise when running in local mode.
         backend: The backend for which the program is prepared. Only required when dynamical
             decoupling is enabled.
 
--- a/qiskit_ibm_runtime/executor_estimator/prepare.py
+++ b/qiskit_ibm_runtime/executor_estimator/prepare.py
@@ -59,9 +59,9 @@ def prepare(
             ``num_randomizations * shots_per_randomization`` when both are specified explicitly
             and twirling is on.
         add_tags: Whether to include tags for the boxes. ``False`` will cause no tags to be added
-            (will pass the `"none"` value to the relevant attribute), while ``True`` will cause tags
-            with the twirled boxes hash to be added (using the `"unique_box"` value of the relevant
-            attribute). These tags are used to inject noise when running in local mode.
+            (will pass the ``"none"`` value to the relevant attribute), while ``True`` will cause
+            tags with the twirled boxes hash to be added (using the ``"unique_box"`` value of the
+            relevant attribute). These tags are used to inject noise when running in local mode.
         backend: The backend for which the program is prepared. Only required when dynamical
             decoupling is enabled.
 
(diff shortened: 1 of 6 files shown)


qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -58,11 +58,10 @@ def prepare(
         shots: The number of shots to use. Will be overridden by
             ``num_randomizations * shots_per_randomization`` when both are specified explicitly
             and twirling is on.
-        add_tags: Whether to include tags for the boxes. Relevant mainly for debugging.
-            ``False`` will cause no tags to be added (will pass the "none" value to the relevant
-            attribute), while ``True`` will cause tags with the twirled boxes hash to be added
-            (using the "unique_box" value of the relevant attribute). These tags can help
-            injecting noise in simulators.
+        add_tags: Whether to include tags for the boxes. ``False`` will cause no tags to be added
+            (will pass the "none" value to the relevant attribute), while ``True`` will cause tags
```
```suggestion
            (will pass the `"none"` value to the relevant attribute), while ``True`` will cause tags
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -58,11 +58,10 @@ def prepare(
...
+            with the twirled boxes hash to be added (using the "unique_box" value of the relevant
```
```suggestion
            with the twirled boxes hash to be added (using the `"unique_box"` value of the relevant
```

qiskit_ibm_runtime/executor_sampler/prepare.py
```diff
@@ -64,6 +65,10 @@ def prepare(
         options: The options.
         default_shots: Default number of shots if not specified in PUBs. If ``None``,
             uses the value from ``self.options.default_shots``.
+        add_tags: Whether to include tags for the boxes. ``False`` will cause no tags to be added
+            (will pass the "none" value to the relevant attribute), while ``True`` will cause tags
```
SamFerracin: same here

qiskit_ibm_runtime/executor_sampler/sampler.py
```diff
@@ -217,7 +217,7 @@ def run(self, pubs: Iterable[SamplerPubLike], *, shots: int | None = None) -> Ru
         # Convert pubs to QuantumProgram and map options using the prepare method
         logger.info("Starting pre-processing")
         quantum_program, executor_options = prepare(
-            coerced_pubs, options, default_shots, backend=self._backend
+            coerced_pubs, options, default_shots, add_tags=local_mode, backend=self._backend
```
SamFerracin, line 220: Is there a reason for this VS this:
```suggestion
            coerced_pubs, options, default_shots, add_tags=True, backend=self._backend
```
?

Not saying we need to do it one way or the other, but I'd like to know if there are pros and cons

apchytr: I don't think there are any pros/cons one way or the other (at least as far as I can tell) but mostly just avoiding unnecessary population of tags here where we know we really only want to populate for local mode

SamFerracin (dismissed):

qiskit_ibm_runtime/executor_sampler/prepare.py
```diff
@@ -64,6 +65,10 @@ def prepare(
         options: The options.
         default_shots: Default number of shots if not specified in PUBs. If ``None``,
             uses the value from ``self.options.default_shots``.
+        add_tags: Whether to include tags for the boxes. ``False`` will cause no tags to be added
+            (will pass the `"none"` value to the relevant attribute), while ``True`` will cause tags
+            with the twirled boxes hash to be added (using the `"unique_box"` value of the relevant
```
SamFerracin: I suspect what you have will not be interpreted as code, you need the double thing
```suggestion
            (will pass the ``"none"`` value to the relevant attribute), while ``True`` will cause tags
            with the twirled boxes hash to be added (using the ``"unique_box"`` value of the relevant
```

SamFerracin (approved):

PR #3224, opened 2026-08-13
Introduce XSlowGate

yaelbh:
### Summary

Adding an `xslow` gate, which serves for detecting leakage, useful for post selection.

Usage:
```python
circuit.append(XSlowGate(), [qubit])
```

### Details and comments

The same gate will probably continue to live also in [Qiskit Addons](https://github.com/Qiskit/qiskit-addon-utils/blob/main/qiskit_addon_utils/noise_management/bit_flip_checks/xslow_gate.py), because qiskit-addon-utils does not depend on qiskit-ibm-runtime. There's a benefit from writing it again in qiskit-ibm-runtime, instead of only importing it, so that qiskit-ibm-runtime is independent and does not rely on decisions made about the gate made in Qiskit Addons.

### AI/LLM disclosure


yaelbh:
I'll add a release note.

SamFerracin (approved):
Thanks!

PR #3237, opened 2026-08-14
 When a PUB contains a circuit with boxes, the executor-estimator should fail gracefully

### LINKED ISSUE #3236: When a PUB contains a circuit with boxes, the executor-estimator should fail gracefully
Executor-based Sampler and Estimator do not currently support circuit with boxes. However, while Sampler raises a clean error, Estimator does not, and sometimes fails gracefully.

**Acceptance criteria**
- Estimator raises an error when a PUB contains a circuit with boxes

SamFerracin:
### Summary
This PR assures that a PUB with a box op causes error.

Fixes #3236

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/prepare.py
+++ b/qiskit_ibm_runtime/executor_estimator/prepare.py
@@ -20,6 +20,7 @@ from typing import TYPE_CHECKING
 from ..exceptions import IBMInputValueError
 from ..executor.dynamical_decoupling import apply_dynamical_decoupling
 from ..options_models.converters import estimator_options_to_executor_options
+from ..utils.utils import validate_no_boxes
 from .pec.prepare_pec import prepare_pec
 from .prepare_pea import prepare_pea
 from .prepare_vanilla import prepare_vanilla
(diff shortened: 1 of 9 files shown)


qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -72,19 +73,25 @@ def prepare(
             objects for each pub, with passthrough_data configured for post-processing.
         - :class:`~.ExecutorOptions` mapped from the estimator's options.
     """
-    if options.dynamical_decoupling.enable:
-        for pub in pubs:
+    if options.resilience.pec_mitigation and options.resilience.zne_mitigation:
```
SamFerracin, line 76: I moved this here, close to all the other validation we do

qiskit_ibm_runtime/executor_sampler/prepare.py
```diff
@@ -89,34 +89,27 @@ def prepare(
...
-            )
+            if backend is None:
+                raise IBMInputValueError(
+                    "A backend must be provided when dynamical decoupling is enabled."
+                )
 
-    # Create items based on whether twirling is enabled
-    items: list[QuantumProgramItem] = []
-    program_shots = shots  # Default: use pub shots
+    program_shots = (shots := extract_shots_from_pubs(pubs, default_shots))
```
apchytr, line 106: Is there a reason to have both `program_shots` and `shots`?

SamFerracin: There is a reason, namely that:
- If we are not doing twirling, that's the value of `quantum_program.shots`
- If we are doing twirling, we need to overwrite it (we still do the same number of total shots, but distributed between randomizations, so `quantum_program.shots` becomes something like `program_shots/num_randomizations`)

Not the cleanest though, feel free to clean it up in a follow up PR

apchytr (approved):

PR #3239, opened 2026-08-14
Fix bug occurring when TREX is used to estimate observables

### LINKED ISSUE #3225: TREX mitigation fails for projector operators in the wrapper estimator
**Describe the bug**

**Steps to reproduce**

**Expected behavior**

SamFerracin:
### Details and comments

Fixes #3225

### AI/LLM disclosure

### DIFF
--- a/test/unit/executor_estimator/simulations/utils.py
+++ b/test/unit/executor_estimator/simulations/utils.py
@@ -110,6 +110,7 @@ def create_estimator_test_data_extended(backend, preset_pass_manager):
             {"-IrI": 2.0, "1IYI": -1.0},
             2.0 * proj_q2 * r_q1 - 1.0 * proj_q2 * y_q1,
         ),  # ≈ 0.854
+        ({"I+r0": 1.0}, x_q2),  # ≈ 0.707
     ]
 
     observables = ObservablesArray([obs for obs, _ in observable_ideal_ev_pairs])
--- a/test/unit/executor_estimator/simulations/utils.py
+++ b/test/unit/executor_estimator/simulations/utils.py
@@ -16,7 +16,8 @@ from __future__ import annotations
 
 import numpy as np
 from qiskit.circuit import Parameter
-from qiskit.quantum_info import Operator, PauliLindbladMap, SparsePauliOp
+from qiskit.primitives.containers.estimator_pub import ObservablesArray
+from qiskit.quantum_info import PauliLindbladMap
 from samplomatic import InjectNoise
 from samplomatic.utils import get_annotation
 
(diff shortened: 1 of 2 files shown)


test/unit/executor_estimator/simulations/utils.py
```diff
@@ -104,33 +97,29 @@ def create_estimator_test_data_extended(backend, preset_pass_manager):
...
+        ({"IIY0": 1.0}, y_q1 * z0_q0),  # ≈ 0.783
+        ({"I1YI": 1.0}, proj_q2 * y_q1),  # ≈ 0.740
+        ({"IXII": 1.0}, x_q2),  # ≈ 0.707
         # Weighted linear combination:
         (
-            2.0 * _obs("-IrI") - 1.0 * _obs("1IYI"),
+            {"-IrI": 2.0, "1IYI": -1.0},
             2.0 * proj_q2 * r_q1 - 1.0 * proj_q2 * y_q1,
         ),  # ≈ 0.854
+        ({"I+r0": 1.0}, x_q2),  # ≈ 0.707
```
SamFerracin, line 113: This one is new

TsafrirA (changes requested):
I believe this fix is insufficient.
As far as I understand, this PR tries to add the usual TREX factor for each qubit with a projector operator. Essentially, estimating the observable $\left<0\right>$ as $P_{0}/\lambda$ instead of $P_{0}$ (where $P_{0}$ is the fraction of shots yielding zero).

If we break down the projector into paulis we get:
$\left<0\right> = \frac{I + \left<Z\right>}{2}$
Which has to be scaled as
$\left<0\right> = \frac{I + \frac{\left<Z\right>}{\lambda}}{2}$
If we rephrase this in terms of $P_{0}$ we get
$\left<0\right> = \frac{I + \frac{2P_{0}-1}{\lambda}}{2}$
or
$\left<0\right> = \frac{\lambda - 1 + 2P_{0}}{2\lambda}$
Note that this is generally not the same as $P_{0}/\lambda$.

The issue is, so far we've been treating the TREX factors as a multiplicative scale - and we can no longer do this here. To correctly apply TREX for projection operators, we would need to switch to scaling each operator on its own, and multiply after the scales have been applied.

dekelmeirom:
(Edited the comment to hopefully make it clearer) -
I think (and emphasis Tsafrir comment) that the calculation of the mitigated expectation value can't be in the same way that we calculate the projector expectation value without mitigation, with just multiplying by some factor (as we do in the Pauli case).

The current code calculates the expectation value of an observable that contains projectors by removing from the calculation any shot that will contribute zero to the expectation value (due to the 0 eigenvalue of the projector part). In a noisy case this approach is not accurate anymore as the TREX mitigation calculates the expectation value based on statistics of an ensemble of shots. The mitigation do not fix each individual binary string, therefore "throwing shots from the calculation" based on each binary string separately makes the calculation inaccurate.

In order to calculate the mitigated expectation value accurately, I think we must expand the operator in the Pauli basis, and calculate the expanded observable. Each Pauli term in the expanded observable can be mitigated, and because of linearity, the whole expectation value will be mitigated.
Important note is that such expansion might be exponential in the number of projector elements in the observable. This will result in an exponential classical computation.
In the future, we might think of approximations that can keep the classical calculation polynomial and still have a good approximation to the mitigated result. In the meanwhile, we can warn the user if his observable has too many (we can think what should be the threshold for "too many") projector elements while using TREX mitigation.

Note - also in the original TREX paper, in order to do a mitigation to a non Pauli operator, it is written that one should expand the operator to a Pauli basis and mitigate in the Pauli basis.

SamFerracin:
Superseded by other PRs

PR #3243, opened 2026-08-17
Decompose observables into Paulis when TREX is on

### LINKED ISSUE #3225: TREX mitigation fails for projector operators in the wrapper estimator

TsafrirA:
### Summary
Fixes #3225 

As far as I understand, decomposing the projection operators into paulis is the only way to get proper TREX. The cost is a (potentially) exponential post-processing compute time. Warning is issued when the number of decomposed terms more than doubles the original count. However, it is possible we want to add further guardrails - like an opt-in flag for the user and\or a prepare-side warning.

### AI/LLM disclosure


TsafrirA:
Closing in favor of #3253.

Issue #3208, opened 2026-08-12
Add simulation-based integration tests for the wrapper estimator


TsafrirA:
**What is the expected enhancement?**
Simulators provide a good tool to verify the correctness of the estimator results. We want to use them for integration-style tests (though they currently reside in the unit tests) that come between the test-device integration tests (go e2e but don't verify correctness) and the full-blown e2e testing against real HW (outide the scope of this repo).

We want to cover the noise-less case, and also verify the efficacy of error mitigation schems.

**Acceptance criteria**
Tests are added and cover the desired scenarios. Decision is made regarding the status of the tests (unit vs integration and run frequency)

rainerschoe:
We have basic estimator noise mitigation effectiveness tests and also noise-less correctness tests for some combinations of mitigation techniques
- PEC
- ZNE/PEA
- ZNE/GF
- Twirling
- TREX
(no DD yet)

Depending on if this satisfies the acceptance criteria, we can close this.

diego-plan9:
Marking as completed

PR #3246, opened 2026-08-17
Estimator sim-test cleanup: Use constants for options

rainerschoe:
### Summary
part of #3208

@SamFerracin This adds the indirection for the estimator options. I cleaned up a bit around the estimator factory to make it more natural.
We use different shot counts in noise and noise-less tests. Do we want to keep it this way? If we want to sync on this, we could simplify the arguments a bit.

Also there is still a TODO about DD. Not sure if this is worth an extra test, as it is hard to plug it into one of the others.

### Details and comments

Fixes #

### AI/LLM disclosure

### DIFF
--- a/test/unit/executor_estimator/simulations/test_estimator.py
+++ b/test/unit/executor_estimator/simulations/test_estimator.py
@@ -39,6 +39,44 @@ if TYPE_CHECKING:
     import numpy.typing as npt
 
 
+RESILIENCE_LEVEL_0 = {"resilience_level": 0}
+
+RESILIENCE_LEVEL_1 = {"resilience_level": 1}
+
+RESILIENCE_LEVEL_2_LINEAR = {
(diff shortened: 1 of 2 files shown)


SamFerracin (approved):
Thanks, Rainer! The file looks very clean and readable now, I really liked how this turned out to be!

> We use different shot counts in noise and noise-less tests. Do we want to keep it this way?

I would leave it as is. To elaborate on this:
- Resilience level tests use the same number of shots (200) and randomizations (100) across the files, with or without noise.
- Noisy PEA and PEC use 10x randomizations than noiseless PEA and PEC. I think that's fine, especially for PEC (remember that PEC needs a number of shots that grows exponentially with the level of noise in the circuit?); for PEA, I could argue that this is well justified as well (we are fitting several points, we want them to be stable)

PR #3250, opened 2026-08-18
[WIP] Modify noise-less wrapper estimator simualtions to be statistically rigour

### LINKED ISSUE #3248: Wrapper estimator underestimates reported uncertainties with multi-term observables
**Describe the bug**
In the post processor we sum the contributions to the variance term by term
```python
ensemble_variance += (coeff**2) * term_ensemble_variance * (term_scale_factor**2)
twirl_variance   += (coeff**2) * term_twirl_variance    * (term_scale_factor**2)
```
However, this is only justified when each term is an independent variable. This is not the case with commuting terms - these are based on the same shots. Therefore, the covariance contributions are not accounted for. To correctly estimate the uncertainty we need to group terms by their `config_idx` and compute the variance for the group together.

The bug applies to multi-term observables with commuting terms - for example `Z + ZZ`. Currently the simulation tests decompose projectors into paulis, automatically creating such commuting terms (related to #3225 and #3243).

TsafrirA:
### Summary
Related to #3209 
Fixes #3248 

### Details and comments
This PR is WIP meant to spark discussion. It adds simulation tests with clear statistical foundations. It also adds tests to the uncertainty estimates, which uncovered the bug in #3248 (which is fixed here as well).

Pending discussion, the new tests can replace the old ones for vanilla and PEC.

### AI/LLM disclosure


test/unit/executor_estimator/simulations/utils.py
```diff
@@ -80,6 +85,105 @@ def create_estimator_test_data(backend, preset_pass_manager):
     return pub, [ev for _, ev in observable_ideal_ev_pairs]
 
 
+def create_estimator_test_data_statistical(backend, preset_pass_manager):
```
TsafrirA: Just temporary. Could be combined into one with the old tests.

test/unit/executor_estimator/simulations/utils.py
```diff
@@ -89,7 +193,7 @@ def create_estimator_test_data_extended(backend, preset_pass_manager):
     especially in noisy simulations.
     """
     circuit = make_mirror_circuit_with_phases(
-        backend, num_qubits=4, add_measurement=False, add_rx=False
+        backend, num_qubits=4, layers=1, add_measurement=False, add_rx=False
```
TsafrirA, line 173: Unrelated to the PR. Just to get the old noiseless tests to run a bit faster.

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
```diff
@@ -127,36 +126,70 @@ def compute_exp_val(
...
-        )  # Shape: (num_randomizations,)
+        twirl_exp_vals = np.sum(evals, axis=-1) / shots_per_randomization
         twirl_variance = np.var(twirl_exp_vals, ddof=0)
 
-    # Ensure we always return numpy arrays (even for scalar results)
-    return np.asarray(exp_val), np.asarray(ensemble_variance), np.asarray(twirl_variance)
+    return np.asarray(ensemble_variance), np.asarray(twirl_variance)
+
+
+def compute_exp_val(
```
dekelmeirom, line 165: Does the `compute_exp_val` function is still useful (at least internally)?
Are there cases that we should call it directly and not call separately to `compute_evals` and `variances_from_evals`?

dekelmeirom (commented):
Good catch!

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
TsafrirA: That's a good question. It's currently used by the ZNE paths. But I need to think about it to see if the same fix should be applied there as well.

TsafrirA:
This now requires a fix in qiskit-mitigation, so it is blocked by [21](https://github.com/Qiskit/qiskit-mitigation/issues/21).

PR #3252, opened 2026-08-18
Standardize metadata of executor-based primitives

### LINKED ISSUE #3251: Standardize metadata of executor-based primitives
Currently, the executor-based primitives return metadata in a format that is slightly inconsistent. We want the two primitives to match more closely on this, to be able to document them more easily.

**Acceptance criteria**
- [ ] The results of the executor-based sampler and estimator should return metadata in the format:
```python
# here, 'result' is an executor result
metadata = {"executor": result.metadata, "options": user_options, "shots": shots}
```
- [ ] The metadata of Estimator should also contain info about the target precision
- [ ] The sampler should not compute the execution spans in the legacy format, it should simply report those that are already in `result.metadata`

SamFerracin:
### Summary
Currently, the executor-based primitives return metadata in a format that is slightly inconsistent. We want the two primitives to match more closely on this, to be able to document them more easily.

Fixes #3251

### AI/LLM disclosure


qiskit_ibm_runtime/decoders/executor_sampler/converters.py
```diff
@@ -72,12 +72,9 @@ def quantum_program_item_result_to_sampler_pub_result(
     data_bin = DataBin(**arrays, shape=pub_shape)
 
     # Construct the metadata for the result.
-    pub_metadata: dict[str, Any] = {"circuit_metadata": {}}
-    if circuit_metadata:
-        pub_metadata["circuit_metadata"] = circuit_metadata
-    if num_randomizations > 0:
-        pub_metadata["num_randomizations"] = num_randomizations
```
TsafrirA, line 79: I would consider leaving this in. The value in the options would be the finalized value, but not the resolved one (e.g. it could be `auto`).

TsafrirA (dismissed):
Thank you!

I left one comment for your consideration about `num_randomizations`.

qiskit_ibm_runtime/decoders/executor_sampler/converters.py
SamFerracin: Thanks: [1096710](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3252/commits/1096710d469a6e2a71d505cf60e1323a3f986eb7)

TsafrirA (approved):
Thank you!

PR #3253, opened 2026-08-19
Error when TREX is requested with projection observables

### LINKED ISSUE #3225: TREX mitigation fails for projector operators in the wrapper estimator

TsafrirA:
### Summary
Fixes #3225 

Following discussions it was decided to avoid the exponential cost of the solution in #3243 and for now error when TREX is requested with projection observables.

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/utils.py
+++ b/qiskit_ibm_runtime/executor_estimator/utils.py
@@ -105,6 +105,16 @@ def get_pauli_basis(basis: str) -> Pauli:
     return Pauli(basis)
 
 
+def has_projection_operators(pub: EstimatorPub) -> bool:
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -105,6 +105,16 @@ def get_pauli_basis(basis: str) -> Pauli:
     return Pauli(basis)
 
 
+def has_projection_operators(pub: EstimatorPub) -> bool:
```
SamFerracin, line 108: I wish there was a built-in function, but I couldn't find one either

SamFerracin (approved):
Thanks!

Issue #3211, opened 2026-08-12
Benchmark RTT to spot regressions


TsafrirA:
**What is the expected enhancement?**
We need a benchmarking and monitoring strategy to spot regressions in RTT.

**Acceptance criteria**
The team can monitor RTT, and get notifications on regressions.

diego-plan9:
Marking as completed

PR #3254, opened 2026-08-19
A first WrapperEstimator benchmark

rainerschoe:
### Summary
part of #3211 

Adds a first Benchmark for Wrapper Estimator. Note that this is now resilience level 0 only. Expect more to come.
I see this mostly as a first step, setting up the patterns.

### Details and comments

I added pytest configuration to `pyproject.toml` to disable benchmarks in unit-test mode for CI runs.
However I see that we run the benchmarks as part of regular integration testing without looking at the results. This is something we might want to disable, once we record the benchmarks accordingly.

Fixes #

### AI/LLM disclosure


TsafrirA (commented):
Nice starting point!

I think we are running here into the same issues we are facing with unit testing of the post processing stage. It's hard to mock the data. If you're dealing with a small circuit you can simulate, but at the scale you're looking at, even that can be difficult. I think we should look into some alternatives - 

- A large clifford circuit which could be simulated even at that scale?
- Saving real\test results and reusing them, instead of creating them on the fly? (how do we update? periodically? Upon changes to the prepare code? how do we pick those automatically?)

Mocking could end up very annoying later.

test/benchmarks/test_benchmark_executor_estimator.py
```diff
@@ -0,0 +1,174 @@
...
+    post_processor_data = passthrough["post_processor"]
+    param_basis_pairs_lists = post_processor_data["param_basis_pairs"]
+
+    for i, item in enumerate(quantum_program.items):
+        num_configs = len(param_basis_pairs_lists[i])
+        num_randomizations = 1
+        num_shots = quantum_program.shots
+        num_bits = item.circuit.num_qubits
+
+        meas_shape = (num_randomizations, num_configs, num_shots, num_bits)
```
SamFerracin:
```suggestion
        meas_shape = program.items[0].shape + (num_shots, num_bits)
```

test/benchmarks/test_benchmark_executor_estimator.py
```diff
@@ -0,0 +1,153 @@
...
+    backend = FakeBrisbane()
+
+    if benchmark.disabled:
+        num_qubits = 5
+        num_layers = 10
+        num_shots = 100
+    else:
+        num_qubits = 100
+        num_layers = 20
+        num_shots = 200000
```
SamFerracin, line 78: When benchmarking the post-processor, you don't gain anything from having a deep circuit. If it helps speeding up the test, I'd just do this:
```suggestion
    if benchmark.disabled:
        num_qubits = 5
        num_layers = 0
        num_shots = 100
    else:
        num_qubits = 100
        num_layers = 0
        num_shots = 200000
```

SamFerracin (commented):
Thanks! Looks good, but I'll take another look when you merge main and most of the diff goes away 😄

test/benchmarks/test_benchmark_executor_estimator.py
rainerschoe: I tried, both for benchmark-mode and non benchmark-mode and did not see a difference 

post-processing average in benchmarking mode: 330ms
Full test suite in benchmarking mode: 16s
Full test suite with benchmarking disabled: 1.6s

With this, I'd vote to keep the numbers the same for pre- and post-processing.
I will likely tune those numbers a bit in follow-up PR when we add mitigation.

pyproject.toml
```diff
@@ -147,6 +148,12 @@ split-on-trailing-comma = false
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
+[tool.pytest.ini_options]
+# By default do not run benchmarks in benchmarking mode, but as normal tests.
+# This will - depending on the benchmark - reduce test time by selecting a reduced problem size,
+# skip warmup and execute only a single round.
+addopts = "--benchmark-disable"
```
diego-plan9: How do you see setting this flag in `pyproject.toml` being needed at this stage - or, as an alternative question, what are you specifically aiming to achieve?

Without this change:
* the `benchmarks` stage of `Integration Tests` runs the benchmark:
    * runs the `test_benchmarks.py` files
    * runs the new `test_benchmark_executor.py` files, in proper benchmarking mode
    * prints the output in CI
* by default (`pytest test/benchmarks`), the benchmarks are run without the user requiring to append any flag

With this change:
* the `benchmark` stage of `Integration Tests` runs the tests but not the benchmarks. no output is printed related to benchmarking
* users need to opt-in via `pytest --benchmark-enable` to override the flag

So setting it in `pyproject.toml` seems a bit more intrusive / restricting (breaks the existing `test_benchmarks.py` benchmarking, forces users to opt-in to benchmark, adds another place where pytest options are being tracked compared to the `PYTEST_ADDOPTS` env var in CI). 

Maybe it depends on how mature you see the tests in this PR (ie. ready for being run, but not ready for being benchmarked)?

pyproject.toml
```diff
@@ -147,6 +148,12 @@ split-on-trailing-comma = false
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
+[tool.pytest.ini_options]
+# By default do not run benchmarks in benchmarking mode, but as normal tests.
```
diego-plan9: A note on the current duality (the test being available as "light regular tests" and them being available as "proper benchmarks").

This line is correct as long as `pytest` is the test runner. If using another test runner, the tests are not collectable:
```bash
$ python -m unittest discover --verbose --top-level-directory . --start-directory test/benchmarks

----------------------------------------------------------------------
Ran 0 tests in 0.000s
```

Which might be reasonable depending on scope and the next steps, but will have some implications if we want to keep the test suite fully stdlib-compatible. At this stage is is probably easily solved by placing them in `TestCases` (as long as `pytest-benchmark` copes with it, I haven't tested), but it might be more complex longer term.

Not a comment for acting on now, more as input on the general strategy. It might not pay off to have that duality and simply assume that benchmarks will always be benchmarks and run as such (and `pytest-benchmark` options, such as doing just one iteration, etc, might be enough for time-effective sanity checks on them).

test/benchmarks/test_benchmark_executor_estimator.py
```diff
@@ -0,0 +1,153 @@
...
+from qiskit_ibm_runtime.options_models.estimator import EstimatorOptions
+from qiskit_ibm_runtime.results.quantum_program import (
+    QuantumProgramItemResult,
+    QuantumProgramResult,
+)
+
+from ..utils import make_mirror_circuit_with_phases
+
+
+@pytest.mark.benchmark
```
diego-plan9: Nit, just double-checking that the markers are intentional (and used as an seemingly unused filtering mechanism rather than a need).

pyproject.toml
rainerschoe: Understand.

However I think as we have them now, the benchmarks do not provide additional value for correctness-testing, so I think it is perfectly fine to have them not discoverable via other test runners.
Basically I was mirroring [benchmarks.py](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/benchmarks/test_benchmarks.py), which is also not using test classes.

The only thing we need to guarantee is that the benchmarks themself do not break, this is why I still want them to be executed through CI (but in `--benchmark-disable` mode, as otherwise the benchmark runs minutes). Instead of the `pyproject.toml` we could also move this closer to GH Actions to make it more specific to our processes.

Separately we might want to discuss a bit about the value of being independent of test runners.

test/benchmarks/test_benchmark_executor_estimator.py
rainerschoe: I will remove, since we do not use the filters right now. I want to push more in the direction of using the `--benchmark-disable` flag.

pyproject.toml
rainerschoe: What is the use of integration test benchmarking stage?
Are we looking at the results?

diego-plan9: Continuing from a private conversation - the current `test_benchmarks.py` does not have too much systematic usage yet, but at least _allows_ for looking at the results. It would be good to preserve that if there is no strong technical hurdle against it.

rainerschoe: removed this, since those tests are unrelated to others (unit etc.) we can just always run benchmarks when benchmarks are requested to run :-)

pyproject.toml
rainerschoe: I Removed this, as we currently control which tests / benchmarks are run by selecting the root folder. So no need to control with the flag as well.
Developers can still use the flag.

SamFerracin (approved):
This looks good to me, content-wise. I see that @diego-plan9 left comments, so perhaps let's merge after he gives another approve

diego-plan9 (commented):
My suggested bits are tackled - there is only one bit derived from an off-line conversation. 

As this PR now does not change the default benchmarking behavior, the workflow it will run the new files as benchmarks. Can you verify that it is run in parallel with the "real" integration tests (that dominate time)? This way we just take advantage of their time for not having to worry (for now) about the amount of time the benchmarks take.

It can be tackled in a separate PR if the time is reasonable.

rainerschoe:
> Can you verify that it is run in parallel with the "real" integration tests (that dominate time)? This way we just take advantage of their time for not having to worry (for now) about the amount of time the benchmarks take.

I checked the workflow config: It is listed as a separate job next to integration tests without any dependencies. So I assume it is run in parallel.

diego-plan9 (approved):

Issue #3255, opened 2026-08-19
`LayerNoiseModel` class


apchytr:
**What is the expected feature?**
Introduce a `LayerNoiseModel` class that acts as a container for `pauli_lindblad_maps` and `layers` representing the Pauli Lindblad noise for each layer. This class should be used by `SimulatorOptions` and `ResilienceOptions` in their respective `noise_model` attributes. 

**Acceptance criteria**

- [x] Add a `LayerNoiseModel` class that accepts a list of `pauli_lindblad_maps` and `layers` 
- [x] Update option models to make use of this new class
- [x] Update the executor-based primitives to grab the relevant `noise_model` dictionary 
- [ ] [stretch] resilience fallback

PR #3259, opened 2026-08-19
Add `LayerNoiseModel` to `ExperimentalSimulatorOptions`

apchytr:
### Summary
Updating `ExperimentalSimulatorOptions` to replace `noise_model` with `layer_noise_model`, a sequence of `CircuitInstruction` and `PauliLindbladMap` tuples.

### Details and comments

Part of #3255

### AI/LLM disclosure

- [ ] I didn't use LLM tooling, or only used it privately.
- [ ] I used the following tool to help write this PR description:
- [x] I used the following tool to generate or modify code: Claude to fix test

PR #3262, opened 2026-08-21
Add `LayerNoiseModel` to `ResilenceOptions`

apchytr:
### Summary
Updating `ResilenceOptions` to replace `noise_model` with `layer_noise_model`, a sequence of `CircuitInstruction` and `PauliLindbladMap` tuples.

### Details and comments

Part of #3255

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -144,10 +144,10 @@ class EstimatorV2(BaseEstimatorV2):
             layers = est.find_unique_layers(pubs, types="gates")
 
             results = NoiseLearnerV3(mode).run(layers).result()
-            noise_model = results.to_dict(layers)
+            pauli_linblad_maps = results.to_pauli_lindblad_maps()
 
             # Assign the learned model so PEC uses it on the next run.
-            est.options.resilience.noise_model = noise_model
+            est.options.resilience.layer_noise_model = list(zip(layers, pauli_linblad_maps))
 
         Args:
             pubs: The list of PUBs to return a list of unique boxes for.
--- a/qiskit_ibm_runtime/executor_estimator/estimator.py
+++ b/qiskit_ibm_runtime/executor_estimator/estimator.py
@@ -147,7 +147,7 @@ class EstimatorV2(BaseEstimatorV2):
             pauli_linblad_maps = results.to_pauli_lindblad_maps()
 
             # Assign the learned model so PEC uses it on the next run.
-            est.options.resilience.layer_noise_model = list(zip(layers, pauli_linblad_maps))
+            est.options.resilience.layer_noise_model = zip(layers, pauli_linblad_maps)
 
         Args:
             pubs: The list of PUBs to return a list of unique boxes for.
--- a/qiskit_ibm_runtime/options_models/resilience.py
+++ b/qiskit_ibm_runtime/options_models/resilience.py
@@ -14,8 +14,6 @@
 
 from __future__ import annotations
 
-from collections.abc import Sequence
-
 from .base import BaseOptionsModel
 from .measure_noise_learning import MeasureNoiseLearningOptions
 from .pec import PecOptions
@@ -67,5 +65,5 @@ class ResilienceOptions(BaseOptionsModel):
     zne: ZneOptions = ZneOptions()
     """Additional zero noise extrapolation mitigation options."""
 
-    layer_noise_model: Sequence[LayerNoiseModel] | None = None
+    layer_noise_model: list[LayerNoiseModel] | None = None
     """Noise model specified by a collection of instructions and the noise that affects them."""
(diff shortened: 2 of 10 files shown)


qiskit_ibm_runtime/options_models/resilience.py
```diff
@@ -69,10 +70,21 @@ class ResilienceOptions(BaseOptionsModel):
...
+        cls, value: Sequence[LayerNoiseModel] | None
+    ) -> Sequence[LayerNoiseModel] | None:
+        if value:
+            for instruction, noise in value:
+                if not isinstance(instruction.operation, BoxOp):
+                    raise ValueError("Found an instruction that does not contain a box.")
+                if len(instruction.qubits) != noise.num_qubits:
+                    raise ValueError(
+                        f"Found instruction with {len(instruction.qubits)}",
+                        f"qubits but a noise model with {noise.num_qubits}.",
```
diego-plan9: Small nit - the `,` after the first line will make `ValueError` handle this as a tuple of two strings (rather than a single concatenated string), which I think is not the intent.

qiskit_ibm_runtime/options_models/resilience.py
```diff
@@ -69,10 +70,21 @@ class ResilienceOptions(BaseOptionsModel):
...
-    """A noise model mapping for PEC mitigation.
-
-    Maps layer references (strings) to :class:`~qiskit.quantum_info.PauliLindbladMap` objects that
-    describe the noise characteristics of that layer. The dict contains layers from all PUBs. This
-    is required when using PEC mitigation, or ZNE with PEA amplifier.
-    """
+    layer_noise_model: Sequence[LayerNoiseModel] | None = None
+    """Noise model specified by a collection of instructions and the noise that affects them."""
+
+    @field_validator("layer_noise_model", mode="after")
```
diego-plan9: Can you try further pursuing the use of pydantic `Annotated` pattern? It is already in use in #3259 with the `Annotated[CircuitInstruction, InstanceOf]` bits, and the encouraged direction in `options_models` - this way we stay consistent.

One of the benefits is reusability, which I think you arguably have the chance of being the first one to take advantage of 😁 If both `ExperimentalSimulatorOptions` and `ResilenceOptions` have the same need for validation of `LayerNoiseModel` (at a glance it seems currently a copy-paste), you might be able to something as:

```python
LayerNoiseModel: TypeAlias = Annotated[
    tuple[
        Annotated[CircuitInstruction, InstanceOf], Annotated[PauliLindbladMap, InstanceOf]
    ],
    AfterValidator(_validate_layer_noise_model),
]
```

so every model using a `LayerNoiseModel` field (now a "reusable field) does get that validation automatically. If this makes sense, it might prompt questions on whether those reusable fields deserve a separate module (we might not have enough data points to decide yet, so no opinion there).

qiskit_ibm_runtime/executor_estimator/estimator.py
```diff
@@ -144,10 +144,10 @@ def find_unique_layers(
             layers = est.find_unique_layers(pubs, types="gates")
 
             results = NoiseLearnerV3(mode).run(layers).result()
-            noise_model = results.to_dict(layers)
+            pauli_linblad_maps = results.to_pauli_lindblad_maps()
 
             # Assign the learned model so PEC uses it on the next run.
-            est.options.resilience.noise_model = noise_model
+            est.options.resilience.layer_noise_model = list(zip(layers, pauli_linblad_maps))
```
SamFerracin: Would it be possible to make the options accept `Iterable`s, and cast them to lists internally? It would save users an extra call to `list`
```suggestion
            est.options.resilience.layer_noise_model =zip(layers, pauli_linblad_maps)
```

qiskit_ibm_runtime/options_models/resilience.py
diego-plan9: Glad it worked!

qiskit_ibm_runtime/options_models/resilience.py
diego-plan9: This seems to still be present (in `qiskit_ibm_runtime/options_models/simulator.py` now).

apchytr: My bad should be addressed with this commit: [6caa37e](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3262/commits/6caa37e3c3ee964a5f88708e1ddbb86737a39b9b)

test/unit/executor/simulations/test_executor.py
```diff
@@ -152,8 +152,8 @@ def test_noisy_simulation(self):
         ]
         fidelities = []
         for pauli_map in pauli_maps:
-            executor.options.experimental["simulator_options"].layer_noise_model = list(
-                zip(unique_instructions, (pauli_map,) * len(unique_instructions))
+            executor.options.experimental["simulator_options"].layer_noise_model = zip(
```
SamFerracin, line 155: And this gets automatically converted to list? Nice!

SamFerracin (approved):

PR #3265, opened 2026-08-24
Change zne.noise_factors validator to annotated pattern

diego-plan9:
### Summary

Update the validator for `ZneOptions.noise_factors` in order to use the annotated pattern, for consistency with the rest of the module.

### Details and comments

Related to #2850

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude Opus 4.8

### DIFF
--- a/qiskit_ibm_runtime/options_models/zne.py
+++ b/qiskit_ibm_runtime/options_models/zne.py
@@ -39,7 +39,7 @@ DEFAULT_NOISE_FACTORS = (1, 3, 5)
 """The values of ``noise_factors`` used by default when ZNE is selected."""
 
 
-def _at_least_two_noise_factors(
+def at_least_two_noise_factors(
     value: Sequence[float],
 ) -> Sequence[float]:
     """Validate that `noise_factors` contains at least two factors.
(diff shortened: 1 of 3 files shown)


qiskit_ibm_runtime/options_models/zne.py
```diff
@@ -39,6 +39,19 @@
 """The values of ``noise_factors`` used by default when ZNE is selected."""
 
 
+def _at_least_two_noise_factors(
```
SamFerracin: Non-blocking comment: I'd remove the underscore
```suggestion
def at_least_two_noise_factors(
```

SamFerracin (dismissed):

qiskit_ibm_runtime/options_models/zne.py
diego-plan9: Sounds reasonable, we are mixing conventions a bit in general. So instead of applying just to that line, revised the module's validator helpers so they are consistent (without underscore) at 0ecf995

TsafrirA (approved):
LGTM

PR #3267, opened 2026-08-25
Pass projection observables when TREX is OFF

TsafrirA:
### Summary
We currently don't allow projection operators when TREX is on. The simulation tests circumvented that by decomposing projectors into Paulis (note that this doesn't add coverage, as it only adds a weighted pauli). This meant that projections were never passed into the simulation tests. This PR optionally adds the projection operators when TREX is off, and ignores them otherwise.

### AI/LLM disclosure

### DIFF
--- a/test/unit/executor_estimator/simulations/test_estimator.py
+++ b/test/unit/executor_estimator/simulations/test_estimator.py
@@ -91,7 +91,6 @@ class TestEstimatorWithNoise(IBMTestCase):
         backend = FakeManilaV2()
         preset_pass_manager = generate_preset_pass_manager(optimization_level=1, backend=backend)
 
-        # FIXME: Skipping projections because TREX is on in resilience level>0 (issue 3225)
         pub, ideal_evs = create_estimator_test_data(backend, preset_pass_manager, False)
 
         # maps resilience level to error (compared to statevector simulation) for each observable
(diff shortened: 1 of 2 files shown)


test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -201,17 +203,21 @@ def setUp(self):
         TWIRLING_TREX_PEC,
         TWIRLING_TREX_PEA,
     )
-    def test_correct_estimates_with_noise_injection(self, option_overrides):
+    def test_correct_estimates(self, option_overrides):
```
rainerschoe, line 204: I like this

test/unit/executor_estimator/simulations/test_estimator.py
```diff
@@ -128,7 +129,8 @@ def test_result_quality_with_noise_injection(self, option_overrides):
         backend = AerSimulator(basis_gates=["cz", "rz", "sx", "x"])
         preset_pass_manager = generate_preset_pass_manager(optimization_level=1, backend=backend)
 
-        pub, ideal_evs = create_estimator_test_data(backend, preset_pass_manager)
+        # FIXME: Skipping projection operators because TREX is on (issue 3225)
```
rainerschoe: Is this still a FIXME, or intended behavior?
If projectors are something we just do not support for now, I would remove the FIXME.

rainerschoe (dismissed):
Looks good, and I think it will also speed up the test!
Just small comment about the FIXME.

test/unit/executor_estimator/simulations/test_estimator.py
TsafrirA: Good question. We have some ideas about how to do it, but at the moment it's not supported. And we did close #3225.

I'll remove it.

rainerschoe (approved):

PR #3268, opened 2026-08-25
Exclude noise model from dump to passthrough data

TsafrirA:
### Summary
#3262 changed the resilience options field `noise_model` to `layer_noise_model`. The corresponding exclusion was not updated, causing serialization to fail. Verified end-to-end.

### AI/LLM disclosure


apchytr (approved):
Good catch! Thanks @TsafrirA

PR #3269, opened 2026-08-25
Add selected extrapolators to pub metadata when using ZNE

### LINKED ISSUE #3144: Decide the fate of `selected_extrapolators`

TsafrirA:
### Summary
Closes #3144 

The selected extrapolators are added to the pub metadata using the same convention as legacy estimator.

### AI/LLM disclosure

### DIFF
--- a/test/integration/test_executor_estimator.py
+++ b/test/integration/test_executor_estimator.py
@@ -126,6 +126,9 @@ class TestEstimator(IBMIntegrationTestCase):
               ``ensemble_stds_noise_factors``: ``(*pub_shape, num_noise_factors)``
             * ``evs_extrapolated``, ``stds_extrapolated``:
               ``(*pub_shape, num_extrapolators, num_extrapolated_noise_factors)``
+
+        - Correct shape and make-up for all ZNE-specific pub metadata fields:
+            * ``extrapolators``: pub shape, only requested extrapolators or `multiple`.
         """
         estimator = EstimatorV2(self.backend)
         estimator.options.resilience.zne_mitigation = True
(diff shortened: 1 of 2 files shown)


test/integration/test_executor_estimator.py
```diff
@@ -126,6 +126,9 @@ def test_zne_estimator(self, amplifier):
               ``ensemble_stds_noise_factors``: ``(*pub_shape, num_noise_factors)``
             * ``evs_extrapolated``, ``stds_extrapolated``:
               ``(*pub_shape, num_extrapolators, num_extrapolated_noise_factors)``
+
```
TsafrirA, line 129: Not sure if the integration test is the best place for this, but I think the simulation tests are even a worse fit, and the unit tests are problematic.

SamFerracin (approved):
This LGTM, thanks!

dekelmeirom (approved):
LGTM

PR #3270, opened 2026-08-25
Improve decoders/ modularization, explicitly avoiding importing from wrapper

diego-plan9:
### Summary

When we started moving (circa #2785) post-processing closer to the decoders, the utility functions for each wrapper were evaluated, deciding to move the ones only used by post-processing to `decoders/...`, keeping the one used for pre-processing in `executor-x/...`. This was motivated by modularity and prevention of import cycles, as both modules are heavy and have a tendency of spreading imports over the codebase.

This PR reinforces such modularity, as it slowly declined as work progressed, by moving:
* `unbroadcast_index()`
* `get_pauli_basis()`
* the zne extrapolation module (technically only `process_extrapolated_expectation_values` was used, but in practice seems to bring the rest of the module)

into `decoders/`. The first two functions are indirectly used by the wrappers, this change helps in breaking the `decoders` -> `wrappers` dependency (deemed as more problematic than `wrappers` -> `decoders`).

### Details and comments

Fixes N/A

### AI/LLM disclosure


qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -30,12 +30,11 @@
...
 from qiskit.quantum_info import Pauli
 
-from ...executor_estimator.utils import get_pauli_basis, unbroadcast_index
-from ...executor_estimator.zne.extrapolation import process_extrapolated_expectation_values
 from ...results.estimator_pub import EstimatorPubResult
 from ...results.quantum_program import QuantumProgramResult
 from .trex_utils import calculate_trex_factor, get_processed_calibration_data
-from .utils import compute_exp_val, identify_measure_basis
+from .utils import compute_exp_val, get_pauli_basis, identify_measure_basis, unbroadcast_index
+from .zne_extrapolation import process_extrapolated_expectation_values
```
diego-plan9, line 40: This is the net result - no imports from `executor_*` from the decoders module.

qiskit_ibm_runtime/executor_local_mode/sim_runtime_job.py
```diff
@@ -81,6 +82,4 @@ def result(self) -> QuantumProgramResult | PrimitiveResult:
         Returns:
             IBM Quantum Compute job result (post-processed if applicable).
         """
-        from ..decoders.quantum_program.decoder import QuantumProgramResultDecoder
-
```
diego-plan9, line 85: ... and this is a practical gain.

SamFerracin: neat!

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
```diff
@@ -160,3 +158,64 @@ def compute_exp_val(
...
+    shape_ndims = len(shape)
+    if shape_ndims == 0:
+        return ()
+
+    pad_shape = _pad_broadcast_shape(shape, len(bc_index))
+    bc_index = tuple(0 if dim == 1 else i for i, dim in zip(bc_index, pad_shape))
+    return bc_index[-shape_ndims:]
+
+
+def get_pauli_basis(basis: str) -> Pauli:
```
SamFerracin, line 198: I thought we were using this in `prepare_*` as well, which would have been an issue. But by looking into it again, it appears I was wrong

SamFerracin (commented):
Thanks, Diego! This looks much better!

SamFerracin (dismissed):

qiskit_ibm_runtime/decoders/executor_estimator/utils.py
diego-plan9: Maybe it was at some point. When I started tackling this, I was expecting that there would be some functions shared by post_process and prepare (if they were, they would require moving them somewhere else outside the import cycle).

TsafrirA: It is used by prepare, but not directly. Only via `compute_samplex_arguments`.

TsafrirA (commented):
`unbroadcast_index` and `get_pauli_basis` moved out of the estimator utils, but are now imported there. If they are used in both places, shouldn't we create a common place for them to reside in?

diego-plan9:
> `unbroadcast_index` and `get_pauli_basis` moved out of the estimator utils, but are now imported there. If they are used in both places, shouldn't we create a common place for them to reside in?

Thanks for catching this (somehow I overlooked the intermediate `compute_samplex_arguments`, which should be obvious looking at the new import in `qiskit_ibm_runtime/executor_estimator/utils.py`)!

Looking at options for such a common place, I don't see a good one:
* `executor/something.py` - originally deemed as the place for "utils share by the two families of prepares". Could be an option, but ends up more wrapper-specifics item into `executor/` (which conceptually is not so sound, and in practice reinforces import cycles on `executor` items)
* utils/something.py` - probably doable now that utils is relatively import-cycle-lean. Goes a bit against the direction of trying to use `utils/` as a last resort (as doing so increases the risk of it being a container for disparate functionality)
* a new module at top-level - it does not seem that helper functions merit such a high/visible location just based on import cycle optimization
* a new module inside an existing package - I cannot see any good candidate

So I'm leaning towards keeping the PR as is: it does increase `decoders/` modularization (ie. decoders no longer depends on wrappers, which was a source of cycles). It does increase the wrappers export surface a bit (it already has a large surface), but that direction does not seem to be a source of cycles at the moment and improving the modularization of the wrappers can be tackled separately.

SamFerracin (approved):
> So I'm leaning towards keeping the PR as is

I approve on the account that those two functions that are currently duplicated will move to qiskit-mitigation. So this duplication provides benefits, while being only temporary

PR #3271, opened 2026-08-25
Executor-based Primitives can execute circuits with RZZ gates

### LINKED ISSUE #3242: Executor_sampler/executor_estimator cannot execute non-Clifford rzz gates
**Describe the bug**
[`box_builder.py:129`](https://github.com/Qiskit/samplomatic/blob/487202826012c15385149fe80e83f3b9cbb97166/samplomatic/builders/box_builder.py#L129) in `samplomatic` raises `BuildError` for any non-Clifford or unbound fractional
entangler (`rzz`) unless the pass manager's twirling group is `local_pauli`. But both Executor
primitives choose the group internally and never pass `local_pauli`:

- `executor_sampler/prepare.py:142` calls `generate_boxing_pass_manager()` with no `twirling_group` arg.
- in `executor_estimator/utils.py` `options_to_boxing_pm_kwargs(..., twirling_group: str =
  "balanced_pauli", ...)`. but the [call sites](https://github.com/Qiskit/qiskit-ibm-runtime/blob/718fb5292eae222c327bf4627e38c31ae1b8f65c/qiskit_ibm_runtime/executor_estimator/utils.py#L323) and [`prepare_vanilla.py`](https://github.com/Qiskit/qiskit-ibm-runtime/blob/718fb5292eae222c327bf4627e38c31ae1b8f65c/qiskit_ibm_runtime/executor_estimator/prepare_vanilla.py#L102) do not override it.

**Steps to reproduce**

```python
from qiskit_ibm_runtime.fake_provider import FakeFractionalBackend
from qiskit_ibm_runtime.executor_sampler import SamplerV2

backend = FakeFractionalBackend()
theta = Parameter("theta")
qc = QuantumCircuit(2, 2)
qc.h(0)
qc.rzz(theta, 0, 1)            # non-Clifford, unbound fractional entangler
qc.h(0)
qc.measure([0, 1], [0, 1])
isa = generate_preset_pass_manager(optimization_level=1, backend=backend).run(qc)

sampler.options.twirling.enable_gates = True
sampler.options.experimental = {"local_mode": True}
sampler.run([(isa, {"theta": 0.9})], shots=256).result()
```

**Expected behavior**

```
...
...
...
samplomatic.exceptions.BuildError: Non-Clifford and unbound fractional entanglers are only
supported for Twirl with 'GroupMode LOCAL_PAULI'.
```
`EstimatorV2` on the equivalent observable pub raises the identical `BuildError` from the same line.

**Suggested solutions**

**1. Example minimal fix: look at the circuit first, then decide.** Before boxing, scan the pubs for
an `rzz` that's still a free parameter or bound to anything other than a quarter-turn (π/2), the one
case that actually needs the special 8-Pauli group. If found, switch to it for that call; if not,
leave everything exactly as it is.
(or)
We may add/retain the explicit `twirling_group` field to `TwirlingOptions` that a user sets themselves, however the auto-detect version is the less invasive of the two for the identical outcome: the field-based version needs a new pydantic field.

Example diff of the first, less invasive approach:

```diff
--- a/qiskit_ibm_runtime/executor_estimator/utils.py
+++ b/qiskit_ibm_runtime/executor_estimator/utils.py
@@
+from samplomatic.constants import SUPPORTED_2Q_FRACTIONAL_GATES
+
+def _needs_local_pauli(pubs: Iterable[EstimatorPub]) -> bool:
+    """Whether any pub contains a fractional gate that only "local_pauli" can twirl."""
+    return any(
+        instr.operation.name in SUPPORTED_2Q_FRACTIONAL_GATES
+        and (
+            instr.operation.is_parameterized()
+            or not np.allclose(np.abs(instr.operation.params), np.pi / 2)
+        )
+        for pub in pubs
+        for instr in pub.circuit.data
+    )
+
 def options_to_boxing_pm_kwargs(
     twirling_options, measure_noise_learning, inject_noise,
     twirling_group: str = "balanced_pauli", add_tags: bool = False,
+    pubs: Iterable[EstimatorPub] | None = None,
 ) -> dict[str, Any]:
+    if pubs is not None and _needs_local_pauli(pubs):
+        twirling_group = "local_pauli"
     return {...}
```
A longer effort may have a clearer division of burdens, or have a clear optional pathway for the same, and shall not involve the recounting of `rzz` gates on the `qiskit-ibm-runtime` end, rather, to recycle the template and pubs as created by the `passmanager` in `Samplomatic` as-is. This will leave only the execution pipeline here, without squashing the template into a new one via the default executor path.

SamFerracin:
### Summary
Right now, executor-based primitives raise an error when a circuit contains an RZZ gate with an unset `Parameter`. This PR fixes it.

Fixes #3242 

### AI/LLM disclosure

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -188,7 +188,7 @@ dependencies = [
     "pydantic>=2.13.0",
     "qiskit>=2.3.0",
     "pybase64>=1.4.0",
-    "samplomatic>=0.18.0"
+    "samplomatic>=0.19.0"
 ]
 
 [project.entry-points."qiskit.transpiler.translation"]
--- a/qiskit_ibm_runtime/executor_estimator/utils.py
+++ b/qiskit_ibm_runtime/executor_estimator/utils.py
@@ -281,7 +281,6 @@ def options_to_boxing_pm_kwargs(  # type: ignore[no-untyped-def]
     twirling_options: TwirlingOptions,
     measure_noise_learning: MeasureNoiseLearningOptions | None,
     inject_noise: bool,
-    twirling_group: str = "local_pauli",
     add_tags: bool = False,
 ) -> dict[str, Any]:
     """A helper to map options to kwargs for the boxing passmanager.
@@ -306,7 +305,7 @@ def options_to_boxing_pm_kwargs(  # type: ignore[no-untyped-def]
         if twirling_options.enable_measure or (measure_noise_learning is not None)
         else "change_basis",
         "twirling_strategy": twirling_options.strategy.replace("-", "_"),
-        "twirling_group": twirling_group,
+        "twirling_group": twirling_options.group,
         "inject_noise": inject_noise,
         "add_tags": "unique_box" if add_tags else "none",
     }
--- a/qiskit_ibm_runtime/executor_estimator/utils.py
+++ b/qiskit_ibm_runtime/executor_estimator/utils.py
@@ -281,7 +281,7 @@ def options_to_boxing_pm_kwargs(  # type: ignore[no-untyped-def]
     twirling_options: TwirlingOptions,
     measure_noise_learning: MeasureNoiseLearningOptions | None,
     inject_noise: bool,
-    twirling_group: str = "balanced_pauli",
+    twirling_group: str = "local_pauli",
     add_tags: bool = False,
 ) -> dict[str, Any]:
     """A helper to map options to kwargs for the boxing passmanager.
--- a/qiskit_ibm_runtime/executor_sampler/prepare.py
+++ b/qiskit_ibm_runtime/executor_sampler/prepare.py
@@ -213,6 +213,7 @@ def _build_quantum_program(
             twirling_strategy=finalized_options.twirling.strategy.replace("-", "_"),
             inject_noise_site="after",
             add_tags="unique_box" if add_tags else "none",
+            twirling_group=finalized_options.twirling.group,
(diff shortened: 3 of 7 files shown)


SamFerracin:
As I see it, we have two options to fix the issue.

## Option 1
The change in this PR, namely changing `twirling_group` from `"balanced_pauli"` to `"local_pauli"`.

Advantages:
- It fixes the issue

Disadvantages:
- Talking to Josh, it came up that "local_pauli"  twirls Clifford gates in the same way as "pauli" does, so it does it's job. However, it's less efficient than "pauli" or "balanced_pauli", because it needs to look at the content of the layer beforehand 
- It hardcodes the default to a twirling group that's probably not what we want to have--I imagine that when we expose this via the options, we'd want to have "pauli" as default, but then that'd create a breaking change (circuits with RZZs that were passing before do not pass anymore)

## Option 2
Expose `twirling_group` in the twirling options, set it to `"pauli"` by default, and letting users change it as they want.

Advantages:
- Fixes the issue, actually in a better way

Disadvantages:
- We need to make an exception to the rule that EstimatorV2 needs to support legacy features, and we'll support new features later.

---
Of the two, I prefer option 2. Any thoughts?

SamFerracin:
@TsafrirA I have implemented option 2

TsafrirA (approved):
Thanks @SamFerracin. I think this is the correct decision here.

PR #3273, opened 2026-08-26
Adjust `TestBoxCircuit` to correctly compare box annotations

TsafrirA:
### Summary
Qiskit PR [15490](https://github.com/Qiskit/qiskit/pull/15490) changed `QuantumCircuit` equality check to account for annotations. Against qiskit main the tests in `TestBoxCircuit` started failing after that fix, because the annotations were not identical. In some places it was just the counter being off (`basis_change0` vs `basis_change1` in subsequent pass manager runs) but in other places it  was mismatched boxing options.

This PR modifies the tests such that they pass even against qiskit main.

### AI/LLM disclosure


test/unit/executor_estimator/test_estimator_v2_utils.py
```diff
@@ -288,6 +291,16 @@ def test_run_level_zero_precision_raises(self):
...
 
+    def _reset_ref_counter(self):
+        """Reset the GroupMeasIntoBoxes class-level ref counter to zero.
+
+        GroupMeasIntoBoxes uses a class-level counter to generate unique ``ref`` strings for
+        ChangeBasis annotations. Resetting it before each pass-manager run ensures that
+        box_circuit and the test's own pm.run both start from the same counter value,
+        making the two circuits directly comparable with assertEqual.
+        """
+        GroupMeasIntoBoxes._REF_COUNTER = itertools.count()
```
TsafrirA: This is going a bit too deep into Samplomatic than I would have preferred. What happens if another counter is added next week? Or if this counter changes its name?

It would have been nice if Samplomatic provided a way to reset all counters and guarantee stable result from their pass manager, but currently no such option exists as far as I could find.

We could, alternatively, modify the test to compare the circuit "manually", ignoring the difference in ref counts, but I think it complicates the test too much.

SamFerracin: I agree with your feeling that this is a bit hacky, and I would rather not mess with internals of samplomatic, but rather be more explicit in what we test.

So, I would prefer it if we introduced two helper methods inside of this `TestBoxCircuit` that looks more or less like this:
```python
    def assert_circuits_equal_ignoring_annotations(self, circuit_1, circuit_2):
        """Assert two circuits are equal, ignoring any annotations on box operations."""
        from qiskit.circuit import BoxOp

        def strip_annotations(circuit):
            circuit_copy = circuit.copy()
            for instr in circuit_copy.data:
                if isinstance(instr.operation, BoxOp):
                    instr.operation.annotations = []
            return circuit_copy

        self.assertEqual(strip_annotations(circuit_1), strip_annotations(circuit_2))

    def assert_circuits_annotations_are_equal(self, circuit_1, circuit_2):
        """Assert annotations are equal.
        
        For twirl annotation, it compares this this and that. For change basis, this and that.
        Etc.
        """
        from qiskit.circuit import BoxOp
        from samplomatic import Twirl
        from samplomatic.utils import get_annotations

        self.assertEqual(len(circuit_1), len(circuit_2)) 

        for instr1, instr2 in zip(circuit_1, circuit_2):
            if isinstance(instr1.operation, BoxOp):
                self.assertIsInstance(instr2.operation, BoxOp)

                tw1 = get_annotations(instr1, Twirl)
                tw2 = get_annotations(instr2, Twirl)
                assert (tw1 is None) == (tw2 is None)
                if tw1 is not None:
                    # assert equality over the fields you care about

                # do the same for the other annotation types
```

And then we can go on and replace
```python
self.assertEqual(circuit_out, expected_circuit)
```
with
```python
self.assert_circuits_equal_ignoring_annotations(circuit_out, expected_circuit)
self.assert_circuits_annotations_are_equal(circuit_out, expected_circuit)
```

This would not complicate the tests, imo, and it would just make clearer the extent to which we test circuits and annotations.

TsafrirA: Done. Note that annotations cannot be mutated in python space.

SamFerracin (dismissed):

SamFerracin:
> Note that annotations cannot be mutated in python space.

The "safe" way to do it in Python space is to do it the way Samplomatic does it:
- Generate an empty dag
- Loop through the circuit, and append instructions to the dag with `annotations=[]`
- Convert the dag to a circuit

So that gives you a new circuit, without mutating the input one.

I have approved your PR, but feel free to switch to this if you think what we have is unsafe 😄

TsafrirA:
Thanks @SamFerracin 
The code in the PR is fine as far as I understand, because it modifies the annotations before creating the circuit. You cannot mutate the annotations after they've been added to the circuit. I was just pointing out that the snippet posted before
```python
        def strip_annotations(circuit):
            circuit_copy = circuit.copy()
            for instr in circuit_copy.data:
                if isinstance(instr.operation, BoxOp):
                    instr.operation.annotations = []
            return circuit_copy
```
will not work - the mutation happens on an existing circuit.

SamFerracin (approved):

Issue #2874, opened 2026-06-02
Refactor get_mode_service_backend()


diego-plan9:
**What is the expected enhancement?**

[`get_mode_service_backend()`](https://github.com/Qiskit/qiskit-ibm-runtime/blob/0.47.0/qiskit_ibm_runtime/base_primitive.py#L47) serves as the mechanism for a Primitive (or a program) to retrieve a number of elements based on the "mode" it receives at instantiation, typically via:
```python
self._mode, self._service, self._backend = get_mode_service_backend(mode)
```

It has a number of drawbacks in its current form:
* typing is coarse: consumers have no way of knowing which alternative is returned. This makes reasoning on `._mode`, `._service` and `._backend` hard, preventing to infer the right types.
* its location in `base_primitive.py` is problematic:
   * as it imports from several key items (`QiskitRuntimeService` among others), and `BasePrimitiveV2` is also central, contributes strongly to the import cycles problem
   * other programs are forced to bring `BasePrimitiveV2` in (and the rest of the imports in the file) even if they are not using `BasePrimitiveV2`
* programs or primitives still need to have specific code paths depending on `._mode`, `._service`, `._backend`, usually just checking via `isinstance()`. This forces a hard import dependency on importing from those items, and combined with the typing issues makes maintaining / reasoning hard.

This issue is about rethinking/refactoring how to model `mode`, and minimize the drawbacks. Some options to explore:
* convert the function into a `ModeManager` of sorts that encapsulates managing a mode
* consider moving `._mode`, `._service`, `._backend` to responsibilities of such manager
* move the function to prevent bundling it with `BasePrimitiveV2`
* using `TypeGuard` or other typing features for precise typing information

**Acceptance criteria**

At least one of the drawbacks above is no longer present.

PR #3282, opened 2026-08-27
Extend LocalRuntimeJob to handle executor jobs

### LINKED ISSUE #3281: Revise executor local mode harmony with non-executor local mode

Executor local mode is arguably very semantically similar to legacy primitives local mode: the ability to run a job locally when the primitive was created using a local backend. 

Currently, both implementations have some differences:
* non-executor relies on `QiskitRuntimeLocalService`
   * the primitive is not responsible for creating the jobs
   * instead, the primitive calls a `run` method on the service, that is responsible for job creation and execution
* executor creates the job directly without involving a service, which is submitted in constructor

This story is about aiming to have both implementations converge.

**Acceptance criteria**

The responsibilities in Executor local mode are closer to non-executor local mode, ideally with some reusing of components involved

diego-plan9:
### Summary

Bring the `Executor` local mode and legacy local mode closer by:
* Extend `LocalRuntimeJob` so it handles `Executor` jobs. It gains the "post-processing" of results, a bit of type hints to be able to cope with both kinds of jobs, and some mangling of the constructor for consistency.
* Add a new method to `QiskitRuntimeLocalService` in order to gain the responsibility of executing Executor local jobs. Ideally should get into `_run()`, but the argument differences seemed large enough not to convolute `_run` (the goal is to have the responsibility in the service).
* Adjust `Executor`, so job creation is delegated to the `QiskitRuntimeLocalService`.
* Some adjustments:
   * lower `executor_local_mode/` package to `fake_provider/executor`, as the functionality is now specific to the local service
   * remove `SimRuntimeJob`
   * reorganize the tests accordingly 

This makes `Executor` follow the usual pattern (when running something, delegate on a service, and the service creates the job), as well as allows for having a single entity that represents "jobs in local mode" (`LocalRuntimeJob`).

### Details and comments

Some details inline.

Fixes #3281

### AI/LLM disclosure


qiskit_ibm_runtime/fake_provider/local_runtime_job.py
```diff
@@ -20,33 +20,81 @@
...
+
+    - **Keys** — layer name tags (strings, e.g. ``"r0"``, ``"my_tag"``).  Each key must match
+      the ``ref`` of a ``Tag`` annotation used when building the ``QuantumProgram``.
+      A warning is emitted (if ``warn_absent=True``) when a tagged barrier's tag is absent
+      from the dict; the barrier is left as-is (no noise inserted for that layer).
+    - **Values** — :class:`~qiskit.quantum_info.PauliLindbladMap` instances describing
+      the Pauli-Lindblad noise channel for that gate.  The map's ``num_qubits`` must
+      equal the number of qubits on the corresponding barrier in the circuit.
+    - **Qubit indexing** — indices inside the map are *local* to the barrier's qubit
+      set, independent of global circuit qubit numbering.
```
diego-plan9: This was previously part of `SimRuntimeJob` - and is good content. It was not too visible before, and likely is less visible now. Perhaps it could be moved into some other class or module, not necessarily job-specific?

qiskit_ibm_runtime/fake_provider/local_runtime_job.py
```diff
@@ -20,33 +20,81 @@
...
     def __init__(  # type: ignore[no-untyped-def]
         self,
-        future,
+        function: Callable,
         backend: FakeBackendV2,
-        primitive: Literal["sampler", "estimator"],
-        inputs: dict,
+        primitive: Literal["sampler", "estimator", "executor"],
+        inputs: dict | QuantumProgram,
+        future: Future | None = None,
```
diego-plan9, line 66: The re-orgs here are mostly:
* taking advantage that `function` is already required by `PrimitiveJob`, and we always pass it (both in the old code and new code)
* moving `future` down as is set to `None` in `PrimitiveJob`, and we use that default for Executor jobs.
* consistency

qiskit_ibm_runtime/fake_provider/local_service.py
```diff
@@ -284,3 +287,31 @@ def _run_backend_primitive_v2(
...
+        Args:
+            backend: The backend to run the executor program on.
+            options: Simulator options to use.
+            inputs: The executor program to run.
+
+        Returns:
+            The job object that runs the program.
+        """
+        job = LocalRuntimeJob(
+            function=run_quantum_program,
```
diego-plan9, line 307: We lose some explicitness here, compared to having a dedicated `SimRuntimeJob` class: `LocalRuntimeJob` for the `executor` primitive need to pass a function that matches `run_quantum_program` signature. I tried to solve that lack of explicitness by documentation, and overall it felt that what we gain by being able to use `LocalRuntimeJob` and the regular machinery compensates for it.

test/unit/local_mode/test_local_runtime_job.py
```diff
@@ -0,0 +1,69 @@
...
+                    "simulator_options": ExperimentalSimulatorOptions(),
+                }
+            },
+        )
+        job = executor.run(QuantumProgram(1))
+        self.assertIsInstance(job, LocalRuntimeJob)
+        self.assertTrue(job.metrics())
+        self.assertTrue(job.backend())
+        self.assertTrue(job.inputs)
+        self.assertEqual(job.usage(), 0)
```
diego-plan9, line 65: This tests takes the spirit of the previous `test_sim_runtime_job.py` test, and these asserts help conveying some of the things we gain (same methods as regular `LocalRuntimeJob`, state, etc).

qiskit_ibm_runtime/fake_provider/local_runtime_job.py
apchytr: With the recent updates to `layer_noise_model` I actually think most of this is now not required to convey to the user as this is all handled under the hood. If anything I might just suggest keeping some of the additional information here and moving it to `InsertNoisePass` as that is specifically where this ``noise_model`` dictionary is used.

qiskit_ibm_runtime/executor/executor.py
```diff
@@ -118,7 +118,7 @@ def __setattr__(self, name: str, value: Any) -> None:
 
         super().__setattr__(name, value)
 
-    def run(self, program: QuantumProgram) -> RuntimeJobV2:
+    def run(self, program: QuantumProgram) -> RuntimeJobV2 | LocalRuntimeJob:
```
apchytr, line 121: Should the type hint for the wrapper estimator + sampler also change to reflect this?

apchytr (commented):
Overall I think this makes sense thanks @diego-plan9! I was also wondering if you have an idea of what can be simplified once we remove the legacy local mode code?

qiskit_ibm_runtime/executor/executor.py
diego-plan9: Yes, it makes a lot of sense, thanks for catching! 13b9493dfd4dddbb323f87dd99ee785ebfd1b089. Includes revising `QiskitRuntimeLocalService` methods for narrowing the return types.

qiskit_ibm_runtime/fake_provider/local_runtime_job.py
diego-plan9: Thanks - attempted at 6e030646dd37663fb5d8d314a337fa2698b4cf61. It has been a bulk move (ie. what I kept in this file is implementation specifics), and at the module docstring level.

diego-plan9:
> Overall I think this makes sense thanks @diego-plan9! I was also wondering if you have an idea of what can be simplified once we remove the legacy local mode code?

Thanks for the review!

It might depend which removal you are referring to:
* the one that allows wrapper to use local mode but not through sim-executor: it seems it might be just a matter of deleting the branching and `_run_legacy_simulation` functions.
* the one that allows non-wrapper primitives to be run in local mode: it would make `LocalRuntimeJob` have to deal only with executor jobs, and `QiskitRuntimeLocalService` to only understand executor jobs. This likely will lead to quite some simplification of their logic, but the pattern of "a primitive delegates on a service, and the service create the job" will likely stay (and arguably simplified by this PR).

Regardless, some simplification of the primitives `run` methods and general complexity can likely be achieved via #2874 (ie. if we model a "mode" differently, all the nuances the primitives have and fine-grain peeking into how that "mode" is implemented will result in cleaner primitives). But that is more of a wishlist item with low priority.

apchytr (approved):

Issue #3275, opened 2026-08-26
Backend data not re-fetched when falling back to default cal


jyu00:
**Describe the bug**

Reported by Maika:

When I do:
```py
from qiskit_ibm_runtime import QiskitRuntimeService
service = QiskitRuntimeService()
backend=service.backend('ibm_boston')
backend.supported_instructions
```
I get the correct list.

then I call a `calibration_id`:
```py
backend_extra=service.backend('ibm_boston',calibration_id = 'extra_gates')
backend_extra.supported_instructions
```
this has more gates in supported_instruction. This is also correct.

Now, I define "default" backend again:
```py
backend_new=service.backend('ibm_boston')
backend_new.supported_instructions
```
And this time, it prints supported instruction from previous calibration_id that was used.

**Steps to reproduce**

See above.

**Expected behavior**

`backend_new.supported_instructions` should return the default instructions.

**Suggested solutions**

The last `service.backend` should recognize the calibration_id has changed (i.e. switching from `extra_gates` to `None`) and re-fetch the backend data. 

**Additional Information**

- **qiskit-ibm-runtime version**:
- **Python version**:
- **Operating system**:

diego-plan9:
Closed via #3287

PR #3287, opened 2026-08-30
Avoid caching calibration specific backend configurations

niketh-keshavan:
### Summary

prevent calibration specific backend data from getting into default config by
- updating `QiskitRuntimeService._create_backend_obj()` so configurations retrieved via `calibration_id` remain scoped to request and are not stored in `_backend_confics` cache
- preserve existing caching behavior for default backend configurations and fractional-gate selection
- add coverage for:
    - default -> calibrated -> default retrieval
    - calibrated -> default retrieval with an initially empty cache
    - independence of previously created `IBMBackend` instances
    - repeated calibrated requests remaining fresh
    - default configuration caching remaining effective

add a bug-fix release note.

### Details and comments

calibration specific configs are intentionally not cached (follows `RuntimeClient.backend_configuration()`)

Fixes #

https://github.com/Qiskit/qiskit-ibm-runtime/issues/3275

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: (Qwen 3.8 27b running locally was used to review final diff)


test/unit/test_backend_retrieval.py
```diff
@@ -25,6 +31,36 @@
...
+        calibration_id = query_params.get("calibration_id", [None])[0]
+
+        config = deepcopy(backend.configuration)
+        if calibration_id == "abc1234":
+            supported = list(config.get("supported_instructions", []))
+            if "while_loop" not in supported:
+                supported.append("while_loop")
+            config["supported_instructions"] = supported
+
+        return (200, {"Content-Type": "application/json"}, json.dumps(config))
```
diego-plan9: Can you look into incorporating the ability for this endpoint to cope with a custom calibration into `BaseRegistry`? It is a useful feature that we can make use of in other tests.

A suggested approach:
* Add a `calibrations` attribute to `Backend`, defaulting to empty, and containing configuration "overrides"
* Tweak `BaseRegistry.callback_backends_configuration` so:
    * the query param is extracted, same as here
    * if `calibration_id` is passed, it looks into `backend.calibrations`. if present, merges the backend configuration with the calibration configuration, and returns it
    * likely returning a 404 (or mimicking the API behavior) if `calibration_id` is not valid

Ideally then we could just do:
```python
registry.add_backend(Backend("ibm_foo", calibrations=...))
```

and have the test easily describe what the calibration does, and the rest of the machinery take it into account.

test/unit/test_backend_retrieval.py
```diff
@@ -299,13 +335,63 @@ def test_backend_with_and_without_fractional_from_same_service(self, registry):
...
+        self.assertNotIn("ibm_torino", service._backend_configs)
+
+        default_backend = service.backend("ibm_torino")
+        self.assertIsNone(default_backend.calibration_id)
+        self.assertNotIn("while_loop", default_backend.supported_instructions)
+
+        self.assertIn("ibm_torino", service._backend_configs)
+        self.assertNotIn(
+            "while_loop", service._backend_configs["ibm_torino"].supported_instructions
+        )
```
diego-plan9: While I appreciate the effort in both tests, I wonder if the two tests can be merged into a single one, with this one as an inspiration, mostly exercising:
* retrieving a backend with a custom calibration and asserting that the resulting backend configuration comes from it (feel free to use an entry simpler than `supported_instructions` if it simplifies the logic), as well as it having the right `calibration_id`
* retrieving the same backend without a custom calibration, and asserting that the resulting backend configuration did not take into account the calibration, and does not have `calibration_id`

As it stands, while the tests do provide more explicit guarantees, they involve some layers that are better off handled by different tests (or that are very tied to the implementation of the service/test, and we can relax them):
* peeking into `request_mock.calls` - was needed prior to this PR as a way that the attribute reached the client layer. but not needed anymore, we should now focus on testing the feature
* peeking into `service._backend_configs` - useful for explicitly checking the caching, but is an internal attribute of the service. we can now just rely on the 2nd call not having its config modified
* `second_backend_with_calibration` - unsure if this is providing value, the first call should exercise the feature
* comparing backends against each other - worth for catching side effects, but we can rely on them being different objects (and on the main asserts catching issues in each of them)


Can you check if the tests can be simplified/merged along those lines?

release-notes/unreleased/3275.bug.rst
```diff
@@ -0,0 +1,4 @@
+Fixed :meth:`.QiskitRuntimeService.backend` so that retrieving a backend without a
+``calibration_id`` after retrieving it with a custom calibration returns the default backend
+configuration and supported instructions. Custom calibration data no longer replaces the cached
+default backend configuration.
```
diego-plan9:
```suggestion
configuration correctly. The calibration-specific backend configurations are now not cached.
```

Not super happy with my suggestion, but it would be good to:
* place the focus in configuration overall (considering supported instructions as a subset of it)
* be explicit about the expectations when using the method, as this is a change (ie. users might have been expecting that the cached backend was calibration-id aware)

diego-plan9 (commented):
Thanks Niketh!

This is looking good, and the right fix. I left some comments for aiming to generalize the testing approach (nice usage of the registries, which is notable as it is a relatively recent testing addition!).

Couple of smaller "meta" bits:
* the test did not run, probably confused by the closing/opening of the PR. They should run next time (might require some manual intervention from our side)
* if possible, when interested in working in an issue, please ping in the issue before starting the PR, for easier coordination. No harm at all this time!

test/unit/test_backend_retrieval.py
niketh-keshavan: > * if possible, when interested in working in an issue, please ping in the issue before starting the PR, for easier coordination. No harm at all this time!

Hey, sorry about that =D

> While I appreciate the effort in both tests, I wonder if the two tests can be merged into a single one, with this one as an inspiration, mostly exercising:

The two options that I can think of here are:
1. a single behavioral test that keeps invalid-ID coverage seperate
2. seperating service behavior from the registry behavior

LMK what you prefer, and i'll move forward

niketh-keshavan:
~~@diego-plan9 seems like manual intervention may be required for the tests, sorry~~
nvm

test/unit/test_backend_retrieval.py
diego-plan9: Good catch about covering the invalid-id! Let's go with:

* a single test covering what happens when the calibration id is invalid (ie. `backend()` would raise a `QiskitBackendNotFoundError`)
* a single test covering:
  * correct behavior of `backends()` when the calibration id is valid (backend configuration contains the calibration override)
  * correct behavior of `backends()` when invoked after the first call, this time not passing a calibration id (backend configuration does not contain the calibration override)

Combined with the suggestions in the initial comment, hopefully this leads to leaner, more targeted tests. In the end, it is about guaranteeing `backends()` behavior when passing the `calibration_id` argument.

diego-plan9 (approved):
Thanks @niketh-keshavan for this nice contribution - neatly handled, and directly helping the users by solving the bug!

PR #3290, opened 2026-09-01
Revise QuantumProgram-related imports and deprecations

diego-plan9:
### Summary

Follow-up on the first 2 next steps from https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206#issuecomment-5430872667:

* Update imports
   * for items moved to `samplomatic` (and not inherited): replace local imports with `samplomatic` imports
   * for items moved to `samplomatic` (and inherited): keep the local imports, as we explicitly declare that `qiskit_ibm_runtime` expects our customized objects
* Handle deprecations
   * for items in modules not imported by `__init__.py` files (ie. not reachable unless importing directly): emit a deprecation warning (leaf `datatree`, leaf `draw_chunk_timings`)
   * for other items - do not emit deprecation warnings. The approach in early versions in #3206 using `__getattr__` covers most the use cases but cannot guarantee to be 100% full of side effects, and other approaches (for example, providing a wrapper and updating `sys.modules` on the fly) do not convey such assurances either. I'm opting for ensuring non-breakage of the import mechanisms, at the expense of less visibility and relying on release notes.

### Details and comments

Follow up from #3206

### AI/LLM disclosure


TsafrirA (approved):
LGTM.

It's a little strange that we don't need to bump schemas. The `DataTree` and `DataTreeModel` are seemingly two independent sources of truth for basically the same thing.

PR #3200, opened 2026-08-11
Complete testing of SamplexItem shape in wrapper estimator

TsafrirA:
### Summary
Fixes #3194.
The shape elements we want to test are num_randomizations, shots per randomization (though not strictly a SampleItem property) and the number of param-basis pairs. The last one was already covered by #3087, so this PR completes the other two elements in a unified way for all `prepare_x` functions. 

### AI/LLM disclosure

- [X] I used the following tool to generate or modify code: IBM BOB

PR #3291, opened 2026-09-01
Wrapper estimator prepare unit tests cleanup

TsafrirA:
### Summary
There was an effort to unify the unit tests for the wrapper estimator's prepare stage (for example #3223, #3200). This PR removes some tests that are no longer needed (and moves one into a more appropriate location).

### AI/LLM disclosure


dekelmeirom (dismissed):
I personally think that one should limit the effort of making the tests short and unified, and mainly make sure they are robust and have good coverage.
Anyway, if the effort was already done - it should be merged :)

diego-plan9:
Thanks! Incurred into a conflict while merging #3290 - looks good to go once solved.

diego-plan9 (approved):

PR #3294, opened 2026-09-02
Revise QuantumProgram (and other bits) typing

### LINKED ISSUE #3279: Revise QuantumProgram vs samplomatic.QuantumProgram typing

From discussions during #3206:
* https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206#discussion_r3862089244
* https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206#discussion_r3862259517

In that PR, we customize some of the base classes (now in `samplomatic`). Some of the container-like ones (such as `QuantumProgramResult`) use other base classes in the function signatures (such as `QuantumProgramItemResult`), which we also customize. As our functions expect our customized ones, in the original rounds of the PR some typing mismatches arose. This seems fixed in the latest commits of the PR, but since we don't have a very strict `mypy` configuration, it would be worth taking a look once the feature is nearly completed (3206 merged, imports updated).

**Acceptance criteria**

The codebase correctly declares and understand the customized classes vs the base classes from a typing perspective.

diego-plan9:
### Summary

Revise some of the typing hints related to the `QuantumProgram` move in #3206 (only one notable item, see comments), and take the chance to revise other Pyright warnings (not emitted by `mypy`).

### Details and comments

Fixes #3279

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude Opus 4.8 (for the registries)

### DIFF
--- a/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
+++ b/qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
@@ -76,7 +76,7 @@ def _build_program_result_metadata(post_processor_data: dict) -> dict:
     if options is None:
         return {}
 
-    metadata = {"options": dict(options)}
+    metadata: dict[str, Any] = {"options": dict(options)}
     if "resilience" in metadata["options"]:
         resilience = dict(metadata["options"]["resilience"])
         for flag_key, options_key in [
@@ -111,12 +111,13 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
     if len(result) == 0:
         return PrimitiveResult([])
 
-    if not isinstance(passthrough := result.passthrough_data, dict):
+    if not isinstance(result.passthrough_data, dict):
         raise ValueError(
             "Wrong type for passthrough data: Expected a 'dict', found "
             f"'{type(result.passthrough_data)}'."
         )
 
+    passthrough: dict[str, Any] = result.passthrough_data or {}
     if (post_processor_data := passthrough.get("post_processor", None)) is None:
         raise ValueError("Missing 'post_processor' in passthrough data.")
 
--- a/qiskit_ibm_runtime/decoders/quantum_program/converters.py
+++ b/qiskit_ibm_runtime/decoders/quantum_program/converters.py
@@ -29,7 +29,22 @@ from ...results.quantum_program import (
 )
 
 if TYPE_CHECKING:
-    from ibm_quantum_schemas.executor.version_0_1 import QuantumProgramResultModel
+    from ibm_quantum_schemas.executor.version_0_1 import (
+        QuantumProgramResultModel as QuantumProgramResultModel_0_1,
+    )
+    from ibm_quantum_schemas.executor.version_0_2 import (
+        QuantumProgramResultModel as QuantumProgramResultModel_0_2,
+    )
+    from ibm_quantum_schemas.executor.version_1_0 import (
+        QuantumProgramResultModel as QuantumProgramResultModel_1_0,
+    )
+    from ibm_quantum_schemas.executor.version_1_1 import (
+        QuantumProgramResultModel as QuantumProgramResultModel_1_1,
+    )
+    from ibm_quantum_schemas.executor.version_2_0 import (
+        QuantumProgramResultModel as QuantumProgramResultModel_2_0,
+    )
+
 
 from ...quantum_program.converters.converters_0_2 import passthrough_data_from_0_2
 from ...quantum_program.converters.converters_1_0 import passthrough_data_from_1_0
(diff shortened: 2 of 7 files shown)


qiskit_ibm_runtime/results/quantum_program.py
```diff
@@ -204,15 +204,12 @@ def __getitem__(self, idx: int) -> QuantumProgramItemResult: ...
     def __getitem__(self, idx: slice) -> list[QuantumProgramItemResult]: ...
 
     def __getitem__(
-        self, index: int | slice
+        self, idx: int | slice
```
diego-plan9, line 207: This is the meaningful fix (aligning with the rest of overloads signatures) - the function changes are mostly for simplification.

qiskit_ibm_runtime/decoders/quantum_program/converters.py
```diff
@@ -29,15 +29,30 @@
...
+    )
+    from ibm_quantum_schemas.executor.version_1_0 import (
+        QuantumProgramResultModel as QuantumProgramResultModel_1_0,
+    )
+    from ibm_quantum_schemas.executor.version_1_1 import (
+        QuantumProgramResultModel as QuantumProgramResultModel_1_1,
+    )
+    from ibm_quantum_schemas.executor.version_2_0 import (
+        QuantumProgramResultModel as QuantumProgramResultModel_2_0,
+    )
```
diego-plan9, line 46: These have been a bit of an oversight as the file grew (all conversion functions were incorrectly declaring that they would handle `QuantumProgramResultModel_0_1`).

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -28,7 +28,8 @@
     import numpy.typing as npt
     from qiskit import QuantumCircuit
     from qiskit.circuit import BoxOp, CircuitInstruction
-    from qiskit.primitives import EstimatorPub, SamplerPub
+    from qiskit.primitives.containers.estimator_pub import EstimatorPub
+    from qiskit.primitives.containers.sampler_pub import SamplerPub
```
diego-plan9, line 32: The original import is only available in latest Qiskit versions.

qiskit_ibm_runtime/noise_learner_v3/params_converters.py
```diff
@@ -38,16 +39,20 @@
     from ..options_models import NoiseLearnerV3Options
 
 
-class ParamsConverter(NamedTuple):
+ModelT = TypeVar("ModelT", bound="BaseParamsModel")
+
+
+@dataclass(frozen=True)
+class ParamsConverter(Generic[ModelT]):
```
diego-plan9, line 46: Slightly overkill, but making this (and `ParamsConverter`) generic avoids some typing warnings when the registry dicts are constructed.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -111,12 +111,13 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
         return PrimitiveResult([])
 
-    if not isinstance(passthrough := result.passthrough_data, dict):
+    if not isinstance(result.passthrough_data, dict):
         raise ValueError(
             "Wrong type for passthrough data: Expected a 'dict', found "
             f"'{type(result.passthrough_data)}'."
         )
 
+    passthrough: dict[str, Any] = result.passthrough_data or {}
```
diego-plan9, line 120: These help pyright infer the right types, which were proven useful for catching (visually, not via `mypy` yet) type hinting issues in https://github.com/Qiskit/qiskit-ibm-runtime/pull/3206#discussion_r3862259517

qiskit_ibm_runtime/quantum_program/params_converters.py
```diff
@@ -44,16 +45,20 @@
     from .quantum_program import QuantumProgram
 
 
-class ParamsConverter(NamedTuple):
+ModelT = TypeVar("ModelT", bound="BaseParamsModel")
```
apchytr, line 48: This seems to be identical to the one in `noise_learner_v3/params_converter.py` is that right? Would it make sense to instead import to avoid redefining this type?

apchytr (approved):
One small comment but I think this looks good otherwise!

TsafrirA (approved):
Besides Anthony's comment - looks good.

qiskit_ibm_runtime/quantum_program/params_converters.py
diego-plan9: It serves the same purpose (ie. allowing the decoder/encoder types to be correctly inferred based on the model) indeed, in a different context. As their definitions are really simple (ie. a one-liner), I think we are better off living with that bit of code duplication - as re-using would create coupling between both modules (and there is no "good place higher up completely free of cycles" to put the definition).

PR #3297, opened 2026-09-03
Add basic `AGENTS.md`

### LINKED ISSUE #3292: Add initial agent instructions

As the usage of agentic code tools in the repo increases, it would be nice to provide them with instructions for optimizing their usage and contributions to this library.

This story is about setting up a minimal starting point, that we can grow as the project evolves. Forces to consider:
* size and token consumption of the instructions
* tool usage (ie. `AGENTS.md` vs `CLAUDE.md` and relation)
* best practices for such instructions

The suggestion is starting lean and postpone finer grained decisions and content to later issues - this is about setting a baseline.

**Acceptance criteria**

The repository contains instructions for agents, and those instructions lead to more optimal contributions.

TsafrirA:
### Summary
Closes #3292 

This PR adds a basic `AGENTS.md` file. I tried to keep it lean, while also addressing issues I've encountered while working on the repo with agents. I drew inspiration from [here](https://agents.md/#examples) and [here](https://www.aihero.dev/a-complete-guide-to-agents-md).

It is expected that agent-facing infrastructure will evolve over time.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,9 @@
 # Agent Instructions
 
+## Project description
+The package provides the interface to access the IBM Quantum Compute service on IBM Quantum Platform,
+to run quantum computations.
+
 ## Project goals
 
 Maintain correctness, backward compatibility, and consistency with
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,40 @@
+# Agent Instructions
+
+## Project goals
+
+Maintain correctness, backward compatibility, and consistency with
+existing Qiskit patterns.
+
+## Before making changes
+
+- Read surrounding code before introducing new patterns.
+- Prefer extending existing implementations over creating new abstractions.
+- Keep changes narrowly scoped to the requested issue.
+- Keep changes to client-facing interfaces to a minimum.
+
+## Testing
+
+- Add or update tests when behavior changes.
+- Prefer parameterized tests over individual tests.
+- Run the smallest relevant test set first.
+- Avoid running the entire test suite unless necessary.
+- Integration tests require live IBM Quantum credentials; request permission before running and verify existence of required env variables (`QISKIT_IBM_TOKEN`, `QISKIT_IBM_INSTANCE`, `QISKIT_IBM_URL`, `QISKIT_IBM_QPU`).
+- Do not use `unittest.TestCase` directly; use the project base classes in `test/ibm_test_case.py`.
+- Unit tests must not contact any external service; use `FakeBackendV2` or `QiskitRuntimeLocalService` for backend interactions.
+
+## Constraints
+
+- Preserve public API compatibility.
+- Changes to client-facing interfaces must adhere to the deprecation policy (see `DEPRECATION.md` for the full policy), and be accompanied by a release note (see `CONTRIBUTION.md` for more details).
+- The package must maintain backward compatibility when it comes to loading old jobs.
+- Do not introduce new dependencies without justification.
+- Follow repository linting and formatting rules (use `pre-commit run` to run checks).
+
+## Legacy vs. executor based primitives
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,44 @@
+# Agent Instructions
+
+## Project description
+The package provides the interface to access the IBM Quantum Compute service on IBM Quantum Platform,
+to run quantum computations.
```
diego-plan9, line 5: Minor nits, commenting here but applying to the whole file, and aimed at human maintainers of the file.

* Can you ensure there is a newline after a header?
* Can you go with a unified line wrap (ie. 100 chars)? There are some bits over the limit (in this section, at 101) and other sections under the limit (Project goals)

AGENTS.md
```diff
@@ -0,0 +1,44 @@
...
+- Prefer extending existing implementations over creating new abstractions.
+- Keep changes narrowly scoped to the requested issue.
+- Keep changes to client-facing interfaces to a minimum.
+
+## Testing
+
+- Add or update tests when behavior changes.
+- Prefer parameterized tests over individual tests.
+- Run the smallest relevant test set first.
+- Avoid running the entire test suite unless necessary.
```
diego-plan9, line 24: Unsure if we need both? The first one might semi-imply the second one.

I don't want to over-optimize in this PR, but I'm thinking we probably will want more items in the testing section as the file gets revised (things that come to mind: use pytest as runner but not as framework, mention of the existing decorators, something about the usage of mocking, other conventions). Not saying they are worth inclusions here, just having the feeling we might want to make room already.

AGENTS.md
```diff
@@ -0,0 +1,44 @@
...
+- Integration tests require live IBM Quantum credentials; request permission before running and verify existence of required env variables (`QISKIT_IBM_TOKEN`, `QISKIT_IBM_INSTANCE`, `QISKIT_IBM_URL`, `QISKIT_IBM_QPU`).
+- Do not use `unittest.TestCase` directly; use the project base classes in `test/ibm_test_case.py`.
+- Unit tests must not contact any external service; use `FakeBackendV2` or `QiskitRuntimeLocalService` for backend interactions.
```
diego-plan9, line 27: Unsure if the second part will be successful in conveying info or cause some extra work (`QiskitRuntimeLocalService` is used through decorators and not standalone, and we are moving some of the tests to the "registries" approach). Happy to trial-and-error after merging!

AGENTS.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Constraints
+
+- Preserve public API compatibility.
+- Changes to client-facing interfaces must adhere to the deprecation policy (see `DEPRECATION.md` for the full policy), and be accompanied by a release note (see `CONTRIBUTION.md` for more details).
+- The package must maintain backward compatibility when it comes to loading old jobs.
+- Do not introduce new dependencies without justification.
+- Follow repository linting and formatting rules (use `pre-commit run` to run checks).
+
+## Legacy vs. executor based primitives
```
diego-plan9, line 37: Maybe also a premature optimization: would it make sense to make this a sub-entry on a `## Architecture` section already? There is potentially a good chunk of instructions (general description of the different layers or features, non-evident information for specific components) that might be a fit here in the future.

diego-plan9 (commented):
Thanks Tsafrir!

It looks like a very sensible start to me. Other than the nit comment on the formatting, the rest of them are mostly thoughts and non-blocking, as it will be easier to reason and agree (or disagree!) about them once we put the file to use.

AGENTS.md
TsafrirA: Yes, that makes sense. That's the sort of things I was aiming for.

AGENTS.md
TsafrirA: I actually haven't had an agent trying to do something like that, so I agree it will be better to remove, and add later if we think it's needed.

AGENTS.md
TsafrirA: Strictly speaking the two are not the same. One deals with how you start, and the other with when you finish. But I don't have a strong preference here.

Mocking has been an issue in the team, so I think it makes sense to include it right off the start.

Obviously, we will iterate and optimize as we move along. By no mean do I view this as a complete list.

TsafrirA:
@diego-plan9 
I addressed your comments, and looking for a re-review.

diego-plan9 (approved):

PR #3299, opened 2026-09-04
Update registries focused on usage, add new endpoint

### LINKED ISSUE #3296: Prepare for upcoming API changes

In the upcoming development cycle, new features will be added related to job usage and job information. This story is about doing some groundwork for easier accommodating such changes:
* update the `registries` testing mechanism for easily handling different job metrics (currently, tests mock the client response at the Python level)
* provisionally add support for the workloads endpoint to the REST client (not reachable through `QiskitRuntimeService` yet, but for allowing exploring the usage of such endpoint)

**Acceptance criteria**

The tests related to job/usage metrics use mocked HTTP calls that can be configured via `Job`, and the library rest layer has support for the workloads endpoint.

diego-plan9:
### Summary

Groundwork for making it easier to iterate on usage-related features, as such work is expected in the upcoming weeks. It is centered in preparing the `registries` for easier customizing of the responses related to those endpoints - moving some tests to using the `@mock_responses` decorator (as this way we do get guarantees we are exercising the real `QiskitRuntimeService`, and having the callbacks take into account new information.

The second part is adding "experimental" (as in, still not used by `QiskitRuntimeService`) supports to the `/workloads` endpoint. While it might not end up being used by the upcoming changes, it is a versatile endpoint that could simplify and unify several parts of the service in the future, and having this minimal support available allows for exploring it already.

### Details and comments

Fixes #3296

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/api/rest/runtime.py
+++ b/qiskit_ibm_runtime/api/rest/runtime.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import json
 import logging
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 from ...json import RuntimeEncoder
 from ...utils import local_to_utc
@@ -38,6 +38,7 @@ class Runtime(RestAdapterBase):
         "jobs": "/jobs",
         "backends": "/backends",
         "cloud_usage": "/instances/usage",
+        "workloads": "/workloads",
     }
 
     def program_job(self, job_id: str) -> ProgramJob:
@@ -228,3 +229,67 @@ class Runtime(RestAdapterBase):
         """
         url = self.get_url("cloud_usage")
         return self.session.get(url, headers=self._HEADER_JSON_ACCEPT).json()
+
+    def workloads_get(
+        self,
+        user: str | None = None,
+        sort: Literal["createdAt", "-createdAt"] | None = None,
+        limit: int | None = None,
+        previous: str | None = None,
+        next: str | None = None,
+        backend: str | None = None,
+        search: str | None = None,
+        status: list[str] | None = [],
+        mode: Literal["job", "session", "batch"] | None = None,
+        created_after: datetime | None = None,
+        created_before: datetime | None = None,
+        tags: list[str] | None = None,
+    ) -> dict:
+        """Get a list of user instance workloads.
+
+        Args:
+            user: User identifier.
+            sort: Field to sort the workloads.
+            limit: Number of workloads to return at a time.
+            previous: Cursor to previous workloads results page.
+            next: Cursor to next workloads results page.
+            backend: Backend name.
+            search: Search string, used to filter workloads by id or tags.
+            status: status type to filter workloads by.
+            mode: Workload mode.
+            created_after: Initial date of workloads to be included.
+            created_before: Last date of workloads to be included.
+            tags: List of tags for the worload.
+
+        Returns:
+            JSON response.
+        """
+        url = self.get_url("workloads")
+        payload: dict[str, int | str | list[str]] = {}
+
+        if user:
+            payload["user"] = user
+        if sort:
+            payload["sort"] = sort
+        if limit:
+            payload["limit"] = limit
+        if previous:
+            payload["previous"] = previous
+        if next:
+            payload["next"] = next
+        if backend:
+            payload["backend"] = backend
+        if search:
+            payload["search"] = search
+        if status:
+            payload["status"] = status
+        if mode:
+            payload["mode"] = mode
+        if created_before:
+            payload["created_before"] = local_to_utc(created_before).isoformat()
+        if created_after:
+            payload["created_after"] = local_to_utc(created_after).isoformat()
+        if tags:
+            payload["tags"] = tags
+
+        return self.session.get(url, params=payload, headers=self._HEADER_JSON_ACCEPT).json()
--- a/qiskit_ibm_runtime/api/rest/runtime.py
+++ b/qiskit_ibm_runtime/api/rest/runtime.py
@@ -265,31 +265,22 @@ class Runtime(RestAdapterBase):
             JSON response.
         """
         url = self.get_url("workloads")
-        payload: dict[str, int | str | list[str]] = {}
-
-        if user:
-            payload["user"] = user
-        if sort:
-            payload["sort"] = sort
-        if limit:
-            payload["limit"] = limit
-        if previous:
-            payload["previous"] = previous
-        if next:
-            payload["next"] = next
-        if backend:
-            payload["backend"] = backend
-        if search:
-            payload["search"] = search
-        if status:
-            payload["status"] = status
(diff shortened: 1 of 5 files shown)


test/unit/test_jobs.py
```diff
@@ -281,12 +283,14 @@ def test_result_decoders_is_always_sequence(self, decoders, service):
 
         self.assertIsInstance(job._result_decoders, list)
 
-    @run_cloud_fake
-    def test_result_chains_decoders(self, service):
+    @mock_responses
+    def test_result_chains_decoders(self, registry):
```
diego-plan9, line 287: This one does not exercise the new features: it was mostly a conversion of this test while revising this file.

test/unit/test_jobs.py
```diff
@@ -177,38 +179,36 @@ def test_delete_job(self, service):
         with self.assertRaises(RuntimeJobNotFound):
             service.job(job.job_id())
 
-    @run_cloud_fake
-    def test_instance_limit_warning(self, service):
+    @mock_responses
+    def test_instance_limit_warning(self, registry):
```
diego-plan9, line 183: The changes in this test (and in `test_usage_partial` / `test_usage_no_partial`) can be a good starting point when reviewing (ie. showcase the `Job` changes and how they are taken into account by the callbacks).

qiskit_ibm_runtime/api/rest/runtime.py
```diff
@@ -228,3 +229,67 @@ def cloud_usage(self) -> dict[str, Any]:
...
+
+        Args:
+            user: User identifier.
+            sort: Field to sort the workloads.
+            limit: Number of workloads to return at a time.
+            previous: Cursor to previous workloads results page.
+            next: Cursor to next workloads results page.
+            backend: Backend name.
+            search: Search string, used to filter workloads by id or tags.
+            status: status type to filter workloads by.
```
yaelbh:
```suggestion
            status: Status type to filter workloads by.
```

qiskit_ibm_runtime/api/rest/runtime.py
```diff
@@ -228,3 +229,67 @@ def cloud_usage(self) -> dict[str, Any]:
...
+            mode: Workload mode.
+            created_after: Initial date of workloads to be included.
+            created_before: Last date of workloads to be included.
```
yaelbh: It is not clear whether the specified dates are included.

test/registries.py
```diff
@@ -141,6 +153,16 @@ class Job:
     raw_results: str | None = None
     """Response for the job results."""
 
+    usage: dict | None = None
+    """Job usage dictionary that overrides the default `usage` field in responses."""
```
rainerschoe:
```suggestion
    usage: dict | None = None
    """Job usage dictionary."""
```
Where is the "default" `usage` field which would be overridden by this?

test/unit/test_jobs.py
```diff
@@ -244,27 +244,29 @@ def test_wait_for_final_state_poll_inverval_custom(self, value, service):
...
+        usage = {"qpu_charge_time_seconds": 123, "status": status}
+        registry.add_job(Job("my_job", "common_backend", usage=usage), "a")
+        service = QiskitRuntimeService(token="my_token")
 
-        with patch.object(BaseFakeRuntimeClient, "job_metadata", return_value=api_response):
-            usage = job.usage()
-            self.assertEqual(usage, 0 if status != "completed" else 123)
+        job = service.job("my_job")
+        usage = job.usage()
+        self.assertEqual(usage, 0 if status == "pending" else 123)
```
rainerschoe, line 257: I like this:
- the `registry` via `mock_responses` mves mocking out of the test implementations
- We are testing more specific status
- We can test directly on service instead of lower-level functions

Even better if we could utilize pytest fixtures, then we would not need the decorator. But at least a step into the right direction :-)

test/registries.py
diego-plan9: A bit further down, in the `__post_init__`.

It is the same pattern as with `Instance.usage` - in this case, we do need other fields for computing the "default", so `__post_init__` seems the most natural way.

qiskit_ibm_runtime/api/rest/runtime.py
```diff
@@ -228,3 +229,67 @@ def cloud_usage(self) -> dict[str, Any]:
...
+        if status:
+            payload["status"] = status
+        if mode:
+            payload["mode"] = mode
+        if created_before:
+            payload["created_before"] = local_to_utc(created_before).isoformat()
+        if created_after:
+            payload["created_after"] = local_to_utc(created_after).isoformat()
+        if tags:
+            payload["tags"] = tags
```
rainerschoe: This is reading a bit weirdly and also might be affected by boolean casts (like what if limit is `0` )
I'd propose to
- get rid if the `if`...`if` construct
- check against `is None` to avoid accidents.

Some idea from Copilot AI:
```suggestion
payload = {
    "user": user,
    "sort": sort,
    "limit": limit,
    "previous": previous,
    "next": next,
    "backend": backend,
    "search": search,
    "status": status,
    "mode": mode,
    "created_before": (
        local_to_utc(created_before).isoformat()
        if created_before
        else None
    ),
    "created_after": (
        local_to_utc(created_after).isoformat()
        if created_after
        else None
    ),
    "tags": tags,
}

payload = {k: v for k, v in payload.items() if v is not None}
```

Edit: Sorry the comment should be covering the whole if.... code block, but somehow GH won't let me...

test/unit/test_jobs.py
diego-plan9: > We are testing more specific status

Technically this is due to some potential mismatch between the API spec (declares the possible values to be `completed`, and empirically it seems to return `complete`) that will be pursued separately (our implementation is based just on `pending`).

rainerschoe (changes requested):
Looks good. Some comments, but only one around the `if ...` I consider blocking.

test/registries.py
rainerschoe: but this is field documentation. the text around defaults should if required be part of `__init__` right?

qiskit_ibm_runtime/api/rest/runtime.py
diego-plan9: Indeed - it was mostly keeping things inline with the existing functions of the file, which use that pattern. I agree it doesn't hurt to try something cleaner with this one. c76ba612908f2df0d36fa941c60baf505dcb128f

test/registries.py
diego-plan9: Ah, I see what you mean now! I'm not super concerned about abusing a bit the field docstrings for these (we the testers are the consumers, and provide lots of visibility on how to use the fields) - at the same time, I like keeping them closer to documenting just what the field contains. Tweaked at 3ac2fe5a32c2012fb703ad22a35bcd4d44ca87a9

rainerschoe (approved):
Thanks for incorporating both proposals. LGTM now :-)

Issue #3295, opened 2026-09-03
Make mypy configuration more strict and aware of third-party dependencies


diego-plan9:
**What is the expected enhancement?**

Currently, the `mypy` configuration is quite lax. In particular, it does not take third-party dependencies into account, which places some limits on the amount of type infering that can be enforced, as well as having the potential for tine tuning further. This causes some typing issues only available when using other checkers (ie. IDEs using pyright) that can hide potential bugs.

This story is not about achieving 100% precise typing of the library, but about reaching a state where we can enable third-party dependency checking and a better balance when it comes to opaquing typing issues (ideally being caught by the automated `mypy` run). There are some relatively complex items to tackle:
* our usage of `| None`  in multiple places with `None` as a placeholder that is populated and whether coercion can take place
* constraints from third-party libraries that might not be declaring type hints fully
* imprecise or historical typing (`Backend` vs `BackendV2`; the `mode` handling)

The plan is to tackle this issue incrementally, and reevaluate the balance progress is made.

Some relation to #2514 (some ignores are likely to be made unneeded as more items are properly fixed - and some new ones are likely to be needed for typing issues from upstream).

**Acceptance criteria**

Our `mypy` configuration takes into account third-party dependencies, and the configuration is able to detect more legitimate issues.

PR #3300, opened 2026-09-04
Revise some typing hints (excluding wrappers)

diego-plan9:
### Summary

A round of type hints revising, focused on the ones that appear when the current configuration is changed from `no_site_packages =  true` to `no_site_packages = false`. This changeset excludes the errors that show up related to the wrappers, as the assumption is that those are in modules that will be moved away soon.

With these changes, we would be one step closer to disabling `no_site_packages` (when the wrapper-related modules that contain errors move, in theory we could disable it). There are still plenty of errors that are not caught by the `mypy` configuration, but this way we provide `mypy` with a small bump in its inference abilities.

### Details and comments

Related to #3295

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude opus 4.8

### DIFF
--- a/qiskit_ibm_runtime/visualization/circuit_schedule.py
+++ b/qiskit_ibm_runtime/visualization/circuit_schedule.py
@@ -22,6 +22,7 @@ import numpy as np
 from .utils import plotly_module
 
 if TYPE_CHECKING:
+    import numpy.typing as npt
     from plotly.graph_objects import Figure as PlotlyFigure
     from plotly.graph_objects import Scatter
 
@@ -70,12 +71,12 @@ class CircuitSchedule:
     ):
         self.channels: list = None
         self.type_to_idx: dict[str, int] = None
-        self.circuit_scheduling: np.array = None
+        self.circuit_scheduling: npt.NDArray[np.str_] = None
 
         raw_data = self._load(circuit_schedule)
         self._parse(raw_data)
 
-        self.instruction_set: set[str] = set()
+        self.instruction_set: set[str] | npt.NDArray[np.str_] = set()
         self.max_time: int = None
         self.color_map: dict[str, str] = {}
         self.annotations: list[dict] = []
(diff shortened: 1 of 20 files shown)


diego-plan9:
Please note some of the changes are not just annotation changes, but code changes (mostly due to the original code re-using variables for new purposes). The changes seem harmless and contained, but an extra look is appreciated.

qiskit_ibm_runtime/visualization/circuit_schedule.py
```diff
@@ -70,12 +71,12 @@ def __init__(
     ):
         self.channels: list = None
         self.type_to_idx: dict[str, int] = None
-        self.circuit_scheduling: np.array = None
+        self.circuit_scheduling: npt.NDArray[np.str_] = None
```
TsafrirA, line 74:
```suggestion
        self.circuit_scheduling: npt.NDArray[np.str_] | None = None
```

?

yaelbh (approved):
LGTM

qiskit_ibm_runtime/visualization/circuit_schedule.py
TsafrirA: Also applies to the previous two lines.

TsafrirA (commented):
Just a minor comment on `| None`.

qiskit_ibm_runtime/visualization/circuit_schedule.py
diego-plan9: It is a good change (ie. one we want) - but one more related to `strict_optional = false` (rather than related to `no_site_packages`). Disabling that configuration surfaces 202 errors, which in general require different measures (and some of them will require code changes or arguably interface changes) and deeper evaluation and reviewing.

So let's rather tackle the strict optionals (I do really want to tackle them!) as another PR related to #3295, keeping the scope of this one about `no_site_packages`.

Issue #3193, opened 2026-08-10
Write a detailed guide for the wrapper estimator


TsafrirA:
**What is the expected enhancement?**
We need to extend our documentation with a detailed guide for the wrapper estimator. The guide should include:

- Overview
- Code examples
- Inputs
- Outputs (data fields and metadata, per noise mitigation technique)
- Comparison with legacy estimator (differences, migration guide)

**Acceptance criteria**
The guide was added to the repo's documentation.

jyu00:
> Comparison with legacy estimator (differences, migration guide)

fwiw I'm also writing a migration guide and an AI skill for legacy primitives -> wrapper primitives migration. The guide is for our docs site and shouldn't conflict with what you have in the PR.

PR #3301, opened 2026-09-04
Sphinx updates to prepare for wrappers release

diego-plan9:
### Summary

Round of sphinx updates, in preparation for eventual release of wrappers:
* qualify references to `SamplerV2` and `EstimatorV2`, to prevent ambiguity and multiple references warnings
* some syntax updates

There are still some warnings left related to the clash between classes, seemingly related to `autodoc` and the use of generics.

### Details and comments

f570016527c5e10b870720d9daeb19fa02572c80 and f61f3995364e30610e05b8648d9113e384556f49 temporary added the new content to the documentation - they can be used as a guidance for catching errors (ie. adding those changes, running sphinx, checking output).

Related to #3193 (in spirit: documentation)

### AI/LLM disclosure

### DIFF
--- a/docs/apidocs/executor_estimator.rst
+++ b/docs/apidocs/executor_estimator.rst
@@ -1,0 +1,4 @@
+.. automodule:: qiskit_ibm_runtime.executor_estimator
+   :no-members:
+   :no-inherited-members:
+   :no-special-members:
--- a/docs/apidocs/executor_estimator.rst
+++ b/docs/apidocs/executor_estimator.rst
@@ -1,4 +1,0 @@
-.. automodule:: qiskit_ibm_runtime.executor_estimator
-   :no-members:
-   :no-inherited-members:
-   :no-special-members:
--- a/docs/apidocs/executor_sampler.rst
+++ b/docs/apidocs/executor_sampler.rst
@@ -1,4 +1,0 @@
-.. automodule:: qiskit_ibm_runtime.executor_sampler
-   :no-members:
-   :no-inherited-members:
-   :no-special-members:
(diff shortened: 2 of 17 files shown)


apchytr (approved):
As an aside: is there a general rule to follow for when we want to use the full path in sphinx versus e.g. `~.EstimatorOptions`? I know this PR is meant to address reference warnings but just wondering what should be the best practice for these in general

diego-plan9:
> As an aside: is there a general rule to follow for when we want to use the full path in sphinx versus e.g. `~.EstimatorOptions`? I know this PR is meant to address reference warnings but just wondering what should be the best practice for these in general

On conventions in general - the sphinx syntax is at https://www.sphinx-doc.org/en/master/usage/domains/python.html#target-specification (target specification and target resolution). We don't have an explicit convention in the project - in general, we tend to specify via:
```
 :class:`~.MyClass` 
```
as this will:
* look first in the current module (via the `.`), which is a common case and saves us some characters
* just display the class name as the rendered link text (via the `~`), which is more reading-friendly

Regarding using the full path: in this PR, I used the full path in the cases where we have classes with the same name (ie. `EstimatorOptions`, `SamplerV2`) and sphinx is unable to decide between them. The PR is "reactive" - the first two commits add classes to the documentation and sphinx index, this triggers the warnings, the rest of the commits fix them, and the final commits removes them from the documentation (as we don't want them to be rendered yet, but will once we declare wrappers public). 

So an approach might be: keep using the without-path form until sphinx complains, and if so, use the full-path form for resolving ambiguity.

PR #3303, opened 2026-09-08
Avoid using FakeRuntimeService in tests directly

diego-plan9:
### Summary

Avoid using `FakeRuntimeService` (direct usage - importing an instantiating) in the unit tests, by converting these tests to a registry-based approach:
* registries gain the ability of using `Session`s (superficially, same "add just what is needed for the tests" philosophy)
* some tweaks to the decorator (for playing nicely with `combine()` and smaller improvements (instance tags, better handling of catalog URLs)
* convert tests to `@mock_requests` and to use `@ddt`. Some tests had some stronger or less evident tweaks, spirit should stay the same in all of them

This gets us closer to avoid `FakeRuntimeService` - the remaining usage is through the `@run_cloud_fake` decorator.

### Details and comments

Related to #3296 (continuing the approach, strictly not needed for new features though)

### AI/LLM disclosure

### DIFF
--- a/test/unit/test_account.py
+++ b/test/unit/test_account.py
@@ -19,7 +19,6 @@ import os
 import uuid
 from typing import Any
 from unittest import skipIf
-from unittest.mock import patch
 
 from ddt import data, ddt
 
@@ -37,6 +36,7 @@ from qiskit_ibm_runtime.accounts.management import (
     _DEFAULT_ACCOUNT_NAME_IBM_QUANTUM_PLATFORM,
 )
 from qiskit_ibm_runtime.proxies import ProxyConfiguration
+from qiskit_ibm_runtime.qiskit_runtime_service import QiskitRuntimeService
 
 from ..account import (
     custom_envs,
@@ -44,8 +44,9 @@ from ..account import (
     no_envs,
     temporary_account_config_file,
 )
+from ..decorators import mock_responses
 from ..ibm_test_case import IBMTestCase
-from .mock.fake_runtime_service import FakeRuntimeService
+from ..utils import combine
 
 _TEST_IBM_CLOUD_ACCOUNT = Account.create_account(
     channel="ibm_cloud",
@@ -407,13 +408,14 @@ class TestAccountManager(IBMTestCase):
         )
         self.assertEqual(account.token, dummy_token)
 
+    @mock_responses
     @temporary_account_config_file()
-    def test_default_env_channel(self):
+    def test_default_env_channel(self, registry):
         """Test that if QISKIT_IBM_CHANNEL is set in the environment, this channel will be used."""
         token = uuid.uuid4().hex
         # unset default_channel in the environment
         with temporary_account_config_file(token=token), no_envs("QISKIT_IBM_CHANNEL"):
-            service = FakeRuntimeService()
+            service = QiskitRuntimeService()
             self.assertEqual(service.channel, "ibm_quantum_platform")
 
         # set channel to default channel in the environment
@@ -424,7 +426,7 @@ class TestAccountManager(IBMTestCase):
                 temporary_account_config_file(channel=channel, token=token),
                 custom_envs(channel_env),
             ):
-                service = FakeRuntimeService()
+                service = QiskitRuntimeService()
                 self.assertEqual(service.channel, channel)
 
     def test_save_private_endpoint(self):
@@ -461,8 +463,9 @@ class TestAccountManager(IBMTestCase):
         self.assertEqual(account.channel, "ibm_quantum_platform")
         self.assertEqual(account.token, _TEST_IBM_CLOUD_ACCOUNT.token)
 
+    @mock_responses
     @temporary_account_config_file()
-    def test_set_channel_precedence(self):
+    def test_set_channel_precedence(self, registry):
         """Test the precedence of the various methods to set the account.
 
         account name > env_variables > channel parameter default account
@@ -500,7 +503,7 @@ class TestAccountManager(IBMTestCase):
             custom_envs(channel_env),
             no_envs("QISKIT_IBM_TOKEN"),
         ):
-            service = FakeRuntimeService(name="any-quantum")
+            service = QiskitRuntimeService(name="any-quantum")
             self.assertEqual(service.channel, "ibm_quantum_platform")
             self.assertEqual(service._account.token, any_token)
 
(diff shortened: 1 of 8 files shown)


test/unit/test_account.py
```diff
@@ -722,214 +719,224 @@ def test_enable_account_both_channel(self):
...
-                url = uuid.uuid4().hex
-                envs = {
-                    "QISKIT_IBM_TOKEN": token,
-                    "QISKIT_IBM_URL": url,
-                    "QISKIT_IBM_INSTANCE": _DEFAULT_CRN,
-                }
-                with custom_envs(envs), no_envs("QISKIT_IBM_CHANNEL"):
-                    service = FakeRuntimeService(channel=channel)
+        token = uuid.uuid4().hex
+        url = "https://foo.cloud.ibm.com"
```
diego-plan9, line 733: In general, in this file `url` tended to be a random string, which resulted in legitimate errors. I used this URL (combined with tweaks in registries) as proof of "a different url than the standard one can be passed".

test/unit/test_account.py
```diff
@@ -722,214 +719,224 @@ def test_enable_account_both_channel(self):
...
+        envs = {
+            "QISKIT_IBM_TOKEN": token,
+            "QISKIT_IBM_URL": url,
+            "QISKIT_IBM_CHANNEL": channel,
+            "QISKIT_IBM_INSTANCE": _DEFAULT_CRN,
+        }
+        with custom_envs(envs):
+            service = QiskitRuntimeService()
+        self.assertEqual(service._account.channel, channel or "ibm_quantum_platform")
```
diego-plan9, line 762: This `or` was not explicit before, but I think it was a byproduct of `FakeRuntime`. `test_enable_account_by_env_channel` (a bit higher up) _did_ have the check, which seem to make sense (service account will always have a channel, and will default to the default one).

test/unit/test_data_serialization.py
```diff
@@ -258,24 +258,22 @@ def test_decoder_import(self):
                 )
                 self.assertIn(operator.__class__.__name__, proc.stdout)
 
-    def test_result_decoder(self):
+    @mock_responses
+    def test_result_decoder(self, registry):
```
diego-plan9, line 262: Some loss in this test - the original one got up to the `_run()` level in order to specify a decoder, I imagine that for checking that the `Job` instance would use the specified decoder if explicitly passed. That is not reachable by users (`RuntimeJobV2` is never instantiated directly), and we have some tests that instantiate `RuntimeJobV2` a bit hackishly elsewhere: so I think we are not losing something meaningful (`results(decoder=x)` _is_ exercised).

test/unit/test_session.py
```diff
@@ -108,33 +110,38 @@ def test_context_manager(self):
             session.cancel()
         self.assertFalse(session._active)
 
-    @data([None, "my_id"])
-    def test_session_from_id(self, calibration_id):
+    @data(None, "my_id")
```
diego-plan9, line 113: Actually, some small test bug here - original test was receiving the full list as `calibration_id`, which the Fake service happily accepted.

test/unit/test_session.py
```diff
@@ -108,33 +110,38 @@ def test_context_manager(self):
...
+    @mock_responses
+    def test_correct_execution_mode(self, registry):
         """Test that the execution mode is correctly set."""
-        _ = FakeRuntimeService(channel="ibm_quantum_platform", token="abc")
-        backend = get_mocked_backend("ibm_gotham")
+        service = QiskitRuntimeService(token="my_token")
+        backend = service.backend("common_backend")
         session = Session(backend=backend)
+
+        registry.add_session(RegistrySession(session.session_id, "common_backend"), "a")
```
diego-plan9, line 136: We add the session to the registry, as the call to the mocked endpoint will not automatically create it. Part of me is tempted to do so (avoids explicit registry calls in some cases, and can be very useful specially for submitting jobs and populate the registry), but meriting some more thought and a separate look (might complicate maintenance).

rainerschoe: I think it is fine to see in the test a bit how the mocking is done.
I am starting to think about why the decorator does provide registry, but not the service instance, it would make it more clear that the service is patched.

test/unit/test_sampler.py
```diff
@@ -191,37 +190,30 @@ def test_run_dynamic_circuit_with_fractional_opted(self):
...
-    def test_run_fractional_dynamic_mix(self, use_fractional):
+    @mock_responses
+    def test_run_fractional_dynamic_mix(self, registry):
         """Any backend cannot run mixture of fractional and dynamic circuits."""
-        service = FakeRuntimeService(
-            channel="ibm_quantum_platform",
-            token="my_token",
-            backend_specs=[FakeApiBackendSpecs(backend_name="FakeFractionalBackend")],
-        )
-        backend = service.backends("fake_fractional", use_fractional_gates=use_fractional)[0]
```
rainerschoe, line 224: I like this cleanup :-)

test/unit/test_data_serialization.py
rainerschoe: what is the use of specifying a decoder in `run_program` ?

test/unit/test_session.py
diego-plan9: > I am starting to think about why the decorator does provide registry, but not the service instance, it would make it more clear that the service is patched.

Mostly as the registry is the one that it is more likely to be touched by a particular test (ie `registry.add_x()` when preparing what the test needs), and as some test require / exercise the `QiskitRuntimeService` constructor arguments (`instance` in particular). So keeping the service out of the decorator allows for some explicit, same code as if the user was creating the service approach - at the expense of needing to build the service in nearly every test.

rainerschoe (approved):
LGTM :-)
I did not individually review each test, after I understood the general pattern you do here for some of them.

test/unit/test_data_serialization.py
diego-plan9: Passing it to `QiskitRuntimeService._run`, which is used for setting the default when building `RuntimeJobV2`(around [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/qiskit_runtime_service.py#L1109)). That default is taking into account when calling `job.result()` around [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/runtime_job_v2.py#L142-L144).

So we do lose a bit (the ability to check that a hand-crafted job uses its default decoder if no decoder is specified when calling `result()`) - but actually:
* `run_program` is to be removed/reworked soon in a follow-up PR, it is a testing tool
* `test_result_decoders_is_always_sequence()` [here](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/test/unit/test_jobs.py#L273) does check for the default setting explicitly (and the tests calling `result()` elsewhere in the codebase, and `result(decoder=x)` in this test and probably others, check for how the result uses the decoders implicitly)

PR #3308, opened 2026-09-10
Attempt direct backend creation in service.backend()

### LINKED ISSUE #3306: Allow service.backend() to return a backend directly

Currently, `service.backend()` requires that the backend name is returned by the `/v1/backends` endpoints in order for the backend to be returned. This prevents instantiating backends that might not be present if the endpoint default filtering excludes them

**Acceptance criteria**

`service.backend()` is able to return a backend without depending on the mentioned endpoint.

diego-plan9:
### Summary

Allow `service.backend()` to attempt to create a backend directly (as in, constructing the `backend` object from its properties and configuration) even if it is not returned in the list of `service.backends()`.

### Details and comments

This prepares for an upcoming API change where the defaults for the backends endpoint might filter out some devices. 

Fixes #3306

### AI/LLM disclosure

### DIFF
--- a/qiskit_ibm_runtime/qiskit_runtime_service.py
+++ b/qiskit_ibm_runtime/qiskit_runtime_service.py
@@ -968,7 +968,7 @@ class QiskitRuntimeService:
             from qiskit_ibm_runtime import QiskitRuntimeService
 
             service = QiskitRuntimeService()
-            backend = service.backend()
+            backend = service.backend("ibm_kingston")
 
             status = backend.status()
             assert status.operational and status.status_msg == "active"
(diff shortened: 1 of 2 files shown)


qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -726,6 +727,7 @@ def _create_backend_obj(
         instance: str,
         use_fractional_gates: bool | None,
         calibration_id: str | None = None,
+        cache: bool = True,
```
diego-plan9, line 730: This change allows for "whenever a backend object needs to be created, have the caller decide whether its configuration is cached in the state variables of the service".

Without this change, backends that were created in the new code in `.backend()` would remain tied to the instance that created them (as they would be cached to that instance). Since the new feature is meant to be used in very specific cases (ie. retrieving a backend manually) and to be superseeded soon, relinquishing caching of that backend seems like a sane trade-off (still allows for ensuring that the created backend belongs to the right instance, at the expense of some extra API calls when the same backend is requested multiple times).

SamFerracin (approved):
The PR looks good!
DM'ed you about the bigger picture

PR #3309, opened 2026-09-10
Remove run_cloud_fake and FakeRuntimeFoo items

diego-plan9:
### Summary

Follow-up #3303 and other PRs, in order to replace the usages of `FakeRuntimeService` and their related items (gated under the `@cloud_fake` decorator) with `@mock_responses`.

Involves:
* registries now understand the cancel job / delete job endpoints
* registries now have some filtering capabilities for the jobs endpoint
* registries now have more features for dealing with a job status (including "advancing" the status when the job details endpoint is hit)

Test are updated in order to use `@data` and `@mock_responses` - some inline comments for the tests that required something else. 

### Details and comments

Follow-up to #3303

### AI/LLM disclosure


test/registries.py
```diff
@@ -173,13 +178,25 @@ class Job:
     usage: dict | None = None
     """Job usage dictionary."""
 
+    statuses: list[JobStatus] = field(default_factory=lambda: ["completed"])
```
SamFerracin, line 181: Not hyped about the word "statuses," but I have no better alternatives. 

About the default: Any reasons about this choice, VS for example `["running", "completed"]` or `["queued", "running", "completed"]`? I'm not trying to advocate for one or another, just to understand the choice

test/unit/test_backend_retrieval.py
```diff
@@ -265,7 +265,7 @@ def test_get_backend_with_fractional_optin(self, use_fractional, registry):
 
         This test is originally written in 2024.05.31
         """
-        registry.add_backend(Backend.from_(FakeFractionalBackend, queue_length=10))
```
SamFerracin, line 283: I understand that removing `queue_length` only happens because it was unneded, right?

test/registries.py
```diff
@@ -557,6 +588,22 @@ def callback_jobs_get(self, request: PreparedRequest) -> CallbackResult:
...
 
+        # Get filtering parameters.
+        query_params = parse_qs(urlparse(request.url).query)
+        limit_param = query_params.get("limit", [None])[0]
+        limit = int(limit_param) if limit_param else 20
+        offset_param = query_params.get("offset", [None])[0]
+        offset = int(offset_param) if offset_param else 0
+        pending = query_params.get("pending", [None])[0]
+
+        statuses = ["queued", "running", "completed", "cancelled", "failed"]
```
SamFerracin: Minor comment:

This creates a code-duplication with the alias above ([link](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3309/changes#diff-f50489335f14cfa88386830d754216b7ec57de78ff3dc847eec6fbd3dd1f4cd4R41)). You could consider adding a
```python
AVAILABLE_STATUSES = frozenset(["queued", "running", "completed", "cancelled", "failed"])
```
and then use this both here and in the alias, so that we have a single place that defines the source of truth for this

SamFerracin (commented):
Took me a bit to review. Left a few minor comments. Looks good overall

test/registries.py
diego-plan9: About the word: not thrilled either, but could not find a better alternative.

The default is for matching the default for `status`:

```python
    status: Literal[JobStatus] = "completed"
```

Since the default status for a job is completed, and it is a final state (ie. no transitions to other state happens), both combined have the effect of "a `Job()`, by default, will be in `completed` state and remain there".

test/unit/test_backend_retrieval.py
diego-plan9: Yes - just some cleanup.

test/registries.py
```diff
@@ -31,10 +31,15 @@
 
     from qiskit_ibm_runtime.fake_provider.fake_backend import FakeBackendV2
 
-# ``responses`` callback return value: (status code, headers, body).
+
 CallbackResult: TypeAlias = tuple[int, dict[str, str], str]
+"""``responses`` callback return value: (statis code, headers, body)."""
```
apchytr:
```suggestion
"""``responses`` callback return value: (status code, headers, body)."""
```
?

test/registries.py
```diff
@@ -31,10 +31,15 @@
...
 
 PricingType: TypeAlias = Literal["free", "trial", "paygo", "paid", "subscription", "unknown"]
+"""Princing types for an instance."""
```
apchytr:
```suggestion
"""Pricing types for an instance."""
```

test/registries.py
diego-plan9: I played a bit with it - I think we need Python 3.11 for defining the states as a tuple/list, and be able to use as `Literal[*AVAILABLE_STATUSES]`. So resolved the duplication the other way around (getting the list at this line by inspecting the literal) at 9ec02d516836cf67e07ab0050153efdc16fc3fc3

apchytr (approved):

PR #3310, opened 2026-09-10
Integrating qiskit-mitigation into the wrapper estimator

TsafrirA:
### Summary
The estimation and error mitigation code which was created in `qiskit-ibm-runtime` has finally been moved to `qiskit-mitigation`. This PR integrates `qiskit-mitigation` into the wrapper estimator, and removes the code from our repo.

### AI/LLM disclosure


qiskit_ibm_runtime/executor_estimator/pec_utils.py
```diff
@@ -14,53 +14,8 @@
...
-            ref = annot.ref
-            try:
-                model = noise_model[ref]
-            except KeyError:
-                raise IBMInputValueError(f"Noise model is missing for layer with reference {ref}")
-            # scale the noise by noise_factor
-            model = model.scale_rates(noise_factor)
-            gamma *= model.inverse().gamma()
-    return gamma
+import sys
```
diego-plan9, line 18: Minor - with the `zne/` folder gone and one just this file left in this `pec/` folder, it would be nice to just have either:
* this file moved to `executor_estimator/pec_utils.py`
* or the functions moved to `executor_estimator/utils.py`

No strong preference (supporting argument 1: we do have `trex_setup.py`, which I see as an example of "functions specific to a method that are grouped in a file that follows `method_description.py` pattern; and `utils.py` is a bit too large already / supporting argument 2, I have not checked if everything in `utils.py` currently is shared by _all_ methods, and having more into it would not hurt). 

Can be tackled as follow-up.

qiskit_ibm_runtime/executor_estimator/utils.py
diego-plan9, line 256: I'm not sure if related to this PR - but seems that `find_unique_layers()` (and some methods only used by it) is now specific to `executor_sampler`, rather than `executor_estimator`?

If so, it would be good to move to `executor_sampler/utils.py` instead. Also something that can be tackled post-merge.

qiskit_ibm_runtime/options_models/zne.py
```diff
@@ -165,3 +166,21 @@ class ZneOptions(BaseOptionsModel):
     points at which the ``extrapolator``\s are evaluated to be returned in the data
     fields called ``evs_extrapolated`` and ``stds_extrapolated``.
     """
+
+    def resolve_noise_factors(self) -> tuple[np.ndarray, np.ndarray]:
+        """Resolve ``noise_factors`` and ``extrapolated_noise_factors`` into float arrays.
+
+        Returns:
+            A tuple ``(noise_factors, extrapolated_noise_factors)`` as float arrays.
+        """
```
diego-plan9: This goes against the idea of keeping the option models as tiny as possible (just containers of data, with no extra responsibilities), and seems to be "just" needed due to the needs of argument conversion in an internal, not user-facing step (ie. estimator prepare).

Can it be moved as a function somewhere in prepare? Can also be a follow-up.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -162,86 +160,137 @@ def _validate(
                 )
 
 
+def _choose_task_class(resilience: ResilienceOptions) -> type:
```
diego-plan9: Can the return type be narrowed down (perhaps `type[MitigationTask]`, as it seems that they do have a hierarchy? Mostly for easier maintainability and probably helping the typing tools inference. It would also be nice to rename the function to something a bit more explicit too (`_task_class_based_on_resilience_options_but_shorter_name_than_this_one`).

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -122,34 +116,38 @@ def _validate(
...
-
-    Args:
-        coerced_pubs: The coerced estimator pubs.
-        finalized_options: The finalized estimator options.
-        backend: The backend for which the program is prepared.
-
-    Raises:
-        IBMInputValueError: If no pubs are provided, if unsupported option combinations
-            are detected, or if a required backend is missing.
-    """
```
diego-plan9, line 135: I wonder about the loss of docstrings (here, some bits in `prepare()`, also in `Dispatch to the appropriate prepare function and apply dynamical decoupling.`). Is it intentional?

If it is a matter of the methods diverging from their original meaning or the implementation diverging, it would be preferrable to keep the docstring contents updated rather than removing. They are internal functions, but keeping them documented (including arguments if possible) helps maintainability and readability a lot - and something much appreciated when it comes to the wrappers and the volume of functions involved.

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -162,86 +160,137 @@ def _validate(
...
+        if resilience.layer_noise_model is not None
+        else {}
+    )
+    measure_noise_learning = (
+        resilience.measure_noise_learning if resilience.measure_mitigation else None
+    )
+    trex: TREX | None = TREX() if measure_noise_learning is not None else None
+
+    # Build the QP
+    qp = QuantumProgram(shots=shots_per_randomization)
```
diego-plan9: Minor, but can longer names be used `quantum_program`, `data_bin` instead of abbreviations?

qiskit_ibm_runtime/executor_estimator/trex_setup.py
```diff
@@ -0,0 +1,74 @@
...
+"""Shared TREX finalization helper for the executor-based EstimatorV2."""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from qiskit_mitigation import TREX
+    from samplomatic.quantum_program import QuantumProgram
```
diego-plan9: In this and other instances of this occurrence: can you please double check if `samplomatic.QuantumProgram` is strictly required?

In #3290 we went with "we use `qiskit_ibm_runtime.QuantumProgram` in all places by default", as a strong signal that this library uses our customized version of `QuantumProgram`. While a number of methods are very likely work correctly with `samplomatic.QuantumProgram`, mix-and-matching them breaks the consistency, and opens us up for typing inconsistencies.

Could you check each instance specifically, and see if we can continue using `qiskit_ibm_runtime.QuantumProgram` by default? We can eventually revise _all_ methods (not just the ones changed in this PR) and update the hints of the ones that work with generic `samplomatic.QuantumProgram`, but is a bigger journey (and likely not desirable).

test/unit/executor_estimator/test_estimator_v2_pea_utils.py
```diff
@@ -40,16 +42,71 @@
...
 )
 
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from qiskit.primitives.containers.estimator_pub import EstimatorPubLike
+
+# ---------------------------------------------------------------------------
+# Helper: build EstimatorOptions from old-style args and call prepare()
+# ---------------------------------------------------------------------------
```
diego-plan9: Small ping for removing this temporary comments and revising some introduced `TODO`s .

diego-plan9 (commented):
This is looking really good, thanks @TsafrirA ! Most of my comments are organizational / maintainability asks and most of them can be deferred: the direction of the integration is looking good to me.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -117,424 +62,87 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
+        # Build per-item compilation metadata.
+        pub_meta: dict[str, Any]
+        if isinstance(task, GateFolding) and task.noise_factors is not None:
+            # ZNE has one result item per noise factor; aggregate their metadata.
+            num_nf = len(task.noise_factors)
+            items_meta = [
+                _create_pub_result_metadata(result[task._program_item_index + j].metadata)
+                for j in range(num_nf)
+            ]
+            # TODO: Is there a way to avoid the private attribute?
```
SamFerracin, line 115: Does this fix the TODO?
```suggestion
            items_meta = [
                _create_pub_result_metadata(result[i+j].metadata)
                for j in range(num_nf)
            ]
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -117,424 +62,87 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
-            )
-        elif mitigation is not None:
-            raise ValueError(f"Unknown mitigation technique {mitigation}")
-        else:
-            logger.info("Post processing pub %d.", pub_idx)
-            pub_result = create_pub_result(
-                item_result, observables, param_shape, param_basis_pairs, readout_noise_data
-            )
+        if i < len(circuits_metadata) and (cm := circuits_metadata[i]) is not None:
+            pub_meta["circuit_metadata"] = cm
```
SamFerracin: I don't understand the meaning of this `if`. In `prepare`, we do:
```python
    # Annotate our own metadata block (separate namespace from qiskit_mitigation)
    quantum_program.passthrough_data["post_processor"] = { 
        ...
        "circuits_metadata": [
            pub.circuit.metadata for pub in coerced_pubs
        ],
    }
```
So `circuits_metadata[i]` is always guaranteed to exist. I guess that you're concerned by things going wrong somehow, and getting an index error? Then I'd suggest to check that `len(circuits_metadata) == len(tasks)` and raising if this is False, as we were doing before

pyproject.toml
```diff
@@ -189,7 +190,8 @@ dependencies = [
     "pydantic>=2.13.0",
     "qiskit>=2.3.0",
     "pybase64>=1.4.0",
-    "samplomatic>=0.21.0"
+    "samplomatic>=0.21.0",
+    "qiskit-mitigation @ git+https://github.com/Qiskit/qiskit-mitigation.git",
```
apchytr: Would it make sense to pin this to a specific commit of `qiskit-mitigation` until we have proper versioning?

qiskit_ibm_runtime/executor_estimator/options_to_mitigation.py
```diff
@@ -0,0 +1,91 @@
...
+    This dict is passed directly to ``MitigationTask.prepare()`` (and its subclasses)
+    as the ``custom_boxing_options`` argument.
+
+    Noise-injection options (``inject_noise_*``) are **not** set here — ``PEC._box_circuit()``
+    and ``PEA._box_circuit()`` enforce their own required values for those fields.
+
+    ``enable_measures`` and ``measure_annotations`` are also **not** set here. When TREX
+    is passed to a task, ``trex._edit_boxing_options()`` forces ``enable_measures=True``
+    and ``measure_annotations="all"``. When TREX is absent, ``MitigationTask._box_circuit()``
+    defaults to ``measure_annotations="change_basis"``, which is correct.
```
SamFerracin, line 39: Here and elsewhere, I would avoid mentioning private attributes/methods of another repo as they may change in time
```suggestion
    This dict is passed directly to ``MitigationTask.prepare()`` (and its subclasses)
    as the ``custom_boxing_options`` argument.

    Noise-injection options (``inject_noise_*``) are **not** set here — ``PEC``
    and ``PEA`` enforce their own required values for those fields.

    ``enable_measures`` and ``measure_annotations`` are also **not** set here. When TREX
    is passed to a task, ``trex`` forces ``enable_measures=True``
    and ``measure_annotations="all"``. When TREX is absent, ``MitigationTask``
    defaults to ``measure_annotations="change_basis"``, which is correct.
```

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -162,86 +160,137 @@ def _validate(
...
-            add_tags=add_tags,
+        num_randomizations, shots_per_randomization = 1, shots
+
+    # Setup
+    boxing_opts = estimator_options_to_boxing_options(twirling, inject_noise, add_tags)
+    noise_model = (
+        layer_noise_model_to_dict(resilience.layer_noise_model)
+        if resilience.layer_noise_model is not None
+        else {}
+    )
```
SamFerracin: You're only using `layer_noise_model_to_dict` once, i'd kill it and just do this
```suggestion
    noise_model = {}
    if resilience.layer_noise_model is not None:
        noise_model = {
            annotation.ref: pauli_map
            for instr, pauli_map in layer_noise_model
            if annotation := get_annotation(instr.operation, InjectNoise)
        }
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -117,424 +62,87 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
+    """Dict factory that converts ``expanded_values`` tuples to lists.
 
-    Raises:
-        ValueError: If ``item_result`` has no ``'_meas'`` key.
-        ValueError: If ``item_result['_meas']`` has a number of axis not equal to ``4``.
-        ValueError: If ``item_result`` has no ``'pauli_signs'`` key.
-        ValueError: If ``param_shape`` and ``observables.shape`` cannot be broadcasted against
-            each other.
+    Used as the ``dict_factory`` argument when calling ``dataclasses.asdict``
+    on stretch-value objects that contain ``expanded_values`` fields.
```
SamFerracin: In general, I'd avoid describing ow we use these utility functions, as we may end up using them in more places in the future
```suggestion
```

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -545,694 +153,88 @@ def create_pub_result_metadata(item_metadata: ItemMetadata | dict) -> dict[str,
...
-            Supported models (each fits the named function of the noise factor ``x``):
-            - ``"linear"``: ``a + b*x``
-            - ``"polynomial_degree_k"`` (1 <= k <= 7): a degree-k polynomial
-            - ``"exponential"``: ``a*exp(b*x)``
-            - ``"double_exponential"``: ``a*exp(b*x) + c*exp(d*x)`` (rates constrained to decay)
-            - ``"fallback"``: no fit; the measured value at the lowest noise factor
-        measure_noise_data: Measurement noise calibration data for TREX mitigation. Can be either a
-            PauliLindbladMap of a noise model learned upfront, or a result of a calibration circuit.
+    qiskit-mitigation's broadcast path for ``MitigationTask`` and ``PEC``
+    produces a ``DataBin`` with::
```
```suggestion
    produces a ``DataBin`` with:
```

qiskit_ibm_runtime/executor_estimator/options_to_mitigation.py
SamFerracin, line 39: Same as above
```suggestion
```

qiskit_ibm_runtime/executor_estimator/prepare.py
SamFerracin: I was going to comment something similar. Including `options` instead of `opts`. We also have a bunch of var names like `i`, `j`, and `cm` in the PR that make it harder to read

qiskit_ibm_runtime/executor_estimator/prepare.py
```diff
@@ -162,86 +160,137 @@ def _validate(
...
-    return quantum_program
+    return qp
+
+
+def _normalise_extrapolator(extrapolator: str | Sequence[str]) -> list[str]:
+    """Return ``extrapolator`` as a plain list of strings."""
+    return [extrapolator] if isinstance(extrapolator, str) else list(extrapolator)
+
+
+def _method_kwargs(
```
SamFerracin: Can we find a more descriptive name for this one?

qiskit_ibm_runtime/executor_estimator/trex_setup.py
```diff
@@ -0,0 +1,74 @@
...
+    trex: TREX,
+    quantum_program: QuantumProgram,
+    measure_noise_learning: MeasureNoiseLearningOptions,
+    num_randomizations: int,
+) -> None:
+    """Append the TREX calibration circuit to ``quantum_program``.
+
+    Must be called **after** all per-pub ``task.prepare(trex=trex, ...)`` calls.
+    By that point every task has registered itself on ``trex`` (via
+    ``trex._edit_boxing_options``), so ``trex.tasks`` is fully populated.
```
SamFerracin:
```suggestion
    By that point every task has registered itself on ``trex``,
    so ``trex.tasks`` is fully populated.
```

test/unit/executor_estimator/test_estimator_v2_pea_utils.py
```diff
@@ -12,25 +12,27 @@
 
 """Unit tests for EstimatorV2 PEA helper functions."""
```
SamFerracin, line 13: We're not testing PEA utils anymore, but prepare PEA. We should consider changing the file name, same for PEC

SamFerracin (commented):
Thanks, Tsafrir! Looks good! I left some minor comments 😄 

Another comment:
This PR introduces a few utils with an underscore, such as `_create_pub_result_metadata`, and a few utils without underscore, such as `expanded_values_to_lists`. I can't seem to understand what's motivating adding an underscore or not adding it. Would it make sense to remove all underscores, in line with what we (used to?) do in the rest of the repo?

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: This should work. I was just wondering about the coupling between the packages. This solution assumes something about the way qiskit-mitigation orders the items.

pyproject.toml
TsafrirA: Yes, but we now have a released version. (I kept it unpinned so it would catch changes in qiskit-mitigation)

qiskit_ibm_runtime/executor_estimator/pec_utils.py
TsafrirA: Moved to `executor_estimator/pec_utils.py`.

qiskit_ibm_runtime/executor_estimator/utils.py
TsafrirA: Nice catch. I will move in a separate PR for an easier review.

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: `type[MitigationTask]` is the correct choice here. Thank you.

test/unit/executor_estimator/test_estimator_v2_pea_utils.py
TsafrirA: Will change names in a follow up PR to simplify the review process.

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: Thank you. Revised a bunch of shorthands in [c1cadb2](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3310/commits/c1cadb29c9818d3f8654ba2a052bfe6f1c63cb35).

qiskit_ibm_runtime/options_models/zne.py
TsafrirA: Moved to `executor_estimator/utils.py`.

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: Thank you, seems to be an oversight.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: Thank you.

qiskit_ibm_runtime/executor_estimator/trex_setup.py
TsafrirA: Thank you. I verified we can use our subclass. I think there was a issue that led me to import directly from samplomatic, but I don't see any problem now.

qiskit_ibm_runtime/executor_estimator/options_to_mitigation.py
TsafrirA: Seems to overlap with the previous comment.

qiskit_ibm_runtime/executor_estimator/prepare.py
TsafrirA: I went for the clunky `build_task_prepare_kwargs`, but I am open for other ideas.

qiskit_ibm_runtime/executor_estimator/prepare.py
diego-plan9: Thanks! From my side, the only ones I would love to have tackled are two `qp` mentions in the auxiliary functions in the tests (`_prepare_foo()`). I leave it to @SamFerracin to double check if there are remaining ones.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
SamFerracin: Thanks, Tsafrir! As discussed privately, my preference would be to sync with Dekel to tell him about our usecase and ask the best way to handle it. We can also fix in another PR, given that this is not yet going public. It seems natural to wanting to know the ordering, so perhaps we could see if they could expose this?

pyproject.toml
```diff
@@ -189,7 +190,8 @@ dependencies = [
     "pydantic>=2.13.0",
     "qiskit>=2.3.0",
     "pybase64>=1.4.0",
-    "samplomatic>=0.21.0"
+    "samplomatic>=0.21.0",
+    "qiskit-mitigation>=0.1.0",
```
SamFerracin, line 194: do we trust this, or do we want to be cautious at least initially, and set it to ==?

qiskit_ibm_runtime/executor_estimator/prepare.py
diego-plan9: Thanks!

qiskit_ibm_runtime/executor_estimator/utils.py
diego-plan9: Great, formalized at #3318

SamFerracin:
This PR looks good to me now, thanks for addressing the comments 😄 Let's see if other reviewers have more

From my perspective, two points that can be tackled here or in follow-up PRs:
- The TODO [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3310#discussion_r3992249413).
- The version of qiskit-mitigation, comment [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/3310#discussion_r4006282351). Until the next release, this will only inconvenience us (tests may break if version 0.2 has issues) but not users, so we can think about this some more

pyproject.toml
TsafrirA: Good point. Pinned only to 0.1.0 now.

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
TsafrirA: I am leaving the todo there, but I changed the code to use `pub_index` instead of the private attribute. And will open an issue in qiskit-mitigation.

diego-plan9 (dismissed):
Thanks Tsafrir! Nice outcome for a complex integration.

apchytr (dismissed):

qiskit_ibm_runtime/decoders/executor_estimator/post_processor_v0_1.py
```diff
@@ -117,424 +62,94 @@ def estimator_v2_post_processor_v0_1(result: QuantumProgramResult) -> PrimitiveR
...
-        raise ValueError("Missing 'param_shapes' in post_processor data.")
-
-    # Extract circuit metadata if present
-    circuits_metadata = post_processor_data.get("circuits_metadata", None)
-
-    # Extract mitigation data
-    mitigation = post_processor_data.get("mitigation", None)
-    pec_gammas = post_processor_data.get("pec_gammas", None)
+    num_pubs: int = post_processor_data["num_pubs"]
+    circuits_metadata: list[Any] = post_processor_data.get("circuits_metadata") or []
```
```suggestion
    circuits_metadata: list[Any] = post_processor_data.get("circuits_metadata") or [] * num_pubs
```

TsafrirA: Thanks. Fixed.
(missing the `None` in the suggestion)

SamFerracin (approved):

PR #3314, opened 2026-09-11
Support dry-run mode when executing a primitive

### LINKED ISSUE #3312: Support job dry-run mode

A new feature is to become available in the API that allows for more precise job estimation by using specific backends. In order to take advantage of it, we want to expose it as a `dry_run` flag in the existing Primitive execution methods, that when `True`, it takes advantage of the new feature by transparently using the specific backends instead of the regular ones.

Please note this will be rolled out gradually, and will not be immediately available globally once this package's implementation is merged.

**Acceptance criteria**

The Primitives `run()` methods accept a `dry_run` flags that does not run a regular job but instead uses the backends specific to job estimation.

diego-plan9:
### Summary

Add a `dry-run` flag to existing primitives `run()` methods that, when passed, causes the job to be an estimation/validation job, as opposed to a regular job. This is achieved by routing the job to a different backend based on the original one:
* all public `run()` method gain a `dry_run` flag, defaulting to `False`
* internal methods that do the execution (`QiskitRuntimeService._run()` and the intermediate ones) gain the flag.
* `QiskitRuntimeService._run()` contains the logic

### Details and comments

Fixes #3312

### AI/LLM disclosure


qiskit_ibm_runtime/fake_provider/local_service.py
```diff
@@ -162,6 +163,8 @@ def _run(
                 to the IBM Quantum Compute program.
             options: Runtime options that control the execution environment.
             calibration_id: The calibration id to use with the program execution
+            dry_run: If ``True``, submit the job for estimation and validation, not for job
+                execution. This is not supported in a local service.
```
diego-plan9, line 167: Added to `QiskitRuntimeLocalService` interface too to keep parity with function signatures even if is a no-op, as the spirit of "local mode" is for the user to be able to make minimal changes to their script (ie. just change the backend).

qiskit_ibm_runtime/executor/executor.py
```diff
@@ -107,17 +107,21 @@ def __setattr__(self, name: str, value: Any) -> None:
 
         super().__setattr__(name, value)
 
-    def run(self, program: QuantumProgram) -> RuntimeJobV2 | LocalRuntimeJob:
+    def run(self, program: QuantumProgram, dry_run: bool = False) -> RuntimeJobV2 | LocalRuntimeJob:
         """Run a quantum program.
 
         Args:
             program: The program to run.
+            dry_run: If ``True``, submit the job for estimation and validation, not for job
```
SamFerracin: Do you know if there's going to be a guide for dry-run mode? As a user, I would have tons of questions, for example about billing. Eg if I do dry-run mode in a session, do I get charged?

qiskit_ibm_runtime/fake_provider/local_service.py
```diff
@@ -292,17 +298,23 @@ def _run_executor(
...
+        dry_run: bool = False,
     ) -> LocalRuntimeJob:
         """Run an executor program.
 
         Args:
             backend: The backend to run the executor program on.
             options: Simulator options to use.
             inputs: The executor program to run.
+            dry_run: If ``True``, submit the job for estimation and validation, not for job
+                execution. This is not supported in a local service.
```
SamFerracin: Maybe this?
```suggestion
                execution. This is ignored in a local service.
```

qiskit_ibm_runtime/qiskit_runtime_service.py
```diff
@@ -1090,6 +1096,18 @@ def _run(
         if isinstance(backend, str):
             backend = self.backend(name=qrt_options.get_backend_name())
 
+        # Take into account `dry_run`, replacing the backend with its mock counterpart.
+        if dry_run:
+            backend_name = re.sub(r"^[^_]+", "mock", backend.name)
```
SamFerracin, line 1101: I guess you're replacing everything before the first `_`, and not simply `ibm_`, because dry-run mode is supposed to work for `test_` devices too?

test/unit/executor_estimator/test_estimator_v2.py
```diff
@@ -395,6 +398,25 @@ def test_run_raises_error_when_pec_and_zne_both_enabled(self):
...
 
 
+class TestEstimatorV2RunNoPatching(IBMTestCase):
+    """Tests for the EstimatorV2.run() method (with no Python methods patching)."""
+
+    @mock_responses(OneInstanceDryRunRegistry)
+    def test_run_dry_run(self, registry):
+        """Estimator can run in `dry-run` mode."""
+        service = QiskitRuntimeService(token="my_token")
+        backend = service.backend("ibm_foo")
```
SamFerracin, line 408: See my previous comment, may be worth parametrizing over `startswith` in `['ibm_', 'test_']`

qiskit_ibm_runtime/executor/executor.py
jyu00: +1 on this. We already got questions about the difference between mock backends and fake backends. 

In addition to a guide, the docstring should explain what this is a bit more.

diego-plan9: Tweaked the docstring at 2029bffafd9f0b253db83a751f06cff7566dd9ba (credit to @francabrera for the content).

Overall, this PR is mostly the client providing access to the feature already for a subset of users before it is rolled out widely, so it can already be present in the upcoming `0.50` release.. More comprehensive documentation and details on usage are expected as part of the later rollout (which will also require changes in the client related to API changes and a new release later on anyway, so we have the chance to refine further accordingly).

qiskit_ibm_runtime/fake_provider/local_service.py
diego-plan9: Bundled with 2029bffafd9f0b253db83a751f06cff7566dd9ba

qiskit_ibm_runtime/qiskit_runtime_service.py
diego-plan9: It's more of an attempt to being generic (and only leave the "hard-coding" to the `mock` bit). It felt like an artificial restriction to explicitly capture `ibm_` (even if in practice, all public backends name start with `ibm_` so it should not make a difference), and the follow-up code should fail gracefully if the substitution results in a backend that does not exist. I do not have clarity on the specific for test devices: so in a way it does feel safer/more generic to just capture the first block before `_`.

diego-plan9:
Added an integration test, and verified it by locally. It currently requires the additional backend to be `active`, which might not be the case yet in all environments. Verified locally that the test does run and pass by relaxing the condition at 352 (it does not need to be fully active technically in order to receive a job, and be able to check the right backend) - but re-enabled it in order to prevent extra jobs being queued until backends are active.

test/integration/test_backend.py
```diff
@@ -333,3 +334,29 @@ def test_backend_calibration_id(self):
...
+                if backend.name.startswith("mock_") and backend.status().status_msg == "active"
+            )
+            backend_name = re.sub(r"^[^_]+", "ibm", dry_run_backend.name)
+            backend = self.service.backend(backend_name)
+        except (StopIteration, QiskitBackendNotFoundError):
+            self.skipTest("No dry_run backends available.")
+
+        isa_circuit = transpile(bell(), backend)
+        sampler = Sampler(mode=backend)
+        job = sampler.run([isa_circuit], dry_run=True)
```
SamFerracin, line 361: We never check that a result comes back, but given what we're trying to test here, it makes sense to me

SamFerracin (commented):
This looks good to me, with the additional fact that a guide will be coming later. 
I'll let @jyu00 take one more look and approve when she's ready

Thanks, Diego! 😄

qiskit_ibm_runtime/executor/executor.py
```diff
@@ -107,17 +107,24 @@ def __setattr__(self, name: str, value: Any) -> None:
...
+    def run(self, program: QuantumProgram, dry_run: bool = False) -> RuntimeJobV2 | LocalRuntimeJob:
         """Run a quantum program.
 
         Args:
             program: The program to run.
+            dry_run: If ``True``, performs a dry run without executing the job on a QPU. This mode
+                can be used to validate the job, estimate usage consumption, and retrieve circuit
+                timing metadata before submission. Returned results preserve the expected schema
+                but contain randomized mock data rather than actual or simulated measurement
+                results. Access to this feature may be restricted.
```
jyu00:
```suggestion
            dry_run: If ``True``, performs a dry run without executing the job on a QPU. This mode
                can be used to validate the job, estimate usage consumption, and retrieve circuit
                timing metadata. Returned results preserve the expected schema
                but contain **randomized mock data** rather than actual or simulated measurement
                results. Unlike the fake backends, the processing of this dry run happens on the server-side, so the job may not finish immediately and access to this feature may be restricted.
```

A few minor suggestions
- Remove `before submission` since it does make a submission to the server. 
- Highlight that it returns random data
- Added a sentence about the job may not return immediately

diego-plan9: Updated in 5fef5cb.

SamFerracin (approved):
The wording looks good to me now, so from my side it's good to go in!

PR #3326, opened 2026-09-16
Fix boxing options related to measure mitigation

TsafrirA:
### Summary
Starting with #3310, a user warning started appearing indicating a difference between the layers identified by the prepare stage and the stand alone `EstimatorV2.find_unique_layers`. This difference boils down to the `estimator_options_to_boxing_options` function assuming that some options will be set by the `MitigationTask`. This assumption was true for the prepare stage, but false for the `find_unique_layers` function.

Here we simply set the options explicitly instead of relying on `qiskit-mitigation`.

### AI/LLM disclosure


test/unit/executor_estimator/simulations/utils.py
```diff
@@ -136,7 +136,7 @@ def create_estimator_test_data_extended(backend, preset_pass_manager, include_pr
 
 def create_noise_model_without_noise(estimator, pub):
     """Creates a noise-model, mapping each layer to the identity (no noise)."""
-    layers = estimator.find_unique_layers([pub], types="gates")
+    layers = estimator.find_unique_layers([pub], types="all")
```
TsafrirA, line 139: Needs to change because the estimator is set up with `all` in its simulator noise model.

qiskit_ibm_runtime/executor_estimator/options_to_mitigation.py
```diff
@@ -51,7 +50,9 @@ def estimator_options_to_boxing_options(
     return {
         # Gate twirling is on when requested OR when noise injection is needed.
         "enable_gates": twirling_options.enable_gates or inject_noise,
+        "enable_measures": True,
```
apchytr, line 56: Does this need to be always `True` or should this depend on `twirling_options.enable_measure`?

SamFerracin: This needs to always be `True` because in samplomatic, it is not actually requesting to twirl measurements, but simply to box them up. The annotations (whether it is ChangeBasis, Twirl, or both) is controlled by other inputs of the pass manager. See docs of boxing pass manager in samplomatic

qiskit_ibm_runtime/executor_estimator/options_to_mitigation.py
```diff
@@ -51,7 +50,9 @@ def estimator_options_to_boxing_options(
...
         "twirling_strategy": twirling_options.strategy.replace("-", "_"),
         "twirling_group": twirling_options.group,
         "add_tags": "unique_box" if add_tags else "none",
+        "measure_annotations": "all" if measure_mitigation else "change_basis",
```
SamFerracin: should't this be like this?
```suggestion
        "measure_annotations": "all" if (measure_mitigation is not None or twirling.enable_measure) else "change_basis",
```

TsafrirA: You're right (up to `measure_mitigation` being bool). Fixed.

SamFerracin (approved):

PR #3328, opened 2026-09-16
Disable no_site_packages in mypy configuration

diego-plan9:
### Summary

Update the mypy configuration in order to:
* take into account installed dependencies that provide type hints.
* revise specific rules for `test/`, so it steers us towards type hinting auxiliary functions, and unify how imports are treated (now the same as for the library)

### Details and comments

Some dependencies are still not taken into account (in `[tool.mypy.overrides]`). With this PR, `mypy` will check the dependencies that provided type stubs in the upstream package directly. A number of the remaining ones also provide support through them in `typeshed` or as other dedicated packages: I went with not including for this round as it needs checking for versions compatibility and accuracy/issues.

This also involves reverting the direction from #3156:
* with this change, `mypy` needs to have access to the dependencies installed. With the current approach in `main`, it will run in an isolated venv with just `mypy` installed - and no way for `pre-commit.yml` to install the package and its dependencies.
* reverting to the "assume that `mypy` will be run in a venv managed by the user" solves it, as users/CI control what to install before invoking. 
* `uv` users are likely to be able to replicate `pre-commit` behavior via using `exec uv run pre-commit ...` when manually invoking, or customizing their locally installed git hook
* the alternative to reverting is copy-pasting the dependencies into `pre-commit.yml` - which is prone to keeping them out of sync

Related to #3295

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude Opus 4.8 for some of the typing


pyproject.toml
```diff
@@ -60,34 +60,29 @@ disallow_untyped_defs = true
...
     "qiskit.*",
     "qiskit_mitigation.*",
     "requests_ntlm.*",
-    "responses.*",
-    "rustworkx",
     "samplomatic.*",
-    "urllib3.*",
-    "pytest",
+    "scipy.*",
 ]
```
rainerschoe, line 89: I like it. will we eventually be able to have this list empty?

.pre-commit-config.yaml
```diff
@@ -23,9 +23,8 @@ repos:
         name: mypy
         entry: mypy
         pass_filenames: false
-        language: python
+        language: system
         require_serial: true
-        additional_dependencies: ["mypy==2.3.0"]
```
rainerschoe, line 28: makes sense if we now require the full dependencies installed, it will include correct mypy version :-)

rainerschoe (approved):
Looks good! Let's merge this.

But will it make sense to update `CONTRIBUTING.md` to explicitly state that pre-commit needs to be run with all project dependencies installed?

pyproject.toml
diego-plan9: We can trim it easily for the packages that provide type stubs as an additional dependency (the ones removed already include them in the main package: others include it as an optional package to install, either in https://github.com/python/typeshed or as a fully separate package). I did not prune further in this round as they need to be vetted for license and version compatibility (they should be fine but require some digging and exploring - I think only the one for `scipy` seemed a bit problematic, as there was some incompatibility with Python versions and support for older versions).

Some of them do not provide type stubs at all, though - those are the ones we would be forced to keep. Unless switching to other type checker that does not depend on the type stubs being present. Actually it is something I'd like to explore at some point (there are more performant alternatives than `mypy` and with different set of features, although ultimately we likely will need to clean up our type hints prior to considering switching).

diego-plan9:
> But will it make sense to update `CONTRIBUTING.md` to explicitly state that pre-commit needs to be run with all project dependencies installed?

Thanks - tweaked at 85cf995

rainerschoe (approved):

PR #3333, opened 2026-09-17
Add backend to client-side primitives and mode property to all primitives

diego-plan9:
### Summary

* Add a `.backend()` method for client-side `Estimator` and `Sampler`, to get the interface on par with `Executor` and `Sampler` - as well to a closer degree with the legacy primitives.
* Add a `mode` property for client-side `Estimator` and `Sampler`
* Add a `mode` property to the existing primitives (`Executor`, `NoiseLearnerV3`) that were missing it

### Details and comments

Related to #3073 

### AI/LLM disclosure


SamFerracin (dismissed):

qiskit_ibm_runtime/executor/executor.py
```diff
@@ -90,7 +90,7 @@ def __init__(
         # Coerced to `ExecutorOptions` via `__setattr__()`.
         self.options = options if options is not None else ExecutorOptions()  # type: ignore[assignment]
 
-        self._session, self._service, self._backend = get_mode_service_backend(mode)
```
SamFerracin, line 93: was this just a typo?

qiskit_ibm_runtime/executor/executor.py
```diff
@@ -107,6 +107,19 @@ def __setattr__(self, name: str, value: Any) -> None:
 
         super().__setattr__(name, value)
 
+    def backend(self) -> BackendV2 | None:
```
SamFerracin, line 110: Kind of weird that backend is a method, but mode is a property. Not this PR's fault, but I wonder why we made those choices (likely, years ago)

SamFerracin, line 110: We should do some investigation to try and understand what `None` represents here. I see that in theory, one can also do `executor = Executor(mode=None)`, but we don't document what that means.

SamFerracin (approved):
Left a few comments that may lead to follow-up PRs, but this looks good!

qiskit_ibm_runtime/executor/executor.py
diego-plan9: The change in this PR is intended. I think it was a slight (harmless, as the var is internal) introduction in a29deaa that did not surface until now (when using that variable for exposing it as `mode`). The first returned item is _usually_ a session, but can also be a Batch (or `None`) - I assume when it was first introduced (nlv2 or Executor) most of the manual testing (and the logic that acted on it, as `Batch` is a subclass of `Session`, was seemingly-session-specific) just assumed session instead of the wider name.

qiskit_ibm_runtime/executor/executor.py
diego-plan9: Something to investigate for the future indeed (the weight of historical choice in this one might be high, less on  `mode`).

diego-plan9: Yes - small ping to https://github.com/Qiskit/qiskit-ibm-runtime/issues/2874. I _think_ we should always be able to retrieve a backend (either as the third item returned by the function, or likely querying the first argument for its backend).

And also to https://github.com/Qiskit/qiskit-ibm-runtime/pull/3340:
> There is probably some duplication we can take care of already (interfaces for the local service and real service might be close enough already to reduce the branching)

Both of them resolve over going carefully into `get_mode_service_backend` - I'll follow up on both.

PR #3334, opened 2026-09-17
Make `InsertNoisePass` support `noise_after` on a per layer basis 

### LINKED ISSUE #3329: `InsertNoisePass` always inserts before/after a barrier
**Describe the bug**
The current implementation of `InsertNoisePass` will always inject noise after the tagged barrier unless `noise_after` is set to `False` (note: this option is not exposed to the user). This causes issues e.g. for simulating readout errors as the noise is inserted only after the measurements barrier meaning nothing is applied. 

**Steps to reproduce**
MWE:
```
from qiskit.quantum_info import PauliLindbladMap, SparsePauliOp
from qiskit_aer import AerSimulator

from qiskit_ibm_runtime.executor_estimator import EstimatorV2

backend = AerSimulator()
circuit.measure_all()

preset_pass_manager = generate_preset_pass_manager(backend=backend)

isa_circuit = preset_pass_manager.run(circuit)

observable = SparsePauliOp("Z" * backend.num_qubits)
pub = (isa_circuit, observable)

estimator = EstimatorV2(mode=backend, options={"resilience_level": 0})
estimator.options.simulator.layer_noise_model = [
    (layer, PauliLindbladMap.from_list([("X" * layer.operation.num_qubits, 0.005)])) 
    for layer in estimator.find_unique_layers([pub], types="all")
]

job = estimator.run([pub])
job.result()
print(job.result()[0].data.evs)
```
when `noise_after=True` (default) the expectation value is 1 but when `noise_after=False` it behaves as expected and returns a value <1.

**Expected behavior**
Noise injection via the `InsertNoisePass` should correctly insert noise before or after the barrier depending on the operation e.g. after the ideal gate operation or before the ideal measurement operation.

**Suggested solutions**
One possible solution is to modify the `InsertNoisePass` to work by explicitly identifying barriers but this may require adding and additional `noise_site` argument. 

**Additional Information**

- **Python version**:
- **Operating system**:

apchytr:
### Summary
Adding support to `InsertNoisePass` so that `noise_after` can be specified on a per layer basis. This resolves the issue where noise was being injected after measurements. Rather than specifying a global `noise_after` arg, `InsertNoisePass` now accepts a `noise_pos` dictionary mapping layer `Tag` references to their respective barrier positions.

### Details and comments

Fixes #3329

### AI/LLM disclosure


qiskit_ibm_runtime/fake_provider/executor/run_quantum_program.py
```diff
@@ -50,6 +53,59 @@ def _round_to_clifford(values: np.ndarray, decimals: int) -> np.ndarray:
...
+    box_type = find_box_type(layer)
+
+    if box_type == "measurement":
+        return "M"
+
+    if box_type == "gates" and (twirl := get_annotation(layer.operation, Twirl)) is not None:
+        if twirl.dressing.value == "right":
+            return "L"
+
+    return "R"
```
SamFerracin, line 114: mmmh.. 🤔 
this will return `"R"` for `"unknown"`.. do we want to be more conservative and raise an error ("not supported yet")? essentially, this would amount to enabling local-mode for circuits that contain boxes of gates or measurements, not for circuits with mixed boxes and/or circuits that have resets

apchytr: Yeah, I think that's fair. Then when we follow up we can think about what the best way to handle the "unknown" case 👍

qiskit_ibm_runtime/executor_estimator/utils.py
```diff
@@ -279,35 +279,3 @@ def find_unique_layers(
     return find_unique_box_instructions(
         instructions=instructions, normalize_annotations=None, undress_boxes=True
     )
-
-
-def find_box_type(instruction: BoxOp) -> BoxType:
```
apchytr, line 284: Just moving this as it avoids a circular import in `run_quantum_program.py`

qiskit_ibm_runtime/fake_provider/executor/run_quantum_program.py
```diff
@@ -50,6 +55,65 @@ def _round_to_clifford(values: np.ndarray, decimals: int) -> np.ndarray:
...
+        raise TypeError(f"Unsupported box type: {box_type}")
+
+    if box_type == "measurement":
+        return "M"
+
+    if box_type == "gates" and (twirl := get_annotation(layer.operation, Twirl)) is not None:
+        if twirl.dressing.value == "right":
+            return "L"
+
+    return "R"
```
SamFerracin: Minor change that saves a few lines and an if statement:
```suggestion
    if box_type == "measurement":
        return "M"

    if box_type == "gates":
        if (twirl := get_annotation(layer.operation, Twirl)) is not None and twirl.dressing.value == "right":
            return "L"
        return "R"
            
    raise TypeError(
         "Unsupported box type: Simulations are only supported for "
         "circuits that contain gate boxes and/or measurement boxes,"
         "but not mixed boxes containing both gates and measurements "
         "or boxes containing instructions of different type."
     )
```

I also changed the error message to make it clear. A user that sees an error that mentions an "unknown box" will be very confused

SamFerracin (approved):
Thanks!

Issue #3323, opened 2026-09-15
Reorganize test files for executor-based estimator


SamFerracin:
**What is the expected enhancement?**
Following PR #3310, we are in a situation where we can tidy up tests.
A few tasks that we can take:

- [x] Move unit tests in such a way that the structure of tests files matches that of the files in the source code. For example, given that we no longer have a `estimator_v2_pea_utils.py` file, we can move the tests in `test_estimator_v2_pea_utils.py` to `test_estimator_v2_prepare.py`, and remove that test file.
- [x] After we have achieved a 1:1 matching between source and test files, we can add unit tests for the functions that are missing one, for example those in `pec_utils.py`
- [x] Finally, we should take a look at the warnings raised by the various tests, and find out how we can test them (if the intention is to test the warning) or make sure that the test doesn't warn

PR #3335, opened 2026-09-17
Modify tests for client-based estimator so that they do not emit warnings

SamFerracin:
### Summary
I run through all of our tests for client-based primitives, and found that 6 of them are emitting warnings that can be avoided. I have removed the test cases that trigger warning, replacing them with dedicated tests that assert that a warning is emitted.

Third step of issue #3323

### AI/LLM disclosure


diego-plan9 (approved):
Thanks, this is a nice pattern for gaining accountability over "when do we expect warnings to be emitted in tests" - and leaving the default mechanism in place (ie. pytest collecting other warnings) for the time when we are able to have a more sane suite output.

PR #3339, opened 2026-09-17
More general control of simulator noise map insertion

DanPuzzuoli:
### Summary

This is a proposed more general solution #3334.

The main goal is to expand the simulator options to enable complete control of which barrier a noise model is inserted in each box, and to make sure that there is a robust/general way to specify preparation noise (which sometimes has no associated box).

### Details and comments

The main change is to modify `InsertNoisePass`: to accept noise insertion locations as:
- It now accepts the following locations for the noise insertion: `"L"`, `"M"`, `"R"`, `"before"`, or `"after"`. The first three indicate the barriers when a boxed circuit is turned into a template circuit. The last two are higher level options meant to replicate the behaviour of the `noise_after` argument. The goal with `"before"`/`"after"` is to insert the noise immediately before or after the box body, and these get turned into one of the barrier locations based on the properties of the box (e.g. which side the dressing is on).
   - The goal of the direct barrier-insertion control options is to enable the user to have complete control over where the noise is inserted if the default behaviour is not sufficient for their use case.
   - The default is to specify `"after"`, unless a measurement is in the box, in which case it is set to `"before"`.
- It can now supply multiple noise models for a single box by specifying different barriers.
   - This is pretty much a free enhancement, though probably has very limited scope until we start worrying about MCM circuits. While it isn't completely general, one can model a noisy MCM with a Pauli channel both before AND after the operation (hence requiring the insertion of noise channels at barriers coming from the same box).

This PR also adds a `preparation_noise` option to the `SimulatorOptions`. The purpose of this option is to enable the insertion of a noise model at the very beginning of a circuit (before any basis changes happen), in a case where there is no box to attach it to. 

### Issues to resolve

If its decided to move forward with this PR, there are some remaining issues to be aware of:
- This PR introduces a `PositionedLayerNoiseModel`, which is basically like `LayerNoiseModel` but with the position specification. I don't think there's any reason to have both, but wasn't sure what the preferred choice would be here. I think I would just add the position location to `LayerNoiseModel`.
- A general issue with modelling noise in this context is that samplomatic will sometimes move single qubit gates out of box bodies and into the dressings. E.g. a left-dressed box that has any leading single qubit gates will have them merged into the dressing. This means that if I add a noise model and specify `"before"`, it will be inserted at the `"M"` barrier, but the single qubit gates in the box definition are now occurring BEFORE the `"M"` barrier, and as such the noise map is being applied part way through the original box body definition. I don't know if there's anything to "resolve" here, it's just a limitation that should be documented.
- Related to the above point: any box that _only_ has single qubit gates will end up having an empty body, as everything gets moved into the dressing. In this case, two of the barriers will appear in an identical location in the circuit, making `"before"`/`"after"` redundant/ambiguous. For now I've set up this branch to raise an error if this case is detected, telling the user to explicitly specify one of the three barriers. I don't think the resultant behaviour of not raising is necessarily _wrong_ - it's more of an edge case that I think the user should be made aware of.

### AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Opus 5


qiskit_ibm_runtime/options_models/simulator.py
```diff
@@ -42,6 +45,86 @@ def validate_layer_noise_model(value: LayerNoiseModel | None) -> LayerNoiseModel
...
+    nothing else is equivalent to an empty one: the barriers on either side of it end up adjacent.
+
+    Args:
+        body: The body of a boxed layer.
+    """
+    # `measure` and `reset` are single-qubit but are not standard gates, so they are not absorbed.
+    return all(
+        instruction.is_standard_gate() and instruction.operation.num_qubits == 1
+        for instruction in body.data
+    )
```
diego-plan9, line 71: Acknowledging it is a draft, just a small "stylistic choice" /  "option_models structure" comment: if this function is just a comparison, and invoked only in one place, can you just make the logic part of `validate_positioned_layer_noise_model`?

This would make the validation self-contained, and reduce a bit the level of indirection when maintaining the code.

joshuasn:
I think this PR should be greatly simplified by exclusively having a method that reads the annotations from a noise model to determine whether to insert at L, M, or R.

The way you're doing preparation noise also makes sense but probably makes sense to explore in a separate PR. I think we could get it to work with little effort with its own box.

PR #3340, opened 2026-09-18
Fix integration tests related to dry-run

diego-plan9:
### Summary

Follow up to #3314, fixing the passing of `dry_run` (missing in some branches, set to `True` in others) throughout the run methods. This caused the integration tests to fail (luckily, as those were actual bugs in the previous PR).

### Details and comments

Small ping to #2874, as it is one of the factor that contribute to having multiple `if`s and branches inside the `_run` methods. There is probably some duplication we can take care of already (interfaces for the local service and real service might be close enough already to reduce the branching), but I'm not tackling it on this PR in order to prioritize getting the integration tests up and running and look at branching with different priority.

Related to #3312

### AI/LLM disclosure


diego-plan9:
Issuing this against main repo to take advantage of running the integration tests in CI.

qiskit_ibm_runtime/noise_learner/noise_learner.py
```diff
@@ -217,15 +218,15 @@ def run(
                 options=runtime_options,
                 inputs=inputs,
                 calibration_id=calibration_id,
-                dry_run=True,
```
SamFerracin, line 220: oh shoot! great we caught this!

SamFerracin (approved):

PR #3343, opened 2026-09-18
Ensure Primitives backend method disallow none

diego-plan9:
### Summary

Follow up to comments in #3333, revising type hints and `get_mode_service_backend`. All paths do return a `backend` object and a `service` object, with the `mode` item of the tuple being able to return `None`.

Adjusted the `.backend()` returns annotation accordingly, and revised some type hints in general.

### Details and comments

The original comments also suggested taking the chance and avoid some `if` logic that takes into account `service`, `mode`, and `backend` for routing to the right `_run()` method. Simplification _can_ be done in `base_primitive.py` and `noise_learner.py`, but since we are on the verge of considering them deprecated, I opted not to include the changes.

Fixes N/A

### AI/LLM disclosure


qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
```diff
@@ -116,7 +117,7 @@ def mode(self) -> Session | Batch | None:
 
     def run(
         self, instructions: Iterable[CircuitInstruction], dry_run: bool = False
-    ) -> RuntimeJobV2:
+    ) -> RuntimeJobV2 | LocalRuntimeJob:
```
SamFerracin: I don't think that this is true (yet), because NLV3 does not support local mode
```suggestion
    ) -> RuntimeJobV2:
```

SamFerracin (approved):

qiskit_ibm_runtime/noise_learner_v3/noise_learner_v3.py
diego-plan9: True - reverted at d5612ae45221e5667de63d46b6a4782290452c51 (one more reason for revising the mode approach, as we cannot detect faithfully in `run()` that `QiskitRuntimeLocalService` is not reachable).

SamFerracin (approved):

PR #3342, opened 2026-09-18
Replace SamplerEnvironmentOptions with EnvironmentOptions

SamFerracin:
### Summary
This PR moves `max_execution_time` to `SamplerEnvironmentOptions`, making it become identical to `EnviromentOptions`. It the replaces the former by the latter.

Fixes #3341

### AI/LLM disclosure

- [x] I didn't use LLM tooling, or only used it privately.
- [ ] I used the following tool to help write this PR description:
- [ ] I used the following tool to generate or modify code:

PR #3345, opened 2026-09-18
Move `max_execution_time` to top-level options

SamFerracin:
### Summary
Legacy Sampler and Estimator have `max_execution_time` at top level in the options. On the contrary, NLV3 and Executor (and therefor client-side primitives) have it in `EnvironmentOptions`.

This PR deprecates it from `EnvironmentOptions`, and instead moves it to top-level, so we have it in a single place for all our programs.

### NOTES TO THE REVIER

- The changes should be clearer if you go down commit by commit
- I tested tweaking locally the integration tests (executor, NLV3, client-side primitives). All changes seem to work fine, including the warning

### Details and comments

Fixes a comment that followed #3342 

### AI/LLM disclosure


SamFerracin:
Note to the reviewer: The changes should be clearer if you go down commit by commit

qiskit_ibm_runtime/options_models/converters.py
```diff
@@ -27,17 +27,25 @@
     from .sampler import SamplerOptions
 
 
-def to_runtime_options(options: EnvironmentOptions, backend: IBMBackend) -> dict:
+def to_runtime_options(
+    options: EnvironmentOptions, backend: IBMBackend, max_execution_time: int | None = None
```
SamFerracin, line 31: The signature of this function seems quite ugly and ad-hoc. Next week, under less pressure, I may open a PR to clean it a bit

TsafrirA (approved):
LGTM

diego-plan9 (commented):
In a way I feel this is a bit of a regression from a cleaner design (considering `Executor` the example of "clean" in my eyes: no top-level atomic options, good categorization of the rest of them via the sub-options), and it already surfaces up a bit in `to_runtime_options`:

```python
            options=to_runtime_options(
                self.options.environment, self._backend, self.options.max_execution_time
            ),
```

which now needs to treat `max_execution` as a bit of an exception (similar to `backend`).

I understand the compatibility angle, though, and is a tricky balance (also taking into account the "homogeneous interface for all 4 primitives" angle). I think a different balance might end up being saner:
* keeping `max_execution_time` as a top-level option for `Sampler` and `Estimator`, considering it a sampler/estimator specific choice
* keeping it in `EnvironmentOptions` for `Executor` and `NLV3`, considering it the now "canonical" placement
* as a separate discussion, consider marking it as deprecated (in favor or `EnvironmentOptions`,  even in `Sampler` and `Estimator`) if striving for parity between all the 4 primitives

This sacrifices a bit of parity, but increase the consideration of `Executor` as the "reference design" - which I think it is acceptable, given that in the end, it is to become the "real" interface for all intents and purposes (with other primitives built upon it that can have their specific quirks).

qiskit_ibm_runtime/options_models/base.py
```diff
@@ -28,6 +28,14 @@
 class BaseOptionsModel(BaseModel):
     """Base class for options models."""
 
+    max_execution_time: int | None = None
+    """Maximum execution time in seconds.
+
+    This value bounds system execution time (not wall clock time). System execution time is the
+    amount of time that the system is dedicated to processing your job. If a job exceeds this time
+    limit, it is forcibly cancelled.
+    """
```
diego-plan9, line 37: Surfacing a comment in private, to ensure it does not get lost: adding it to `BaseOptionModels` makes it present in _all_ models, as this is meant to be a base class (and agnostic to domain). Can you instead add the field in every model that needs it instead?

qiskit_ibm_runtime/options_models/environment.py
```diff
@@ -57,6 +58,19 @@ class EnvironmentOptions(BaseOptionsModel):
     limit, it is forcibly cancelled.
     """
 
+    @field_validator("max_execution_time")
+    @classmethod
+    def _warn_max_execution_time_deprecated(cls, value: int | None) -> int | None:
```
diego-plan9, line 63: Can you please opt for the `Annotated` pattern instead of `field_validator`, for consistency (`PostSelectionOptions` has some inspiration).

qiskit_ibm_runtime/options_models/environment.py
```diff
@@ -57,6 +58,19 @@ class EnvironmentOptions(BaseOptionsModel):
...
+        """Warn when ``max_execution_time`` is explicitly set on ``EnvironmentOptions``."""
+        if value is not None:
+            issue_deprecation_msg(
```
diego-plan9, line 66: Can you add a release note to this PR too?

SamFerracin:
Superseded by #3361

